PR #224, opened 2020-04-21
Fix generation of escaped C++ string literals.

### LINKED ISSUE #217: in a bytes constant, had to double backslash the two printable characters
To provide a bytes constant, I had to double backslash the two printable characters, i.e:
```
module tzDatetime;

# public type input0210324115000A21 = unit {
callWith0210324115000A21();

function callWith0210324115000A21() {
local contentExceptLast : bytes = b"\x02\x10\\x32\\x41\x15\x00\x0A"
}
```
else it reports:
```
3323 /usr/local/etc/tmp$printf "12\n" | spicy/build/bin/spicyc -j Issue_217_fwd_ref.spicy
tzDatetime.cc:24:47: error: hex escape sequence out of range
    hilti::rt::Bytes contentExceptLast = "\x02\x102A\x15\x00\n"_b;
                                              ^~~~~~
1 error generated.
[error] jit: failed to execute compilation action.
[error] jit: failed to compile C++ code unit tzDatetime to bitcode
```

rsmmr:
Turns out we can't just encode non-printable characters as \xNN
because C++ compilers look for more than 2 digits when decoding
"\x...". Now using octal \NNN instead, that's always 3 digits.

Closes #217.

### DIFF
--- a/hilti/include/base/util.h
+++ b/hilti/include/base/util.h
@@ -409,6 +409,15 @@ using hilti::rt::escapeUTF8;
 using hilti::rt::expandEscapes;
 
 /**
+ * Wrapper for `escapeBytes` that produces a valid C++ string literal.
+ *
+ * @param s string to escape
+ * @return escaped std::string
+ *
+ */
+inline std::string escapeBytesForCxx(std::string_view s) { return escapeBytes(std::move(s), true, true, true); }
+
+/**
  * Turns an arbitrary string into something that can be used as C-level
  * identifier.
  *
--- a/hilti/include/rt/util.h
+++ b/hilti/include/rt/util.h
@@ -229,9 +229,13 @@ std::string expandEscapes(std::string s);
  * @param str string to escape
  * @param escape_quotes if true, also escapes quotes characters
  * @param escape_control if false, do not escape control characters
+ * @param use_octal use `\NNN` instead of `\XX` (needed for C++)
  * @return escaped string
+ *
+ * \todo This is getting messy; should use enums instead of booleans.
  */
-std::string escapeBytes(std::string_view s, bool escape_quotes = false, bool escape_control = true);
+std::string escapeBytes(std::string_view s, bool escape_quotes = false, bool escape_control = true,
+                        bool use_octal = false);
 
 /*
  * Escapes non-printable and control characters in an UTF8 string. This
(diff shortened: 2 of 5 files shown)


rsmmr:
Will wait for CI with this too.

bbannier:
> Will wait for CI with this too.

To pick up the latest CI setup you'd need push an updated version rebased on `master`.

hilti/include/base/util.h
```diff
@@ -408,6 +408,15 @@ using hilti::rt::escapeBytes;
 using hilti::rt::escapeUTF8;
 using hilti::rt::expandEscapes;
 
+/**
+ * Wrapper for `escapeBytes` that produces a valid C++ string literal.
+ *
+ * @param str string to escape
+ * @return escaped std::string (without the quotes)
```
bbannier, line 415: The `without the quotes` part made me pause. Are you talking about how literals are formatted in (input) C++ source code? -- in that case it should be clear that these quotes are not part of the literal itself, and I wouldn't expect them to be preserved when calling anything but possibly very arcane macros.
```suggestion
 * @return escaped std::string
```

hilti/include/base/util.h
```diff
@@ -408,6 +408,15 @@ using hilti::rt::escapeBytes;
 using hilti::rt::escapeUTF8;
 using hilti::rt::expandEscapes;
 
+/**
+ * Wrapper for `escapeBytes` that produces a valid C++ string literal.
+ *
+ * @param str string to escape
```
bbannier, line 414:
```suggestion
 * @param s string to escape
```

Btw, there's a clang warning for errors like that, #240.

duffy-corelight:
Using octal is not necessarily the solution. At https://en.cppreference.com/w/cpp/language/string_literal I read that: "If a valid hex digit follows a hex escape in a string literal, it would fail to compile as an invalid escape sequence. String concatenation can be used as a workaround."
const char* p = "\xfff"; // error: hex escape sequence out of range
const char* p = "\xff""f"; // OK: the literal is const char[3] holding {'\xff','f','\0'}

rsmmr:
Yeah, I get the part on hex escapes (which I didn't know), but where do see that octal isn't going to work either?  From what I read, octal sequences are always the digits and hence don't run into the same problem.

(String concatenation would be another solution, but that seems more difficult to implemented, and it also just looks funny)

duffy-corelight:
Oh if you have octal working, then go with that. I made this suggestion, because I envisioned that your problem space was for example, the .spicy being literally: b"\x7\4C", something like vector's to_string turning that into "\x07C" and then trying to use that in the Cxx source. Going byte-at-a-time, and always outputting the enclosing double-quotes, e.g. "\x07""C" would succeed in the Cxx source, and be a straightforward transformation. But if you have octal working, stick with that. I've had excellent success for years using the octal strategy. I just learned the literal string concatenation trick on the cppreference page at the above URL, yesterday.

PR #233, opened 2020-04-22
Make sure generated doc items have stable ordering.

bbannier:
When generating type operator and method documentation entries we
previously sorted by `id`. This could lead to non-stable ordering if
`id` was not unique.

We now sort entries by their string representation. This ensures that
the entries generated documentation always have a stable order.


rsmmr:
This in in. (Ended up doing a rebase which probably broke the link to the PR)

PR #236, opened 2020-04-22
Fix broken doc links

bbannier:
This PR fixes broken links in documentation and a `check` target to `doc/Makefile`. We execute `check` when building documentation in CI.


rsmmr:
This looks good but looks like the check target is failing  CI?

bbannier:
> This looks good but looks like the check target is failing CI?

I had changes fixing that failure staged, but uncommitted. Took me way to long to figure out why this worked locally :facepalm:

PR #239, opened 2020-04-22
Catch & report broken escape sequences in HILTI/Spicy code at parse t…

### LINKED ISSUE #210: escapes need three digits, but diagnostic lacks line number or other information with which to locate
```
module escapes_need_three_digits;
   print "\22";
```
makes its report without line number or really anything to locate it.
```
build/bin/spicyc -j ../Issue_211_escape_sequence.spicy
libc++abi.dylib: terminating with uncaught exception of type hilti::rt::Exception: unknown escape sequence
Abort trap: 6
```

rsmmr:
…ime.

(Skipping test, it's only about catchng these.)

Closes #210.

### DIFF
--- a/hilti/src/compiler/parser/scanner.ll
+++ b/hilti/src/compiler/parser/scanner.ll
@@ -33,6 +33,15 @@ static hilti::Meta toMeta(hilti::detail::parser::location l) {
     return hilti::Meta(hilti::Location(*l.begin.filename, l.begin.line, l.end.line));
 }
 
+static std::string expandEscapes(Driver* driver, std::string s, hilti::detail::parser::location l) {
+    try {
+        return util::expandEscapes(s);
+    } catch ( const hilti::rt::Exception& ) {
+        driver->error("invalid escape sequence", toMeta(l));
+        return "<error>";
+    }
+}
+
 %}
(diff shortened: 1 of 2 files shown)


hilti/src/compiler/parser/scanner.ll
```diff
@@ -33,6 +33,15 @@ static hilti::Meta toMeta(hilti::detail::parser::location l) {
     return hilti::Meta(hilti::Location(*l.begin.filename, l.begin.line, l.end.line));
 }
 
+static std::string expandEscapes(Driver* driver, std::string s, hilti::detail::parser::location l) {
+    try {
```
bbannier, line 37: Is there any reason to not perform more generic exception handling in the scope invoking the scanner/parser (which I think could be in HILTI's `parser/driver`? I am asking since the fix here just addresses #210, but not the more general issue of exceptions escaping from parsing.

Similarly in Spicy's scanner.

rsmmr: Yeah, this way we can continue parsing and potentially catch & report more such errors before aborting. 

But you're right, the parser should have an outer catch on top of this.

PR #255, opened 2020-04-27
Retain column information in parsing.

bbannier:
We previously were not storing column information when parsing source
files which makes it much harder to correlate input to output. This
patch initializes already existing data structures with this information
when available. Locations are emitted as `<line>:<character>` so that
e.g., the second character on the fourth line would be displayed as
`4:2`. This extends to ranges as well which look e.g., like `1:1-2:30`.


zeek/compiler/glue-compiler.cc
```diff
@@ -249,11 +249,10 @@ bool GlueCompiler::loadEvtFile(std::filesystem::path& path) {
     std::vector<glue::Event> new_events;
 
     try {
-        int lineno = 0;
         std::string chunk;
 
         while ( ! in.eof() ) {
-            _locations.emplace_back(path, ++lineno);
+            _locations.emplace_back(path);
```
rsmmr, line 255: Isn't this loosing the line information?

bbannier: Yes, it is. I rolled back this change.

Issue #230, opened 2020-04-21
Documentation missing / unclear / wrong (aggregate ticket #2)


duffy-corelight:
Documentation items to fix are each in a separate comment.

duffy-corelight:
`cast<int>(type<uint>) → uint`   
has its parameter and the type in the cast<> backwards relative to the prose.
Converts the value to an unsigned integer type, accepting any loss of information.  
and  
`cast<uint>(type<int>) → int`   
Converts the value to an signed integer type, accepting any loss of information.  
and  
`cast<real>(type<int>) → int`  
Converts the value to a signed integer type, accepting any loss of information.  
and  
`cast<real>(type<uint>) → uint`
Converts the value to an unsigned integer type, accepting any loss of information.  
also have it backwards.

duffy-corelight:
Returns the number of stream the view contains.
 ->
Returns the number of streams the view contains.

duffy-corelight:
Wrong operator in the , 
`uint & uint → uint`  
Computes the bit-wise ‘or’ of the two integers.

duffy-corelight:
`-uint → uint`
Computes the bit-wise negation of the integer.
is not usually bit-wise. - is usually arithmetic while ~ is bit-wise.

duffy-corelight:
Make change to one phrasing, for the inconsistency between two very similar concepts.
`uint + uint → uint`
Returns the sum of the integers.  
`uint + uint → uint`
Computes the sum of the integers.

Also (for the inconsistency)
`int += int → int`
Increments the first integer by the second, assigning the new value.  
`uint += uint → uint`
Increments the first value by the second.

The first of these is the same operator being documented twice, which is another aspect. Might also want to eliminate the duplication.

rsmmr:
The list type isn't documented, and in particular list comprehension is not.

duffy-corelight:
In 5.1.10.2 Leveraging MIME Types

a typo:
(if that’s multiple, all we connected and all will receive the same data).
 ->
(if that’s multiple, all will be connected and all will receive the same data).

duffy-corelight:
In 5.1.10.2 Leveraging MIME Types, the literal example doesn't have the whole phrase for the keyword:
equivalent of connect_mime("x-port/%d" % resp_port_of_connection).
 ->
equivalent of connect_mime_type("x-port/%d" % resp_port_of_connection).

duffy-corelight:
In 5.1.9 Filters
[...]

PR #256, opened 2020-04-28
fix usage of spicy-build -S

JustinAzoff:


doc/getting-started.rst
```diff
@@ -332,7 +332,7 @@ specific parser. Here's a small C++ program that parses input with our
 
 .. code::
 
-    # spicy-build -o a.out my-http.cc my-http.spicy
+    # spicy-build -o a.out -S my-http.cc my-http.spicy
```
rsmmr, line 335: Did you find you need the `-S` to make this work? It's should actually be as shown: no option to automatically include the spicy-driver C++ skeleton; and with `-S` to *not* include it, which would mean that you can't actually have the resulting `a.out` parse anything. (It used to be opposite in an older version, but that ended up just being confusing because it made the most common use case require an option).

doc/getting-started.rst
```diff
@@ -348,7 +348,7 @@ does, if we ignore the dynamic JIT compilation.
     - ``my-http.cc`` is a very stripped down version of a file
       installed along with Spicy into
       ``share/spicy/spicy-driver-host.cc``. That code is compiled into
-      an executable when executing ``spicy-build`` with the ``-s``
+      an executable when executing ``spicy-build`` with the ``-S``
       option, as we did above. If you look at
```
rsmmr, line 352: This is a leftover from the old version, should now read:


```suggestion
      an executable when executing ``spicy-build``
      as we did above. If you look at
```

doc/getting-started.rst
JustinAzoff: yeah.. I believe I was getting duplicate 'main' functions during build if I didn't use -S.. i'll have to test it again though.

rsmmr: I'm wondering if that's another linker issue, there're a weak main() symbols involved ...

JustinAzoff:
```
justin@ap1k:~/src/spicy/doc/examples$ spicy-build -o a.out my-http.cc my-http.spicy
/nix/store/nyhj00w339gk2gaj3faz70gjrnbmam8v-binutils-2.31.1/bin/ld: /run/user/1000/ccVnoX2q.o: in function `main':
my-http.cc:(.text.startup+0x0): multiple definition of `main'; /run/user/1000/ccMa3gxn.o:spicy-driver-host.cc:(.text.startup+0x0): first defined here
collect2: error: ld returned 1 exit status
```

but with -S it works

```
justin@ap1k:~/src/spicy/doc/examples$ spicy-build -o a.out -S my-http.cc my-http.spicy
justin@ap1k:~/src/spicy/doc/examples$ echo "GET /index.html HTTP/1.0"  | ./a.out
GET, /index.html, 1.0
```

rsmmr: I can't reproduce this. it's pretty certainly an artifact of the linking, and the docs describe how it's supposed to work, so I'll close this.

doc/getting-started.rst
rsmmr: I'll move this over to the doc ticket #230.

PR #258, opened 2020-04-28
Add a section to the development manual on the structure of the runtime libraries

rsmmr:


doc/development/architecture.rst
```diff
@@ -4,4 +4,113 @@
...
+    Private Spicy-side library functionality that the HILTI code
+    coming out of Spicy compilation can ``import`` (e.g., functions to
+    access parsing input, such as ``spicy_rt::waitForInput()``;
+    HILTI-side type definitions for Spicy-specific types, such as for
+    a ``sink``). This functionality is declared in
+    ``spicy/lib/spicy_rt.spicy`` and implemented in C++ in
+    ``libspicy-rt.a``.
+
+Category 4
+    Public library functionality for C++ host applications to
```
bbannier, line 59: Can we change to ordering of these? With that 1-3 would all be public functionality and 4-5 private.

doc/development/architecture.rst
```diff
@@ -4,4 +4,113 @@
...
+        categories.
+
+.. rubric:: What goes where?
+
+Think of Category 1 as the "Spicy standard library": functionality for
+user-side Spicy code to leverage.
+
+Category 2 is the same for HILTI, except that the universe of HILTI
+users remains extremely small right now (it’s just Spicy and people
+writing tests :-).
```
bbannier, line 97: Are we missing the closing parenthesis here?

doc/development/architecture.rst
rsmmr: Ok, I was debating about that, will do.

rsmmr: Done

PR #259, opened 2020-04-28
Fix spurious semicolons in C++ output.

### LINKED ISSUE #178: Bug with if/else code generation
```
module Issue_178;

type whatever = unit {
  data: uint8;
  moreTag : uint8 if (self.data== 31);

  on %error {
    print "error handler data:", self.data;
    if (self.data== 31) {
#      print "moreTag:", self.moreTag;
    }
    else {
#      print "    tag:", self.data;
    }
  }
};
```
reports:
```
2041 /usr/local/etc/tmp/spicy$cat ../InitialDP.pcapng | build/bin/spicy-driver ../Issue_178_empty_if.spicy 
Issue_178.cc:52:134: warning: if statement has empty body
    if ( hilti::rt::optional::value((*__self).data, "../Issue_178_empty_if.spicy:9") == hilti::rt::integer::safe<std::uint8_t>{31u} );
                                                                                                                                     ^
Issue_178.cc:52:134: note: put the semicolon on a separate line to silence this warning
Issue_178.cc:55:5: error: expected expression
    else;
    ^
1 warning and 1 error generated.
[error] jit: failed to execute compilation action.
[error] jit: failed to compile C++ code unit Issue_178 to bitcode
```
But if one does follow that instruction and  puts semicolon on a separate line to silence this warning, i.e.
```
module Issue_178;

type whatever = unit {
  data: uint8;
  moreTag : uint8 if (self.data== 31);

  on %error {
    print "error handler data:", self.data;
    if (self.data== 31) {
        ;
#      print "moreTag:", self.moreTag;
    }
    else {
        ;
#      print "    tag:", self.data;
    }
  }
};
```
it instead errors differently, with:
```
2039 /usr/local/etc/tmp/spicy$cat ../InitialDP.pcapng | build/bin/spicy-driver ../Issue_178_empty_if.spicy 
[error] ../Issue_178_empty_if.spicy:10: syntax error, unexpected ';', expecting CONST or TYPE
[error] spicy-driver: spicy-driver: parse error
```

rsmmr:
Our C++ printer would insert spurious semicolons for empty blocks.

Closes #178.


rsmmr:
This is probably tricky to review, the code for the printer is pretty messy, don't worry too much.

hilti/src/compiler/cxx/elements.cc
```diff
@@ -546,8 +546,14 @@ cxx::Formatter& cxx::operator<<(cxx::Formatter& f, const cxx::Block& x) {
             else {
                 if ( compact_block && ! b )
                     f << s << ';';
```
bbannier, line 548: We need a similar fix here.

rsmmr: Good catch. I'll see if we can actually combine those two blocks.

PR #264, opened 2020-04-28
Reject lists with inconsistent types.

### LINKED ISSUE #196: When iterating over a list constructor, all elements must have same type
```
module Issue_196_negative_for_idx;

for ( indx in [1, -2, 3] )
    print indx;
```
```
2461 /usr/local/etc/tmp/spicy$printf "whatever\n" | build/bin/spicy-driver ../Issue_196_negative_for_idx.spicy
uncaught exception hilti::rt::Overflow: integer overflow
```

rsmmr:
For a list constructor [x_1,...,x_n], we used to take the type of the
first element as the list's element type. Now we check all elements if
their types match; only if so, we use that type. Otherwise we treat
the element type as unknown, which may then lead to validation errors.

We also unify this logic across the list/map/set constructors,
although map/set did already have custom validation logic to catch
type mismatches.

Closes #196.

### DIFF
--- a/hilti/include/ast/builder/type.h
+++ b/hilti/include/ast/builder/type.h
@@ -2,10 +2,23 @@
 
 #pragma once
 
+#include <vector>
+
+#include <hilti/ast/expression.h>
+#include <hilti/ast/type.h>
 #include <hilti/ast/types/id.h>
 
 namespace hilti::builder {
 
 inline Type typeByID(::hilti::ID id, Meta m = Meta()) { return hilti::type::UnresolvedID(std::move(id), std::move(m)); }
+
+/**
+ * Determines a common type for a list of expressions.
+ *
+ * @param e expressions
+ * @return if *e* is non-empty and all expressions have the same type,
+ * returns that type; otherwise returns ``type::Unknown``.
+ */
+Type typeOfExpressions(const std::vector<Expression>& e);
 
 } // namespace hilti::builder
--- a/hilti/include/ast/ctors/map.h
+++ b/hilti/include/ast/ctors/map.h
@@ -2,6 +2,7 @@
 
 #pragma once
 
+#include <hilti/ast/builder/type.h>
 #include <hilti/ast/ctor.h>
 #include <hilti/ast/expression.h>
 #include <hilti/ast/types/map.h>
@@ -50,10 +51,9 @@ public:
 
 private:
     std::vector<Type> _inferTypes(const std::vector<Element>& e, const Meta& /* m */) {
-        if ( e.size() )
-            return {e.front().first.type(), e.front().second.type()};
-
-        return {type::unknown, type::unknown};
+        auto keys = util::transform(e, [](const auto& e) { return e.first; });
+        auto values = util::transform(e, [](const auto& e) { return e.second; });
+        return {builder::typeOfExpressions(keys), builder::typeOfExpressions(values)};
     }
 
     std::vector<Expression> _flatten(const std::vector<Element>& elems) {
--- a/hilti/include/ast/ctors/vector.h
+++ b/hilti/include/ast/ctors/vector.h
@@ -2,6 +2,7 @@
 
 #pragma once
 
+#include <hilti/ast/builder/type.h>
 #include <hilti/ast/ctor.h>
 #include <hilti/ast/expression.h>
 #include <hilti/ast/types/unknown.h>
@@ -15,7 +16,8 @@ namespace ctor {
 /** AST node for a vector constructor. */
 class Vector : public NodeBase, public hilti::trait::isCtor {
 public:
-    Vector(const std::vector<Expression>& e, const Meta& m = Meta()) : NodeBase(nodes(_inferType(e, m), e), m) {}
+    Vector(const std::vector<Expression>& e, const Meta& m = Meta())
+        : NodeBase(nodes(builder::typeOfExpressions(e), e), m) {}
     Vector(Type t, std::vector<Expression> e, Meta m = Meta())
         : NodeBase(nodes(std::move(t), std::move(e)), std::move(m)) {}
 
@@ -38,11 +40,6 @@ public:
     auto isEqual(const Ctor& other) const { return node::isEqual(this, other); }
     /** Implements `Node` interface. */
     auto properties() const { return node::Properties{}; }
-
-private:
-    Type _inferType(const std::vector<Expression>& e, const Meta& /* m */) {
-        return e.size() ? e.front().type() : type::unknown;
-    }
 };
 
 } // namespace ctor
--- a/hilti/src/ast/builder/type.cc
+++ b/hilti/src/ast/builder/type.cc
@@ -1,0 +1,24 @@
+// Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
+
+#include <vector>
+
+#include <hilti/ast/builder/type.h>
+#include <hilti/ast/ctors/list.h>
+#include <hilti/ast/expression.h>
+#include <hilti/ast/types/unknown.h>
+
+using namespace hilti;
+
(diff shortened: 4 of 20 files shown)


hilti/include/ast/ctors/map.h
```diff
@@ -50,10 +50,9 @@ class Map : public NodeBase, public hilti::trait::isCtor {
 
 private:
     std::vector<Type> _inferTypes(const std::vector<Element>& e, const Meta& /* m */) {
-        if ( e.size() )
-            return {e.front().first.type(), e.front().second.type()};
-
-        return {type::unknown, type::unknown};
+        auto keys = util::transform(e, [](const auto& e) { return e.first; });
+        auto values = util::transform(e, [](const auto& e) { return e.second; });
+        return {builder::typeOfExpressions(keys), builder::typeOfExpressions(values)};
```
bbannier, line 55: Let's add an `#include <hilti/ast/builder/type.h>` for `typeOfExpression`.

hilti/src/ast/builder/type.cc
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
+
+#include <hilti/ast/ctors/list.h>
+#include <hilti/ast/expression.h>
+#include <hilti/ast/types/unknown.h>
+
+using namespace hilti;
+
+Type builder::typeOfExpressions(const std::vector<Expression>& e) {
```
bbannier, line 9: I am surprised this actually implements the correct function 😅 . Maybe instead do the following (requires manual adjustment at the end of the namespace).
```suggestion
namespace hilti::builder {
Type typeOfExpressions(const std::vector<Expression>& e) {
```

hilti/src/compiler/visitors/validator.cc
```diff
@@ -139,27 +139,15 @@ struct Visitor : public visitor::PostOrder<void, Visitor> {
...
-                break;
-            }
-
-            if ( v.type() != et ) {
-                error("type mismatch in map values", p);
-                break;
-            }
-        }
+            error("map elements have inconsistent types", p);
     }
```
bbannier, line 151: All this does now for `Map` and also `Set` is to check for unresolved types. Do you think it would make sense to introduce a dedicated validator for that and to remove these cases here? -- if one were to remove e.g., this function one would currently hit a hard error in codegen which seems errorprone,
```
$ btest hilti.types.map.ctor-fail -d
[  0%] hilti.types.map.ctor-fail ... failed
  % 'btest-diff output' failed unexpectedly (exit code 1)
  % cat .diag
  == File ===============================
  [internal-error] <singleton>: codegen: type <unknown type> does not have a visitor

  --- Aborting
  # 0   libhilti.dylib                      0x000000010de18101 _ZN5hilti2rt9BacktraceC2Ev + 81
  # 1   libhilti.dylib                      0x000000010de18765 _ZN5hilti2rt9BacktraceC1Ev + 21
  # 2   libhilti.dylib                      0x000000010df9cdbe _ZN4util20abort_with_backtraceEv + 62
  # 3   libhilti.dylib                      0x000000010df8d116 _ZN5hilti6Logger13internalErrorERKNSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEERKNS_8LocationE + 118
  # 4   libhilti.dylib                      0x000000010e1820bd _ZN5hilti6Logger13internalErrorINS_4type6detail4TypeELPv0EEEvNSt3__112basic_stringIcNS6_11char_traitsIcEENS6_9allocatorIcEEEERKT_ + 125
  # 5   libhilti.dylib                      0x000000010e180e94 _ZN5hilti6detail7CodeGen7compileERKNS_4type6detail4TypeENS0_7codegen9TypeUsageE + 292
  # 6   libhilti.dylib                      0x000000010e0a0d86 _ZN12_GLOBAL__N_17VisitorclERKN5hilti4ctor3MapE + 246
  # 7   libhilti.dylib                      0x000000010e09ca8e _ZN5hilti6detail7visitor15do_dispatch_oneINSt3__112basic_stringIcNS3_11char_traitsIcEENS3_9allocatorIcEEEENS_4ctor3MapENSA_6detail4CtorEN12_GLOBAL__N_17VisitorENS1_8IteratorINS_4NodeELNS1_5Or
derE0ELb0EEEEENS3_11conditionalIXsr3stdE9is_void_vIT_EEbNS3_8optionalISL_EEE4typeERT1_RKSt9type_infoRT2_RNT3_8PositionERb + 206
  ..
```

hilti/src/ast/builder/type.cc
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
+
+#include <hilti/ast/ctors/list.h>
+#include <hilti/ast/expression.h>
+#include <hilti/ast/types/unknown.h>
```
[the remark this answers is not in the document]
rsmmr: (updated code slightly)

hilti/src/ast/builder/type.cc
rsmmr: This pattern is used quite a bit across the code (the *.cc files usually don't open a namespace), so I'll leave as is for consistency. It as some point we wanted to do a bulk change, I wouldn't object.

hilti/src/compiler/visitors/validator.cc
rsmmr: I  don't quite follow what you mean here. It checks for unresolved types, but that includes types inside the containers just being inconsistent, which is the same the previous check did explicitly. Removing the function would of course skip this validation and then run into trouble later.

bbannier: I was hoping that the _unknown type_ property would propagate up into the container `Ctor` so that we could check that, but it seems I misunderstood since the following cannot replace all the checks of the individual type `Ctor`s, 
```
void operator()(const Ctor& c, position_t p) {
    if ( c.type() == type::unknown )
        error("cannot construct value of unspecified type", p);
}
```

rsmmr: In some cases it's actually ok to have a type::unknown coming from a ctor, in particular in case of list because they may be used to intialize other types where the elements can be coerced appropriately, like in `vector<real> x = [3.14, 4];`. The list itself doesn't have a consistent type, but all elements coerce into floats, so it should be fine.

bbannier: I think I understand what you are saying, but constructs like `local auto x = vector<real>(1, 1.1)` should now be rejected.

rsmmr: Well, turns out this patch missed vectors anyways, so I'm adding that. With that in place, the above wasn't rejected, but also didn't work as expected: the element coercion didn't happen. I'm working on a new version of the patch that gets that in place.

hilti/include/ast/ctors/vector.h
```diff
@@ -15,7 +16,8 @@ namespace ctor {
 /** AST node for a vector constructor. */
 class Vector : public NodeBase, public hilti::trait::isCtor {
 public:
-    Vector(const std::vector<Expression>& e, const Meta& m = Meta()) : NodeBase(nodes(_inferType(e, m), e), m) {}
+    Vector(const std::vector<Expression>& e, const Meta& m = Meta())
+        : NodeBase(nodes(builder::typeOfExpressions(e), e), m) {}
```
bbannier, line 20: Let's remove the now unused `_inferType` function below.

hilti/include/ast/builder/type.h
```diff
@@ -2,10 +2,21 @@
 
 #pragma once
```
bbannier, line 4:
```suggestion

#include <vector>

```

PR #283, opened 2020-04-30
Fixes for parsing & validating unit properties

rsmmr:


duffy-corelight:
line 515 has the grammar-o:  host applications may use of the
                                                     ->                       may make use
if you might want to fix that while changing the text anyway.

Issue #271, opened 2020-04-29
vector of bytes does not match a filter method; reports call does not match any method: <TheUppercaser::Foo>.__on_x(<bytes>)


duffy-corelight:
vector of bytes seems incompatible with using a filter
```
module Issue_270;

type Filter = unit {
    %filter
    : bytes &eod &chunked { self.forward($$.upper()); }
};

public type Foo = unit {
    on %init { self.connect_filter(new Filter); }
    x: bytes[] &size=1 { print $$; }
    on %done {
      print "\n";
    }
};
```
reports
```
printf "sDeTr\n" | build/bin/spicy-driver /usr/local/etc/tmp/Issue_270_on_x.spicy
[error] /usr/local/etc/tmp/Issue_270_on_x.spicy:10: call does not match any method: <Issue_270::Foo>.__on_x(<bytes>)
[error] /usr/local/etc/tmp/Issue_270_on_x.spicy:10:   candidate methods:
[error] /usr/local/etc/tmp/Issue_270_on_x.spicy:10:   - <Issue_270::Foo>.__on_x(<const vector<bytes>>)
[error] spicy-driver: aborting after errors
```
When trying instead, removing the &size=1, instead reports
```
spicy$        printf "sDeTr\n" | build/bin/spicy-driver /usr/local/etc/tmp/Issue_270_on_x.spicy
Bus error: 10
```

rsmmr:
This is going to be fixed with #284 (update coming there). However, that Spicy code is probably not going to do what you intended: `x: bytes[] &size=1 ` limits the vector to just one bye total, so you'll get a single-element vector, with that element being one byte.

I'm guessing what you meant here is `x: (bytes &size=1)[];` (i.e., associate the size with the elements). #284 will support that as well.

duffy-corelight:
Where you say that Spicy code x: bytes[] &size=1 limits the vector to just one byte total, is that behavior interacting the way we would like, with respect the distinction between &size and &length? Is &size supposed to impose an absolute limit at which the sudden conclusion of the final being-accumulated element of the vector is dictated?

duffy-corelight:
And do we support (or want to support) &size inside parenthesis and another &size outside after the parenthesis?

rsmmr:
Yes `&size` imposes a hard limit on the field, as if it had reached end of data. It works for all types of fields, just happens to be most common with `bytes` (which without any further attributes just greedily parses everything it can).

[...]

PR #284, opened 2020-04-30
Add support for &until_including when parsing bytes.

### LINKED ISSUE #269: reports &until_including is only valid for fields of type bytes or vector
Using bytes &until_including, it nonetheless reports &until_including is only valid for fields of type bytes or vector 
```
module Issue_269;

type Filter = unit {
    %filter
    : bytes &eod &chunked { self.forward($$.upper()); }
};

public type Foo = unit {
    on %init { self.connect_filter(new Filter); }
#    x: bytes &size=5 { print $$; }
    : bytes &until_including=(b"\n") { print $$; }
};
```
reports
```
printf "aBcD\n" | build/bin/spicy-driver  /usr/local/etc/tmp/Error_4561_only_bytes.spicy
[error] /usr/local/etc/tmp/Error_4561_only_bytes.spicy:11: &until_including is only valid for fields of type bytes or vector
```

rsmmr:
The error message one got when trying to use &until_including with a
field of type "bytes" would claim that "bytes" was indeed supported.
It was lying. To fix the error message, this commit adds support for
that feature:

    # cat test.spicy

    module Test;

    public type X = unit {
        x: bytes &until_including=b"567";
        on %done { print self; }
    };

    # printf '1234567890' | spicy-driver test.spicy
    [$x=b"1234567"]

The commits also cleans up validation for attributes on bytes fields.

Closes #269.


duffy-corelight:
The combination of
`    : bytes &size=2 &parse-at=`
is likely still useful, but the new error clause at line 235 in spicy/src/compiler/visitors/validator.cc will preclude that. When I looked in
spicy/tests/spicy/types/unit/attr-parse-at.spicy 
I notice that there is no test of an attempt at combining.

rsmmr:
Good catch, you're right, I missed that. Will make `&size=2 &parse-at=` supported.

rsmmr:
Actually, &size should always works, I'll pull that out of the validation code here.

spicy/src/compiler/visitors/validator.cc
```diff
@@ -224,14 +224,21 @@ struct PreTransformVisitor : public hilti::visitor::PreOrder<void, PreTransformV
...
+                if ( i )
+                    ++num_attrs;
             }
 
-            else if ( ! until_attr && ! size_attr && ! parse_from_attr && ! parse_at_attr )
-                error("bytes field requires one of &size, &eod, or &until", p);
+            // &size can be combined with any other attribute, or be used standalone.
+
+            if ( num_attrs > 1 )
+                error("incompatible attributes for bytes field", p);
```
bbannier, line 238: I'd imagine it might be hard for users to figure out their error here.

How instead of `num_attrs` we store the present attributes (e.g., in a `vector<Attribute>` or `vector<Attribute*>` if you want to prevent allocations), so we can show them to the user, e.g.,
```
if (attrs.size() > 1)
  error(fmt("attributes cannot be combined: %s", join(attrs, ","));
```

doc/programming/parsing.rst
```diff
@@ -187,6 +187,40 @@ There's also a programmatic way to change a field's type to something
...
+        y: bytes &eod;
+        on %done { print self; }
+    };
+
+.. spicy-output:: size.spicy
+    :exec: printf '12345678X90' | spicy-driver %INPUT
+    :show-with: foo.spicy
+
+As you can see, parsing `x` stops at the size limit, not at the
+delimiter coming only later.
```
bbannier, line 213: Is the logic here to consume the minimum number of bytes to eiter see `&until`'s arg or `&size` bytes? If it is I think it would be worth spelling that out.

I see one can also attach `&size` to e.g., an `uint8` field. If our explanation here could explain what happens in that case as well the even better.

spicy/src/compiler/visitors/validator.cc
rsmmr: I like that, will do.

doc/programming/parsing.rst
rsmmr: Yeah, that's what it means. Thinking about it, though, I think the semantics are still not quite right. The example in the doc should actually trigger a parse error because it's not finding the `X` before the imposed size limit (that's how `until` normal works when encountering EOD). I'll give that another try.

rsmmr:
Turns out this also overlaps with the work for #271, I'll tackle them together.

duffy-corelight:
It is great that x: (bytes &size=1)[]; (i.e., associate the size with the elements), but I think that merits a documentation item too. I would not have had the inclination to think that the language supported &attributes inside the parenthesis, so when 284 will support that, mention it with an illustrative example.

rsmmr:
CI isn't running for some reason.

rsmmr:
@duffy-corelight I'm adding the note on documentation to the doc ticket.

bbannier:
> CI isn't running for some reason.

Just loading the CI page, it looks like Cirrus CI has problems with their GraphQL backend again. Nothing official at this point, though.

tests/hilti/types/struct/attributes.hlt
```diff
@@ -52,4 +52,8 @@ hilti::print(x);
 hilti::print(y);
 hilti::print(z);
 
+hilti::print(z?.s1);
+delete z.s1;
```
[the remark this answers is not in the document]
rsmmr: For background, the `del` operator comes originally from Zeek: `del s[x]` removes an element from a set in Zeek. HILTI & Spicy are using that as well. I now reused that existing operator to implement the reset for optional. And yeah, I can see how that might look scary ... Let me see if I can make `... = Null` work. That was my initial thought as well, but turned out a little tricky. But might not be a real problem.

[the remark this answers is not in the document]
rsmmr: Well, it remains tricky. :-) I'll introduce a new operator with a different name instead.

rsmmr:
Just pushed an update: 

- The first commit is replaced by a new one introducing an 'unset' operator instead of reusing `delete`.

- There's a new commit at the end addressing #289.

No other substantial changes (other than now using "unset" instead "delete" later).

tests/hilti/types/struct/attributes.hlt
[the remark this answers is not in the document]
rsmmr: "unset" it is.

Issue #200, opened 2020-04-16
Add option to Spicy driver to read input from file


duffy-corelight:
Support spicy-driver taking a -f argument, where if present spicy would open that as a file, using that command line parameter as the name of a file, and take the stream to parse from there.

rsmmr:
I suggest we do the following for now:

- Add `-f <path>`. let's not support `@` for now, it's easy to do with a shell loop around it.
- Make a function available to Spicy scripts to read environment variables: `getenv(VAR)`. That way, you can pass in a file name that way from an outer script wrapper; and that's a more generally useful capability, too.

duffy-corelight:
I concur. Making a function available to Spicy scripts to read environment variables: getenv(VAR) is sufficient to satisfy this request to have the passed-in file name that way. And is capable of more functionality, and flexibility for the investment, compared to having just the cmd line string.

PR #294, opened 2020-05-06
Add option -f to spicy-driver; getenv() function for Spicy scripts.

rsmmr:
Per #200.


rsmmr:
@bbannier this looking ok?

spicy/include/rt/util.h
```diff
@@ -9,4 +9,7 @@ namespace spicy::rt {
 /** Returns a string identifying the version of the runtime library. */
 extern std::string version();
 
+/** Returns the value of an environment variable, if set. */
+extern std::optional<std::string> getenv(const std::string& name);
```
bbannier, line 13: Let's include `<optional>` for `std::optional`.

More a question, this function would look like a fine candidate to accept a `std::string_view` instead of a `std::string` ref, do we prefer one over the other?

spicy/src/rt/util.cc
```diff
@@ -18,3 +18,10 @@ std::string spicy::rt::version() {
 #error "Neither HILTI_RT_BUILD_TYPE_DEBUG nor HILTI_RT_BUILD_TYPE_RELEASE define."
 #endif
 }
+
+std::optional<std::string> spicy::rt::getenv(const std::string& name) {
+    if ( auto x = ::getenv(name.c_str()) )
```
bbannier, line 23: Let's include `<cstdlib>` for `::getenv`.

tests/spicy/tools/spicy-driver-file.spicy
```diff
@@ -0,0 +1,17 @@
...
+# @TEST-EXEC: btest-diff output
+
+module Test;
+
+public type Foo = unit {
+    x: bytes &eod;
+    on %done { print self; }
+};
+
+@TEST-START-FILE input.dat
```
bbannier, line 13:
```suggestion
# @TEST-START-FILE input.dat
```

tests/spicy/tools/spicy-driver-file.spicy
```diff
@@ -0,0 +1,17 @@
...
+1
+2
+3
+@TEST-END-FILE
```
bbannier, line 17:
```suggestion
# @TEST-END-FILE
```

spicy/include/rt/util.h
rsmmr: `string_view` doesn't have `c_str()`, so I don't think it would buy us anything here.

More generally, I haven't become friends with `string_view` yet :-). I find it hard to use consistently as too often conversion from/to `std::string and` `const char*` runs into trouble.

tests/spicy/tools/spicy-driver-file.spicy
rsmmr: btest will take out the line, so won't make a difference. Just looked at existing tests, and we are inconsistent in whether we're using  '#" with `@TEST-START-FILE` or not ...

bbannier: Putting it in a comment would make it easier to distinguish non-Spicy meta data from actual code e.g., for syntax highlighting. But OTOH, this file isn't valid Spicy code anyway due to `@TEST-START-FILE` etc.

rsmmr: yeah, right. I'll put it in, and we should unify it sometime

PR #295, opened 2020-05-06
Make iterators over rt::Vector safe

bbannier:


hilti/src/rt/tests/to_string.cc
```diff
@@ -32,9 +33,21 @@ TEST_CASE("string") { CHECK_EQ(to_string(std::string("abc")), "\"abc\""); }
 
 TEST_CASE("vector") {
     CHECK_EQ(to_string(std::vector<int8_t>()), "[]");
-    CHECK_EQ(to_string(std::vector{int8_t{1}}), "[1]");
-    CHECK_EQ(to_string(std::vector{int8_t{1}, 2}), "[1, 2]");
-    CHECK_EQ(to_string(std::vector{std::vector{int8_t{1}, 2}}), "[1, 2]");
+    CHECK_EQ(to_string(std::vector<int8_t>({1})), "[1]");
+    CHECK_EQ(to_string(std::vector<int8_t>({1, 2})), "[1, 2]");
+    CHECK_EQ(to_string(std::vector<std::vector<int8_t>>({{1, 2}})), "[[1, 2]]");
```
rsmmr, line 38: I'm wondering if/why we still need such support for "normal" vectors. They are probably still produced somewhere, but could we aim for using `Vector` everywhere?

hilti/include/rt/types/vector.h
```diff
@@ -195,46 +349,45 @@ namespace detail::adl {
...
 }
 
 template<typename T, typename Allocator>
-inline std::string to_string(const vector::SafeConstIterator<T, Allocator>& /*unused*/, adl::tag /*unused*/) {
+inline std::string to_string(const vector::ConstIterator<T, Allocator>& /*unused*/, adl::tag /*unused*/) {
     return "<const vector iterator>";
 }
 
 template<typename T, typename Allocator>
 inline auto safe_begin(const Vector<T, Allocator>& x, adl::tag /*unused*/) {
```
rsmmr, line 374: Do we need the `safe_*()` versions still just because the code gen still generates them>

hilti/include/rt/types/vector.h
```diff
@@ -251,15 +404,56 @@ inline std::ostream& operator<<(std::ostream& out, const vector::Empty& x) {
...
+bool operator!=(const Vector<T, Allocator>& a, const Vector<T, Allocator>& b) {
+    return ! (a == b);
 }
 
 template<typename T, typename Allocator>
-inline std::ostream& operator<<(std::ostream& out, const vector::SafeConstIterator<T, Allocator>& x) {
-    out << to_string(x);
-    return out;
+typename vector::Iterator<T, Allocator>::reference vector::Iterator<T, Allocator>::operator*() {
+    if ( auto&& c = _container(); c ) {
```
rsmmr, line 413: Nit: why the `; c` )?

hilti/include/rt/types/vector.h
```diff
@@ -251,15 +404,56 @@ inline std::ostream& operator<<(std::ostream& out, const vector::Empty& x) {
 }
 
 template<typename T, typename Allocator>
-inline std::ostream& operator<<(std::ostream& out, const vector::SafeIterator<T, Allocator>& x) {
```
rsmmr, line 254: `fmt()` uses the `operator<<` for string formatting, so unless I'm missing something, we should keep them.

hilti/include/rt/types/vector.h
```diff
@@ -75,98 +76,251 @@ bool operator!=(Allocator<T, D1> const&, Allocator<U, D2> const&) noexcept {
...
+
+    using reference = T&;
+
 public:
-    using Base =
-        hilti::rt::detail::iterator::SafeIterator<Vector<T, Allocator>, typename Vector<T, Allocator>::SafeIterator,
-                                                  SafeIterator<T, Allocator>>;
-    using Base::Base;
+    Iterator() = default;
+    Iterator(typename V::size_type&& index, const typename V::C& control)
```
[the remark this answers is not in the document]
bbannier: I personally think that if a class does not declare any copy ctrs or assignment operators, and one uses >=C++17, not explicitly defaulting is fine. We needed a custom constructor here to pass in index and the control block from the `Vector`; this does influence whether e.g., copy ctrs are implicitly generated.

I think that would be in line with the [rule of three/five/zero](https://en.cppreference.com/w/cpp/language/rule_of_three) (we are in the _zero_ case here).

hilti/include/rt/types/vector.h
```diff
@@ -75,98 +76,251 @@ bool operator!=(Allocator<T, D1> const&, Allocator<U, D2> const&) noexcept {
...
+     *
+     * @param i position of the element to return
+     * @return a reference to the element
+     * @throw `IndexError` if the `Vector` cannot be resized
+     */
     T& operator[](uint64_t i) & {
-        if ( i >= V::size() )
-            V::resize(i + 1);
+        if ( i >= V::size() ) {
+            if constexpr ( std::is_default_constructible_v<T> )
```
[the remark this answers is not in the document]
bbannier: Yes, I saw issues with `rt::MIMEType` which is not default-constructible. The code here generalizes that, but maybe we don't need that if we fix that class (though it doesn't hurt I guess). As of now one e.g., shouldn't be able to use a `rt::MIMEType` as a `global`.

hilti/include/rt/types/vector.h
bbannier: Not for `rt::Vector`, but I'd say let's leave them in for now until we have all containers cleaned up. After that we can remove it, and do not need to introduce workarounds for intrinsically safe containers.

hilti/include/rt/types/vector.h
bbannier: Urg, that's horrible. Looks like brain damage from looking at Golang code. I'll fix it.

hilti/include/rt/types/vector.h
bbannier: I missed `fmt`, added them.

hilti/src/rt/tests/to_string.cc
```diff
@@ -32,9 +33,21 @@ TEST_CASE("string") { CHECK_EQ(to_string(std::string("abc")), "\"abc\""); }
...
+
+TEST_CASE("Vector") {
+    CHECK_EQ(to_string(vector::Empty()), "[]");
+
+    CHECK_EQ(to_string(Vector<int8_t>()), "[]");
+    CHECK_EQ(to_string(Vector<int8_t>({1})), "[1]");
+    CHECK_EQ(to_string(Vector<int8_t>({1, 2})), "[1, 2]");
+    CHECK_EQ(to_string(Vector<Vector<int8_t>>({{1, 2}})), "[[1, 2]]");
+
+    CHECK_EQ(to_string(Vector<Vector<int8_t>>({{1, 2}}).begin()), "<vector iterator>");
```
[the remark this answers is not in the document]
bbannier: Yeah, thought about that as well. I'd say let's not expose that for now until we also want to commit to providing a getter for the index.

hilti/include/rt/types/vector.h
```diff
@@ -75,98 +76,251 @@ bool operator!=(Allocator<T, D1> const&, Allocator<U, D2> const&) noexcept {
...
+        return *this;
+    }
+
+    Iterator operator++(int) {
+        auto ret = *this;
+        ++(*this);
+        return ret;
+    }
+
+    friend bool operator==(const Iterator& a, const Iterator& b) { return a._index == b._index; }
```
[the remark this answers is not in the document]
bbannier: Was wondering about this as well. Comparing iterators into different `std::vector` is strictly illegal, so it wouldn't implement that. Our situation is slightly different here as our iterators behave like indices (e.g., they remain valid beyond the lifetime of the iterated data). With that _not comparing `_control`_ is a valid and usable implementation (we'd compare indices in that case); comparing `_control` would just add as a sanity check that the user did only compare iterators into the same `Vector` though we wouldn't need that check for internal safety. WDYT?

hilti/src/rt/tests/to_string.cc
bbannier: The only use case I could think of would be for debugging output in internal code involving `std::vector`. I'll remove `to_string` for `std::vector` for now; if a need arises, we could add an implementation back, but somewhere outside of `hilti::rt`.

hilti/include/rt/types/vector.h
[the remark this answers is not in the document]
rsmmr: yeah, I think that sanity check would be good. In my mind an iterator is tied to a specific container, and it seems more intuitive to me to use that as the semantics for comparison as well so that two iterators pointing to different containers will never be considered equal. 

In fact, I can see going further and throw an exception if somebody even tries to compare iterators pointing to different containers because most likely it's evidence of a logic error somewhere. Let's do that I would say.

hilti/include/rt/types/vector.h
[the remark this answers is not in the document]
rsmmr: Ack, we should fix MIMEYype, can just enter some null state that never matches any actual MIME type (`unknown/unknown` maybe)

hilti/include/rt/types/vector.h
rsmmr: Ack.

hilti/include/rt/types/vector.h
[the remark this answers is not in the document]
bbannier: Done.

hilti/include/rt/types/vector.h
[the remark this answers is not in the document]
bbannier: I added validation that the iterators point to the same vector. I started writing our own control block, but it turns out I needed to allocate additional heap data at least once, so let's stay with `shared_ptr<Vector>` as control block for now. With that iterator comparisons will cause two read/write operation on the atomic use count inside the shared_ptr, but this should probably be fine. We can replace this implementation with a custom non-atomic shared_ptr implementation should this ever become an issue.

hilti/include/rt/types/vector.h
```diff
@@ -75,98 +76,251 @@ bool operator!=(Allocator<T, D1> const&, Allocator<U, D2> const&) noexcept {
...
-    Vector(Vector&&) noexcept = default;
+
+    // Constructing from other `Vector` updates the data, but keeps the control block alive.
+    Vector(const Vector& other) : V(other) {}
+    Vector(Vector&& other) noexcept : V(std::move(other)) {}
+
+    Vector(std::initializer_list<T> init, const Allocator& alloc = Allocator()) : V(std::move(init), alloc) {}
+
     Vector(const std::list<T>& l) : std::vector<T>(l.begin(), l.end()) {}
     Vector(std::list<T>&& l) : std::vector<T>(std::move_iterator(l.begin()), std::move_iterator(l.end())) {}
```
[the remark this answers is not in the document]
rsmmr: Ah, ok, that's a valid use case (constructing one HILTI type from another).

hilti/include/rt/types/vector.h
```diff
@@ -75,98 +79,256 @@ bool operator!=(Allocator<T, D1> const&, Allocator<U, D2> const&) noexcept {
...
-    : public hilti::rt::detail::iterator::SafeIterator<
-          Vector<T, Allocator>, typename Vector<T, Allocator>::SafeIterator, SafeIterator<T, Allocator>> {
+template<typename T, typename Allocator>
+class Iterator {
+    using V = Vector<T, Allocator>;
+
+    std::weak_ptr<Vector<T, Allocator>*> _control;
+    typename V::size_type _index = 0;
+
+    using reference = T&;
```
bbannier, line 89: Let's move these definitions into the container where they belong. Ditto for `ConstIterator::reference`.

PR #297, opened 2020-05-07
Check Doxygen comments at compile time

### LINKED ISSUE #240: Consider checking Doxygen comments at compile time
Even though we currently do not render Doxygen output, we still try to document some interfaces with it. We should consider validating that these comments are correct, e.g., by adding ` -Wdocumentation` to the compiler flags if compiling with Clang.

bbannier:
Closes #240.


rsmmr:
I'm going to close this PR for now, we'll keep the branch around.

PR #298, opened 2020-05-07
Document that even with HLTO analyzers Zeek still needs to be passed …

bbannier:
…interface defintions.


bbannier:
The documented way should actually work, closing.

PR #302, opened 2020-05-11
Make iterators of rt::Set safe

bbannier:


hilti/include/rt/types/set.h
```diff
@@ -25,33 +25,82 @@ namespace hilti::rt {
...
 
-template<typename T>
-class SafeConstIterator : public hilti::rt::detail::iterator::SafeIterator<const Set<T>, typename Set<T>::ConstIterator,
-                                                                           SafeConstIterator<T>> {
 public:
-    using Base =
-        hilti::rt::detail::iterator::SafeIterator<const Set<T>, typename Set<T>::ConstIterator, SafeConstIterator<T>>;
-    using Base::Base;
+    Iterator() = default;
+    Iterator(typename S::V::iterator iterator, const typename S::C& control)
```
rsmmr, line 37: Seems like an internal API, could we make this protected or private, with `S` being a friend?

hilti/include/rt/types/set.h
```diff
@@ -25,33 +25,82 @@ namespace hilti::rt {
...
+
+    friend bool operator!=(const Iterator& a, const Iterator& b) { return ! (a == b); }
 };
 
 } // namespace set
 
-/** HILTI's `Set` is an extended version `std::set`. */
+/** HILTI's `Set` is a `std::set`-like type with additional safety guarantees.
+ *
+ * In particular it guarantees that iterators are either valid, or throw an
```
rsmmr, line 81:
```suggestion
 * In particular it guarantees that iterators are either valid, or throws an
```

hilti/include/rt/types/set.h
```diff
@@ -63,36 +112,75 @@ class Set : protected std::set<T>, public hilti::rt::detail::iterator::Controlle
...
 
-    /** Returns true if a specific element is part of the set. */
-    bool contains(const T& t) { return this->find(t) != this->end(); }
-
-    // Methods of `std::set`.
-    using V::begin;
-    using V::cbegin;
-    using V::cend;
-    using V::end;
+    /** Checsk whether an element is in the set.
```
rsmmr, line 117:
```suggestion
    /** Checks whether an element is in the set.
```

hilti/include/rt/types/set.h
```diff
@@ -63,36 +112,75 @@ class Set : protected std::set<T>, public hilti::rt::detail::iterator::Controlle
...
+     *
+     * @param `t` the element to check for
+     * @return `true` if the element is part of the set. */
```
rsmmr, line 120:
```suggestion
     * @return `true` if the element is part of the set.
     */
```

hilti/include/rt/types/set.h
```diff
@@ -25,33 +25,82 @@ namespace hilti::rt {
...
+    Iterator() = default;
+    Iterator(typename S::V::iterator iterator, const typename S::C& control)
+        : _control(control), _iterator(std::move(iterator)) {}
+
+    typename S::reference operator*() const {
+        if ( auto&& l = _control.lock() ) {
+            // Iterators to `end` cannot be dereferenced.
+            if ( _iterator == static_cast<const std::set<T>&>(**l).end() ) {
+                throw IndexError("iterator is invalid");
+            }
```
rsmmr, line 45: Style nit: one-line blocks don't need braches

```suggestion
            if ( _iterator == static_cast<const std::set<T>&>(**l).end() )
                throw IndexError("iterator is invalid");
```

hilti/include/rt/types/set.h
```diff
@@ -63,36 +112,75 @@ class Set : protected std::set<T>, public hilti::rt::detail::iterator::Controlle
...
-    using V::cbegin;
-    using V::cend;
-    using V::end;
+    /** Checsk whether an element is in the set.
+     *
+     * @param `t` the element to check for
+     * @return `true` if the element is part of the set. */
+    bool contains(const T& t) { return this->count(t); }
+
+    auto begin() const { return iterator(static_cast<const V&>(*this).begin(), empty() ? nullptr : _control); }
```
rsmmr, line 123: Not sure I fully follow here why we're not setting the control block for empty sets' begin/end. I assume it's so that dereferencing them will fail? However, isn't this changing the semantics in the sense that these iterators won't be tied to the specific set anymore? So would this, e.g., succeed for two empty sets: `empty1.begin() == empty2.begin()`? (not sure how std::iterators compare when pointing to different sets).

hilti/include/rt/types/set.h
```diff
@@ -63,36 +112,75 @@ class Set : protected std::set<T>, public hilti::rt::detail::iterator::Controlle
...
+
+    auto end() const { return iterator(static_cast<const V&>(*this).end(), empty() ? nullptr : _control); }
+
+    /** Removes an element from the set.
+     *
+     * This function invalidates all iterators into the set.
```
rsmmr, line 129: That's a comment worth reiterating more generally in the type's doc header: All modifying operations invalidate outstanding iterators.

hilti/include/rt/types/set.h
bbannier: Hmm, this needs to match _iterators_ so it should be plural.

rsmmr:
Ah, right, I misread.

-- 
Robin Sommer * Corelight, Inc. * robin@corelight.com * www.corelight.com

hilti/include/rt/types/set.h
bbannier: Yes, this is done to make deref of `begin()` of an empty set throw. As for the comparison, with this implementation we cannot see that the iterators point to different sets, and defer to the comparison of the underlying stdlib iterators. These _should_ compare not equal for different sets, but I think this is implementation-defined. As an alternative, one could perform a `size` check to make sure the set is not empty when dereferencing an iterator, or introduce some sentinel iterator type for empty sets, but I feel this implementation here is good enough. Let's definitely add a test.

hilti/include/rt/types/set.h
bbannier: I added something along those lines to the type doc. I also fixed the signature of `contains` which is not a mutating operation.

PR #304, opened 2020-05-12
Add an end-to-end TFTP analyzer for Zeek.

rsmmr:
The Spicy side was easy, the Zeek-side logging took significantly
longer. :)


spicy/lib/protocols/tftp.spicy
```diff
@@ -0,0 +1,91 @@
+#
+# Trivial File Transfer Protocol
+#
+# Specs from https://tools.ietf.org/html/rfc1350
+#
+
+module TFTP;
+
+import spicy;
```
bbannier, line 9: Is this needed?

spicy/lib/protocols/tftp.spicy
```diff
@@ -0,0 +1,91 @@
...
+
+# Common header for all messages:
+#
+#      2 bytes
+# ---------------
+# |  TFTP Opcode  |
+#  ---------------
+
+public type Message = unit {
```
bbannier, line 18: Let's add a comment here why this type is `public` even though we use other types from this module as well.

spicy/lib/protocols/tftp.spicy
rsmmr: It's not, removing.

PR #305, opened 2020-05-12
Make iterators of rt::Map safe

bbannier:


hilti/include/rt/types/map.h
```diff
@@ -87,9 +87,8 @@ class Map : protected std::map<K, V>, public hilti::rt::detail::iterator::Contro
 
         throw IndexError("key does not exist");
     }
-
-    const V& operator[](const K& k) const { return get(k); }
-    auto operator[](const K& k);
+    const V& operator[](const K& k) const { return static_cast<const M&>(*this)[k]; }
```
rsmmr: This is leading to unusual semantics now: reading from an non-existing index inserts the element and returns a default constructed values. That's pretty uncommon behavior for scripting languages. However, I don't think there's much we can do about it without having the AssignProxy, and I still favor getting rid of that. But I think we need to document this, should go into the Index operator docs for HILTI's map (and we probably want to expand docs for vector/set there, too; hadn't thought above that earlier).

hilti/include/ast/operators/map.h
```diff
@@ -69,13 +69,14 @@ STANDARD_OPERATOR_2x_lhs(map, IndexNonConst, Index, operator_::elementType(0), t
...
+                                 return type::Optional(operator_::elementType(0)(e1, e2).value_or(type::unknown));
+                             },
                          .id = "get",
-                         .args = {{.id = "key", .type = type::Any()},
-                                  {.id = "default", .type = type::Any(), .optional = true}},
+                         .args = {{.id = "key", .type = type::Any()}},
                          .doc = R"(
-Returns the map's element for the given key. If the key does not exist, returns
-the default value if provided; otherwise throws a runtime error.
+Returns the map's element for the given key.
```
rsmmr, line 79: If we keep the change, expand a bit on behavior with non-existing elements to make it explicit.

hilti/include/ast/operators/map.h
```diff
@@ -69,13 +69,14 @@ STANDARD_OPERATOR_2x_lhs(map, IndexNonConst, Index, operator_::elementType(0), t
 BEGIN_METHOD(map, Get)
```
rsmmr, line 69: The change on the C++ side makes sense. However, I'm wondering if we could keep the signature & semantics for the HILTI-side `get()` method the same as they were and handle the default argument during code generation. 
As an example, ``m.get("key", "default")`` could produce some version of `(m.has("key") ? m["key"] : "default")`.

hilti/include/rt/types/map.h
```diff
@@ -31,76 +31,240 @@ class Map;
...
             return i->second;
 
         return std::nullopt;
     }
 
-    const V& operator[](const K& k) const { return static_cast<const M&>(*this)[k]; }
-    V& operator[](const K& k) { return static_cast<M&>(*this)[k]; }
+    /** Access an element by key
+     *
+     * This function invalidates all iterators into the map.
```
rsmmr, line 210: Hmm ... I'm realizing this is actually not great in terms of semantics. First, we can't guarantee that the map will actually be written to, right? So iterators could be invalidated even if no change happens. And second, for an existing key `x`, when just updating its value (e.g., `m["x"] = 42`), it would be pretty surprising to have all iterators be invalidated. Seems that's pretty coming thing to do during a loop over a map for example.

hilti/include/rt/types/map.h
```diff
@@ -27,107 +31,240 @@ class Map;
...
+    /** Removes an element from the map.
+     *
+     * This function invalidates all iterators into the map.
+     *
+     * @param key key of the element to remove
+     * @return 1 if the element was in the set, 0 otherwise
+     */
+    auto erase(const key_type& key) {
+        // Update control block to invalidate all iterators previously created from it.
+        _control = std::make_shared<Map<K, V>*>();
```
bbannier, line 254: Let's only invalidate iff we actually remove an element.

hilti/include/rt/types/map.h
bbannier: Let's make this return an `AssignProxy` which should only invalidate iterators if it actually (lazyily) inserts a new element.

hilti/include/rt/types/map.h
bbannier: I brought back a slightly rewritten `AssignProxy`. PTAL.

hilti/include/rt/types/map.h
bbannier: A slightly rewritten `AssignProxy` is back now.

hilti/src/compiler/codegen/operators.cc
```diff
@@ -307,15 +307,13 @@ struct Visitor : hilti::visitor::PreOrder<std::string, Visitor> {
...
         ;
 
-        if ( auto default_ = optionalArgument(args, 1); ! default_.empty() ) {
-            x += ", ";
-            x += default_;
-        }
-
-        return fmt("%s.get(%s)", self, x);
+        if ( auto default_ = optionalArgument(args, 1); ! default_.empty() )
+            return fmt("(%s.contains(%s)? %s.get(%s) : %s)", self, k, self, k, default_);
```
rsmmr, line 314: `k` is evaluated twice here now, which can be problematic in cases where it has a side effect (it can end up being any kind of C++ expression). Not sure what's the best way around that, we can't use a tmp local either because we're inside an expression. It might need a helper function of some kind.

hilti/include/rt/types/map.h
```diff
@@ -141,6 +141,51 @@ class ConstIterator {
         : _control(control), _iterator(std::move(iterator)) {}
 };
 
+namespace detail {
+
+/** Proxy class for performing save assignments to `Map` entries.
+ *
+ * @note All method accessing the underlying map are only defined for r-values
+ * since this class only holds a reference to it. That makes it in general
+ * unsafe to use these methods when the instance was bound to a later expired
```
rsmmr, line 150: That leads to the question: are they situations where generated code could end up using this in an unsafe way? 

Should we disable copy/move ctor and self-assignment operators to enforce that one can't do much with an AssignProxy?

Alternative would be using control blocks here too.

hilti/include/rt/types/map.h
```diff
@@ -141,6 +141,51 @@ class ConstIterator {
         : _control(control), _iterator(std::move(iterator)) {}
 };
 
+namespace detail {
+
+/** Proxy class for performing save assignments to `Map` entries.
+ *
+ * @note All method accessing the underlying map are only defined for r-values
```
rsmmr, line 148:
```suggestion
 * @note All methods accessing the underlying map are only defined for r-values
```

hilti/include/rt/types/map.h
bbannier: I think the code we generate should be safe. I think there is no way in C++ to disallow users to bind a name to instances of this class. The assignment operators should not be generated here as the class contains a reference member which cannot be reseated.

Re:control block, that was my first attempt, but I ran into issues in constructs like `map["entry"] = "foo"` which do not involve any sequence points between construction and destruction of the proxy which lead to control blocks not being destructed as I naively expected.

hilti/src/compiler/codegen/operators.cc
bbannier: Good catch. I will introduce a lambda here which binds `k` once.

PR #308, opened 2020-05-13
More doc updates

rsmmr:
A couple of independent commits.


doc/scripts/autogen-docs
```diff
@@ -10,19 +10,32 @@ ROOTDIR="$(cd "$( dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)/../.."
 BUILDDIR="${ROOTDIR}/build"
 SPICYDOC="${BUILDDIR}/bin/spicy-doc"
 SPICYDTR="${ROOTDIR}/doc/scripts/spicy-doc-to-rst"
-AUTOGEN="${ROOTDIR}/doc/autogen"
+AUTOGEN_FINAL="${ROOTDIR}/doc/autogen"
+AUTOGEN_STAGE="${ROOTDIR}/doc/autogen.$$.tmp"
```
bbannier, line 14: Can we use `mktemp` here? This would remove most errors around the directory existing, and also give us a robust way to put the directory into the system temp dir.

    mktemp -d -t spicy-autogen-docs

duffy-corelight:
```
module Issue_308_DivisionAssign;

Foo();

function Foo() {
    local wantShifted: int32 = 234;
      wantShifted /= 10;
      print wantShifted;
      wantShifted /= 10;
#      wantShifted // 10;
# if uncommented reports a quite directive error

#      wantShifted //= 10;
# if uncommented reports a reasonable error
      print wantShifted;
}

public type FooUnit = unit {
    var wantShifted: int32 = 567;
    on %init {
      self.wantShifted /= 10;
      print self.wantShifted;
    }
    on %done {
#      self.wantShifted // 10;
# if uncommented here, reports as DIVIDEASSIGN error

      self.wantShifted /= 10;
#      self.wantShifted //= 10;
# if uncommented reports a reasonable error

#      self.wantShifted // 10;
# if uncommented here, reports a pretty obscure error
      print self.wantShifted;
    }
};
```
in case handy to experiment with the // interaction

PR #313, opened 2020-05-14
Add `strftime` to `hilti::rt`.

### LINKED ISSUE #237: Add library function spicy::strftime
... so that people can customize the formatting of timestamps.

bbannier:
This function is exposed in Spicy as `strftime`.

Closes #237.

### DIFF
--- a/hilti/src/rt/util.cc
+++ b/hilti/src/rt/util.cc
@@ -3,6 +3,9 @@
 #include <sys/resource.h>
 #include <unistd.h>
 
+#include <cerrno>
+#include <cstring>
+
 #include <hilti/3rdparty/utf8proc/utf8proc.h>
 
 #include <hilti/rt/autogen/version.h>
@@ -12,6 +15,7 @@
 #include <hilti/rt/backtrace.h>
 #include <hilti/rt/exception.h>
 #include <hilti/rt/fiber.h>
+#include <hilti/rt/fmt.h>
 #include <hilti/rt/util.h>
 
 std::string hilti::rt::version() {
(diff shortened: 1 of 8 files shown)


hilti/src/rt/util.cc
```diff
@@ -309,3 +313,24 @@ hilti::rt::ByteOrder hilti::rt::systemByteOrder() {
...
+std::string hilti::rt::strftime(const std::string& format, const hilti::rt::Time& time) {
+    auto seconds = static_cast<time_t>(time.seconds());
+
+    std::tm tm;
+
+    constexpr size_t size = 128;
+    char mbstr[size];
+
+    auto localtime = ::localtime_r(&seconds, &tm);
+    if ( ! localtime ) {
```
rsmmr, line 326: For style consistency: No need for braces around single-statement block.

spicy/lib/spicy.spicy
```diff
@@ -80,4 +80,7 @@ public function current_time() : time &cxxname="::hilti::rt::time::current_time"
 public function bytes_to_hexstring(value: bytes) : string &cxxname="::spicy::rt::bytes_to_hexstring";
 
 ## Returns the value of an environment variable, if set.
-public function getenv(name: string) : optional<string> &cxxname="spicy::rt::getenv";
+public function getenv(name: string) : optional<string> &cxxname="::spicy::rt::getenv";
+
+## Formats a timestamp.
```
rsmmr, line 85: This is the text that will make it into the documentation, would be good to expand a bit. I realize the other functions in this file don't have much info either, but this one would benefit a least from saying how a format string looks like (similar to the C++ API docs: POSIX style, with link to a page with reference of all format options).

bbannier: I expanded the comment here, but it does not look like `autogen-spicy-lib` pics up multiline docstrings. I filed #315 for that.

PR #314, opened 2020-05-14
Fix comment parsing in EVT files.

### LINKED ISSUE #312: Comments in EVT files aren't parsed correctly
This is not finding the `;` terminating the block.

```
protocol analyzer spicy::TFTP over UDP:
    parse with TFTP::Message,   # entry point for parsing TFTP payload
    port 69/udp;                # use analyzer for sessions on UDP port 69
```

rsmmr:
The fix is quite involved because the parser is really just a hack
right now, and we need to add state machine to track whether we are
inside a comment. We need to switch to a flex/bison parser here
eventually.

We use new comments inside tftp.evt as our test, that used to be not
working.

Closes #312.

### DIFF
--- a/zeek/compiler/glue-compiler.cc
+++ b/zeek/compiler/glue-compiler.cc
@@ -235,6 +235,72 @@ GlueCompiler::GlueCompiler(Driver* driver) : _driver(driver) {}
 
 GlueCompiler::~GlueCompiler() {}
 
+hilti::Result<std::string> GlueCompiler::getNextEvtBlock(std::istream& in, int* lineno) const {
+    std::string chunk;
+
+    // Parser need to track whether we are inside a string or a comment.
+    enum State { Default, InComment, InString } state = Default;
+    char prev = '\0';
+
+    while ( true ) {
+        char cur;
+        if ( in.get(cur) ) {
+            chunk = util::trim(std::move(chunk));
+            if ( chunk.empty() )
+                // Legitimate end of data.
+                return std::string();
+            else
+                // End of input before semicolon.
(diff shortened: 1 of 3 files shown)


zeek/compiler/glue-compiler.cc
```diff
@@ -235,6 +235,73 @@ GlueCompiler::GlueCompiler(Driver* driver) : _driver(driver) {}
...
+    std::string chunk;
+
+    // Parser need to track whether we are inside a string or a comment.
+    enum State { Default, InComment, InString } state = Default;
+    char prev = '\0';
+
+    while ( true ) {
+        char cur;
+        in.get(cur);
+        if ( in.eof() ) {
```
bbannier, line 248: nit: can write this shorter

```suggestion
        if ( in.get(cur) ) {
```

zeek/compiler/glue-compiler.cc
```diff
@@ -235,6 +235,73 @@ GlueCompiler::GlueCompiler(Driver* driver) : _driver(driver) {}
 
 GlueCompiler::~GlueCompiler() {}
 
+hilti::Result<std::string> GlueCompiler::getNextEvtBlock(std::istream& in, int* lineno) {
```
bbannier, line 238: Can be `const` (also in decl).
```suggestion
hilti::Result<std::string> GlueCompiler::getNextEvtBlock(std::istream& in, int* lineno) const {
```

zeek/compiler/glue-compiler.cc
rsmmr: Turns out this doesn't work, reverted.

PR #317, opened 2020-05-18
Make the Zeek plugin build with JIT, and make spicyz standalone

rsmmr:


hilti/src/compiler/driver.cc
```diff
@@ -730,11 +744,16 @@ Result<Nothing> Driver::outputUnits() {
     for ( auto& unit : _hlts ) {
         if ( auto cxx = unit.cxxCode() ) {
             if ( _driver_options.output_cxx ) {
-                auto output = openOutput(output_path, false);
+                auto cxx_path = output_path;
+
+                if ( _driver_options.output_cxx_prefix.size() )
```
bbannier, line 749: nit: prefer `empty` to check for emptiness.
```suggestion
                if ( ! _driver_options.output_cxx_prefix.empty() )
```

Do we need a similar check (or assert) of `cxx->id()` for non-emptiness? It could in principle be empty as well which could lead to collisions.

zeek/bin/spicyz.cc
```diff
@@ -0,0 +1,226 @@
...
+            case 'X': {
+                auto arg = std::string(optarg);
+
+                if ( arg == "help" ) {
+                    std::cerr << "Additional debug instrumentation:\n";
+                    std::cerr << "   flow:     log function calls to debug stream \"hilti-flow\"\n";
+                    std::cerr << "   location: track current source code location for error reporting\n";
+                    std::cerr << "   trace:    log statements to debug stream \"hilti-trace\"\n";
+                    std::cerr << "\n";
+                    exit(0);
```
bbannier, line 95: This is inconsistent with how we e.g., handle `'v'` or `'h'`. I'd say prefer not `exit`'ing from here, but on the top-level.
```suggestion
                    return Nothing();
```

zeek/bin/spicyz.cc
```diff
@@ -0,0 +1,226 @@
...
+                auto arg = std::string(optarg);
+
+                if ( arg == "help" ) {
+                    std::cerr << "Debug streams:\n";
+
+                    for ( const auto& s : hilti::logging::DebugStream::all() )
+                        std::cerr << "  " << s << "\n";
+
+                    std::cerr << "\n";
+                    exit(0);
```
bbannier, line 116:
```suggestion
                    return Nothing();
```

zeek/bin/spicyz.cc
```diff
@@ -0,0 +1,226 @@
...
+}
+
+using hilti::Nothing;
+
+static hilti::Result<Nothing> parseOptions(int argc, char** argv, hilti::driver::Options* driver_options,
+                                           hilti::Options* compiler_options) {
+    while ( true ) {
+        int c = getopt_long(argc, argv, "ABc:CdX:D:L:o:ORTvh", long_driver_options, nullptr);
+
+        if ( c < 0 )
```
bbannier, line 63: Let's be more explicit.
```suggestion
        if ( c == -1 )
```

zeek/bin/spicyz.cc
```diff
@@ -0,0 +1,226 @@
...
+            case 'T': driver_options->keep_tmps = true; break;
+
+            case 'v':
+                std::cerr << "spicyz"
+                          << " v" << hilti::configuration().version_string_long << std::endl;
+                return Nothing();
+
+            case 'h': usage(); return Nothing();
+
+            default: usage(); return hilti::result::Error(util::fmt("option %c not implemented", c));
```
bbannier, line 144: This will trigger for the normal error handling of `getopt`, i.e.,

- `':'` for a missing option argument
- `'?'` for an unsupported option

Let's either handle them explicitly, or probably simpler not propagate `c`, but just e.g., an `Error("could not parse options")`.

zeek/bin/spicyz.cc
```diff
@@ -0,0 +1,226 @@
...
+
+    while ( optind < argc )
+        driver_options->inputs.emplace_back(argv[optind++]);
+
+    if ( driver_options->inputs.empty() )
+        return hilti::result::Error("no input file given");
+
+    if ( ! compiler_options->debug ) {
+        if ( compiler_options->debug_trace || compiler_options->debug_flow )
+            return hilti::result::Error("cannot use --optimize with --cgdebug");
```
bbannier, line 156: Shouldn't this be `--debug-addl`?

zeek/bin/spicyz.cc
```diff
@@ -0,0 +1,226 @@
...
+#endif
+
+    driver.setDriverOptions(std::move(driver_options));
+    driver.setCompilerOptions(std::move(compiler_options));
+
+    driver.initialize();
+
+    for ( auto p : driver.driverOptions().inputs ) {
+        if ( ! driver.loadFile(p) )
+            exit(1); // error already reported
```
bbannier, line 219: Maybe this function should instead return a `Result<Nothing>` so we could report here into the right logger.

Similar for `compile` below.

zeek/bin/spicyz.cc
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    if ( ! driver.compile() )
+        exit(1); // error already reported
+
+    exit(0);
```
bbannier, line 225: Redundant. If at all let's `return 0;`, but probably better to just rely on the default behavior of `main`.

hilti/src/rt/library.cc
```diff
@@ -0,0 +1,51 @@
...
+#include <dlfcn.h>
+
+#include <hilti/rt/library.h>
+#include <hilti/rt/logging.h>
+
+using namespace hilti::rt;
+
+hilti::rt::Library::Library(const std::filesystem::path& path) {
+    if ( ! std::filesystem::exists(path) )
+        throw std::runtime_error(fmt("no such library: %s", path));
```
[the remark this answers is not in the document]
rsmmr: I'm adding a new `rt::EnvironmentError`

hilti/src/compiler/driver.cc
rsmmr: I'll add an assert, if it's empty we're in trouble.

I need to add to the style guide that using `size()` instead if `! empty()` is acceptable. :)

zeek/bin/spicyz.cc
rsmmr: Turns tout the whole if() isn't needed, `debug` is set automatically with these.

hilti/src/compiler/driver.cc
bbannier: Well, in a way we already do by explicitly disabling [this](https://clang.llvm.org/extra/clang-tidy/checks/readability-container-size-empty.html) check in the clang-tidy config. But since the code is almost clean here we could alternatively enable the check to remove the annoying human from the equation ;)

zeek/bin/spicyz.cc
rsmmr: Ack, changing.

zeek/bin/spicyz.cc
rsmmr: Turns out we don't need this if-clause at all, debug is always implicitly set here.

rsmmr:
I don't see anything wrong with the include paths for `compile/debug.h`. Does a clean `configure` help by any chance?

PR #320, opened 2020-05-20
Fix for Bison versions >= 3.6.

rsmmr:


hilti/src/3rdparty/justrx/src/re-parse.y
```diff
@@ -2,15 +2,18 @@
...
+#undef yyerror
+#define yyerror RE_Error
+
 #include <justrx/jrx-intern.h>
 #include <justrx/nfa.h>
 
 #include <stdio.h>
 
-extern void RE_error(void* scanner, jrx_nfa_context* nfactx, jrx_nfa** nfa, const char* msg);
+extern void RE_Error(void* scanner, jrx_nfa_context* nfactx, jrx_nfa** nfa, const char* msg);
```
bbannier, line 13: Let's add a comment here explaining why we deviate from the usual naming convention.
```suggestion
// NOTE: This function is not called `RE_error` to avoid a name collision with a Bison-internal name.
extern void RE_Error(void* scanner, jrx_nfa_context* nfactx, jrx_nfa** nfa, const char* msg);
```

PR #321, opened 2020-05-20
Fix regexp handling to respect &size

### LINKED ISSUE #286: Bug when parsing regexes with &size
Limiting the input of a regex with `&size` doesn't work correctly:

```
# cat test.spicy
module Test;

public type Foo = unit {
    x: /123.*/ &size=5;
    y: bytes &eod;
    on %done { print self; }
}; 

# printf '1234567890' | ./bin/spicy-driver -d test.spicy
[$x=b"1234567890", $y=b"67890"]
```

Expected would be ``x == b"12345"``.

The problem is in `regexp::MatchState::_advance()` in `hilti/src/rt/types/regexp.cc`: that function  iterates through its input data without paying attention to the end of the stream view that it got passed in. It always parses all the way to the end.

bbannier:
Closes #286.


hilti/src/compiler/clang.cc
```diff
@@ -633,6 +633,11 @@ Result<Library> ClangJIT::Implementation::compileModule(llvm::Module&& module) {
                                      "-o",
                                      *library_path};
 
+    const auto& ld_flags = context->options().optimize ? hilti::configuration().runtime_ld_flags_release :
+                                                         hilti::configuration().runtime_cxx_flags_debug;
```
rsmmr, line 637:
```suggestion
    const auto& ld_flags = context->options().optimize ? hilti::configuration().runtime_ld_flags_release :
                                                         hilti::configuration().runtime_ld_flags_debug;
```

Typo, right?

hilti/src/rt/tests/regexp.cc
```diff
@@ -44,3 +44,69 @@ TEST_CASE("advance on limited view") {
 }
 
 TEST_SUITE_END();
+
```
rsmmr, line 47: These should probably move into `tests/stream.cc`?

hilti/src/rt/tests/regexp.cc
```diff
@@ -44,3 +44,69 @@ TEST_CASE("advance on limited view") {
...
+    REQUIRE_EQ(view.size(), input.size());
+
+    // Trimming removes specified amount of data.
+    auto trimmed = view.trim(view.safeBegin() + 3);
+    CHECK_EQ(trimmed.size(), input.size() - 3);
+    CHECK(trimmed.startsWith("4567890"_b));
+
+    // Trimmed view expands when data is added.
+    stream.append("123"_b);
+    CHECK_EQ(trimmed.size(), input.size() - 3 + 3);
```
rsmmr, line 93: worth checking the new content here as well I'd say

hilti/src/compiler/clang.cc
bbannier: Yes, but actually the other way around. We'd need `runtime_cxx_flags_release` here here to trigger the correct link when using `-fprofile-instr-generate -fcoverage-mapping`.

I am removing this patch from this patchset for now since I am not 100% clear on what this could mean for other arbitrary CXX flags users could pass. A more controlled approach for properly including coverage information might be an extra configure time switch which sets correct compiler _and_ linker flags.

PR #322, opened 2020-05-20
Fix JIT-less build of Zeek plugin.

rsmmr:
- Fix include path
    - Build spicyz even if no JIT support is available (it can still generate C++ code).
    - Remove some unnecessary dependencies on HILTI/Spicy libs on plugin code


rsmmr:
Tried manual builds of `--disable-jit`/`--disable-it-for-zeek` and that worked fine. Will wait for CI, but @bbannier you can take a look already.

rsmmr:
CI passed now, I'll go ahead and merge to get the fix in.

PR #334, opened 2020-05-25
Additional unit tests for rt::RegExp

bbannier:
This adds additional unit tests for `rt::RegExp` and stringification tests. This patchset also includes cleanups of existing regexp-related functions.


bbannier:
This chain is on top of #321 which it includes for now.

hilti/src/rt/types/regexp.cc
```diff
@@ -202,8 +203,7 @@ RegExp::RegExp(const std::vector<std::string>& patterns, regexp::Flags flags) :
 }
 
 void RegExp::_newJrx() {
-    if ( _jrx_shared )
```
rsmmr, line 204: The commit message says "Replace unreachable pattern errors with assertions.", but this just removes the check; same below. That on purpose?

hilti/src/rt/tests/regexp.cc
```diff
@@ -83,6 +83,25 @@ TEST_SUITE_END();
 
 TEST_SUITE_BEGIN("MatchState");
 
+TEST_CASE("advance") {
+    CHECK_EQ(RegExp("123", regexp::Flags()).tokenMatcher().advance("123"_b, false), std::make_tuple(-1, 3));
```
rsmmr, line 91: hmm, shouldn't this return `(1, 3)`?

hilti/src/rt/tests/regexp.cc
```diff
@@ -83,6 +83,25 @@ TEST_SUITE_END();
...
+    CHECK_EQ(RegExp("123", regexp::Flags()).tokenMatcher().advance("123"_b, true), std::make_tuple(1, 3));
+
+    CHECK_EQ(RegExp(std::vector<std::string>({"abc", "123"}), regexp::Flags()).tokenMatcher().advance("123"_b, true),
+             std::make_tuple(2, 3));
+
+    CHECK_EQ(RegExp("", regexp::Flags()).tokenMatcher().advance("123"_b, false), std::make_tuple(-1, 3));
```
rsmmr, line 97: an empty regex should always match (or be rejected, either way would be fine)

hilti/src/rt/types/regexp.cc
bbannier: Added them back.

bbannier:
> The tests have quite a few `RegExp("a", regexp::Flags())`. Do we just want to make `regexp::Flags()` the default for the `RegExp` constructor?

Done.

hilti/src/rt/tests/regexp.cc
bbannier: I agree, but I am unsure what is happening here. We put a non-final `Byte` into the matcher which almost immediately forwards it into `jrx_regex_partial`. I believe jrx should return something else here as we (as in `rt::RegExp`) cannot tell if the pattern contained e.g., wildcards towards the end.

This test documents the existing behavior. I added a `TODO` to look into that independently.

hilti/src/rt/tests/regexp.cc
bbannier: Is returning `(-1, 3)` for 3 bytes of input not the correct answer for _always match_? The range seems also valid, but if we match always, we also always can consume more data.

rsmmr: I would expect an empty regex  to immediately match an empty string and stop there, so `(1, 0)`. 

Not a big deal, but worth another TODO.

bbannier:
Can you have another look @rsmmr? I added commits to the end fixing the justrx testsuite and one fixing a leak in justrx the added unit tests triggered.

PR #338, opened 2020-05-25
Add range check to port parsing in EVT files.

### LINKED ISSUE #330: evt format parser does not validate range of port
We currently do not validate that a port parsed from an `evt` file contains a value in the allowed range, e.g., a value `port 123456/tcp` is accepted. We should perform some validation before we create a `hilti::rt::Port` which restricts values to `uint16_t`.

rsmmr:
Closes #330.

### DIFF
--- a/zeek/compiler/glue-compiler.cc
+++ b/zeek/compiler/glue-compiler.cc
@@ -198,10 +198,13 @@ static hilti::rt::Port extract_port(const std::string& chunk, size_t* i) {
         throw ParseError("cannot parse port specification");
 
     hilti::rt::Protocol proto;
-    uint16_t port;
+    uint64_t port;
 
     s = chunk.substr(*i, j - *i);
     util::atoi_n(s.begin(), s.end(), 10, &port);
+
+    if ( port > 65535 )
+        throw ParseError("port outside of valid range");
 
     *i = j;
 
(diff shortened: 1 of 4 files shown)


zeek/compiler/glue-compiler.cc
```diff
@@ -370,7 +373,7 @@ bool GlueCompiler::loadEvtFile(std::filesystem::path& path) {
             _locations.pop_back();
         }
 
-    } catch ( const ParseError& e ) {
+    } catch (const ParseError& e) {
```
rsmmr, line 376: Looks like it picked up the wrong clang-format here. I'll fix that before merge.

zeek/compiler/glue-compiler.cc
```diff
@@ -198,11 +198,14 @@ static hilti::rt::Port extract_port(const std::string& chunk, size_t* i) {
         throw ParseError("cannot parse port specification");
 
     hilti::rt::Protocol proto;
-    uint16_t port;
+    uint64_t port;
```
bbannier, line 201: Line 194 will already lead to us rejecting negative numbers (e.g., `-2/udp`), but it would be great to add a test for that as well so we do not accidentally introduce changes causing well-defined, but unintended unsigned integer overflow here.

Issue #56, opened 2020-04-06
Add reflection API for generated Spicy parsers


bbannier:
This is so a Spicy host applications can discover what's being parsed without specific knowledge about each parser.

We can probably reuse some existing, light-weight C++ reflection library to instrument the generated parsing structs. It will need to go into the HILTI codegen code then.

rsmmr:
Being worked on [in this branch](https://github.com/zeek/spicy/tree/topic/robin/gh-38-spicy-dump).

rsmmr:
Done with #347 .

Issue #38, opened 2020-04-06
Bring back spicy-dump


bbannier:
This is an alternative command line drive for Spicy that just prints fields as they are parsed. The name is meant allude to tcpdump.

Old code here: https://github.com/rsmmr/hilti/tree/master/tools/spicy-dump

rsmmr:
Done with #347 .

PR #347, opened 2020-05-27
Type reflection API & spicy-dump

rsmmr:
This is adding support for #56  and #38. It's a series of commits with the initial ones just adding/fixing some smal stuff, before  a few larger ones then add infrastructure and then finally the new spicy-dump tool. The latter could use a few more tests, still thinking how to do that.


hilti/include/rt/type-info.h
```diff
@@ -0,0 +1,149 @@
...
+        type_info::SignedInteger<int16_t>,
+        type_info::SignedInteger<int32_t>,
+        type_info::SignedInteger<int64_t>,
+        type_info::String,
+        type_info::Struct,
+        type_info::ValueReference,
+        type_info::UnsignedInteger<uint8_t>,
+        type_info::UnsignedInteger<uint16_t>,
+        type_info::UnsignedInteger<uint32_t>,
+        type_info::UnsignedInteger<uint64_t>
```
bbannier, line 144: Such a closed set of types probably works if we can keep the set of language types quasi-constant on the Spicy-side as well. Else this very strong coupling between the language and users of this API could lead to us breaking users often (as the visitors they'd necessarily implement would not be able to handle all cases absent a common base class).

Once this becomes ready we'll need type info for function-like types as well.

spicy/include/rt/parsed-unit.h
```diff
@@ -0,0 +1,69 @@
...
+#include <hilti/rt/types/reference.h>
+
+namespace spicy::rt {
+
+/**
+ * Type-erased wrapper around parsed unit instances.
+ *
+ * Initially, this will be unbound, i.e., not refer to any particular
+ * instance. `init()` then binds it to an instance, which will let
+ * `ParsedUnit` hold a strong reference to it.
```
bbannier, line 15: Not sure I understand the motivation for this. Couldn't we just as well provide a helper function which a `ParsedUnit` which could then use RAII to manage its data? This would lead to different visitor impls though.

hilti/include/rt/type-info.h
rsmmr: Yeah, I imagine the set of types to become very stable. For parameterized types will need only one entry (except for the integers, and that's only to get access to underlying C++ ints).

spicy/include/rt/parsed-unit.h
rsmmr: The main reason here is the API that we have for calling into the Spicy-generated parsing. We can't easily pass complex objects back from there, that's why we pass a reference parameter of type `ParsedUnit` for the function to then initialize. However, the instance needs to survive the initial call into parsing so that resuming will continue to have access to it. And internally parsed instances are manager through a `ValueReference`. Hence the `StrongReference` to keep it around; a strong reference can bind to a value reference.

Does that help explain it? If you would still do it differently, can you give a quick example of how that would look like?

hilti/include/rt/compiler-setup.h
```diff
@@ -5,6 +5,9 @@
 
 #pragma once
 
+// Our type info infrastructure needs this, and it seems to work for our cases.
+#pragma clang diagnostic ignored "-Winvalid-offsetof"
```
bbannier, line 9: This is a pretty big hammer. Any chance to move it closer to the affected code? Something like (i.e., `push` it before the suspect call, `pop` it again after).

hilti/include/rt/type-info.h
```diff
@@ -0,0 +1,994 @@
...
+namespace type_info {
+
+/**
+ * Class representing a HILTI value generically through a pair of (1) a raw
+ * pointer refering the value's storage, and (2) type information describing
+ * how to interpret the raw pointer. An instance may be in an invalid state
+ * if there's no underlying value available (e.g., when dereferencing an
+ * unset `optional`).
+ */
+class Value {
```
bbannier, line 29: I am not a fan of this type as it just unsafely wraps data owned elsewhere, but could e.g., be stored away obscuring the data dependency. It seems we use it to access data inside `ParsedUnit` and to recursively iterate over values.

I think an implementation where we do not store not-owned data away would be much easier. I do not have a full alternative design ready, but I could imagine something which could still store the type info as e.g., a `shared_ptr<TypeInfo>` (might need to look out for reference cycles, but that could be an unlikely problem in a tree structure), but not store a pointer to the underlying value storage, but instead provide a function to visit a `ParsedUnit`'s data (might need to be a `friend` of `ParsedUnit`). Storing an additional offset into the data wouldn't be safe either and it would be better if we could use safe hilti-rt container access instead.

Such a change would largely be confined to our implementations of types in `detail` here and apart from accessing the underlying storage `Value` could keep most of its interface.

rsmmr: As discussed offline, not sure we could build the API much else without taking a very different approach. But I tried the idea of adding back pointers to the top-level value so that we can catch life-time violations. I just pushed that into a new commit a the very end, see what you think about that.

bbannier: Thanks, comments for inline.

hilti/include/rt/type-info.h
```diff
@@ -0,0 +1,1071 @@
...
+    std::weak_ptr<bool> handle() const {
+        if ( ! _value )
+            throw InvalidValue("type-info traversal not tied to value");
+
+        return _handle;
+    }
+
+    std::shared_ptr<bool> _handle;
+    std::optional<hilti::rt::StrongReferenceGeneric> _value;
+};
```
bbannier, line 60: This helper class gets the job done, but feels a little heavy for me. WDYT about instead of inheriting `ParsedUnit` from `Parent` to just use a `shared_ptr<ParsedUnit*>` when constructing a `Value` (assertions are that the smart pointer contains a value and that the contained value is not nil)?

That would avoid `friend`ing `Value`, the additional inheritance level and from having to distinguish between the different states of `_handle` and `_value`. The downside would be that `ParsedUnit` would have to make a slightly nastier call passing in `make_shared(this)` and would need to check that `value` is not called before `initialize` (maybe just using one of its other members as a proxy, e.g., asserting `_unit`.

hilti/include/rt/type-info.h
```diff
@@ -0,0 +1,1071 @@
...
+#include <hilti/rt/fmt.h>
+#include <hilti/rt/types/all.h>
+#include <hilti/rt/types/bytes.h>
+#include <hilti/rt/types/stream.h>
+
+namespace hilti::rt {
+
+struct TypeInfo;
+
+namespace type_info {
```
bbannier, line 20: Couple general issues in this file:

- We usually require rt types to be default constructable so they can e.g., be used in global variables. Most of these classes do not have default constructors. Feel free to drop this if you think this is not an issue.
- Many classes in this file have `const` member variables so they are e.g., not assignable. Unless this is deliberate I'd suggest we drop the `const` qualifiers.

hilti/include/rt/type-info.h
```diff
@@ -0,0 +1,1071 @@
...
+};
+
+}; // namespace union_
+
+/** Auxiliary type information for type ``union`. */
+class Union {
+public:
+    /**
+     * Type of a function that, given a union value, returns the index of the
+     * currently set field, with -1 indicating no field being set.
```
bbannier, line 810: Since this function returns a `size_t` it should be impossible to return `-1` from it. Can we change this to return e.g., an `optional<size_t>` instead?

hilti/src/compiler/codegen/types.cc
```diff
@@ -787,12 +993,74 @@ std::optional<cxx::Expression> CodeGen::typeDefaultValue(const hilti::Type& t) {
...
+const CxxTypeInfo& CodeGen::_getOrCreateTypeInfo(const hilti::Type& t) {
+    std::stringstream display;
+
+    if ( t.typeID() )
+        // Prefer the bare type name as the display value.
+        display << *t.typeID();
+    else
+        hilti::print(display, t);
+
+    if ( display.str().empty() )
```
bbannier, line 1015: ❤️

hilti/src/compiler/codegen/types.cc
```diff
@@ -787,12 +993,74 @@ std::optional<cxx::Expression> CodeGen::typeDefaultValue(const hilti::Type& t) {
...
+
+            unit()->add(*ti.declaration);
+
+            return ti;
+        });
+}
+
+cxx::Expression CodeGen::typeInfo(const hilti::Type& t) {
+    auto ti = _getOrCreateTypeInfo(t);
+    return ti.reference;
```
bbannier, line 1063: No need for the temporary.
```suggestion
    return _getOrCreateTypeInfo(t).reference;
```

hilti/src/compiler/parser/parser.yy
```diff
@@ -710,6 +711,7 @@ expr_e        : BEGIN_ '(' expr ')'              { $$ = hilti::expression::Unres
               | NEW expr                         { $$ = hilti::expression::UnresolvedOperator(hilti::operator_::Kind::New, {std::move($2), hilti::expression::Ctor(hilti::ctor::Tuple({}, __loc__))}, __loc__); }
               | NEW type                         { $$ = hilti::expression::UnresolvedOperator(hilti::operator_::Kind::New, {hilti::expression::Type_(std::move($2)), hilti::expression::Ctor(hilti::ctor::Tuple({}, __loc__))}, __loc__); }
               | NEW type '(' opt_exprs ')'       { $$ = hilti::expression::UnresolvedOperator(hilti::operator_::Kind::New, {hilti::expression::Type_(std::move($2)), hilti::expression::Ctor(hilti::ctor::Tuple(std::move($4), __loc__))}, __loc__); }
+              | TYPEINFO '(' type ')'            { $$ = hilti::expression::TypeInfo(std::move($3), __loc__); }
```
bbannier, line 714: WDYT about also allowing to pass an expression to `typeinfo`? Since not all types can be instantiated in scripts a type variant is needed, but I could imagine an expression variant being useful to for debugging or defensive programming.

tests/hilti/codegen/type-info.hlt
```diff
@@ -0,0 +1,537 @@
...
+                           auto i = x.value(v);
+                           CHECK_EQ(type_info::value::auxType<type_info::String>(i).get(i), "11");
+                       },
+                       [&](const hilti::rt::type_info::Void& x) { SEEN(); },
+                       [&](const hilti::rt::type_info::WeakReference& x) {
+                           SEEN();
+                           CHECK(x.value(v));
+                           testStruct(x.value(v)); // TODO: failure
+                       },
+                       [&](const auto& x) {}},
```
bbannier, line 428: Remove the default case.

tests/hilti/codegen/type-info.hlt
```diff
@@ -0,0 +1,537 @@
+# @TEST-EXEC: spicy-build -d -S %INPUT type-info.cc && ./a.out
```
bbannier, line 1: With you writing code like this I should again look into why I didn't see compiler warnings emitted during JIT :)

spicy/include/rt/parsed-unit.h
```diff
@@ -0,0 +1,64 @@
...
+        u._unit = hilti::rt::StrongReference(t);
+        u._ptr = t.get();
+        u._ti = ti;
+        u.tie(u._unit);
+    }
+
+private:
+    hilti::rt::StrongReferenceGeneric _unit;
+    const hilti::rt::TypeInfo* _ti;
+    const void* _ptr;
```
bbannier, line 61: Since we use a default constructor let's initialize these.
```suggestion
    const hilti::rt::TypeInfo* _ti = nullptr;
    const void* _ptr = nullptr;
```

spicy/include/rt/parsed-unit.h
```diff
@@ -0,0 +1,64 @@
...
+            return *p;
+        else
+            throw hilti::rt::NullReference("parsed unit not set");
+    }
+
+    /**
+     * Returns the instance and its type in a value representation suitable
+     * to use with the `type-info` API for iteration over the fields.
+     */
+    hilti::rt::type_info::Value value() const { return {_ptr, _ti, *this}; }
```
bbannier, line 32: Does it make sense to create a `Value` if we are not initialized and `_ptr` and `_ti` are nil? Maybe we should assert here or emit an internal error.

spicy/include/rt/typedefs.h
```diff
@@ -21,6 +23,9 @@ template<typename T>
 using Parse2Function = hilti::rt::Resumable (*)(UnitType<T>&, hilti::rt::ValueReference<hilti::rt::Stream>&,
                                                 const std::optional<hilti::rt::stream::View>&);
 
+using Parse3Function = hilti::rt::Resumable (*)(ParsedUnit&, hilti::rt::ValueReference<hilti::rt::Stream>&,
```
bbannier, line 26: `Resumable` needs `hilti/rt/fiber.h`.

hilti/include/rt/types/any.h
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
+
+#pragma once
+
+#include <hilti/rt/extension-points.h>
+
+namespace hilti::rt {
+
+namespace detail::adl {
```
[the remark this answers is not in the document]
rsmmr: prefer to leave as is, just for consistency.

hilti/include/rt/compiler-setup.h
rsmmr: Not easy to move there as it's inside a block of `static ...` declarations, we'd need to add something to the formatter API to prefix/postfix. I'm not too worried, though, it's a pretty rare feature to use.

spicy/bin/spicy-dump/printer-json.cc
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
+
+#include <hilti/rt/libhilti.h>
+
+#include "printer-json.h"
```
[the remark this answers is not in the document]
rsmmr: Let's do that sometime as a general pass over the files, not adhoc for a single one.

hilti/include/rt/type-info.h
rsmmr: Problem with that: it'd be a layering violation. The type-info API is on the HILTI side, but ParsedUnit is on the Spicy side, HILTI doesn't know about that. That's why I did it this way. It's also used inside a HILTI-side test, that's what the 1st constructor is for. Let me know if you have another idea, though.

hilti/include/rt/type-info.h
rsmmr: The type info instances are generated exclusively as `static ...`` globals, and should never change. That's why the members are `const` (maybe it helps the compiler with optimizing something). Similar reasoning re/  the default ctors: They won't be used at all, and it may in fact be better to not have them so that they don't accidentally get used.

hilti/include/rt/type-info.h
rsmmr: It's passing through std::variant_npos, I'll add a local const mapping to that.

hilti/include/rt/types/reference.h
```diff
@@ -244,7 +244,9 @@ class ValueReference {
      */
     ValueReference& operator=(ValueReference&& other) noexcept {
         if ( &other != this ) {
-            *_safeGet() = std::move(*other._safeGet());
+            // We can't move the actual value as other referencez may be
+            // refering to it.
+            *_safeGet() = *other._safeGet();
```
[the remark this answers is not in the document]
rsmmr: A `ValueReference` may be created from a raw pointer, in which case we can't return a `shared_ptr`. That use case is a bit unfortunate, it's needed to create a `ValueReference` from this.  Can't think of a different way of doing this, so leaving as is.

hilti/src/compiler/parser/parser.yy
rsmmr: Good thought, adding `typeinfo(expr)`.

spicy/src/compiler/codegen/unit-builder.cc
```diff
@@ -132,7 +132,7 @@ struct FieldBuilder : public hilti::visitor::PreOrder<void, FieldBuilder> {
 
         addHookImplementation(h.hook(), AttributeSet::find(h.hook().attributes(), "foreach"));
     }
-};
+}; // namespace
```
bbannier, line 135: Was this generated by clang-format (it does generate these comments). This paren terminates a struct, not a namespace.
```suggestion
};
```

rsmmr: (Not sure.)

hilti/include/rt/types/tuple.h
```diff
@@ -17,6 +17,7 @@ template<typename Tuple, size_t Idx>
 ptrdiff_t elementOffset() {
     // This is pretty certainly not well-defined, but seems to work for us ...
     const Tuple* p = nullptr;
+    // NOLINTNEXTLINE(clang-analyzer-core.NonNullParamChecker)
```
bbannier, line 20: Yikes, I certainly agree with clang-tidy that this assured nullptr dereference is nasty, if not certainly UB. WDYT about instead of suppressing this warning to pass in the a presumably valid tuple as arg and then work with that?
```.cpp
template<typename Tuple, size_t Idx>
ptrdiff_t elementOffset(const Tuple& xs) {
    // This is pretty certainly not well-defined, but seems to work for us ...
    return reinterpret_cast<const char*>(&std::get<Idx>(xs)) - reinterpret_cast<const char*>(&xs);
}
```

Looking at the call sites we should always have such a tuple at hand.

Of course even better would be to not require passing this in, but instead work only with types, but I cannot see how that would be possible in general (it might e.g., be impossible to create an instance of the tuple since its fields might not have default constructors).

rsmmr: I changed this.

PR #358, opened 2020-06-02
Tests and fixes for rt::Bytes

bbannier:


hilti/include/rt/types/bytes.h
```diff
@@ -297,7 +337,7 @@ class Bytes : protected std::string, public hilti::rt::detail::iterator::Control
     Bytes join(const Vector<T>& parts) const {
         Bytes rval;
 
-        for ( auto i = 0; i < parts.size(); ++i ) {
+        for ( size_t i = 0; i < parts.size(); ++i ) {
```
rsmmr, line 340: Just curious: Why this change?

tests/hilti/types/bytes/iterator.hlt
```diff
@@ -18,15 +18,6 @@ assert *i++ == '1';
 assert *i++ == '2';
 assert *++i == '4';
 
-assert i > j;
```
rsmmr, line 21: Did we lose the comparison operators? They would be good to keep, so that one can build custom loops.

hilti/include/rt/types/bytes.h
bbannier: As written `i` is signed while `parts.size()` is unsigned. This triggers a diagnostic e.g., with clang that we previously didn't see as we only exercised this code in JIT; now this causes an compilation error from the unit test.

tests/hilti/types/bytes/iterator.hlt
bbannier: This is the only place we ever exercised this kind of operator for operators, and we e.g., do not support it for `rt::Vector` iterators either. General STL-type iterators are typically only compared for inequality (e.g., `it != container.end()`), but if we need iterators to be ordered as well I can add the operators back. In that case lets come up with some code which actually exercises that property.

bbannier: I added the ordering of iterators into `rt::Bytes` back (along with the previously removed test case), and also opened #363 for adding the same property to iterators into `rt::Vector`. As `rt::List` is not index-based I didn't add this property to its iterators, but if it turns out to be important we can change its iterators as well to hold the needed internal state.

PR #359, opened 2020-06-02
Fix operator parsing

rsmmr:


rsmmr:
@bbannier please take another quick look

PR #360, opened 2020-06-03
Reduce resources requested from Cirrus.

rsmmr:
Adapting to Cirrus' changes in limits.


bbannier:
GCC typically uses way more memory then Clang which is probably the reason we got unlucky with https://github.com/zeek/spicy/pull/360/checks?check_run_id=734031305. Let's reduce the build parallelism even more (hopefully we can recoup some of that via a warm ccache); typically execution time is dominated by btest anyway.

rsmmr:
I adapted the memory limits, see if that makes sense to you.

PR #364, opened 2020-06-03
 Fix duplicate evaluation of if-condition when used with &parse-at.

### LINKED ISSUE #343: &parse-at combined with a conditional if on the line, isn't supported
&parse-at combined with a conditional if on the line, isn't supported
```
module Error_6465_parse_at_w_if;

public type oneInstr = unit {
%random-access;
    var rootOfInstr: iterator<stream>;
    var SixSevenActive: bool = False;
    on %init {
      self.rootOfInstr = self.input();
    }
   : evex &parse-at = self.rootOfInstr if (SixSevenActive == False);
};

type evex = unit {
  frst : uint8;
};
```
outputs
```
spicy/build/bin/spicy-driver Error_6465_reduced.spicy < pmovmskb_ldmxcsr.bin
[error] Error_6465_reduced.spicy:3:24-11:2: call does not match any method: <Error_6465_parse_at_w_if::evex>.__parse_stage1(<value_ref<stream>>, <<unknown type>>, <bool>, <int<64>>, <iterator<stream>>)
[error] Error_6465_reduced.spicy:3:24-11:2:   candidate methods:
[error] Error_6465_reduced.spicy:3:24-11:2:   - <Error_6465_parse_at_w_if::evex>.__parse_stage1(<value_ref<stream>>, <view<stream>>, <bool>, <int<64>>, <iterator<stream>>)
```
replacing that conditional if on the line, with just semi-colon, 
```
   : evex &parse-at = self.rootOfInstr;
```
succeeds.

rsmmr:
Closes #343.

### DIFF
--- a/spicy/src/compiler/codegen/parser-builder.cc
+++ b/spicy/src/compiler/codegen/parser-builder.cc
@@ -409,85 +409,89 @@ struct ProductionVisitor
         const auto& field = meta.field();
         assert(field); // Must only be called if we have a field.
 
-        if ( field && is_field_owner && ! meta.container() ) {
-            pb->enableDefaultNewValueForField(true);
+        if ( is_field_owner ) {
+            // We are the field's owner, apply the various attributes.
 
-            if ( field->parseType().isA<type::Void>() ) {
-                // No value to store.
-                pushDestination(hilti::expression::Void());
+            if ( ! meta.container() ) {
+                pb->enableDefaultNewValueForField(true);
+
+                if ( field->parseType().isA<type::Void>() ) {
+                    // No value to store.
+                    pushDestination(hilti::expression::Void());
+                }
+                else if ( AttributeSet::find(field->attributes(), "&convert") ) {
+                    // Need a temporary for the parsed field.
+                    auto dst = builder()->addTmp(fmt("parsed_%s", field->id()), field->parseType());
+                    pushDestination(dst);
+                }
+                else if ( field->isTransient() ) {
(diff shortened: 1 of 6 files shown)


spicy/src/compiler/codegen/parser-builder.cc
```diff
@@ -409,7 +409,10 @@ struct ProductionVisitor
         const auto& field = meta.field();
         assert(field); // Must only be called if we have a field.
 
-        if ( field && is_field_owner && ! meta.container() ) {
+        if ( ! is_field_owner )
```
bbannier, line 412: This disables all attribute handling in this function if `! is_field_owner`? Is that intentional? Since this function does so much it might be easier to follow the code if we'd avoid the early return and instead wrap handling not applicable in e.g., `if ( is_field_owner) ...` or similar.

spicy/src/compiler/codegen/parser-builder.cc
```diff
@@ -495,6 +498,15 @@ struct ProductionVisitor
         const auto& field = meta.field();
         assert(field); // Must only be called if we have a field.
 
+        if ( ! is_field_owner ) {
```
bbannier, line 501: This one is closer to the top so not as confusing, but maybe we should still add a comment here to explain why handling of fields with `! is_field_owner` is not needed and an early return is safe.

spicy/src/compiler/codegen/parser-builder.cc
rsmmr: Yes, on purpose, but let me make it more clear.

PR #366, opened 2020-06-04
Zeek plugin: Catch if more parameters are given than Zeek expects.

### LINKED ISSUE #365: segfault when .evt event doesn't match .zeek event
Re: conversation in Slack on #spicy channel.

Specifically in the attached examples... in .evt:
on SM::DATA -> event SM::data($conn, $is_orig, self.first_byte);

in .zeek
#event SM::data(c: connection, is_orig: bool, first_byte: int)
event SM::data(c: connection, first_byte: int)

[SM.tar.gz](https://github.com/zeek/spicy/files/4726294/SM.tar.gz)

rsmmr:
The check we already had for this came to late.

Skipping a test, as it would require a full compiler cycle + Zeek
execution to trigger the error message.

Closes #365.

### DIFF
--- a/zeek/plugin/src/runtime-support.cc
+++ b/zeek/plugin/src/runtime-support.cc
@@ -86,9 +86,15 @@ void rt::raise_event(EventHandlerPtr handler, const hilti::rt::Vector<Val*>& arg
     ::mgr.QueueEventFast(handler, vl);
 }
 
-BroType* rt::event_arg_type(EventHandlerPtr handler, uint64_t idx) {
+BroType* rt::event_arg_type(EventHandlerPtr handler, uint64_t idx, std::string_view location) {
     assert(handler);
-    return (*handler->FType()->ArgTypes()->Types())[idx];
+
+    auto zeek_args = handler->FType()->ArgTypes()->Types();
+    if ( idx >= static_cast<uint64_t>(zeek_args->length()) )
+        throw TypeMismatch(fmt("more parameters given than the %d that the Zeek event expects", zeek_args->length()),
+                           location);
+
+    return (*zeek_args)[idx];
 }
 
 Val* rt::current_conn(std::string_view location) {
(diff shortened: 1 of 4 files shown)


zeek/plugin/src/runtime-support.cc
```diff
@@ -86,9 +86,15 @@ void rt::raise_event(EventHandlerPtr handler, const hilti::rt::Vector<Val*>& arg
     ::mgr.QueueEventFast(handler, vl);
 }
 
-BroType* rt::event_arg_type(EventHandlerPtr handler, uint64_t idx) {
+BroType* rt::event_arg_type(EventHandlerPtr handler, uint64_t idx, std::string_view location) {
     assert(handler);
-    return (*handler->FType()->ArgTypes()->Types())[idx];
+
+    auto zeek_args = handler->FType()->ArgTypes()->Types();
+    if ( idx >= zeek_args->length() )
```
bbannier, line 93: That's a comparison between signed and unsigned variables. Let's cast on of them to fix the semantics we want. Since `zeek_args->length()` is signed we also need to fix the format string below.

zeek/plugin/src/runtime-support.cc
```diff
@@ -86,9 +86,15 @@ void rt::raise_event(EventHandlerPtr handler, const hilti::rt::Vector<Val*>& arg
...
+        throw TypeMismatch(fmt("more parameters given than the %u that the Zeek event expects", zeek_args->length()),
```
bbannier, line 94: Just wondering, since we now abort processing for an incorrect number of arguments, can we remove turn the check in `rt::raise_event` into an assert/hard failure, or remove it altogether?

rsmmr: No, there are still cases where get in there (less parameters on the Spicy side than Zeek)

zeek/plugin/src/runtime-support.cc
rsmmr: Good catch, will fix.

PR #369, opened 2020-06-08
Fix &convert with vectors.

### LINKED ISSUE #368: &convert cannot deduce type when working on list
While `&convert` can be used to transform scalar values with user-provided functions, it fails to deduce the correct destination type for lists as of fa3e0b22160fa059092782fe3586e2482942dc06.
```
$ cat /tmp/bla.spicy
module Foo;

function fn1(x: uint8): uint8 { return x; }

function fn2(xs: vector<uint8>): vector<uint8> { return xs; }

public type P = unit {
    x: uint8 &convert=fn1($$);
    xs: uint8[] &size=1 &convert=fn2($$); # Fails to deduce.
};
```

```
$ spicyc -j /tmp/bla.spicy
[error] /tmp/bla.spicy:9:9: type '<unknown type>' cannot be used for function parameter
```

rsmmr:
Fixes #368.

### DIFF
--- a/hilti/include/rt/types/vector.h
+++ b/hilti/include/rt/types/vector.h
@@ -233,8 +233,8 @@ public:
 
     Vector(std::initializer_list<T> init, const Allocator& alloc = Allocator()) : V(std::move(init), alloc) {}
 
-    Vector(const rt::List<T>& l) : std::vector<T>(l.begin(), l.end()) {}
-    Vector(rt::List<T>&& l) : std::vector<T>(std::move_iterator(l.begin()), std::move_iterator(l.end())) {}
+    Vector(const rt::List<T>& l) : V(l.begin(), l.end()) {}
+    Vector(rt::List<T>&& l) : V(std::move_iterator(l.begin()), std::move_iterator(l.end())) {}
     ~Vector() = default;
 
     /** Returns the last element of the `vector`.
--- a/hilti/src/ast/node_ref.cc
+++ b/hilti/src/ast/node_ref.cc
@@ -11,7 +11,7 @@ NodeRef::NodeRef(Node& n) : _control(n._control()) {}
 
 Node* NodeRef::_node() const {
     if ( ! _control )
-        throw node_ref::Invalid("accesd to uninitalized node reference");
+        throw node_ref::Invalid("access to uninitialized node reference");
 
     if ( ! _control->_node )
         throw node_ref::Invalid("dangling node reference");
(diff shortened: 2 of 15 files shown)


hilti/include/rt/types/vector.h
```diff
@@ -233,8 +233,8 @@ class Vector : protected std::vector<T, Allocator> {
 
     Vector(std::initializer_list<T> init, const Allocator& alloc = Allocator()) : V(std::move(init), alloc) {}
 
-    Vector(const rt::List<T>& l) : std::vector<T>(l.begin(), l.end()) {}
-    Vector(rt::List<T>&& l) : std::vector<T>(std::move_iterator(l.begin()), std::move_iterator(l.end())) {}
+    Vector(const rt::List<T>& l) : std::vector<T, Allocator>(l.begin(), l.end()) {}
+    Vector(rt::List<T>&& l) : std::vector<T, Allocator>(std::move_iterator(l.begin()), std::move_iterator(l.end())) {}
```
bbannier, line 237:
```suggestion
    Vector(const rt::List<T>& l) : V(l.begin(), l.end()) {}
    Vector(rt::List<T>&& l) : V(std::move_iterator(l.begin()), std::move_iterator(l.end())) {}
```

hilti/src/ast/node_ref.cc
```diff
@@ -11,7 +11,7 @@ NodeRef::NodeRef(Node& n) : _control(n._control()) {}
 
 Node* NodeRef::_node() const {
     if ( ! _control )
-        throw node_ref::Invalid("accesd to uninitalized node reference");
+        throw node_ref::Invalid("access to uninitialized node reference");
```
bbannier, line 14: Not added here, but we should eventually make `node_ref::Invalid` a proper `RuntimeError`.

spicy/include/ast/types/unit-items/field.h
```diff
@@ -103,6 +103,21 @@ class Field : public hilti::NodeBase, public spicy::trait::isUnitItem {
     // Node interface.
     auto properties() const { return node::Properties{{"engine", to_string(_engine)}}; }
 
+    // Helper function for vector fields that returns the type of the
+    // vector's elements. The helper can be used in situations where the
+    // field type might not be fully resolved it. It computes the type
```
bbannier, line 108:
```suggestion
    // field type might not be fully resolved. It computes the type
```

hilti/src/ast/node_ref.cc
rsmmr: It's a purely compiler/AST-side error, so not really a runtime error I would say.

Issue #355, opened 2020-06-02
a passed parameter cannot be used in the initializer on same line as var declaration


duffy-corelight:
When a passed parameter is used in the initializer on same line as the var declaration, vector dereference, and bitwise-and (perhaps there are others) do not work.
```
module Issue_355_initializing_string;

public type outerUint = unit {
    var ohEffOpCodeToUse: uint8 = 1;
  : OhEffSixAndSeven(self.ohEffOpCodeToUse);
};

const OhEffSixAndSevenOpCode = vector<string>("PUNPCKLBW", "PUNPCKLWD");
type OhEffSixAndSeven = unit(initialOp: uint8) {
    var opCodeText: string = OhEffSixAndSevenOpCode[0x1 & initialOp];
    var nonMaskdText: string = OhEffSixAndSevenOpCode[initialOp];
    on %done {
      self.opCodeText = OhEffSixAndSevenOpCode[0x1 & initialOp];
      self.nonMaskdText= OhEffSixAndSevenOpCode[initialOp];
    }
};
```
reports:
```
printf "4\n" | /usr/local/etc/spicy/build/bin/spicy-driver -RB  /usr/local/etc/tmp/Issue_355_initializing_string.spicy
[error] /usr/local/etc/tmp/Issue_355_initializing_string.spicy:10:53: unsupported operator: <uint<64>> & <<unknown type>>
[error] /usr/local/etc/tmp/Issue_355_initializing_string.spicy:11:32: unsupported operator: <const vector<string>>[<<unknown type>>]
[error] spicy-driver: aborting after errors
7062 ~/Documents/work/dev/source/fuzzspicy$
```

By changing to do these declarations with no initializer, i.e.
```
    var opCodeText: string; # = OhEffSixAndSevenOpCode[0x1 & initialOp];
    var nonMaskdText: string; # = OhEffSixAndSevenOpCode[initialOp];
```
it works fine.

rsmmr:
Not supported currently, but the pattern seems natural enough that it feels its worth a "bug" label.

PR #370, opened 2020-06-08
Allow access to unit parameters inside default expressions for unit  variables

rsmmr:
#355


rsmmr:
Best to look at individual commits, there's some reorg at the beginning.

hilti/include/global.h
```diff
@@ -84,35 +84,35 @@ Result<std::pair<NodeRef, ID>> lookupID(const ID& id, const visitor::Position<No
...
         auto [stop, resolved] = detail::lookupID(id, **i);
 
-        if ( ! stop ) {
-            if ( auto t = (*i)->tryAs<Type>(); t && t->hasFlag(type::Flag::NoInheritScope) ) {
-                // Advance to module scope directly.
-                while ( ++i != p.path.rend() ) {
-                    if ( (*i)->isA<Module>() )
-                        break;
+        if ( resolved ) {
+            if ( auto d = (*resolved).first->tryAs<D>() ) {
```
bbannier:
```suggestion
            if ( auto d = resolved->first->tryAs<D>() ) {
```

hilti/include/ast/scope-lookup.h
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
+//
+// Functionality factored out from scope.h to avoid header loops.
+
+#pragma once
+
+#include <optional>
+#include <string>
```
bbannier, line 8: Let's move the include of `<utility>` for `pair` from the `.cc` file here.

```suggestion
#include <utility>
```

hilti/src/ast/scope-lookup.cc
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
+
+#include <utility>
+
+#include <hilti/ast/ctors/enum.h>
+#include <hilti/ast/declarations/constant.h>
+#include <hilti/ast/declarations/module.h>
+#include <hilti/ast/scope-lookup.h>
```
bbannier, line 8: Let's include the header matching this impl file first,
```.cpp
#include "hilti/ast/scope-lookup.h"
```

This is used by some tools to e.g., infer which header should declare entities defined in some impl file. This is also the a widely used style, see e.g., https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes (which I do not agree with in many places, but here it IMHO makes good sense).

hilti/include/ast/scope-lookup.h
```diff
@@ -0,0 +1,75 @@
...
+#include <hilti/ast/module.h>
+#include <hilti/ast/node_ref.h>
+#include <hilti/ast/type.h>
+#include <hilti/base/visitor-types.h>
+
+namespace hilti::scope {
+
+namespace detail {
+/** Internal backend to `hilti::lookupID()`. */
+extern std::pair<bool, Result<std::pair<NodeRef, ID>>> lookupID(const ID& id, const Node& n);
```
bbannier, line 21: No need to declare this `extern`.
```suggestion
std::pair<bool, Result<std::pair<NodeRef, ID>>> lookupID(const ID& id, const Node& n);
```

hilti/src/ast/scope-lookup.cc
```diff
@@ -0,0 +1,73 @@
...
+#include <hilti/base/logger.h>
+
+using namespace hilti;
```
bbannier, line 11: This is the style we use in many other places, but an alternative and slightly less ambiguous style would be to move all these definitions into an explicit namespace instead of relying on the compiler to figure out names matching declarations. Maybe something we could do since this is a new file.
```.cpp
namespace hilti {

namespace logging::debug { ... }

std::pair<bool, Result<std::pair<NodeRef, ID>>> scope::detail::lookupID(...

} // namespace hilti
```

hilti/include/ast/scope-lookup.h
```diff
@@ -34,8 +34,7 @@ extern std::pair<bool, Result<std::pair<NodeRef, ID>>> lookupID(const ID& id, co
  */
 template<typename D>
 Result<std::pair<NodeRef, ID>> lookupID(const ID& id, const visitor::Position<Node&>& p) {
-    auto i = p.path.rbegin();
-    while ( i != p.path.rend() ) {
+    for ( auto i = p.path.rbegin(); i != p.path.rend(); i++ ) {
```
bbannier, line 37: While this is a likely cheap `std::vector` iterator, we have iterators in hilti-rt which are not that cheap to construct, so the preferred pattern if possible should always be precrement.
```suggestion
    for ( auto i = p.path.rbegin(); i != p.path.rend(); ++i ) {
```

hilti/include/ast/scope-lookup.h
```diff
@@ -58,15 +57,32 @@ Result<std::pair<NodeRef, ID>> lookupID(const ID& id, const visitor::Position<No
...
+
         if ( auto t = (*i)->tryAs<Type>(); t && t->hasFlag(type::Flag::NoInheritScope) ) {
+            if ( auto x = i; ++x != p.path.rend() && (*x)->tryAs<declaration::Type>() )
+                // Ignore, we'll cover this in next round in the case below.
+                continue;
+
+            skip_to_module = true;
+        }
+
+        else if ( auto t = (*i)->tryAs<declaration::Type>(); t && t->type().hasFlag(type::Flag::NoInheritScope) )
```
bbannier, line 75:
```suggestion
        }
        else if ( auto t = (*i)->tryAs<declaration::Type>(); t && t->type().hasFlag(type::Flag::NoInheritScope) )
```

hilti/include/ast/scope-lookup.h
```diff
@@ -58,15 +57,32 @@ Result<std::pair<NodeRef, ID>> lookupID(const ID& id, const visitor::Position<No
             // Pass back error.
             return std::move(resolved);
 
+        // If the type has the NoInheritScope flag, we skip everything else
+        // in remainder of the path except for the top-level module, to which
+        // we then jump directly. One exception: If the type is part of a
+        // type declaration, we need to check the declaration's scope still
+        // as well.
```
bbannier, line 64: I am not sure I understand the last sentence here; were does this happen in the code below?

hilti/src/compiler/cxx/elements.cc
```diff
@@ -436,9 +436,18 @@ std::string cxx::type::Struct::inlineCode() const {
...
-    // Create default constructor.
-    inline_code += fmt("inline %s::%s() {\n%s%s%s}\n\n", type_name, type_name, init_parameters(), init_locals_user(),
-                       init_locals_non_user());
+    // Create default constructor. This initializes user-controlled members
+    // only if there are no struct parameters. If there are, we wouldn't have
+    // access to them here yet and hence some init expressions might not
+    // evaluate. However, in that case we will run through the
+    // parameter-based constructors normally anyways, so don't need this
+    // here.
+    if ( args.size() != 0 )
```
bbannier, line 445:
```suggestion
    if ( ! args.empty() )
```

hilti/src/ast/scope-lookup.cc
rsmmr: I'd be fine switching to that style, but not sure I like doing that incrementally just for a individual files, that would just make things inconsistent. One wouldn't even know if for a specific file, header inference works or not. So staying with current style until we switch over.

hilti/src/ast/scope-lookup.cc
rsmmr: similar argument here: we can discuss changing style, but I prefer consistency over changing just individual files, so leaving as is.

hilti/src/compiler/cxx/elements.cc
rsmmr: adding another note to the comment, it's what the if/else does

hilti/include/ast/scope-lookup.h
rsmmr: adding another note to the comment, it's what the if/else does

PR #372, opened 2020-06-09
Make stream-related iterators safe by default

### LINKED ISSUE #54: Iterators aren't safe yet
The goal is for HILTI's iterators to be safe in the sense that accesses to expired content gets caught. This works currently already for containers that go out of scope. For example, for a vector `v`, `begin(v)` returns an iterator that will become invalid once `v` gets destroyed. However, for mutable containers, we don't catch the case yet that changes to the container's content may invalidate iterators due to internal data structure updates. An access to such an iterator needs to throw an exception, under a general policy of "container modifications invalidate existing iterators".

bbannier:
This closes #54.

### DIFF
--- a/hilti/include/rt/types/stream.h
+++ b/hilti/include/rt/types/stream.h
@@ -941,10 +941,6 @@ inline std::string detail::to_string_for_print<stream::View>(const stream::View&
 }
 
 namespace detail::adl {
-inline auto safe_begin(const Stream& x, adl::tag /*unused*/) { return x.safeBegin(); }
-inline auto safe_end(const Stream& x, adl::tag /*unused*/) { return x.safeEnd(); }
-inline auto safe_begin(const stream::View& x, adl::tag /*unused*/) { return x.safeBegin(); }
-inline auto safe_end(const stream::View& x, adl::tag /*unused*/) { return x.safeEnd(); }
 inline std::string to_string(const Stream& x, adl::tag /*unused*/) {
     return fmt("b\"%s\"", escapeBytes(x.data(), true));
 }
--- a/hilti/include/rt/types/stream.h
+++ b/hilti/include/rt/types/stream.h
@@ -613,7 +613,7 @@ public:
      *
      * @param i the number of stream to advance.
      */
-    View advance(integer::safe<uint64_t> i) const { return View(safeBegin() + i, _end); }
+    View advance(integer::safe<uint64_t> i) const { return View(begin() + i, _end); }
 
     /**
      * Extracts a subrange of bytes from the view, returned as a new view.
@@ -629,7 +629,7 @@ public:
      *
      * @param to iterator pointing to just beyond subrange
      */
-    View sub(SafeConstIterator to) const { return View(safeBegin(), std::move(to)); }
+    View sub(SafeConstIterator to) const { return View(begin(), std::move(to)); }
 
     /**
      * Extracts subrange of bytes from the view, returned as a new view.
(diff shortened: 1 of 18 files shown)


hilti/include/rt/types/stream.h
```diff
@@ -683,18 +690,19 @@ class View {
...
         return detail::UnsafeConstIterator(_begin);
     }
 
-    detail::UnsafeConstIterator end() const { return detail::UnsafeConstIterator(safeEnd()); }
+    detail::UnsafeConstIterator unsafeEnd() const { return detail::UnsafeConstIterator(safeEnd()); }
+
+    auto begin() const { return safeBegin(); }
+    auto end() const { return safeEnd(); }
 
     const SafeConstIterator& safeBegin() const { return _begin; }
```
rsmmr, line 703: Do we want just want to remove these and just use begin/end for the safe versions?

hilti/include/rt/types/stream.h
```diff
@@ -565,15 +572,15 @@ class View {
     bool isOpenEnded() const { return ! _end.has_value(); }
```
rsmmr, line 572: How about renaming this to `isExpandable()` now, that seems more intuitive.

hilti/src/rt/tests/stream.cc
```diff
@@ -16,169 +16,471 @@ using hilti::rt::to_string;
...
+
+    SUBCASE("difference") {
+        const auto [s, before_begin] = []() {
+            auto s = Stream(" 123"_b);
+            const auto before_begin = s.begin();
+
+            s.trim(before_begin + 1);
+            REQUIRE_EQ(s, "123"_b);
+
+            return std::make_tuple(s, before_begin);
```
rsmmr, line 348: Isn''t `before_begin` just becoming an expired iterator here because `s` won't survive the return?

bbannier: Good catch! I added a `REQUIRE_FALSE(before_begin.isExpired())` which failed with the code in this form, but succeeds if we `std::move(s)` into the returned tuple.

hilti/include/rt/types/stream.h
bbannier: Hmm, personally `isOpenEnded` seems clearer to me as it a pretty clear property of some range/window to me. `isExpandable` sounds like one could do something with the `View` to make it expandable, but it is really a fixed property.

hilti/src/rt/tests/stream.cc
rsmmr: ok, nevermind then

hilti/include/rt/types/stream.h
```diff
@@ -223,14 +217,18 @@ class SafeConstIterator {
...
+                e = {_content, c->offset() + c->size(), _chunk};
+            else {
+                auto l = c->last();
+                assert(l);
+                assert(l->isLast());
+                e = {_content, l->offset() + l->size(), l};
+            }
         }
+        else
+            cannot_be_reached();
```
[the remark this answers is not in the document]
bbannier: As far as I can tell, these truely cannot be reached (checked by inspection and confirmed by coverage in the surrounding code). I would suggest we leave this as is, and revisit should we ever run into an issue. Previously these cases potentially triggered segfaults, but now they should generate much easier to understand aborts from `cannot_be_reached`.

Resolving for now.

bbannier:
> I would suggest one additional set of tests that exercises methods with a more complex stream instance containing multiple chunks, incl. some with SBO. You could add a helper function that creates such an instance and then reuse that across the tests to see if the various operations produce the expected result on it it as well.

I did that and augmented a number of tests with checks of multichunk streams. What does SBO refer to?

PR #382, opened 2020-06-18
Have the code generator fully declare enum types directly as part of …

### LINKED ISSUE #381: Consider emitting definitions for enums ASAP instead of forward declaring them
We currently forward declare enums and define them at a later location roughly matching the declaration ordering in the original Spicy source file. While this works well enough for uses involving opaque enum values, it breaks down when one declares e.g., `vector`s of enums since they rely on the implicit `Undef` value to set a element default. Inside a module this can be fixed by making sure the enum is declared before any of its users (or users of its users and so on).

We should consider declaring enums whenever we currently forward declare them.

We also should make sure that this works across module boundaries, e.g., a file importing an enum definition from another module should also be able to declare types like `vector` of that enum. It seems this case cannot be fixed by reordering declarations.

* * *
NB. this would fix warnings due to the way we currently forward-declare enums. Forward-decls of the form
```.cpp
using E = enum class E: int64_t;
```
trigger a clang-11 warning of the form `Non-definining declaration of enumeration with a fixed underlying type is only permitted as a standalone declaration; missing list of enumerators? [-Welaborated-enum-base]`. Instead one should forward decl enums with `no_using` so we generate
```.cpp
enum class E: int64_t;
```

rsmmr:
…the forward-declarations block.

In addition, we longer use "using" to declare enum types because
that's apparently not standard compliant.

Closes #381.

### DIFF
--- a/hilti/src/compiler/codegen/types.cc
+++ b/hilti/src/compiler/codegen/types.cc
@@ -261,24 +261,13 @@ struct VisitorDeclaration : hilti::visitor::PreOrder<cxx::declaration::Type, Vis
         if ( sid.namespace_() )
             scope = scope.namespace_();
 
+        // We declare the full enum type as part of the forward declarations block, that makes sure it's always fully
+        // available. This is e.g., needed so we can set default values for vectors of enums.
         auto id = cxx::ID(scope, sid);
-
-        // Also add a forward declaration.
-        auto type_forward = cxx::declaration::Type{
-            .id = id,
-            .type = fmt("enum class %s : int64_t", id.local()),
-            .forward_decl = true,
-            .forward_decl_prio = true,
-        };
-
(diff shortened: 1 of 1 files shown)


hilti/src/compiler/codegen/types.cc
```diff
@@ -261,24 +261,13 @@ struct VisitorDeclaration : hilti::visitor::PreOrder<cxx::declaration::Type, Vis
         if ( sid.namespace_() )
             scope = scope.namespace_();
 
+        // We declare the full enum type as part of the forward declarations block, that makes sure it's always fully
+        // available.
```
bbannier, line 265:
```suggestion
        // available. This is e.g., needed so we can set default values for vectors of enums.
```

PR #384, opened 2020-06-18
Fixes and tests for hilti::rt::util

bbannier:


hilti/src/rt/tests/util.cc
```diff
@@ -95,6 +95,36 @@ TEST_CASE("enumerate") {
...
+        SUBCASE("true") {
+            escape_quotes = true;
+            quote = "\\\"";
+        }
+        SUBCASE("false") {
+            escape_quotes = false;
+            quote = "\"";
+        }
+
+        CHECK_EQ(escapeBytes("", escape_quotes), "");
```
rsmmr, line 181: Could use `R"(...)"` for better readability in this test.

zeek/compiler/glue-compiler.cc
```diff
@@ -124,9 +130,9 @@ static int extract_int(const std::string& chunk, size_t* i) {
     auto x = chunk.substr(*i, j - *i);
     *i = j;
 
-    int integer;
-    hilti::util::atoi_n(sint.begin(), sint.end(), 10, &integer);
-    return integer;
+    int integer = 0;
+    hilti::util::atoi_n(x.begin(), x.end(), 10, &integer);
```
rsmmr, line 134: In the case where we got a sign, `i` is still pointing to that, right? And `atoi()` is examining the sign (at least the `-`) so aren't we now applying it twice?

hilti/include/rt/util.h
```diff
@@ -359,49 +360,51 @@ class OutOfRange;
...
-    Result n = 0;
+    if ( s == e )
+        throw Exception("cannot decode from empty range");
+
+    std::optional<Result> n = std::nullopt;
     bool neg = false;
+    auto it = s;
 
-    if ( s != e && *s == '-' ) {
+    if ( *it == '-' ) {
```
rsmmr, line 378: Let's add another case for `+` here to avoid surprises.

hilti/include/rt/util.h
```diff
@@ -359,49 +360,51 @@ class OutOfRange;
...
+ * @pre Base must be in the inclusive range [2, 36].
  */
 template<class Iter, typename Result>
 inline Iter atoi_n(Iter s, Iter e, int base, Result* result) {
     if ( base < 2 || base > 36 )
         throw OutOfRange("base for numerical conversion must be between 2 and 36");
 
-    Result n = 0;
+    if ( s == e )
+        throw Exception("cannot decode from empty range");
```
rsmmr, line 372: Throw `InvalidArgument`

hilti/include/rt/util.h
```diff
@@ -359,49 +360,51 @@ class OutOfRange;
 /**
  * Parses a numerical value from a character sequence into an integer.
  * `base` must be between 2 and 26.
+ * @pre The input sequence must not be empty, i.e., we require `s != e`.
+ * @pre Base must be in the inclusive range [2, 36].
  */
```
rsmmr, line 365: let's add documentation for parameters and result.

hilti/include/rt/util.h
```diff
@@ -359,49 +360,51 @@ class OutOfRange;
...
 template<class Iter, typename Result>
 inline Iter atoi_n(Iter s, Iter e, int base, Result* result) {
```
rsmmr, line 367: To follow up on our discussion, let's leave this as is. Returning a pair with an optional result would be the only alternative I see and that feels quite cumbersome to work with.

hilti/src/rt/util.cc
```diff
@@ -100,8 +100,13 @@ std::vector<std::string_view> hilti::rt::split(std::string_view s, std::string_v
     if ( delim.empty() )
         return {s};
 
+    if ( s.size() < delim.size() )
+        return {s};
+
     std::vector<std::string_view> l;
 
+    const bool ends_in_delim = s.substr(s.size() - delim.size()) == delim;
```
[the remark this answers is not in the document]
bbannier: Yeah, this seems to work, apparently mostly due to the behavior of `std::string::find`. Looking at https://en.cppreference.com/w/cpp/string/basic_string/find I do not see anything specified for the empty delim case, but we have a test case to catch when this should ever become an issue.

hilti/src/rt/tests/util.cc
```diff
@@ -25,10 +32,281 @@ ostream& operator<<(ostream& stream, const vector<string_view>& xs) {
...
+    CHECK_EQ(pow(2, 5), 32);
+    CHECK_EQ(pow(2, 16), 65536);
+
+    CHECK_EQ(pow(integer::safe<int8_t>(2), 3), 8);
+    CHECK_THROWS_WITH_AS(pow(integer::safe<int8_t>(2), 4), "integer overflow", const Overflow&);
+    CHECK_EQ(pow(integer::safe<int16_t>(2), 4), 16);
+    CHECK_EQ(pow(integer::safe<int16_t>(2), integer::safe<int16_t>(4)), 16);
+}
+
+TEST_CASE("normalizePath") {
```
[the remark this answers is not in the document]
bbannier: I added a test case for this so behavior and also added a `TODO`. I am unsure whether it is at all possible to normalize e.g., paths which cannot be read (think of paths involving `/../` components, not listable subdirectories and symlinks).

hilti/src/rt/tests/util.cc
bbannier: I now use it in a couple places where it helps.

zeek/compiler/glue-compiler.cc
bbannier: Good catch, fixed.

rsmmr:
Looks good!

PR #390, opened 2020-06-22
 Build and host Doxygen documentation

### LINKED ISSUE #388: Build and host Doxygen documentation

bbannier:
Closes #388.


doc/Doxyfile
```diff
@@ -0,0 +1,2553 @@
...
+
+DOXYFILE_ENCODING      = UTF-8
+
+# The PROJECT_NAME tag is a single word (or a sequence of words surrounded by
+# double-quotes, unless you are using Doxywizard) that should identify the
+# project for which the documentation is generated. This name is used in the
+# title of most generated pages and in a few other places.
+# The default value is: My Project.
+
+PROJECT_NAME           = "My Project"
```
rsmmr, line 35:
```suggestion
PROJECT_NAME           = "Spicy"
```

doc/Doxyfile
```diff
@@ -0,0 +1,2553 @@
...
+# could be handy for archiving the generated documentation or if some version
+# control system is used.
+
+PROJECT_NUMBER         =
+
+# Using the PROJECT_BRIEF tag one can provide an optional one line description
+# for a project that appears at the top of each page and should give viewer a
+# quick idea about the purpose of the project. Keep the description short.
+
+PROJECT_BRIEF          =
```
rsmmr, line 47: Can use this to link back to the main documentation?

doc/Doxyfile
```diff
@@ -0,0 +1,2553 @@
...
+
+# With the PROJECT_LOGO tag one can specify a logo or an icon that is included
+# in the documentation. The maximum height of the logo should not exceed 55
+# pixels and the maximum width should not exceed 200 pixels. Doxygen will copy
+# the logo to the output directory.
+
+PROJECT_LOGO           =
```
rsmmr, line 54: Would be nice to show the logo.

doc/Makefile
```diff
@@ -14,7 +14,7 @@ all: html
 clean:
 	rm -rf $(DESTDIR)
 
-html: Makefile autogen-docs
+html: Makefile autogen-docs doxygen
```
rsmmr, line 17: I'm thinking we should skip building this by default. The output is really noisy and may bury stuff from Sphinx that we are interested in. Let people type "make doxygen" if they want it, we can mention that in the installation instructions (and let's mention the Doxygegen dependency there then as well).

On CI we can always kick off, but maybe we can still suppress some of the warnings (e.g, "method has no documentation") generally?

doc/development/doxygen.rst
```diff
@@ -0,0 +1,6 @@
+.. _doxygen:
+
+C++ API documentation
+=====================
+
+The generated C++ API documentation can be found `here </doxygen/html/index.html>`__.
```
rsmmr, line 6: This link doesn't work for me on my local file system, it turns into: `file:///doxygen/html/index.html` in my browser.

bbannier: Yepp, the way I linked to the site root here always needed a server. I now copy the files into the generate site with Sphinx (which might help with readthedocs as well), and used a relative link which should work both with and without a server.

doc/Doxyfile
bbannier: It seems as if (rightly so) sanitizes this so we cannot e.g., emit raw HTML here. I cannot think of a way to achieve this, so users would need to manipulate the URL to go back to the root of the site (this would just involve removal of path components though).

Dropping this, please reopen if you have an idea how to do this.

PR #394, opened 2020-06-23
Add "&requires" attribute

rsmmr:


spicy/src/compiler/codegen/parser-builder.cc
```diff
@@ -1045,6 +1045,17 @@ void ParserBuilder::newValueForField(const type::unit::item::Field& field, const
...
 
+    if ( auto a = AttributeSet::find(field.attributes(), "&requires") ) {
+        // We evaluate "&requires" here so that the field's value has been
+        // set already, and is hence accessible to the condition through
+        // "self.<x>".
+        auto block = builder()->addBlock();
+        block->addLocal(ID("__dd"), field.parseType(), value);
+        auto cond = block->addTmp("requires", *a->valueAs<Expression>());
+        pushBuilder(block->addIf(builder::not_(cond)),
+                    [&]() { parseError("&required failed ($$ == %s)", {value}, a->value().location()); });
```
bbannier, line 1056: We do not check any of the data dependencies of `cond` here and would trigger an `UnsetOptional` exception at runtime here if the condition accesses data which has not (yet) filled, e.g., in

```
type X = unit {
    a: uint8 &requires=(self.a > self.b); # self.b is always unset here.
    b: uint8;
};
```

While this makes sense, it can be nasty when e.g., reordering fields, when composing messages, or when making fields conditional with e.g., `switch`.

If we cannot validate that we only access `$$` and the final value of exactly this field we should at least add a test to ensure we trigger an exception.

doc/programming/parsing.rst
```diff
@@ -305,6 +305,42 @@ representation as a ``uint64``:
     :exec: printf 12345 | spicy-driver %INPUT
     :show-with: foo.spicy
 
+.. _attribute_requires:
+
+.. rubric:: Enforcing Expected Field Values
```
bbannier, line 310: I wonder whether we should also allow `&requires` attributes on units to enforce global unit constraints. The way we handle it eagerly on fields means that constraints can only use data of fields up to and including the field having the `&requires` attribute, so to check a global unit constraint one would now need to e.g., add it to the last field.

I.e., instead of forcing users to write this
```
type X = unit {
    a: uint8;
    b: uint8;
    c: uint8 &requires=(self.c*self.c == self.a*self.a + self.b*self.b);
};
```
we could allow this form which should be less error-prone.

```
type X = unit {
    &requires=(self.c*self.c == self.a*self.a + self.b*self.b); // Maybe even allow this in an arbitrary position?
    
    a: uint8;
    b: uint8;
    c: uint8;
};
```

spicy/src/compiler/codegen/parser-builder.cc
bbannier: Maybe an overall better approach would be to evaluate these conditions only after all unit fields have been parsed. That would not only make the data dependency issue go away, but also allow users to check preconditions where they make sense (e.g., check that an optional field is only set if some other flag field has a certain value). A unit-level `&requires` could still make sense in that world to enforce even higher level constraints.

rsmmr: The semantics here are similar to other field attributes (e.g., `&convert`), and I think it's also the right time to evaluate the constraint so that subsequent fields can rely on condition holding. 

I like the idea of a global `&requires`, I'll add that.

rsmmr:
@bbannier take another look

spicy/src/compiler/codegen/parser-builder.cc
```diff
@@ -1154,6 +1165,16 @@ void ParserBuilder::initializeUnit(const Location& l) {
 void ParserBuilder::finalizeUnit(bool success, const Location& l) {
     const auto& unit = state().unit.get();
 
+    if ( success ) {
+        // We evaluate any "%requires" before running the final "%done" hook
+        // so that (1) that one can rely on the condition, and (2) we keep
+        // running either "%done" or "%error".
+        for ( auto cond : unit.propertyItems("%requires") ) {
+            pushBuilder(builder()->addIf(builder::not_(*cond.expression())),
+                        [&]() { parseError("%requires failed", cond.meta()); });
```
bbannier, line 1174: This will generate a parser error as soon as any constraint fails (i.e., later constraints will not be validated and not emit additional error messages. WDYT about instead aggregating all failed constraints and emitting the error only after all have been checked. I believe this could be a useful debugging tool, but could have other interactions when e.g., a constraint builds on a previous one (e.g., first check fields are set, then check their values).

tests/spicy/types/unit/requires-attribute.spicy
```diff
@@ -0,0 +1,19 @@
...
+
+type E = enum { X = 1, Y = 2 };
+
+public type Foo = unit {
+    x: uint8 &requires=($$ == 1);
+    y: uint8 &convert=E($$) &requires=($$ == 1);
+    z: uint8 &convert=E($$) &requires=(self.z == E::X);
+
+    on %done { print self; }
+    on %error { print "ErRrOr"; }
```
bbannier, line 18: We could print `self` here to capture the semantics of which fields have been parsed when the error is triggered.

spicy/src/compiler/codegen/parser-builder.cc
rsmmr: Would that difference actually be observable? It would fail with a single "parse error" message either way, we can report that only once. If an expression has a side effect, then it would be noticeable but I'd argue it shouldn't. :)

tests/spicy/types/unit/requires-attribute.spicy
rsmmr: will do

spicy/src/compiler/codegen/parser-builder.cc
bbannier: > Would that difference actually be observable?
Yes, if the expectation is that the order in which `requires` attributes are declared has significance, my previous example:
```
type X = unit {
    %requires=(self.a? && self.b);     # (1)
    %requires=(self.a + self.b < 42); # (2)
    a: int8 &requires=($$ > 0); # Using field constraints here to make these fields potentially
    b: int8 &requires=($$ > 0); # unset, but similar possible with `switch`.
};
```
The way the code is written suggests that `(1)` would be checked before `(2)` (like it does), so the side effect of `(1)` is that `(2)` can only trigger a parse error, but no exception from accessing an unset optional.

I think with that example I have convinced myself why we should stop checking global constraints as soon as a single one fails 😄

PR #397, opened 2020-06-23
Add a DHCP analyzer.

bbannier:


zeek/plugin/lib/protocols/zeek_dhcp.spicy
```diff
@@ -0,0 +1,230 @@
...
+            rebind_time_value.push_back(cast<interval>(option.rebind_time_value));
+
+        if (option?.class_identifier)
+            class_identifier.push_back(option.class_identifier);
+
+        if (option?.user_class)
+            user_class.push_back(option.user_class.decode(hilti::Charset::ASCII));
+    }
+
+    return (
```
rsmmr, line 169: On this Zeek side, is this setting all the `&optional` fields? For example, even if no `routers` are specified, the record field wouldl be set to an empty vector (vs being left unset), right? I just checked, we actually do have the `optional<T>` type  in Spicy already, there are tests for it. If the tuple returned all the values as wrapped in `optional`, it could leave records fields unset where we don't have seen an option it all. But it's ok to leave as it is, seems fine for now.

bbannier: Yes, while this does not fail, it produces incorrect results: when accessing an invalid index, `rt::Vector` will add default-initialized elements to make the access legal; this is definitely not what we want here.

As discussed offline we should leverage the Zeek plugin's support for optional values on both the Spicy and Zeek side for this and use `optional` values in the helper functions added here. I fixed the signatures, but the handling of unset optional fields is currently broken in the plugin, see #401. I pushed updates which at least fix the signatures for now.

bbannier: Filed #415 for other issues and adjusted the code accordingly.

PR #399, opened 2020-06-23
Remove duplicate spicy-driver CMake target.

### LINKED ISSUE #398: Target `spicy-driver` defined multiple times
We currently define the executable target `spicy-driver` twice in `CMakeLists.txt`. This seems to be an issue e.g., when building with `HILTI_USE_JIT` defined.
```
CMake Error at spicy/CMakeLists.txt:192 (add_executable):
  add_executable cannot create target "spicy-driver" because another target
  with the same name already exists.  The existing target is an executable
  created in source directory "/build/spicy/spicy".  See documentation for
  policy CMP0002 for more details.
```

We should probably just remove the conditional target under `HILTI_USE_JIT`.

bbannier:
This would have triggered an error when building with `HILTI_USE_JIT`.

Closes #398.


bbannier:
There are still a few paths left which can set that variable. We should check whether they are needed or if we can clean them up separately.

rsmmr:
I'll go ahead and pull this in.

PR #403, opened 2020-06-24
Fix hilti::rt::is_optional so it applies to const and ref of optionals also.

bbannier:


hilti/include/rt/util.h
```diff
@@ -520,10 +520,16 @@ struct is_tuple<std::tuple<T...>> : std::true_type {};
...
+/** Checks if a type is an optional. */
 template<typename... T>
 struct is_optional<std::optional<T...>> : std::true_type {};
 
+template<typename... T>
+struct is_optional<const std::optional<T...>> : std::true_type {};
+
+template<typename... T>
+struct is_optional<const std::optional<T...>&> : std::true_type {};
+
```
bbannier, line 532: I went with manual specializations since we do not use `volatile` in Spicy and the number of permutations is still somewhat manageable. Another approach would be to make `is_optional` a templated constexpr function which can then return different types based on the result of applying `std::remove_cv` and `std::remove_reference`.

duffy-corelight:
In hilti/src/rt/tests/util.cc are lines 313 314 unintentionally identical?
```
static_assert(is_optional<const std::optional<int>>::value); 
static_assert(is_optional<const std::optional<int>>::value);
```

bbannier:
> In hilti/src/rt/tests/util.cc are lines 313 314 unintentionally identical?
> 
> ```
> static_assert(is_optional<const std::optional<int>>::value); 
> static_assert(is_optional<const std::optional<int>>::value);
> ```

Thanks for having a look @duffy-corelight. That's an error from an overeager copy and paste, I removed the duplicate.

PR #407, opened 2020-06-25
Decouple vector element access and resizing

bbannier:


tests/hilti/types/vector/ops.hlt
```diff
@@ -42,8 +42,8 @@ p([1,2,3]);
 
 local vector<real> b;
 
-b[1] = 0;
-b[2] = 42;
+b.assign(1, 0);
```
rsmmr, line 45: Do we have any tests still for assigning to an existing element through the index operator? If not, please add one.

bbannier: I added a C++ unit test for that.

PR #414, opened 2020-06-26
Fix `spicy_plugin` function not called when building with static libs.

Jason--n:
Add `-all_load` compiler option for clang and `--whole-archive` for gcc.

Have tested with clang compiler.

### DIFF
--- a/spicy/CMakeLists.txt
+++ b/spicy/CMakeLists.txt
@@ -193,6 +193,23 @@ if ( HILTI_USE_JIT )
     target_link_libraries(spicy-driver PRIVATE spicy)
 endif ()
 
+set(SPICY_EXEUTABLES "spicyc" "spicy-config" "spicy-driver" "spicy-doc")
+
+# The Compiler would not include static variable located at plugin.cc in executable
+# if we do not tell the compiler. We can't choose which symbol to inlcude so we have to
+# include them all.
+
+if ( (NOT DEFINED BUILD_SHARED_LIBS) OR (NOT BUILD_SHARED_LIBS) )
+    if (${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
+        foreach (EXE ${SPICY_EXEUTABLES})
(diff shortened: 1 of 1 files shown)


bbannier:
Since we do not have dedicated setups for this in our CI matrix, I built Spicy with `-DBUILD_SHARED_LIBS=OFF` and ran the BTest suite in `tests/`. This produces a number of failures. See https://docs.zeek.org/projects/spicy/en/latest/development/testing.html#btest for some pointers on how to use BTest.

That being said, the approach you take here is hidden from anybody trying to link against our libraries and they would need to remember to use these flags. A way which could be surfaced in `spicy-config --ldflags` would be much easier to use. @rsmmr suggested using `-Wl,-u<symbol_name>` to manually expose the needed symbols (this should also be added to e.g., what `spicy-config` displays). Could you look into that?

spicy/CMakeLists.txt
```diff
@@ -193,6 +193,23 @@ if ( HILTI_USE_JIT )
     target_link_libraries(spicy-driver PRIVATE spicy)
 endif ()
 
+set(SPICY_EXEUTABLES "spicyc" "spicy-config" "spicy-driver" "spicy-doc")
```
bbannier, line 196: This misses a number of executable (e.g., `spicyz`), but like I wrote before, manually exposing these symbols via general linker flags would be preferrable.

Jason--n:
Ok, I will. I think that I should be more familiar with the codebase and tools of spicy, if I want to solve this problem in a more suitable approach.

rsmmr:
@Jason--n Have you had a chance to look further into this? We can give it a try to as well if you prefer.

Jason--n:
@rsmmr  I have put aside this issue for a while because I'm getting familiar with the use and code of spicy. If you have time and think it's necessary to solve this issue, that would be great.

rsmmr:
Superseded by #430 430

PR #417, opened 2020-06-29
Remove hilti::rt::List in favor of hilti::rt::Vector

bbannier:


hilti/src/rt/tests/vector.cc
```diff
@@ -85,6 +85,37 @@ TEST_CASE("assignment") {
...
+    SUBCASE("allocator change") {
+        auto xs = Vector<int, vector::Allocator<int, 5>>();
+        xs.assign(2, 5);
+        REQUIRE_EQ(to_string(xs), "[5, 5, 5]");
+
+        auto ys = Vector<int, vector::Allocator<int, 3>>();
+        ys.assign(2, 3);
+        REQUIRE_EQ(to_string(ys), "[3, 3, 3]");
+
+        ys = xs;
```
rsmmr, line 98: I'm wondering if this should be legal. In HILTI/Spicy, a different default value can happen only if it's a different element type  for the vector (because the default value is determined by the element type). So I'd be inclined to enforce that to catch codegen errors.

bbannier: I think I needed this because list comprehension via `list::make` (now `vector::make`) used a different allocator than `Vector`s constructed elsewhere. If the default value depends only on the type we might be able to make these conversion between vectors with different allocators impossible. I'll have a look.

bbannier: I was able to remove this conversion behavior by taking the target allocator type when calling `vector::make`.

PR #420, opened 2020-06-30
Fix handling of vector of bools

### LINKED ISSUE #392: Cannot access elements of vector<bool>
As of 16c19edf264616d83c0e973d849cc584a52a70d0 we invoke calls to `std::vector<bool>::data` which does not exist since `std::vector<bool>` is a proxy class.

```$ cat /tmp/foo.spicy
module Foo;

local xs: vector<bool>;
xs[0];
```

```
$ spicyc /tmp/foo.spicy -j
In file included from Foo.cc:6:
In file included from /usr/local/Cellar/spicy/HEAD-ff5b3c2/include/hilti/rt/libhilti.h:15:
In file included from /usr/local/Cellar/spicy/HEAD-ff5b3c2/include/hilti/rt/types/all.h:5:
In file included from /usr/local/Cellar/spicy/HEAD-ff5b3c2/include/hilti/rt/types/address.h:9:
In file included from /usr/local/Cellar/spicy/HEAD-ff5b3c2/include/hilti/rt/types/bytes.h:12:
In file included from /usr/local/Cellar/spicy/HEAD-ff5b3c2/include/hilti/rt/types/regexp.h:9:
In file included from /usr/local/Cellar/spicy/HEAD-ff5b3c2/include/hilti/rt/types/stream.h:21:
/usr/local/Cellar/spicy/HEAD-ff5b3c2/include/hilti/rt/types/vector.h:328:19: error: no member named 'data' in 'std::__1::vector<bool, std::__1::allocator<bool> >'
        return V::data()[i];
                  ^
Foo.cc:20:7: note: in instantiation of member function 'hilti::rt::Vector<bool, std::__1::allocator<bool> >::operator[]' requested here
    xs[hilti::rt::integer::safe<std::uint64_t>{0u}];
      ^
1 error generated.
```

The calls to `std::vector::data` originate in `operator[]` implementations. We need to check how we can support these for vectors of bool, in particular for assignment for vector elements.

bbannier:
Closes #392.


hilti/include/ast/ctors/bool.h
```diff
@@ -4,14 +4,15 @@
...
+#include <hilti/rt/types/bool.h>
 
 namespace hilti {
 namespace ctor {
 
 /** AST node for a boolean constructor. */
 class Bool : public NodeBase, public hilti::trait::isCtor {
 public:
-    Bool(bool v, Meta m = Meta()) : NodeBase(std::move(m)), _value(v) {}
+    Bool(hilti::rt::Bool v, Meta m = Meta()) : NodeBase(std::move(m)), _value(v) {}
```
rsmmr, line 15: I think I'd leave this as `bool` for the AST side. Don't much of an advantage of switching to `Bool` there, but reduces readability.

hilti/include/ast/node.h
```diff
@@ -33,12 +34,14 @@ namespace node {
 namespace detail {
 
 /** Value of a node property, stored as part of `Properties`. */
-using PropertyValue = std::variant<bool, const char*, double, int, int64_t, unsigned int, uint64_t, std::string>;
+using PropertyValue =
+    std::variant<bool, const char*, double, int, int64_t, unsigned int, uint64_t, std::string, hilti::rt::Bool>;
```
rsmmr, line 38: Same here.

hilti/include/compiler/detail/parser/driver.h
```diff
@@ -34,7 +34,7 @@
 
 /** Bison value type. */
 struct yystype_hilti {
-    bool bool_ = false;
+    hilti::rt::Bool bool_ = false;
```
rsmmr, line 37: ... and here.

hilti/include/rt/types/bool.h
```diff
@@ -6,8 +6,20 @@
 
 namespace hilti::rt {
 
+class Bool {
```
rsmmr, line 9: Just for my C++ education: I suppose there's no strong typedef for builtin types, so that we wouldn't need to create a class?

bbannier: AFAIK as of C++17 there is no strong typedef feature in the core language. Since we are wrapping an integral type we could use the following though (which I missed in an almost mechanical way):
```.cpp
enum Bool : bool {
  True = true,
  False = false,
};
```

That would not only fix the issue with `std::vector<bool>`, but as an added benefit introduce literals `True` and `False` (we could make them scoped by using an `enum class`). I'll do that instead.

bbannier: As I suspected, if we'd use an enum type, we'd need to manually convert `Bool` to `bool` when used e.g., in conditionals as conversion functions need to be member functions which we cannot add to an enum. I'd suggest we use the class form for now.

PR #430, opened 2020-07-08
Fix build with static libraries.

### LINKED ISSUE #400: `Spicyc` doesn't support `.spicy` file when built with static libs
Compiler : LLVM 10 on MacOS、GCC 10 on  CentOS 8
Build Type: Debug

I tried two platform mentioned above，and it ended up with the same result. When built with static libs, the `Spicyc` binary can't compile file with extention `.spicy` and shows an error "unsupported file type".

I used gdb to debug this binary, and found out that the static variable  in `spicy/src/compiler/plugin.cc` which should be initialized with function `spicy_plugin` did not be initialized. There was no symbol `spicy_plugin` in the binary.

When built with dynamic libs, `Spicyc` **can** compile `.spicy` file.

rsmmr:
The binaries weren't pulling in the Spicy plugin because the linker
didn't see any of its symbols referenced and would just strip it all
out, including the global constructor. Even when I forced my way
around that, the same problem would show up in other ways: other
constructors now missing, and runtime functions not being included
because a binary would seemingly not make use of them.

I now fixed that in a similar way as we already do for the runtime
libraries: using object libraries instead of static libraries. That
seems the way to go with CMake. So if the build is configured for
static libraries, we now link our binaries against separately defined
object versions of libhilti and libspicy. That makes sure everything
gets included. We still build the actual static libraries as well for
use by external applications.

Test suite passes for me with both shared and static libraries on
macOS (with one exception that I need to track down still). CI will
show how it's looking on Linux. This commits changes the CI build for
clang9_zeek30_ubuntu_release to use static libraries.

Note: Different from what I thought at first, I don't think we
actually need to do anything about spicy-config output because that's
not refering to libhilti/spicy at all, just to the runtime libraries
(which are already static).

Closes #400.

### DIFF
--- a/hilti/src/compiler/jit.cc
+++ b/hilti/src/compiler/jit.cc
@@ -71,10 +71,7 @@ JIT::JIT(std::shared_ptr<Context> context) : _context(std::move(context)) {
     _jit = std::make_unique<detail::ClangJIT>(_context);
 }
 
-JIT::~JIT() {
-    if ( _jit )
-        finishRuntime();
-}
+JIT::~JIT() {}
 
 bool JIT::compile() {
     util::timing::Collector _("hilti/jit/compile");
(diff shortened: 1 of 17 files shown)


hilti/src/compiler/jit.cc
```diff
@@ -71,10 +71,7 @@ JIT::JIT(std::shared_ptr<Context> context) : _context(std::move(context)) {
     _jit = std::make_unique<detail::ClangJIT>(_context);
 }
 
-JIT::~JIT() {
-    if ( _jit )
-        finishRuntime();
-}
+JIT::~JIT() {}
```
bbannier, line 74: This does nothing now and isn't declared `virtual` either. I'd say let's just remove this custom destructor altogether.

PR #431, opened 2020-07-08
Extend test coverage of hilti-rt library

bbannier:


hilti/src/rt/tests/to_string.cc
```diff
@@ -43,6 +44,16 @@ TEST_CASE("safe-int") {
 
 TEST_CASE("string") { CHECK_EQ(to_string(std::string("abc")), "\"abc\""); }
 
+TEST_CASE("AddressFamily") {
+    CHECK_EQ(to_string(AddressFamily::IPv4), "AddressFamily::IPv4");
```
rsmmr, line 61: I think we should change the output  to skip the `AddressFamily::` prefix. I recently did a similar change for port's `Protocol`. Mind making that change?

PR #435, opened 2020-07-10
Additional hilti-rt tests

bbannier:


bbannier:
> Looks good. The integer flipping is hard to check, though,(as always), did you notice any cases where the old vs new implementation yield different results?

AFAICT the change here is only a refactoring. I haven't come across a case where the behavior changes.

PR #440, opened 2020-07-15
Fix some build failures from building against Zeek 3.2 dev versions

timwoj:
This PR makes some simple changes that fix problems with building against Zeek's master branch. https://github.com/zeek/zeek/pull/1066 is a corresponding PR on the Zeek side that fixes a number of related issues that are causing Spicy builds to fail.

The only issue that this does not fix is the usage of versions of `ConnectionEstablished` and `ConnectionFinished` that take `int` arguments in zeek/plugin/include/protocol-analyzer.h. These methods were changed in Zeek to take `bool` arguments, but deprecated versions of them were not added (because virtual methods aren't flagged as deprecated when overridden anwyays). A NEWS entry for Zeek indicates that this is a breaking change, and Spicy should be updated to use those when appropriate.

### DIFF
--- a/zeek/plugin/include/runtime-support.h
+++ b/zeek/plugin/include/runtime-support.h
@@ -21,6 +21,7 @@
 #include <Type.h>
 #include <Val.h>
 #include <Var.h>
+#include <IPAddr.h>
 #undef DEBUG
 
 #include "cookie.h"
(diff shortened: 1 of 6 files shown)


rsmmr:
Cool, thanks, I'll merge. I'll also put `#ifdefs` around the functions heads.

rsmmr:
Actually `ConnectionEstablished` isn't used, so fixing only `ConnectionFinished`.

PR #464, opened 2020-07-24
Port Spicy plugin to Zeek 3.2

### LINKED ISSUE #444: Port Zeek plugin to 3.2
Need to adapt to API deprecations and add to CI.

rsmmr:
This will close #444. 

This now seems to work with both 3.2 and 3.0. CI is still just doing 3.0/3.1. I'll leave it as that for now until the 3.2 binary packages come out (because that's what we're using).


rsmmr:
Just noticed that some of the HTTP changes got accidentally committed into an earlier commit ("Implement compatibility layer for Zeek < 3.2.", 8be9138). They HTTP changes are really all part of "Fix HTTP for Zeek 3.2." (3c6715)

zeek/plugin/include/protocol-analyzer.h
```diff
@@ -2,18 +2,17 @@
 
 #pragma once
 
+#include "cookie.h"
```
bbannier, line 5: Any reason we include this differently than `zeek-sicy/zeek-compat.h` below? I'd suggest to stick to a single style, e.g.,
```suggestion
#include <zeek-spicy/cookie.h>
```

zeek/plugin/include/runtime-support.h
```diff
@@ -6,31 +6,22 @@
 
 #pragma once
 
+#include "cookie.h"
```
bbannier, line 9: Any reason we include this differently than `zeek-sicy/zeek-compat.h` below? I'd suggest to stick to a single style, e.g.,
```suggestion
#include <zeek-spicy/cookie.h>
```

zeek/plugin/include/protocol-analyzer.h
rsmmr: Ack, fixing.

zeek/plugin/include/runtime-support.h
rsmmr: Ack, fixing.

rsmmr:
> Could you change one CI platform to use zeek-3.2? Right we do not seem to test this at all.

We only have Alpine that uses a non-binary Zeek package, and that's the weird one. I was thinking we can just wait until the 3.2 packages come out. If something breaks in the meantime, I'll fix it at that point; and until then it's all "inofficial" anyways.

Issue #468, opened 2020-07-30
Unit testing for the runtime libraries


rsmmr:
We're in the process of reviewing, and unit testing, the HILTI and Spicy runtime libraries. This ticket tracks which pieces are covered so far:

`hilti-rt-tests`:

- [x] hilti/backtrace.h (#475)
- [x] hilti/compiler-setup.h (probably neither worth nor really possible to test)
- [x] hilti/configuration.h (#475)
- [x] hilti/context.h (#478)
- [x] hilti/debug-logger.h (#475)
- [x] hilti/deferred-expression.h (#478)
- [x] hilti/exception.h (#478)
- [X] hilti/extension-points.h (tested in concrete implementations)
- [ ] hilti/fiber.h (staying with old tests for now; may swap out fiber implementation later and may revisit tests then as well)
- [x] hilti/fmt.h (#480)
- [x] hilti/global-state.h (#480)
- [x] hilti/hilti.h (#480 )
- [x] hilti/init.h (#480)
- [x] hilti/iterator.h (needs a pass removing unused functionality, possibly not needed anymore)  (#475)
- [x] hilti/library.h (#478)
- [x] hilti/logging.h (#480)
- [X] hilti/result.h
- [x] hilti/safe-int.h (#475)
- [x] hilti/threading.h (provides only variables, hard to test)
- [X] hilti/type-info.h (NOTE: not unit tested, hardly any incidental coverage)
- [X] hilti/types/address.h
- [X] hilti/types/any.h
- [X] hilti/types/bool.h
- [X] hilti/types/bytes.h
- [X] hilti/types/error.h
- [X] hilti/types/function.h
- [x] hilti/types/integer.h
- [X] hilti/types/interval.h
- [X] hilti/types/map.h
- [x] hilti/types/network.h
- [X] hilti/types/null.h
- [X] hilti/types/optional.h
- [X] hilti/types/port.h
- [X] hilti/types/real.h
- [x] hilti/types/reference.h (#474)
- [X] hilti/types/regexp.h
- [X] hilti/types/result.h
- [X] hilti/types/set.h
- [X] hilti/types/set_fwd.h
- [X] hilti/types/stream.h
- [x] hilti/types/string.h
- [x] hilti/types/struct.h
- [x] hilti/types/time.h
- [X] hilti/types/tuple.h
- [x] hilti/types/union.h
- [X] hilti/types/vector.h
- [X] hilti/types/vector_fwd.h
- [X] hilti/unpack.h
[...]

PR #470, opened 2020-07-31
Extend test coverage for rt::Network

bbannier:


hilti/src/rt/tests/network.cc
```diff
@@ -0,0 +1,294 @@
...
+    const Address addr4("0:0:0:0:0:0:0:0");
+
+    CHECK_EQ(Network(addr1, 0), Network(addr1, 0));
+    CHECK_EQ(Network(addr1, 12), Network(addr1, 12));
+    CHECK_EQ(Network(addr1, 32), Network(addr1, 32));
+
+    CHECK_NE(Network(addr1, 32), Network(addr1, 0));
+    CHECK_NE(Network(addr1, 32), Network(addr2, 32));
+    CHECK_NE(Network(addr1, 0), Network(addr3, 0));
+    // TODO(bbannier): `Address` cannot distinguish the different families of these addresses.
```
bbannier, line 25: This and the other TODOs in this file need work on `rt::Address` to explicitly store its address family.

hilti/include/rt/types/network.h
```diff
@@ -21,20 +21,36 @@ class Network {
...
+        switch ( _prefix.family() ) {
+            case AddressFamily::IPv4:
+                if ( _length < 0 || _length > 32 )
+                    throw InvalidArgument(fmt("prefix length %s is invalid for IPv4 networks", _length));
+                break;
+            case AddressFamily::IPv6:
+                if ( _length < 0 || _length > 128 )
+                    throw InvalidArgument(fmt("prefix length %s is invalid for IPv6 networks", _length));
+                break;
+            case AddressFamily::Undef: cannot_be_reached();
```
rsmmr, line 40: Let's throw InvalidArgument here as we. `cannot_be_reached()` is meant for places that really logically cannot happen, but here could be a way to create an `Address` with an `Undef` family somehow.

hilti/src/rt/tests/network.cc
rsmmr: Yeah, seems we need to store the family with address. I had been hoping to avoid that, but the results feel non-intuitive.

hilti/src/rt/types/address.cc
```diff
@@ -61,11 +67,11 @@ Address Address::mask(unsigned int width) const {
     else
         a2 = 0;
 
-    return Address{a1, a2};
+    return Address{a1, a2, _family};
 }
 
 std::variant<struct in_addr, struct in6_addr> Address::asInAddr() const {
-    switch ( family() ) {
+    switch ( (_a1 == 0 && (_a2 & 0xffffffff00000000) == 0) ? AddressFamily::IPv4 : AddressFamily::IPv6 ) {
```
rsmmr, line 74: Why's this not using family?

hilti/src/rt/types/address.cc
```diff
@@ -87,12 +93,18 @@ std::variant<struct in_addr, struct in6_addr> Address::asInAddr() const {
     cannot_be_reached();
 }
 
-bool Address::operator==(const Address& other) const { return _a1 == other._a1 && _a2 == other._a2; }
+bool Address::operator==(const Address& other) const {
+    // NOTE: `_family` is not checked here as IPv4 and IPv6 addresses can be equivalent.
```
rsmmr, line 97: I can see doing this either way, so ok.

hilti/src/rt/types/address.cc
bbannier: Good catch, this is a leftover from shuffling things around. Cleaned up.

hilti/src/rt/types/address.cc
bbannier: In the current implementation addresses still have a slight preference towards IPv4 since e.g., `Adress("::")` constructs an address `0.0.0.0`. Not including `_family` here makes it slightly less easier for us to compare values, but it shouldn't be hard to change the behavior here (also why I included the `NOTE`).

bbannier:
Closed by c1d989bfc1e1af3b253a1976782538b7ea71d51d.

PR #471, opened 2020-08-04
Switch current Zeek packages to 3.2-rc.

rsmmr:
Also factoring out the package config in cirrus.yml and Dockerfiles.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,6 +1,6 @@
 environment:
     CCACHE_BASEDIR: $CIRRUS_WORKING_DIR
-    ZEEK_PACKAGES_CURRENT: zeek-rc=3.2.0-* zeek-rc-core-dev=3.2.0-*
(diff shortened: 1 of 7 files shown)


.cirrus.yml
```diff
@@ -1,11 +1,13 @@
 environment:
     CCACHE_BASEDIR: $CIRRUS_WORKING_DIR
+    ZEEK_PACKAGES_CURRENT: zeek-rc zeek-rc-core-dev
+    ZEEK_PACKAGES_LTS:     zeek-lts zeek-lts-core-dev
```
bbannier, line 4: I think it would be better to explicitly pin versions here, i.e., instead of e.g., `zeek-lts` install `zeek-lts=X.Y.Z`. Otherwise CI would be using whatever version was present in the repo when the (cached) image was last built which could lead to non-reproducible builds.

Similarly, we could pin versions of packages installed through pip in the `Dockerfile`s as well, but the syntax is [slightly more arcane](https://github.com/hadolint/hadolint/wiki/DL3013).

PR #472, opened 2020-08-04
Extend test coverage of hilti::rt

bbannier:
This patch is part of #468 and covers `types/string.h`, `types/struct.h`, `types/time.h`, and `types/union.h`.

This series also contains a small number of bug fixes.

### DIFF
--- a/hilti/src/rt/tests/time.cc
+++ b/hilti/src/rt/tests/time.cc
@@ -27,11 +27,6 @@ TEST_CASE("current_time") {
 
 TEST_SUITE_BEGIN("Time");
 
-TEST_CASE("bool") {
-    CHECK(Time());
-    CHECK(Time(0, Time::SecondTag{}));
-}
-
 TEST_CASE("comparisions") {
     const auto t0 = Time(0, Time::NanosecondTag{});
     const auto t1 = Time(1, Time::NanosecondTag{});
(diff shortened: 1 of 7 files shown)


hilti/src/rt/tests/time.cc
```diff
@@ -24,4 +25,93 @@ TEST_CASE("current_time") {
     CHECK_GE(end + 1, current_time.seconds());
 }
 
+TEST_SUITE_BEGIN("Time");
+
+TEST_CASE("bool") {
+    CHECK(Time());
```
rsmmr, line 31: Shouldn't these evaluate to false?

bbannier: That was my initial feeling as well, but this is exactly how this operator is designed, https://github.com/zeek/spicy/blob/0dc760db6206bcf4a41cb14a7835bd950737ca08/hilti/include/rt/types/time.h#L91-L92.

That being said, I am not sure the "unset" semantic is useful as in reality it is impossible to construct a truly unset unset `Time` value. Currently we use the valid null value of `Time` to mean "unset" (same domain). We could remove "unset" completely and leave it up to users to decide what semantics the null value should have. WDYT?

bbannier: Posting that expert made me reread it and I now get your point.

We could either fix comment or impl to be consistent with the other, or remove this completely. Removal seems preferable to me.

bbannier: Removing this operator like I did in f3ebf0 seems to work.

PR #473, opened 2020-08-05
Add batch input mode to spicy-dump

rsmmr:


zeek/plugin/src/protocol-analyzer.cc
```diff
@@ -10,178 +10,103 @@ using namespace spicy::zeek;
...
-        DebugMsg(*endp, s.c_str());
-        reporter::weird(cookie.analyzer->Conn(), s);
+    try {
+        hilti::rt::context::CookieSetter _(&endp->cookie());
+        endp->process(len, reinterpret_cast<const char*>(data));
+    } catch ( const spicy::rt::ParseError& e ) {
+        reporter::weird(endp->cookie().analyzer->Conn(), e.what());
+        endp->skipRemaining();
+    } catch ( const hilti::rt::Exception& e ) {
+        // TODO: For some reason this executes even when the Spicy parser
```
rsmmr, line 62: Any idea in this one?

rsmmr:
@bbannier We'll see if this pass now but otherwise this is ready. I marked one TODO where I'm a bit puzzled what's going on in case you have an idea.

tests/spicy/tools/spicy-driver-properties.spicy
```diff
@@ -0,0 +1,43 @@
...
+    x: bytes &eod;
+    on %done { print "Port (resp)", self; }
+};
+
+public type FooMIME = unit {
+    %mime-type = "application/http";
+
+    x: bytes &eod;
+    on %done { print "MIME", self; }
+};
```
bbannier, line 38: Can we also add one with a wildcard like e.g., `application/*`?

spicy/src/rt/driver.cc
```diff
@@ -17,7 +21,17 @@ using hilti::rt::fmt;
 
 using namespace spicy::rt;
 
-inline void Driver::_debug(const std::string_view& msg) {
+inline static auto _pretty_print(uint64_t n) {
```
bbannier, line 24: Now that this is visible, could we give the function a more descriptive name and also name the argument accordingly, e.g., `auto pretty_print_memory(uint64_t mem)` or similar? I don't think we need to starting underscore as this function is only visible in this TU.

spicy/src/rt/driver.cc
```diff
@@ -92,30 +114,31 @@ Result<Nothing> Driver::listParsers(std::ostream& out) {
...
-    else {
-        for ( const auto& i : parsers ) {
-            if ( i->name == parser_name )
-                return i;
-        }
+    const auto& parsers_by_name = detail::globalState()->parsers_by_name;
 
-        return Error(fmt("spicy-driver: parser '%s' is not available", parser_name));
+    if ( auto p = parsers_by_name.find(name); p != parsers_by_name.end() ) {
+        assert(p->second.size() > 0);
```
bbannier, line 133:
```suggestion
        assert( !p->empty());
```

spicy/lib/spicy-driver-host.cc
```diff
@@ -63,19 +65,25 @@ void SpicyDriver::usage() {
...
            "exceptions.\n"
            "  -B | --show-backtraces          Include backtraces when reporting unhandled exceptions.\n"
+           "  -F | ---batch-file <path>       Read Spicy batch input from <path>; see docs for description of format.\n"
            "\n";
 }
 
 void SpicyDriver::parseOptions(int argc, char** argv) {
     while ( true ) {
-        int c = getopt_long(argc, argv, "ABhdf:lp:i:v", long_driver_options, nullptr);
+        int c = getopt_long(argc, argv, "ABFhdf:lp:i:v", long_driver_options, nullptr);
```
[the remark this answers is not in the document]
rsmmr: ack, good catch

spicy/src/rt/init.cc
```diff
@@ -21,8 +21,43 @@ void spicy::rt::init() {
...
+                    globalState()->parsers_by_name[idx + "%orig"].emplace_back(p);
+                    globalState()->parsers_by_name[idx + "%resp"].emplace_back(p);
+                    break;
+
+                case Direction::Undef: break;
+            }
+        }
+
+        for ( const auto& x : p->mime_types ) {
+            if ( ! x.isWildcard() )
```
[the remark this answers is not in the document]
rsmmr: because when picking a parser, we can't match against them as one would intuitively expect. Guess we could add wildcard matching as well, but I'll leave that for when we need it.

tests/spicy/tools/spicy-driver-properties.spicy
rsmmr: we don't do that matching, so it wouldn't have any noticeable effect

rsmmr:
@bbannier For the exception fix, I changed from a virtual method to a virtual destructor. Had originally hoped to make the virtual method pure, to catch if it wasn't implemented by a derived class; but that doesn't work because we are still creating instances of `Exception`.

hilti/include/rt/exception.h
```diff
@@ -98,6 +103,64 @@ HILTI_EXCEPTION(OutOfRange, RuntimeError)
 /** Exception flagging invalid arguments passed to a function. */
 HILTI_EXCEPTION(InvalidArgument, RuntimeError);
 
+/** Exception flagging access to an iterator that not, or no longer, valid. */
```
bbannier, line 106: I feel that these we loose some granularity by putting all these into the same namespace, at least for some of them WDYT about instead keeping them in their origin namespaces and probably headers? For headers currently without impl file we could add a very simple on to emit defintions for the exceptions.

hilti/src/rt/exception.cc
```diff
@@ -39,7 +63,7 @@ Exception::Exception(const std::string& desc)
 Exception::Exception(std::string_view desc, std::string_view location)
     : Exception(fmt("%s (%s)", desc, location), desc, location) {}
 
-Exception::~Exception() = default;
+Exception::~Exception() {}
```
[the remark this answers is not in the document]
rsmmr: just a left-over from reverting a previous change, will update to use `default`

hilti/include/rt/exception.h
rsmmr: I actually like the new naming  structure better because the previous was inconsistent in terms of what got namespaced vs what not, and because I'm not sure that that level of granularity is actually helpful/needed. I was thinking about going even further and replace some of the more specifics ones with more general exceptions; didn't do that for now, but we could still consider that.

PR #474, opened 2020-08-10
Increase test coverage for reference types in hilti-rt library

bbannier:
This patchset is part of #468, in particular the work on `hilti/rt/types/reference.h`.


bbannier:
While this patchset is broken into individual commits, it probably makes more sense to review the complete diff and only consult a particular patch if more information on a set of changes is needed.

rsmmr:
Nice job!

PR #476, opened 2020-08-11
Add version information to HLTO files

rsmmr:


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "3rdparty/doctest"]
 	path = 3rdparty/doctest
 	url = https://github.com/onqtam/doctest.git
+[submodule "3rdparty/json"]
```
bbannier, line 4: Did you consciously not pick a release for this one? The last one seems to be v3.9.1.

Also, could you update `LICENSE.3rdparty` for the new path?

tests/spicy/rt/release-vs-debug.cc
```diff
@@ -17,6 +17,9 @@
 #include <hilti/rt/libhilti.h>
 #include <spicy/rt/libspicy.h>
 
+// Just dummuy data.
```
bbannier, line 20:
```suggestion
// Just dummy data.
```

hilti/include/rt/library.h
```diff
@@ -2,9 +2,46 @@
...
+    bool optimize;     /**< true if compiled with optimizations enabled */
+
+    std::filesystem::path path; /**< path to file that library was loaded from; not embedded into JSON, but filled in by
+                                   `Library::open()` */
+
+    /** Converts the instances into a JSON string. */
+    std::string toJSON() const;
+
+    /**
+     * Parsed a JSON representation of an instance.
```
bbannier, line 30:
```suggestion
     * Parses a JSON representation of an instance.
```

hilti/include/rt/library.h
```diff
@@ -2,9 +2,46 @@
...
+     *
+     * @return error if parsing of JSON failed
+     */
+    hilti::rt::Result<Nothing> fromJSON(const std::string& json);
+
+    /**
+     * Checks the version for compatiblity with the current runtime system.
```
bbannier, line 37:
```suggestion
     * Checks the version for compatibility with the current runtime system.
```

hilti/include/rt/library.h
```diff
@@ -2,9 +2,46 @@
...
+     * Prints out warnings on mismatches, but doesn't abort.
+     */
+    void checkCompatibility() const;
```
bbannier, line 40: This is hard to test for, could we instead return any errors? Something like
```suggestion
     * Returns an Error on mismatch, Nothing otherwise.
     */
    Result<Nothing> checkCompatibility() const;
```

hilti/src/rt/library.cc
```diff
@@ -1,14 +1,51 @@
 // Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
 
 #include <dlfcn.h>
+#include <nlohmann/json.hpp>
```
bbannier, line 4: Let's instead push `hilti/json.h` into `hilti/rt` and include it here so we can still suppress diagnostics in `nlohmann/json.hpp`. We'd also move the dep on `nlohmann_json` from `hilti` to `hilti-rt` again for that.

hilti/src/rt/library.cc
```diff
@@ -32,13 +69,28 @@ hilti::rt::Library::~Library() {
...
 
-    if ( ! ::dlopen(_path.c_str(), mode) )
+    void* hlto = ::dlopen(_path.c_str(), mode);
+    if ( ! hlto )
         return result::Error(fmt("failed to load library %s: %s", _path, dlerror()));
 
-    return hilti::rt::Nothing();
+    auto version_string = reinterpret_cast<const char**>(::dlsym(hlto, "__hlto_library_version"));
+    if ( ! version_string )
+        return result::Error("no version information");
```
bbannier, line 81: This should go into the changelog as a breaking change.

hilti/src/rt/library.cc
```diff
@@ -32,13 +69,28 @@ hilti::rt::Library::~Library() {
...
+    library::Version version;
+    if ( auto rc = version.fromJSON(*version_string); ! rc )
+        return result::Error(fmt("broken version information (%s)", rc.error()));
+
+    // Check version. We only warn for now, don't abort.
+    if ( version.magic != "v1" )
+        result::Error(fmt("unknown HLTO version '%s'", version.magic));
+
+    version.path = _orig_path;
+    version.checkCompatibility();
```
bbannier, line 92: If this function would return results, we'd print them here.

hilti/src/rt/library.cc
```diff
@@ -1,14 +1,51 @@
...
+        return result::Error(e.what());
+    }
+}
+
+void hilti::rt::library::Version::checkCompatibility() const {
+    if ( hilti_version != PROJECT_VERSION_NUMBER )
+        warning(fmt("module %s was compiled with HILTI version %d, but using HILTI version %d", path.filename(),
+                    hilti_version, PROJECT_VERSION_NUMBER));
+
+    if ( hilti::rt::isDebugVersion() && optimize )
```
bbannier, line 41: This is essentially still the tech debt of us conflating debug and/or optimizations in `debug`. Ideally this would just check if we are unoptimized right now and we should at least capture that in a comment.

Do you think we'd want a similar warning for the inverse (running an unoptimized library while in an optimized run)?

hilti/src/rt/library.cc
```diff
@@ -32,13 +69,28 @@ hilti::rt::Library::~Library() {
...
+    if ( auto rc = version.fromJSON(*version_string); ! rc )
+        return result::Error(fmt("broken version information (%s)", rc.error()));
+
+    // Check version. We only warn for now, don't abort.
+    if ( version.magic != "v1" )
+        result::Error(fmt("unknown HLTO version '%s'", version.magic));
+
+    version.path = _orig_path;
+    version.checkCompatibility();
+    return version;
```
bbannier, line 93: We could move here to avoid the copy in the conversion to the return type.
```suggestion

    return std::move(version);
```

hilti/src/rt/library.cc
rsmmr: Yeah, that debug vs opt is still on my mind, and I was tempted to address it in the same PR, but then didn't want to conflate things.

I opted against a warning for the inverse as it seems a more common use case for users. They'll have the release build of the toolchain by default, and may just be debugging their parsers.

hilti/include/rt/library.h
rsmmr: How would we return what's potentially multiple errors? Also, I'd prefer to keep having the  function report the warnings as that way reporting will be consistent across different callers.

.gitmodules
rsmmr: Nope, missed picking a release, will do.

hilti/include/rt/library.h
rsmmr: As discussed, leaving as is because it's hard to unit test this anyways.

PR #477, opened 2020-08-11
Add option to show resource usage at termination

rsmmr:


hilti/include/rt/global-state.h
```diff
@@ -44,6 +46,9 @@ struct GlobalState {
     /** If not zero, `Configuration::abort_on_exception` is disabled. */
     int disable_abort_on_exceptions = 0;
 
+    /** getrusage() at library initialization time. */
+    struct rusage rusage_init;
```
bbannier, line 50: I'd much more prefer if we used a `ResourceUsage` here, if only to ensure that we can consistently collect and compare data we care about, e.g., in `hilti::rt::init` we could just have
```.cpp
rusage_init = rt::resource_usage();
```

hilti/src/rt/util.cc
```diff
@@ -50,13 +51,19 @@ void hilti::rt::abort_with_backtrace() {
 
 void hilti::rt::cannot_be_reached() { hilti::rt::internalError("code is executing that should not be reachable"); }
 
-hilti::rt::MemoryStatistics hilti::rt::memory_statistics() {
-    MemoryStatistics stats;
+hilti::rt::ResourceUsage hilti::rt::resource_usage() {
+    ResourceUsage stats;
 
     struct rusage r;
     getrusage(RUSAGE_SELF, &r);
```
bbannier, line 58: This was missing error handling, see my comment on `hilti/src/rt/init.cc`.

hilti/src/rt/util.cc
```diff
@@ -50,13 +51,19 @@ void hilti::rt::abort_with_backtrace() {
...
     auto fibers = detail::Fiber::statistics();
 
+    const auto to_double = [](const auto& x) {
+        return static_cast<double>(x.tv_sec) + static_cast<double>(x.tv_usec) / 1e6;
+    };
```
bbannier, line 63: Maybe we could be clearer on types and units here, e.g.,
```suggestion
        const auto to_seconds = [](const timeval& t) {
        return static_cast<double>(t.tv_sec) + static_cast<double>(t.tv_usec) / 1e6;
    };
```

hilti/src/rt/tests/util.cc
```diff
@@ -361,9 +362,11 @@ TEST_CASE("map_tuple") {
     CHECK_EQ(input, std::make_tuple(2u, 2L, std::string("aa")));
 }
 
-TEST_CASE("memory_statistics") {
-    auto ms = memory_statistics();
+TEST_CASE("resource_usage") {
+    auto ms = resource_usage();
 
+    usleep(10000);
```
bbannier, line 368: Why is this needed? I'd imagine this could be useful if we'd test for a change in resource usage, but we don't.

hilti/src/rt/tests/util.cc
```diff
@@ -361,9 +362,11 @@ TEST_CASE("map_tuple") {
     CHECK_EQ(input, std::make_tuple(2u, 2L, std::string("aa")));
 }
 
-TEST_CASE("memory_statistics") {
-    auto ms = memory_statistics();
+TEST_CASE("resource_usage") {
+    auto ms = resource_usage();
```
bbannier, line 366: Could rename this to `ru` now.

spicy/src/rt/driver.cc
```diff
@@ -70,7 +70,7 @@ void Driver::_debug_stats(size_t current_sessions) {
 
     debug(fmt("sessions: current=%s total=%s", num_sessions, total_sessions));
 
-    auto stats = hilti::rt::memory_statistics();
+    auto stats = hilti::rt::resource_usage();
```
bbannier, line 73: Ditto.

spicy/src/rt/driver.cc
```diff
@@ -51,7 +51,7 @@ void Driver::_debug_stats(const hilti::rt::ValueReference<hilti::rt::Stream>& da
     debug(fmt("input : size-current=%s size-total=%s chunks-cur=%s offset-head=%" PRIu64 " offset-tail=%" PRIu64,
               data_size_cur, data_size_total, data_chunks, data_begin, data_end));
 
-    auto stats = hilti::rt::memory_statistics();
+    auto stats = hilti::rt::resource_usage();
```
bbannier, line 54: It would probably make sense to output the added information on l.60.

This would be simplest and easiest to maintain if we'd provide an `operator<<` for `ResourceUsage` and then just output `debug(fmt("%s", &stats))`.

tests/Scripts/canonify-spicy-debug
```diff
@@ -5,4 +5,5 @@
 cat \
   | sed 's/stream=0x[0-9a-z]\{2,\} /stream=0xXXXXXXXX /g' \
   | sed 's/for stream 0x[0-9a-z]\{2,\}/for stream 0xXXXXXXXX/g' \
-  | sed 's/\(memory *: \).*/\1XXX/g'
+  | sed 's/\(memory *: \).*/\1XXX/g' \
+  | grep -v "^# user_time"
```
bbannier, line 9: This would omit any time information or the complete resource usage information if you follow my suggestion above.

Maybe we could insyead canonify time information similar to how we handle `memory` (canonify it, and not just drop it).

tests/spicy/tools/spicy-build-driver.spicy
```diff
@@ -1,7 +1,8 @@
 # @TEST-GROUP: no-jit
 # @TEST-EXEC: ${SPICY_BUILD} %INPUT
-# @TEST-EXEC: ${SCRIPTS}/printf '\x00\x00\x27\x10\x2a' | HILTI_DEBUG=spicy ./a.out -p Mini::Test >output
+# @TEST-EXEC: ${SCRIPTS}/printf '\x00\x00\x27\x10\x2a' | HILTI_DEBUG=spicy ./a.out -p Mini::Test -U >output
 # @TEST-EXEC: btest-diff output
+# @TEST-EXEC: grep -q "^# user_time" .stderr
```
bbannier, line 5: Might not be needed if we canonify time information instead of dropping it.

tests/spicy/tools/spicy-driver-batch.spicy
```diff
@@ -1,6 +1,7 @@
 # @TEST-EXEC: spicyc -d -j -o test.hlto %INPUT
-# @TEST-EXEC: HILTI_DEBUG=spicy-driver:libspicy spicy-driver -F test.dat test.hlto >output
+# @TEST-EXEC: HILTI_DEBUG=spicy-driver:libspicy spicy-driver -F test.dat -U test.hlto >output
 # @TEST-EXEC: btest-diff output
+# @TEST-EXEC: grep -q "^# user_time" .stderr
```
bbannier, line 4: Ditto.

hilti/include/rt/util.h
```diff
@@ -58,17 +58,19 @@ extern void abort_with_backtrace() __attribute__((noreturn));
...
+    // Note when changing this, update `resource_usage()`.
+    double user_time;       //< user time since runtime initialization
+    double system_time;     //< system time since runtime initialization
     uint64_t memory_heap;   //< current size of heap in bytes
     uint64_t num_fibers;    //< number of fibers currently in use
     uint64_t max_fibers;    //< high-water mark for number of fibers in use
     uint64_t cached_fibers; //< number of fibers currently cached for reuse
 };
 
 /** Returns statistics about the current state of memory allocations. */
```
bbannier, line 72:
```suggestion
/** Returns statistics about the current resource usage. */
```

hilti/include/rt/util.h
```diff
@@ -58,17 +58,19 @@ extern void abort_with_backtrace() __attribute__((noreturn));
 /** Aborts with an internal error saying we should not be where we are. */
 extern void cannot_be_reached() __attribute__((noreturn));
 
-/** Statistics about the current state of memory allocations. */
-struct MemoryStatistics {
-    // Note when changing this, update `memory_statistics()`.
+/** Statistics about current resource uage. */
```
bbannier, line 61:
```suggestion
/** Statistics about resource usage. */
```

hilti/src/rt/debug-logger.cc
```diff
@@ -26,24 +26,21 @@ void detail::DebugLogger::print(const std::string& stream, const std::string& ms
...
-        if ( ! out.is_open() )
-            fatalError(fmt("libhilti: cannot open file '%s' for debug output", _path));
-
-        _output = std::move(out);
+        if ( _path == "/dev/stdout" )
+            _output = &std::cout;
+        else if ( _path == "/dev/stderr" )
+            _output = &std::cerr;
+        else {
+            _output_file = std::make_unique<std::ofstream>(_path, std::ios::out | std::ios::trunc);
```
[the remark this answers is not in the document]
rsmmr: the output concludes with `std::endl`, that should take care of the flush, no?

hilti/include/rt/global-state.h
rsmmr: good idea

hilti/src/rt/tests/util.cc
rsmmr: I saw test failures for the new `> 0` check on execution time so far, presumably because the double couldn't represent whatever tiny number of execution had passed at that point.

spicy/src/rt/driver.cc
rsmmr: I was debating about that as well but ended up leaving as it is because I'm not sure that information would be very helpful in that debugging output.

tests/spicy/tools/spicy-build-driver.spicy
rsmmr: I'll try the canonification but will still leave this in there as btest-diff is done only for debug builds.

PR #478, opened 2020-08-12
Extend tests of hilti/rt/{context,deferred-expression,exception,library}.h

bbannier:

### DIFF
--- a/hilti/include/rt/exception.h
+++ b/hilti/include/rt/exception.h
@@ -216,6 +216,7 @@ void printUncaught(const Exception& e, std::ostream& out);
 
 namespace detail::adl {
 inline std::string to_string(const Exception& e, adl::tag /*unused*/) { return fmt("<exception: %s>", e.what()); }
+inline std::string to_string(const WouldBlock& e, adl::tag /*unused*/) { return fmt("<exception: %s>", e.what()); }
 } // namespace detail::adl
 
 } // namespace hilti::rt
--- a/hilti/src/rt/library.cc
+++ b/hilti/src/rt/library.cc
@@ -74,11 +74,16 @@ hilti::rt::Library::~Library() {
 hilti::rt::Result<hilti::rt::library::Version> hilti::rt::Library::open() const {
     constexpr auto mode = RTLD_LAZY | RTLD_GLOBAL;
 
-    void* hlto = ::dlopen(_path.c_str(), mode);
-    if ( ! hlto )
-        return result::Error(fmt("failed to load library %s: %s", _path, dlerror()));
+    if ( ! _handle ) {
+        void* handle = ::dlopen(_path.c_str(), mode);
 
-    auto version_string = reinterpret_cast<const char**>(::dlsym(hlto, "__hlto_library_version"));
+        if ( ! ::dlopen(_path.c_str(), mode) )
+            return result::Error(fmt("failed to load library %s: %s", _path, dlerror()));
+
+        _handle = handle;
+    }
+
+    auto version_string = reinterpret_cast<const char**>(::dlsym(_handle, "__hlto_library_version"));
     if ( ! version_string )
         return result::Error("no version information");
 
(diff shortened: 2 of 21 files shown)


hilti/src/rt/tests/library.cc
```diff
@@ -0,0 +1,225 @@
...
+            CAPTURE(e.description());
+            CHECK(startsWith(e.description(),
+                             fmt("could not add library \"%s\": could not create temporary file: ", dummy.c_str())));
+        } catch ( ... ) {
+            FAIL("unexpected exception thrown");
+        }
+    }
+
+    SUBCASE("cannot write TMPDIR") {
+        // NOTE: This test likely does not work if run as `root` as `root`
```
rsmmr, line 134: want to add uid==0 test and just skip in that case? Tests shouldn't be running as root, but who knows ...

hilti/src/rt/tests/library.cc
```diff
@@ -0,0 +1,225 @@
...
+TEST_CASE("destruct") {
+    TemporaryDirectory tmp;
+    Env _("TMPDIR", tmp.path().c_str());
+
+    auto library = std::unique_ptr<Library>(new Library(dummy));
+
+    // Recursively change the permissions of the directory
+    // so the temporary library cannot be removed anymore.
+    //
+    // NOTE: This test likely does not work if run as `root` as `root`
```
rsmmr, line 162: same. 

Does doctest have a notion of "run test only of X is true"?

hilti/src/rt/tests/library.cc
```diff
@@ -0,0 +1,225 @@
...
+        REQUIRE_FALSE(symbol("foo").hasValue());
+
+        {
+            // Creating `Library` instance does not load it automatically.
+            Library library(dummy);
+            CHECK_FALSE(symbol("foo").hasValue());
+
+            // Explicitly opening the `Library` loads it so the symbol can be found.
+            CHECK_EQ(library.open(), Nothing());
+            CHECK(symbol("foo").hasValue());
```
rsmmr, line 185: one more test that would be useful if easy to add: if we have two libraries that provide the same symbol, we want to (1) be able to open them both, and (2) still get access to each's version of the symbol. We need this for the new HLTO versioning: same symbol, but library-specific content.

hilti/src/rt/tests/library.cc
bbannier: Ideally we'd skip this test if we are in a privileged context where file permissions are ignored, but it is really hard to check for in a generic, platform-independent way so I left it out.

The correct way to check for this under Linux would be to check whether the calling process has the `CAP_DAC_OVERRIDE` capability (which `root` usually has); I am not sure what the equivalent concept of that would be on macos. That being said, I think the most likely scenario where the process has such privileges would be when running in Docker containers without explicitly configured `USER` (like we now do for our CI images). I don't want to worry about other scenarios as I'd like to claim that e.g., a user running these unit tests as `root` on macos is clearly "holding it wrong".

I'd prefer we don't try to do more here and instead have anybody running into issues stumble into the comment I added.

hilti/src/rt/tests/library.cc
bbannier: See above, dropping.

No, I don't think doctest even has a way to explicitly `SKIP` tests. One could probably add a `return` or `continue` statement to skip the rest of a test (probably after emitting a `WARNING`).

What we could do if this becomes a repeated issue is to update the test filter by providing a custom test [main function](https://github.com/onqtam/doctest/blob/master/doc/markdown/main.md).

hilti/src/rt/tests/library.cc
bbannier: This is currently not possible since `Library` does not hold handles to libraries it wraps and users instead call `::dlsym` which might return a unexpected symbol if multiple symbols with the same name exist.

I'll add a member function to obtain a symbol from a specific `Library`; we'd still need to update calling code to use that instead of global `::dlsym` which would also require them to hold on to instances of `Library` (once we fix that we should also add `::dlclose` to `Library` destructors).

hilti/include/rt/exception.h
```diff
@@ -216,6 +216,7 @@ void printUncaught(const Exception& e, std::ostream& out);
 
 namespace detail::adl {
 inline std::string to_string(const Exception& e, adl::tag /*unused*/) { return fmt("<exception: %s>", e.what()); }
+inline std::string to_string(const WouldBlock& e, adl::tag /*unused*/) { return fmt("<exception: %s>", e.what()); }
```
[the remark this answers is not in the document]
bbannier: `WouldBlock` is not an `Exception`, but instead a subclass of `std::runtime_error`.

bbannier:
Could you have a look again @rsmmr? Since your last review I rebased on top of your recent changes to `rt::Library` and also added a function to retrieve a symbol from a library.

hilti/src/rt/library.cc
```diff
@@ -74,25 +79,44 @@ hilti::rt::Library::~Library() {
 hilti::rt::Result<hilti::rt::library::Version> hilti::rt::Library::open() const {
     constexpr auto mode = RTLD_LAZY | RTLD_GLOBAL;
 
-    void* hlto = ::dlopen(_path.c_str(), mode);
-    if ( ! hlto )
-        return result::Error(fmt("failed to load library %s: %s", _path, dlerror()));
+    if ( ! _handle ) {
+        void* handle = ::dlopen(_path.c_str(), mode);
+
+        if ( ! ::dlopen(_path.c_str(), mode) )
```
rsmmr, line 85:
```suggestion
        if ( ! handle )
```

PR #480, opened 2020-08-17
Add tests for hilti/rt/{fmt,global-state,hilti,init,logging}.h

bbannier:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,4 +39,9 @@ repos:
     exclude: '3rdparty/'
     args: ["--quiet"]
 
+- repo: https://github.com/jorisroovers/gitlint
+  rev:  v0.13.1
+  hooks:
+  - id: gitlint
+
 exclude: /3rdparty/|doc/.*examples/|/Baseline/|(\.svg$)|(\.dat$)
--- a/hilti/src/rt/tests/util.cc
+++ b/hilti/src/rt/tests/util.cc
@@ -11,6 +11,8 @@
 #include <vector>
 
 #include <hilti/rt/autogen/version.h>
+#include <hilti/rt/fiber.h>
+#include <hilti/rt/init.h>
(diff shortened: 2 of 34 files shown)


rsmmr:
I'll wait for the rebase here before reviewing.

.pre-commit-config.yaml
```diff
@@ -39,4 +39,9 @@ repos:
     exclude: '3rdparty/'
     args: ["--quiet"]
 
+- repo: https://github.com/jorisroovers/gitlint
```
rsmmr, line 42: So what's this going to enforce? Just the title-match-regex? Can you add a short summary of the constraints to the dev guide? https://docs.zeek.org/projects/spicy/en/latest/development/coding-style.html#commit-messages

(And while you are at it: We can remove "Avoid rebases once pushed upstream", the current model works pretty well.)

hilti/src/rt/tests/util.cc
```diff
@@ -362,23 +364,49 @@ TEST_CASE("map_tuple") {
     CHECK_EQ(input, std::make_tuple(2u, 2L, std::string("aa")));
 }
 
-TEST_CASE("resource_usage") {
-    auto ru = resource_usage();
+namespace hilti::rt {
+std::ostream& operator<<(std::ostream& stream, const ResourceUsage& ru) {
```
rsmmr, line 368: are you using/testing this?

bbannier: I used this in development for printing `ResourceUsage` objects from doctest as otherwise it would use its default formatter (which just outputs `{?}`). I removed this for now.

.pre-commit-config.yaml
bbannier: This will mainly catch overly long lines, and an incorrect title format (e.g., not starting with a capital letter, or not ending in `.`). I called out the tiltle part in the documentation.

rsmmr (approved):

PR #483, opened 2020-08-19
Build system tweaks

rsmmr:


bbannier (approved):
Looks good to me.

You might want to fix the commit message of c6e79d167783872f6a113bfc546e05338f1701d1 to something like _Compile runtime libraries with debug information also when in release mode._ before committing since that's what the commit really does.

PR #484, opened 2020-08-19
Extend unit test coverage of spicy-rt

bbannier:
This patch is part of #468.

### DIFF
--- a/spicy/src/rt/base64.cc
+++ b/spicy/src/rt/base64.cc
@@ -1,8 +1,8 @@
 // Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
 
-#include <hilti/rt/types/bytes.h>
+#include "spicy/rt/base64.h"
 
-#include <spicy/rt/base64.h>
+#include <hilti/rt/types/bytes.h>
 
 extern "C" {
 #include <spicy/3rdparty/libb64/b64/cdecode.h>
@@ -36,8 +36,7 @@ hilti::rt::Bytes Stream::encode(const hilti::rt::Bytes& data) {
         throw Base64Error("encoding already finished");
 
     char buf[static_cast<uint64_t>(data.size() * 2)];
-    int len = base64_encode_block(reinterpret_cast<const char*>(data.data()), static_cast<int>(data.size()), buf,
-                                  &_state->estate);
+    int len = base64_encode_block(data.data(), static_cast<int>(data.size()), buf, &_state->estate);
     return hilti::rt::Bytes(buf, len);
 }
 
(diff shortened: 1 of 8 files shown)


spicy/src/rt/base64.cc
```diff
@@ -61,6 +60,8 @@ hilti::rt::Bytes Stream::decode(const hilti::rt::Bytes& data) {
     if ( ! _state )
         throw Base64Error("decoding already finished");
 
+    base64_init_decodestate(&_state->dstate);
```
bbannier, line 63: This seems to not be what we want as it breaks the BTests. Not doing this leads to what looks like junk in the decoder stream.

spicy/src/rt/base64.cc
```diff
@@ -76,6 +77,8 @@ hilti::rt::Bytes Stream::decode(const hilti::rt::stream::View& data) {
 
     for ( auto block = data.firstBlock(); block; block = data.nextBlock(block) ) {
         char buf[static_cast<uint64_t>(block->size * 2)];
+        base64_init_decodestate(&_state->dstate);
```
bbannier, line 80: Ditto.

spicy/src/rt/tests/base64.cc
```diff
@@ -0,0 +1,127 @@
...
+                Stream data("TW9yZSB0aGFuIDYgYnl0ZXM");
+                CHECK_EQ(base64::decode(stream, data.view()), "More than 6 bytes"_b);
+                data.append("Zm9v");
+                CHECK_EQ(base64::decode(stream, data.view()), "More than 6 bytesfoo"_b);
+            }
+        }
+    }
+}
+
+TEST_CASE("encode") {
```
rsmmr, line 80: Suggest to add a couple of "multiple calls" tests for encode as well.

spicy/src/rt/base64.cc
rsmmr: That resets the state, so that the "multiple calls" will presumably be breaking. If it doesn't work without that, we have a bug somewhere.

bbannier: I removed this again. The behavior I was attempting to "fix" was due to me using the base64 API incorrectly, in particular I was passing incomplete and unpadded data which lead to bytes I was not expecting to appear in the output. I fixed the test now.

I also attempted to upgrade libb64 to https://github.com/libb64/libb64's latest release (we are using 1.2.0 which is 10 years old while the latest release from the github fork is just over 2 years old). This failed since the github version changed its behavior.

rsmmr (approved):

PR #488, opened 2020-08-21
Add unit tests for spicy/rt/{global-state,util,mime,parsed-unit,zlib}.h

bbannier:

### DIFF
--- a/spicy/src/rt/tests/zlib.cc
+++ b/spicy/src/rt/tests/zlib.cc
@@ -1,0 +1,91 @@
+// Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
+
+#include <doctest/doctest.h>
+
+#include <hilti/rt/extension-points.h>
+#include <hilti/rt/types/bytes.h>
+
+#include <spicy/rt/zlib.h>
+
+using namespace hilti::rt;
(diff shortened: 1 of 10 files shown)


spicy/src/rt/tests/zlib.cc
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        SUBCASE("single block") {
+            CHECK_EQ(zlib::decompress(stream, "x\x01\x01\x03\x00\xfc\xff\x00\x01\x02\x00\x07\x00\x04"_b),
+                     "\x00\x01\x02"_b);
+            CHECK_EQ(zlib::finish(stream), ""_b);
+        }
+
+        SUBCASE("multiple blocks") {
+            CHECK_EQ(zlib::decompress(stream, "x\x01\x01\x03\x00\xfc\xff\x00\x01\x02\x00\x07\x00\x04"_b),
```
rsmmr, line 32: This is the same as in the "single block" case?

rsmmr (approved):

spicy/src/rt/tests/zlib.cc
bbannier: Ups it is. I already broke the data up, but forgot to decompress it in two steps. Fixed now.

PR #489, opened 2020-08-26
Add tests for spicy/rt/{init,parser}.h

bbannier:

### DIFF
--- a/spicy/src/rt/parser.cc
+++ b/spicy/src/rt/parser.cc
@@ -1,12 +1,13 @@
 // Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
 
+#include "rt/parser.h"
+
 #include <limits>
 #include <utility>
 
 #include <hilti/rt/types/bytes.h>
 
 #include <spicy/rt/debug.h>
-#include <spicy/rt/parser.h>
 
 using namespace spicy::rt;
 using namespace spicy::rt::detail;
@@ -124,7 +125,6 @@ bool detail::atEod(const hilti::rt::ValueReference<hilti::rt::Stream>& data, con
 }
 
 bool detail::haveEod(const hilti::rt::ValueReference<hilti::rt::Stream>& data, const hilti::rt::stream::View& cur) {
-    return data->isFrozen() || cur.unsafeEnd().offset() < data->unsafeEnd().offset();
     // We've the reached end-of-data if either (1) the bytes object is frozen
     // (then the input won't change anymore), or (2) our view is limited to
     // something before the current end (then even appending more data to the
--- a/spicy/src/rt/parser.cc
+++ b/spicy/src/rt/parser.cc
@@ -94,9 +94,6 @@ bool detail::waitForInputOrEod(hilti::rt::ValueReference<hilti::rt::Stream>& dat
                                               data.get(), cur.size()));
         hilti::rt::detail::yield();
 
-        auto x = cur.end();
-        x += 0;
-
         if ( filters ) {
             SPICY_RT_DEBUG_VERBOSE("resuming filter execution");
             spicy::rt::filter::flush(filters);
--- a/spicy/src/rt/tests/parser.cc
+++ b/spicy/src/rt/tests/parser.cc
@@ -212,6 +212,49 @@ TEST_CASE("waitForInput") {
     }
 }
 
+TEST_CASE("waitForInput with min") {
+    hilti::rt::test::CaptureIO _(std::cerr); // Suppress output.
+
+    hilti::rt::init(); // Noop if already initialized.
+
+    auto data = hilti::rt::ValueReference<hilti::rt::Stream>();
+    auto view = data->view();
+
+    auto filters = hilti::rt::StrongReference<filter::detail::Filters>();
+
+    auto _waitForInput = [&](hilti::rt::resumable::Handle*) {
+        return detail::waitForInput(data, view, 3, "error message", "location", filters);
+    };
+
+    auto waitForInput = [&]() { return hilti::rt::fiber::execute(_waitForInput); };
+
+    SUBCASE("not enough data") {
+        // `waitForInput` yields if not enough data available. We
+        // can only wait from `Resumable`.
+        CHECK_FALSE(waitForInput());
+        CHECK_THROWS_WITH_AS(_waitForInput(nullptr), "'yield' in non-suspendable context",
+                             const hilti::rt::RuntimeError&);
+    }
+
+    SUBCASE("enough data") {
+        // With enough data available we can get a result.
+        data->append("\x01\x02"_b);
+        REQUIRE_EQ(data->size(), 2);
+        CHECK_FALSE(waitForInput()); // Still need one more byte.
+
+        data->append("\x03");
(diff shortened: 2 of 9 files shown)


bbannier:
Could you have a look @rsmmr? I'll probably still need to investigate while `clang9_zeek_ubuntu_debug` failed with a leak warning, but the general structure should be in place.

spicy/src/rt/tests/parser.cc
```diff
@@ -114,4 +125,101 @@ TEST_CASE("registerParser") {
...
+    auto waitForInputOrEod = [&]() { return hilti::rt::fiber::execute(_waitForInputOrEod); };
+
+    SUBCASE("wait for nothing") {
+        // We can always successfully get "no data".
+        CHECK(detail::waitForInputOrEod(data, data->view(), 0, filters));
+    }
+
+    SUBCASE("not enough data") {
+        // `waitForInputOrEod` yields if not enough data is available. We
+        // can only wait from a `Resumable`.
```
rsmmr, line 306:
```suggestion
        // can only wait from inside a `Resumable`.
```

spicy/src/rt/tests/parser.cc
```diff
@@ -114,4 +125,101 @@ TEST_CASE("registerParser") {
...
+        }
+
+        SUBCASE("multiple filters") {
+            filters = Vector<filter::detail::OneFilter>();
+
+            bool called1 = false;
+            bool called2 = false;
+
+            filters->push_back(filter::detail::OneFilter({.resumable = [&](hilti::rt::resumable::Handle*) {
+                std::cerr << "FOO!\n";
```
rsmmr, line 346: Remove?

spicy/src/rt/tests/parser.cc
```diff
@@ -114,4 +125,101 @@ TEST_CASE("registerParser") {
...
+                called1 = true;
+            }}));
+            filters->push_back(filter::detail::OneFilter({.resumable = [&](hilti::rt::resumable::Handle*) {
+                std::cerr << "FOO!\n";
```
rsmmr, line 350: Remove?

spicy/src/rt/tests/parser.cc
```diff
@@ -114,4 +125,101 @@ TEST_CASE("registerParser") {
...
+            data->append("\x01\x02\x03"_b);
+            res.resume();
+
+            CHECK(res);
+            CHECK(called1);
+            CHECK(called2);
+        }
+    }
+
+    SUBCASE("eod") {
```
rsmmr, line 369: suggest to move this to above the filters; I had originally added comment there already to add an "eod" test. :-)

spicy/src/rt/parser.cc
```diff
@@ -94,9 +94,6 @@ bool detail::waitForInputOrEod(hilti::rt::ValueReference<hilti::rt::Stream>& dat
                                               data.get(), cur.size()));
         hilti::rt::detail::yield();
 
-        auto x = cur.end();
```
rsmmr, line 96: I'm pretty sure this was on purpose originally because it triggers normalization of the stream iterator. I can't tell if we still need it. As long as all the btests pass, it should be fine to remove. The new stream implementation should also not need it anyways.

spicy/src/rt/tests/parser.cc
```diff
@@ -217,6 +217,52 @@ TEST_CASE("waitForInput") {
...
+
+    auto _waitForInput = [&](hilti::rt::resumable::Handle*) {
+        return detail::waitForInput(data, data->view(), 3, "error message", "location", filters);
+    };
+
+    auto waitForInput = [&]() { return hilti::rt::fiber::execute(_waitForInput); };
+
+    SUBCASE("not enough data") {
+        // `waitForInput` yields if not enough data available. We
+        // can only wait from `Resumable`.
```
rsmmr, line 255:
```suggestion
        // can only wait from inside a `Resumable`.
```

spicy/src/rt/tests/parser.cc
```diff
@@ -217,6 +217,52 @@ TEST_CASE("waitForInput") {
     }
 }
 
+TEST_CASE("waitForInput with min") {
```
rsmmr, line 234: What's the "with min" part in this test? Not seeing a different to the earlier test, but I may well be missing something.

spicy/src/rt/tests/parser.cc
```diff
@@ -0,0 +1,117 @@
...
+        CHECK_EQ(detail::globalState()->parsers_by_mime_type,
+                 std::map<std::string, std::vector<const Parser*>>({{"foo/bar", {&parser}}, {"foo", {&parser}}}));
+
+        CHECK_FALSE(parser.__parse_sink);
+        CHECK_FALSE(parser.__hook_gap);
+        CHECK_FALSE(parser.__hook_skipped);
+        CHECK_FALSE(parser.__hook_undelivered);
+    }
+
+    SUBCASE("w/ sink support") {
```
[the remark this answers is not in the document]
bbannier: The first subcase does not exercise this block, https://github.com/zeek/spicy/blob/ced6bcec1139a5bd9c7e28f10248ff5bc6581260/spicy/include/rt/parser.h#L207-L215, and checks that none of its side-effects were triggered. The second subcase can observe these side effects since the unit supports sinks.

spicy/src/rt/parser.cc
bbannier: The tests still pass.

spicy/src/rt/tests/parser.cc
bbannier: The difference is in the lambda which passes `min=3` while the previous test tries to read any input. The called function takes a different call path so I'd suggest we keep this case.

spicy/src/rt/tests/parser.cc
```diff
@@ -114,4 +125,101 @@ TEST_CASE("registerParser") {
     }
 }
 
+TEST_CASE("waitForInputOrEod with min") {
+    hilti::rt::test::CaptureIO _(std::cerr); // Suppress output.
+
+    // Reinitialize the runtime to make sure we do not carry over state between test cases.
+    //
+    // TODO(robin): If we comment out this `done` the "enough data" test cases fails. This seems weird.
```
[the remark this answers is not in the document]
bbannier: I believe this was not a real issue. While looking at the memory leaks reported by asan I fixed these by changing the callbacks from e.g.,
```.cpp
    auto _waitForInput = [&](hilti::rt::resumable::Handle*) {
        return detail::waitForInput(data, data->view(), "error message", "location", filters);
    };
```
where `data` has test case lifetime to
```.cpp
    auto _waitForInput = [&](hilti::rt::resumable::Handle*) {
        return detail::waitForInput(data, view, "error message", "location", filters);
    };
```
which also puts the lifetime of the view in the same scope as `data`. Without this fix we seem to run into something which looks like a cycling reference which gets remedied by adjusting the view lifetime.

Incidentally this fix also makes the issue the `TODO` was referring to go away.

I removed the `TODO`s now.

rsmmr (approved):
Looks good. Interesting to see that the data/view change fixes both problems.

PR #491, opened 2020-08-31
Stream reimplementation

rsmmr:
First 3 commits are just small stuff that came up during development. Final patch is the main change.

### DIFF
--- a/hilti/CMakeLists.txt
+++ b/hilti/CMakeLists.txt
@@ -324,7 +324,6 @@ add_executable(hilti-rt-tests
                src/rt/tests/vector.cc)
 target_compile_options(hilti-rt-tests PRIVATE "-Wall")
 target_compile_options(hilti-rt-tests PRIVATE $<$<CONFIG:Debug>:-O0>)
-# TODO: add the following two lines tpo Spicy RT testt
 target_link_options(hilti-rt-tests PRIVATE $<$<CONFIG:Debug>:-O0>)
 target_link_libraries(hilti-rt-tests PRIVATE $<IF:$<CONFIG:Debug>,hilti-rt-debug-objects,hilti-rt-objects> doctest)
 add_dependencies(hilti-rt-tests hilti-rt-tests-library-dummy1 hilti-rt-tests-library-dummy2)
--- a/hilti/CMakeLists.txt
+++ b/hilti/CMakeLists.txt
@@ -299,6 +299,7 @@ add_executable(hilti-rt-tests
                src/rt/tests/integer.cc
                src/rt/tests/interval.cc
                src/rt/tests/iterator.cc
+               src/rt/tests/intrusive-ptr.cc
                src/rt/tests/library.cc
                src/rt/tests/logging.cc
                src/rt/tests/map.cc
--- a/hilti/CMakeLists.txt
+++ b/hilti/CMakeLists.txt
@@ -322,13 +322,18 @@ add_executable(hilti-rt-tests
                src/rt/tests/util.cc
                src/rt/tests/vector.cc)
 target_compile_options(hilti-rt-tests PRIVATE "-Wall")
-target_link_libraries(hilti-rt-tests PRIVATE hilti-rt-objects doctest)
+target_compile_options(hilti-rt-tests PRIVATE $<$<CONFIG:Debug>:-O0>)
+# TODO: add the following two lines tpo Spicy RT testt
+target_link_options(hilti-rt-tests PRIVATE $<$<CONFIG:Debug>:-O0>)
+target_link_libraries(hilti-rt-tests PRIVATE $<IF:$<CONFIG:Debug>,hilti-rt-debug-objects,hilti-rt-objects> doctest)
 add_dependencies(hilti-rt-tests hilti-rt-tests-library-dummy1 hilti-rt-tests-library-dummy2)
 add_test(NAME hilti-rt-tests COMMAND ${CMAKE_BINARY_DIR}/bin/hilti-rt-tests)
 
 add_executable(hilti-rt-configuration-tests src/rt/tests/configuration.cc)
 target_compile_options(hilti-rt-configuration-tests PRIVATE "-Wall")
-target_link_libraries(hilti-rt-configuration-tests PRIVATE hilti-rt-objects doctest)
+target_compile_options(hilti-rt-configuration-tests PRIVATE $<$<CONFIG:Debug>:-O0>)
+target_link_options(hilti-rt-configuration-tests PRIVATE $<$<CONFIG:Debug>:-O0>)
+target_link_libraries(hilti-rt-configuration-tests PRIVATE $<IF:$<CONFIG:Debug>,hilti-rt-debug-objects,hilti-rt-objects> doctest)
 add_test(NAME hilti-rt-configuration-tests COMMAND ${CMAKE_BINARY_DIR}/bin/hilti-rt-configuration-tests)
 
 add_executable(hilti-tests
--- a/hilti/include/rt/backtrace.h
+++ b/hilti/include/rt/backtrace.h
@@ -20,8 +20,8 @@ public:
     friend bool operator!=(const Backtrace& a, const Backtrace& b) { return ! (a == b); }
 
 private:
-    int _frames = -1;
     void* _callstack[64];
+    int _frames = -1;
 };
 
 bool operator==(const Backtrace& a, const Backtrace& b);
--- a/hilti/include/rt/backtrace.h
+++ b/hilti/include/rt/backtrace.h
@@ -14,14 +14,17 @@ class Backtrace {
 public:
     Backtrace();
 
-    const auto& backtrace() const { return _backtrace; }
+    std::vector<std::string> backtrace() const;
 
-    friend bool operator==(const Backtrace& a, const Backtrace& b) { return a._backtrace == b._backtrace; }
+    friend bool operator==(const Backtrace& a, const Backtrace& b);
     friend bool operator!=(const Backtrace& a, const Backtrace& b) { return ! (a == b); }
 
 private:
-    std::vector<std::string> _backtrace;
+    int _frames = -1;
+    void* _callstack[64];
 };
+
+bool operator==(const Backtrace& a, const Backtrace& b);
 
 /** Wrapper around the ABI's C++ demangle function. */
 inline std::string demangle(const std::string& symbol) {
--- a/hilti/include/rt/types/stream.h
+++ b/hilti/include/rt/types/stream.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
+/// Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
 
 #pragma once
 
@@ -18,6 +18,7 @@
 #include <vector>
 
 #include <hilti/rt/exception.h>
+#include <hilti/rt/intrusive-ptr.h>
 #include <hilti/rt/result.h>
 #include <hilti/rt/safe-int.h>
 #include <hilti/rt/types/time.h>
@@ -32,16 +33,20 @@ class Stream;
 namespace stream {
 class View;
 class SafeConstIterator;
+
+namespace detail {
+class UnsafeConstIterator;
+}
 } // namespace stream
 
 namespace detail::adl {
 std::string to_string(const Stream& x, adl::tag /*unused*/);
 std::string to_string(const stream::View& x, adl::tag /*unused*/);
 std::string to_string(const stream::SafeConstIterator& x, adl::tag /*unused*/);
+std::string to_string(const stream::detail::UnsafeConstIterator& x, adl::tag /*unused*/);
 } // namespace detail::adl
 
 namespace stream {
-
 /** A single element inside a stream instance. */
 using Byte = uint8_t;
 
@@ -53,11 +58,22 @@ using Size = integer::safe<uint64_t>;
 
 namespace detail {
 
+class Chain;
+using ChainPtr = IntrusivePtr<Chain>;
+using ConstChainPtr = IntrusivePtr<const Chain>;
 class UnsafeConstIterator;
 
 /**
- * One block of continuous data inside a stream instance. A stream instance
- * chains these to represent all of its content.
+ * Represents one block of continuous data inside a stream instance. A
+ * stream's *Chain* links multiple of these chunks to represent all of its
+ * content.
+ *
+ * A chunk internally employs small-buffer optimization for very small
+ * amounts of data, storing it directly inside the instance instead of using
+ * heap-allocated memory.
+ *
+ * All public methods of Chunk are constant. Modifications can be done only
+ * be through the owning Chain (so that we can track changes there).
  */
 class Chunk {
 public:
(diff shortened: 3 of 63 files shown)


hilti/src/rt/types/stream.cc
```diff
@@ -69,53 +53,181 @@ void Chunk::trim(Offset o) {
     _offset = o;
 }
 
-SafeConstIterator View::find(Byte b, const SafeConstIterator& n) const {
-    for ( auto i = UnsafeConstIterator(n ? n : _begin); i != UnsafeConstIterator(end()); ++i ) {
+const Chunk* Chain::findChunk(Offset offset, const Chunk* hint_prev) const {
```
bbannier, line 56: Can you add a test for this?

hilti/src/rt/types/stream.cc
```diff
@@ -69,53 +53,181 @@ void Chunk::trim(Offset o) {
...
+    _ensureValid();
+    _ensureMutable();
+    other._ensureValid();
+
+    if ( ! other._head )
+        return;
+
+    _tail->setNext(std::move(other._head));
+    _tail = other._tail;
+    other._tail = nullptr;
```
bbannier, line 125: Should we instead call `other.invalidate()`?

hilti/src/rt/types/stream.cc
```diff
@@ -69,53 +53,181 @@ void Chunk::trim(Offset o) {
...
+        _tail = _tail->last();
+    }
+    else {
+        assert(! _head);
+        chunk->setOffset(_head_offset);
+        chunk->setChain(this);
+        _head = std::move(chunk);
+        _tail = _head->last();
+    }
+}
```
bbannier, line 113: We are not at all invalidating `chunk` here even though this is not much different from any function taking a `Chain&&`. Explicitly invalidating allows callers to perform some operations on `move`'d from chunks, so if we _need to_ do it elsewhere I think we'd also need to do it here. Alternatively, make less promises for the behavior of `move`'d from `Chain` objects and invalidate nowhere.

hilti/src/rt/types/stream.cc
```diff
@@ -69,53 +53,181 @@ void Chunk::trim(Offset o) {
...
+        if ( offset >= _head->endOffset() ) {
+            // Delete chunk.
+            _head = std::move(_head->_next);
+            if ( ! _head || _head->isLast() )
+                _tail = _head.get();
+        }
+
+        else if ( _head->inRange(offset) ) {
+            _head->trim(offset);
+            assert(_head->offset() == offset);
```
bbannier, line 144: We can probably get rid of this assert.

hilti/include/rt/types/stream.h
```diff
@@ -112,65 +142,230 @@ class Chunk {
...
-struct Chain {
-    std::shared_ptr<Chunk> head;
-    std::shared_ptr<Chunk> tail;
+/**
+ * Main data structure for the content of a stream object. A chain is
+ * heap-allocated by the stream and retains ownership of the linked chunks. A
+ * Chain may survive its stream in the case that iterators to any of its
+ * chunks are still around.
+ */
+class Chain : public intrusive_ptr::ManagedObject {
```
bbannier, line 248: This is effectively a list with extra offset and state information now, and it should be possible to delegate all the list behavior to some `std::list` or `std::deque` member. That would get rid of a lot of low-level twiddling, but involve some more work disentangling what is happening here.

hilti/src/rt/types/stream.cc
```diff
@@ -417,40 +453,59 @@ std::string hilti::rt::detail::adl::to_string(const stream::SafeConstIterator& x
...
     if ( ! c )
         // Can happen if trimmed off.
         chunk = -1;
 
-    out << fmt("iterator %p: parent=%p chunk=#%d offset=%llu is_end=%d\n", this, _content.lock().get(), chunk, _offset,
+    out << fmt("iterator %p: chain=%p chunk=#%d/%p offset=%llu is_end=%d\n", this, _chain.get(), chunk, c, _offset,
                static_cast<int>(isEnd()));
 }
 
+std::string hilti::rt::detail::adl::to_string(const UnsafeConstIterator& x, adl::tag /*unused*/) {
```
bbannier, line 473: Can you add a small test for this covering the different scenarios (short & long `View`, expired & unset iterator)?

bbannier (commented):
Looks pretty good, I don't see big things.

As we now `Chain` behaves largely like a list type we could think about whether it makes sense to finish this refactoring by completely delegating the management of that aspect to some stdlib type. After that we'd be left with a small class performing offset manipulation and state management.

hilti/src/rt/types/stream.cc
```diff
@@ -156,209 +269,124 @@ void View::copyRaw(Byte* dst) const {
...
+std::string Stream::data() const { return view().data(); }
 
 std::string stream::View::data() const {
     std::string s;
     s.reserve(size());
 
-    for ( auto i = unsafeBegin(); i != unsafeEnd(); ++i )
-        s += static_cast<char>(*i);
+    for ( auto block = firstBlock(); block; block = nextBlock(block) )
+        s += std::string(reinterpret_cast<const char*>(block->start), block->size);
```
bbannier, line 389: We need to do a `reinterpret_cast` here since `Byte` is a `uint8` (aka a `unsigned char`) while a `std::string` wants to be constructed from a signed `char`. This seems to do what we want, but looks like tech debt to me.

I think ideally this function wouldn't return a `std::string` but a `Bytes` so that we could directly append a range of `Byte` to it (would need a `void Bytes::append(Byte*, size_t)` and probably also a way to `reserve` capacity in a `Bytes`). Such a change percolates a little further than this small fix, but it might be worth it.

hilti/src/rt/types/stream.cc
```diff
@@ -156,209 +269,124 @@ void View::copyRaw(Byte* dst) const {
...
-    for ( auto i = unsafeBegin(); i != unsafeEnd(); ++i )
-        s += static_cast<char>(*i);
-
-    return s;
-}
+std::string Stream::data() const { return view().data(); }
 
 std::string stream::View::data() const {
     std::string s;
     s.reserve(size());
```
bbannier, line 386: I'd totally expect us soon to see `size` as a prominent runtime contributor. ;)

hilti/src/rt/library.cc
```diff
@@ -77,7 +77,7 @@ hilti::rt::Library::~Library() {
 }
 
 hilti::rt::Result<hilti::rt::library::Version> hilti::rt::Library::open() const {
-    constexpr auto mode = RTLD_LAZY | RTLD_GLOBAL;
+    constexpr auto mode = RTLD_NOW | RTLD_GLOBAL;
```
bbannier, line 80: Why this change? I could imagine `RTLD_NOW` to introduce issues for underlinked libraries. While we do not produce such libraries during JIT I am unsure we should disallow it here, at this low level.

EDIT: Does this change belong to the changes in dd571e95249d0dea6a1754bfa46983477caf3b27?

hilti/src/rt/fiber.cc
```diff
@@ -174,6 +174,9 @@ std::unique_ptr<Fiber> Fiber::create() {
 }
 
 void Fiber::destroy(std::unique_ptr<Fiber> f) {
+    if ( f->_state == State::Running || f->_state == State::Yielded )
+        f->abort();
```
bbannier, line 178: `abort` asserts that the fiber is `Yielded`, but we could also abort a `Running` fiber here.

hilti/src/rt/backtrace.cc
```diff
@@ -1,38 +1,61 @@
...
+std::vector<std::string> Backtrace::backtrace() const {
+    std::vector<std::string> bt;
 
-    strings = backtrace_symbols(callstack, frames);
-    assert(strings);
+#ifdef HILTI_HAVE_BACKTRACE
+    assert(_frames >= 0);
+    char** strings = backtrace_symbols(_callstack, _frames);
+    if ( ! strings ) {
+        bt.push_back("# <trouble resolving backtrace symbols>");
```
bbannier, line 24: This sounds like something a user could do something about. Could we either tell them what to do (not sure what the error scenarios are), or else show something like `could not resolve backtrace symbols` or similar?

hilti/include/rt/backtrace.h
```diff
@@ -14,15 +14,18 @@ class Backtrace {
...
+    std::vector<std::string> backtrace() const;
 
-    friend bool operator==(const Backtrace& a, const Backtrace& b) { return a._backtrace == b._backtrace; }
+    friend bool operator==(const Backtrace& a, const Backtrace& b);
     friend bool operator!=(const Backtrace& a, const Backtrace& b) { return ! (a == b); }
 
 private:
-    std::vector<std::string> _backtrace;
+    int _frames = -1;
+    void* _callstack[64];
```
bbannier, line 24: Let's remove the extra padding.
```suggestion
    void* _callstack[64];
    int _frames = -1;
```

hilti/src/rt/backtrace.cc
```diff
@@ -1,38 +1,61 @@
 // Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
 
-#include "rt/backtrace.h"
-
 #include <cstring>
 
 #include <hilti/rt/autogen/config.h>
+#include <hilti/rt/backtrace.h>
```
bbannier, line 6: Let's continue to include the matching header first, https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes. We are not fully consistent on this, but I tried to fix this in files I touched.

hilti/src/rt/tests/intrusive-ptr.cc
```diff
@@ -0,0 +1,38 @@
...
+class Managed : public intrusive_ptr::ManagedObject {
+public:
+    Managed() { ++instances; }
+    ~Managed() { --instances; }
+    static inline int instances = 0;
+};
+
+using ManagedPtr = IntrusivePtr<Managed>;
+
+TEST_CASE("managed objects") {
```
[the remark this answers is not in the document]
rsmmr: I'll postpone this and create a ticket for it.

hilti/src/rt/tests/stream.cc
```diff
@@ -659,10 +658,35 @@ TEST_CASE("Trim with existing iterator and append") {
...
+
+    i += 10;
+    j += 2;
+    x.trim(j);
+    x.append("2345678901"_b);
+
+    CHECK_EQ(*i, '0');
+}
+
+TEST_CASE("Trim to beyond end") {
```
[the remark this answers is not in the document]
rsmmr: it trims further into the space we don't have yet, and then ensure the offsets are right. could probably be combined but I'll just leave as it is.

hilti/CMakeLists.txt
```diff
@@ -322,13 +323,18 @@ add_executable(hilti-rt-tests
                src/rt/tests/util.cc
                src/rt/tests/vector.cc)
 target_compile_options(hilti-rt-tests PRIVATE "-Wall")
-target_link_libraries(hilti-rt-tests PRIVATE hilti-rt-objects doctest)
+target_compile_options(hilti-rt-tests PRIVATE $<$<CONFIG:Debug>:-O0>)
+# TODO: add the following two lines tpo Spicy RT testt
```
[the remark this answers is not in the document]
rsmmr: Ah, that's an oversight, was supposed to be added!

hilti/include/rt/types/stream.h
```diff
@@ -112,65 +142,230 @@ class Chunk {
...
+    Chunk _fromArray(Offset o, std::array<Byte, N> d) {
+        if constexpr ( N <= Chunk::SmallBufferSize )
+            return Chunk(o, std::move(d));
+
+        return Chunk(o, Chunk::Vector(d.begin(), d.end()));
+    }
+
+
+    inline Chunk _fromArray(Offset o, const char* d, Size n) {
+        auto ud = reinterpret_cast<const Byte*>(d);
```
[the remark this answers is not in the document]
rsmmr: will leave as is

hilti/include/rt/types/stream.h
```diff
@@ -742,113 +1260,133 @@ inline std::ostream& operator<<(std::ostream& out, const View& x) {
...
+     * Assigns from another stream instance. This invalidates all existing iterators.
      * @param other the stream instance to assign from
      */
     Stream& operator=(Stream&& other) noexcept {
         if ( &other == this )
             return *this;
 
-        _content = std::move(other._content);
-        _frozen = other._frozen;
+        _chain->invalidate();
```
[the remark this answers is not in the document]
rsmmr: reason is that we need to invalidate the *previous* chain in case there are still iterators.

hilti/include/rt/types/stream.h
```diff
@@ -742,113 +1260,133 @@ inline std::ostream& operator<<(std::ostream& out, const View& x) {
...
+        _chain = other._chain->deepCopy();
         return *this;
     }
 
-    ~Stream() = default;
+    /** Destructor. */
+    ~Stream() {
+        assert(_chain);
+        _chain->invalidate();
```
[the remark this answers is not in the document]
rsmmr: _chain will stick around longer than stream, we need to mark it as invalid here.

hilti/src/rt/types/stream.cc
rsmmr: no, invalidate needs to be called only when the parent stream gets destroyed.

rsmmr: actually, I think things were a bit mixed up here. I'm adding a a dedicated `reset()` method that sets chains back to their initial state.

hilti/include/rt/types/stream.h
rsmmr: Prefer to leave as it as that gives us more control.

hilti/src/rt/types/stream.cc
rsmmr: returns a `bytes` is good, but also a larger change. Will leave as is for now, but will create a ticket for it.

hilti/src/rt/library.cc
rsmmr: The change isn't strictly necessary, but I ended up trying this 3 separate times while doing performance measurement to see if it could make a difference. It didn't, but just so that we don't keep wondering if it might, I left it at `NOW`.

hilti/src/rt/backtrace.cc
rsmmr: seems rare enough to leave as is

hilti/src/rt/fiber.cc
rsmmr: yeah, Running can't really happen here I think, changing to leave out

hilti/src/rt/backtrace.cc
rsmmr: accidental change, will fix

hilti/include/rt/types/stream.h
```diff
@@ -112,65 +142,230 @@ class Chunk {
...
+    const Chunk* next() const { return _next.get(); }
 
-    const Byte* data(Offset o) const {
-        auto c = this;
+    auto last() const {
+        const Chunk* i = this;
+        while ( i && i->_next )
+            i = i->_next.get();
+        return i;
+    }
```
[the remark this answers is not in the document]
rsmmr: The Chunk API is generally not tested by itself (only indirectly through the other classes). I'll leave it at that, but we could add a set of tests there at some point.

hilti/include/rt/types/stream.h
```diff
@@ -112,65 +142,230 @@ class Chunk {
...
+            _state = State::Mutable;
+    }
+
+    // Returns the number of dynamic chunks allocated.
+    int numberOfChunks() const;
+
+private:
+    void _ensureValid() const {
+        if ( ! isValid() )
+            throw InvalidIterator("stream object no longer available");
```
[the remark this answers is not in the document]
rsmmr: It's hard to trigger directly, but I think it could indirectly through iterator/view. Prefer to keep the checks.

hilti/include/rt/types/stream.h
```diff
@@ -112,65 +142,230 @@ class Chunk {
...
+    }
+
+    void _ensureMutable() const {
+        if ( isFrozen() )
+            throw InvalidIterator("stream object can no longer be modified");
```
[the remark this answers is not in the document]
rsmmr: I removed the checks in Stream, we did it twice and adapted the exception.

hilti/src/rt/tests/stream.cc
```diff
@@ -886,6 +910,188 @@ TEST_CASE("View") {
...
+        auto i = view.begin();
+        i += 5;
+        view = view.trim(i);
+        CHECK_EQ(view, ""_b);
+        CHECK_EQ(stream, "012"_b);
+        stream.append("3456789"_b);
+        CHECK_EQ(view, "56789"_b);
+    }
+
+    SUBCASE("find - SafeIterator") {
```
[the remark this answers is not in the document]
rsmmr: This seems tricky enough to unify that I'll punt. :)

hilti/src/rt/types/stream.cc
rsmmr: Another a Chunk API piece, and it's implicitly tested through a bunch of other methods.

hilti/src/rt/types/stream.cc
rsmmr: done

bbannier (approved):

PR #502, opened 2020-09-21
Preparations for creating a 'spicy-runtime' package for Zeek

rsmmr:


hilti/runtime/include/json.h
```diff
@@ -10,7 +10,7 @@
 #pragma GCC diagnostic push
 #endif
 
-#include <nlohmann/json.hpp>
+#include <hilti/rt/3rdparty/nlohmann/json.hpp>
```
bbannier, line 13: Shouldn't this include the header from `3rdparty/json/include/nlohmann` instead?

More generally, I think directly including with our own 3rdparty headers makes this harder to repackage. Ideally We would set up CMake targets in `3rdparty/CMakeLists.txt` or directly `add_subdirectory` to use their predefined targets, and then make our own targets depend on them. This works for both compiled and header-only dependencies, and is exactly what we did for `json` before the changes here.

One benefit of that approach is that building against system-provided dependencies becomes much easier (patch CMake setup instead of patching a number of files). While I suspect right now nobody is interested in that, being able to unbundle dependencies is pretty much required to get into distributions via official package repos.

hilti/toolchain/src/base/util.cc
```diff
@@ -6,17 +6,27 @@
...
+// There's a bizarr "printf" inside the PathFind code. So we include the
+// source file here to be able to #define that away.
+#define __ignore__(x,y)
+#define printf __ignore__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-variable"
+#include <pathfind/src/PathFind.cpp>
+#pragma GCC diagnostic pop
+#undef printf
+#undef __ignore__
```
bbannier, line 28: Let's instead introduce a `pathfind.hpp` header which does this magic and then use that one exclusively.

3rdparty/LICENSE.3rdparty
```diff
@@ -62,7 +60,7 @@ SOFTWARE.
 
 
 ================================================================================
-For EnumClass (hilti/include/3rdparty/enum-class):
+EnumClass
```
bbannier, line 63: Should this instead be named `ArticleEnumClass-v2`?

cmake/FindZeekSpicy.cmake
```diff
@@ -0,0 +1,81 @@
...
+    string(REPLACE " " ";" spicy_include_directories "${spicy_include_directories}")
+    list(TRANSFORM spicy_include_directories PREPEND "-I")
+    string(REPLACE ";" " " spicy_include_directories "${spicy_include_directories}")
+
+    message(STATUS "Spicy compiler       : ${spicyz}")
+
+elseif ( NOT "${SPICY_RUNTIME_DIR}" STREQUAL "" )
+    # No Spicy toolchain available, but looks like we have the runtime installed.
+    if ( IS_DIRECTORY "${SPICY_RUNTIME_DIR}/runtime/include/spicy" )
+        message(STATUS "No spicyz, cannot recreate Spicy parsers")
```
bbannier, line 34: Is this a leftover? We do not check for `spicyz` here.

cmake/Util.cmake
```diff
@@ -30,6 +30,7 @@ function(install_headers src dst)
                           FILES_MATCHING PATTERN "*.h"
                                          PATTERN "*.hpp"
                                          PATTERN "*.hh"
+                                         PATTERN "3rdparty*" EXCLUDE
```
bbannier, line 33: Not sure how/if it was possible to include e.g., installed runtime headers in user code, and nothing seems to have changed here with this commit. Will we eventually need code here to install 3rdparty deps, or should we instead never use 3rdparty headers in public headers?

hilti/runtime/include/json.h
rsmmr: As we discussed, I'm going to leave this as it is. The `3rdparty` includes are now used only for functionality that is needed by the public runtime headers, and they are installed alongside those. That way we make sure the use our own versions, and not accidentally any system-wide installed versions. We could do this through CMake magic as well, but not sure it's worth the trouble right now given that usage is contained to just a handful of places.

3rdparty/LICENSE.3rdparty
rsmmr: ack

hilti/toolchain/src/base/util.cc
rsmmr: We use this only at this one place, it's really just an implementation detail.

cmake/FindZeekSpicy.cmake
rsmmr: Not a left-over, but a warning that we don't have the toolchain installed. I'll change wording to that, saying `spicyz` is confusing.

rsmmr:
@bbannier this is finally ready, take another look, just a few additional small commits at the end

bbannier (approved):

PR #523, opened 2020-09-29
Track data offsets during parsing

### LINKED ISSUE #494: Track data offsets during parsing
For some host applications (Wireshark) it can be useful to know the offsets into the data where fields were parsed from. We can store those inside the structs. Might need to be an optional feature that one would  enable on demand during parser generation.

`ParseError` can likewise provide the offset where parsing failed.

bbannier:
This closes #494.


bbannier:
I am unsure what's the issue with the `zeek_packages` CI step.

hilti/runtime/include/type-info.h
```diff
@@ -878,23 +882,34 @@ class Struct {
...
+    /**
+     * Returns the struct's field.
+     *
+     * @param includeInternal include internal fields
+     * */
+
+    auto fields(bool includeInternal = false) const {
+        std::vector<std::reference_wrapper<const struct_::Field>> fields;
+        std::copy_if(_fields.begin(), _fields.end(), std::back_inserter(fields),
+                     [=](const struct_::Field& f) { return includeInternal || ! f.isInternal(); });
```
rsmmr, line 894: style-nit 

```suggestion
    auto fields(bool include_internal = false) const {
        std::vector<std::reference_wrapper<const struct_::Field>> fields;
        std::copy_if(_fields.begin(), _fields.end(), std::back_inserter(fields),
                     [=](const struct_::Field& f) { return include_internal || ! f.isInternal(); });
```

hilti/runtime/src/tests/type-info.cc
```diff
@@ -54,14 +54,14 @@ namespace {
 const hilti::rt::TypeInfo __ti_Test_X =
     {"Test::X", "Test::X",
      hilti::rt::type_info::Struct(std::vector<hilti::rt::type_info::struct_::Field>(
-         {hilti::rt::type_info::struct_::Field{"i", &hilti::rt::type_info::int32, offsetof(Test::X, i)},
-          hilti::rt::type_info::struct_::Field{"s", &hilti::rt::type_info::string, offsetof(Test::X, s)},
-          hilti::rt::type_info::struct_::Field{"y", &type_info::__ti_Test_Y, offsetof(Test::X, y)}}))};
+         {hilti::rt::type_info::struct_::Field{"i", &hilti::rt::type_info::int32, offsetof(Test::X, i), false},
```
rsmmr, line 57: Want to set some to `true` and add a test for the `is_internal` filtering?

hilti/toolchain/src/compiler/codegen/types.cc
```diff
@@ -845,9 +845,9 @@ struct VisitorTypeInfoDynamic : hilti::visitor::PreOrder<cxx::Expression, Visito
                 accessor = fmt(", hilti::rt::type_info::struct_::Field::accessor_optional<%s>()",
                                cg->compile(f.type(), codegen::TypeUsage::Storage));
 
-            fields.push_back(fmt("hilti::rt::type_info::struct_::Field{ \"%s\", %s, offsetof(%s, %s)%s }",
+            fields.push_back(fmt("hilti::rt::type_info::struct_::Field{ \"%s\", %s, offsetof(%s, %s), %s%s }",
                                  cxx::ID(f.id()), cg->typeInfo(f.type()), cxx::ID(*n.typeID()), cxx::ID(f.id()),
-                                 accessor));
+                                 f.isInternal().has_value(), accessor));
```
rsmmr, line 850: That looks like a problem with the `isInternal()` API, it should really just return a boolean (instead of `auto`ing the AttributeSet::find()` result`. You could just make that change as part of this commit if you want; also for the other `is*` methods in `Field`. Or we can do some other time.

spicy/toolchain/include/ast/types/unit.h
```diff
@@ -60,13 +61,20 @@ class Unit : public hilti::TypeBase,
      */
     std::optional<unit::Item> field(const ID& id) const;
 
-#if 0
-    /** Returns all items of a given name. */
-    std::vector<unit::Item> items(std::string) const;
+    // FIXME(bbannier): docstring.
```
rsmmr, line 64: FIXME

spicy/toolchain/include/ast/types/unit-items/field.h
```diff
@@ -29,41 +29,44 @@ static inline Type adaptType(Type type, const std::optional<Expression>& repeat)
...
+          _index(index),
           _is_anonynmous(! id.has_value()),
           _engine(e),
           _args_start(6),
           _args_end(_args_start + static_cast<int>(args.size())),
           _sinks_start(_args_end),
           _sinks_end(_sinks_start + static_cast<int>(sinks.size())) {}
 
-    Field(const std::optional<ID>& id, Item item, Engine e, const std::vector<Expression>& args,
+    Field(const std::optional<ID>& id, uint64_t index, Item item, Engine e, const std::vector<Expression>& args,
```
rsmmr, line 62: It feels error-prone to have the caller put in charge of the indices; would be safer if the unit always did the numbering automatically behind the scenes. I see that it does that for `addItems()`, but not for the constructor. If I'm reading the patch right, that latter is because we'll be replacing fields later and need to set the index at that point, correct? Could we instead (1) leave the index out of the `Field` constructor and default it to zero, (2) change to Unit ctor to assign indices similar to `addItems()`, and (3) change `resolvefField` to preserve the original index for the new fields (through `new_field.setIndex(new_field, old_field.index())`?

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -493,12 +493,30 @@ struct ProductionVisitor
...
+
+        auto __offsets = builder::member(state().self, "__offsets");
+        auto cur_offset = builder::memberCall(state().cur, "offset", {});
+
+        // Since the offset list is created empty make sure the element for the currrent field is present.
+        auto index = builder()->addTmp("index", builder::integer(field->index()));
+        auto resize = builder()->addWhile(builder::lowerEqual(builder::size(__offsets), index));
+        pushBuilder(resize);
+        builder()->addMemberCall(__offsets, "push_back", {builder::null()});
+        popBuilder();
```
rsmmr, line 505: Shorter way to do this in this case:

```suggestion
        resize->addMemberCall(__offsets, "push_back", {builder::null()});
```

There's no need for `pushBuilder()` if we don't pass control elsewhere that doesn't know about the specific builder we want to use.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -493,12 +493,30 @@ struct ProductionVisitor
                 pushState(std::move(pstate));
             }
         }
+
+        auto __offsets = builder::member(state().self, "__offsets");
+        auto cur_offset = builder::memberCall(state().cur, "offset", {});
+
+        // Since the offset list is created empty make sure the element for the currrent field is present.
+        auto index = builder()->addTmp("index", builder::integer(field->index()));
+        auto resize = builder()->addWhile(builder::lowerEqual(builder::size(__offsets), index));
```
rsmmr, line 502: (Looks like a `vector::resize` method would be helpful that preinitializes fields with defaults. But no need to add that right now, this is fine.)

rsmmr (commented):
Some more suggestions:

(1) We now have the internal __offsets member in the final struct. Rather than having host applications access that directly,it'd be nicer to provide an accessor function as part of the Spicy runtime library that, given a struct and its type info, provides a reference to the array.

(2) a btest validating generated offsets would be good to add. Not trivial unfortunately, because we can't get to the offsets from inside Spicy. Ideas:

   1. Create a mini host application that prints out offsets
   2. Extend spicy-dump with an option to show offsets (this would be nice generally)
   3. Add ``spicy::unit_offsets`` library functions that can be called from  Spicy to retrieve the offsets() array for a unit (wouldn't be very helpful beyond testing because Spicy programs won't know the indices)

I'd vote for 2 I think.

(3) This introduces some overhead that not every host application will need. It would be neat if we could make that optional: have the host say if it wants offsets recorded. However, we don't have a way right now to pass such options into the Spicy-side. For HILTI, we have `hilti::driver::Options`, but there's nothing comparable for the Spicy codegen. Given that, this is probably an action item for some other time, but wanted to mention in case you have an idea.

spicy/toolchain/include/ast/types/unit.h
bbannier: Fixed, was one dead code.

hilti/runtime/src/tests/type-info.cc
bbannier: I added a dedicated test for that.

rsmmr:
@bbannier take a look at the two new commits

spicy/toolchain/src/ast/types/unit.cc
```diff
@@ -14,3 +16,35 @@ std::optional<type::unit::Item> type::Unit::field(const ID& id) const {
 
     return {};
 }
+
+struct AssignIndicesVisitor : public hilti::visitor::PreOrder<void, AssignIndicesVisitor> {
```
bbannier, line 20: Can we go with e.g., `AssignFieldIndicesVisitor` instead to make clear these are indices of fields?

spicy/toolchain/src/ast/types/unit.cc
```diff
@@ -14,3 +16,35 @@ std::optional<type::unit::Item> type::Unit::field(const ID& id) const {
...
+
+std::vector<type::unit::Item> type::Unit::assignIndices(std::vector<unit::Item> items) {
+    std::vector<unit::Item> new_items;
+    new_items.reserve(items.size());
+
+    AssignIndicesVisitor v(_next_index);
+
+    for ( auto&& item : items ) {
+        auto nitem = Node(std::move(item));
+        for ( auto&& c : v.walk(&nitem) )
```
bbannier, line 42: That's a much nicer API that the manual recursion.

bbannier (commented):
Thanks for the cleanups @rsmmr, this LGTM, but I cannot :+1: it since I opened the PR.

hilti/toolchain/include/compiler/driver.h
```diff
@@ -341,6 +341,20 @@ class Driver {
      */
     void printHiltiException(const hilti::rt::Exception& e);
 
+    /**
+     * Hook for derived classes to add more options to the getopt() option
+     * string.
+     */
+    virtual std::string hookAddCommandLineOptions() { return ""; }
```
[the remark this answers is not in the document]
rsmmr: I think eventually we'll do this differently, probably with a struct describing command line options. Let's leave fine-tuning this all until then.

hilti/toolchain/src/compiler/driver.cc
```diff
@@ -103,7 +107,8 @@ void Driver::usage() {
            "  -V | --skip-validation          Don't validate ASTs (for debugging only).\n"
            "  -X | --debug-addl <addl>        Implies -d and adds selected additional instrumentation "
            "(comma-separated; see 'help' for list).\n"
-           "\n"
+        << addl_usage
```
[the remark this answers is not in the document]
rsmmr: It's adding an empty-line separator for that reason. We can clean this up more some other time (see other comment below).

rsmmr:
@bbannier Had you looked already into why the packages test is failing?

rsmmr (approved):
Approving and assigned back to you for merge once CI clears.

bbannier:
> @bbannier Had you looked already into why the packages test is failing?

The issue is like suspected that given the API changes in this PR `spicy-tftp` does not build anymore, https://cirrus-ci.com/task/6598721674674176. It seems we would need to merge this PR with that CI task failing and then update the package out of band.

Going forward we should have a deployment task on `master` to trigger updates of downstream packages.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
rsmmr: Filed ticket for this.

rsmmr:
I just updated the spicy-tftp package, let's see if it passes now.

Yeah, we need to figure out a way to update both in parallel.

rsmmr (approved):
CI passes now, ready to merge.

PR #536, opened 2020-10-05
Change `Stream::data` to return `Bytes`.

### LINKED ISSUE #493: Change Stream::data() to return bytes
`Stream::data()` and `stream::View::data()` should return `Bytes` instead of `std::string`. The change will propagate upwards a bit, and it looks like it'll also need some header file shuffling due to cyclic dependencies.

bbannier:
Closes #493.

### DIFF
--- a/hilti/runtime/include/types/stream.h
+++ b/hilti/runtime/include/types/stream.h
@@ -1431,7 +1431,7 @@ public:
      * Returns a copy of the data the stream refers to. Depending on size of
      * the stream, this can be an expensive operation.
      */
-    std::string data() const;
+    Bytes data() const;
 
     bool operator==(const Bytes& other) const { return view() == other; }
     bool operator==(const Stream& other) const { return view() == other.view(); }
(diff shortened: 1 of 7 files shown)


hilti/runtime/include/types/stream.h
```diff
@@ -778,7 +778,19 @@ class UnsafeConstIterator {
...
 
-    Byte _dereference() const { return *_chunk->data(_offset); }
+    Byte _dereference() const {
+        // TODO(bbannier): Ideally we would `assert(_chunk)` here so clang-tidy
+        // knows that `_chunk` is always not null. Unfortunately it looks like
+        // that fails to it pruning that CFG edge so we have to return instead
+        // here. This should be inconsequential to users as this function must
+        // not be called if the data is invalid.
+        if ( ! _chunk )
+            return 0;
```
rsmmr, line 788: How about throwing an internal error exception instead?

rsmmr (dismissed):

rsmmr (approved):

PR #537, opened 2020-10-06
Do not copy libraries on JIT or loading

### LINKED ISSUE #521: HLTO libraries are always copied to temp dir, even if compiled ahead of time
When creating `Library` instances, we always copy the libraries to a temp file in the system temp directory and load the library in the copied to location. This allows us to treat JIT'ed and libraries compiled ahead of time the same way since JIT'ed libraries might currently go away after being created.

This can be cumbersome to use for a number of reasons:

1. if the system temp directory is mounted `noexec`, the library cannot be loaded and executed from the temp file. Users are forced to overwrite `TMPDIR` or similar to be able to execute any Spicy code
1. for code compiled ahead of time we create copies we do not really need.

(1) could be fixed by making it possible to customize the location where the temporary libraries would be stored. On most non-fatal execution paths we already clean up created files, but we would still need some additional garbage collection of leftover files on that path.

(2) could be fixed by properly tracking `Library` instances created during JIT instead of creating and loading them, and then indiscriminately dropping handles to them. This would also allow us to properly unload unused libraries. Once properly tracked the need to create copies for JIT'ed libraries would go away (and was never needed for libraries compiled ahead of time). 

Fixing (2) might be the proper, long-term fix which incidentally also avoids introducing another knob.

bbannier:
Closes #521.

### DIFF
--- a/zeek/plugin/src/plugin.cc
+++ b/zeek/plugin/src/plugin.cc
@@ -322,8 +322,11 @@ int plugin::Zeek_Spicy::Plugin::HookLoadFile(const LoadType type, const std::str
 
     if ( ext == ".hlto" ) {
         try {
-            if ( auto load = hilti::rt::Library(file).open(); ! load )
-                hilti::rt::fatalError(hilti::rt::fmt("could not open library file %s: %s", file, load.error()));
+            if ( ! _libraries.count(file) ) {
+                _libraries.insert({file, hilti::rt::Library(file)});
+                if ( auto load = _libraries.at(file).open(); ! load )
+                    hilti::rt::fatalError(hilti::rt::fmt("could not open library file %s: %s", file, load.error()));
+            }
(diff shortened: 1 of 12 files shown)


zeek/plugin/src/plugin.cc
```diff
@@ -322,7 +322,9 @@ int plugin::Zeek_Spicy::Plugin::HookLoadFile(const LoadType type, const std::str
 
     if ( ext == ".hlto" ) {
         try {
-            if ( auto load = hilti::rt::Library(file).open(); ! load )
+            if ( ! _libraries.count(file) )
+                _libraries.insert({file, hilti::rt::Library(file)});
+            if ( auto load = _libraries.at(file).open(); ! load )
```
rsmmr, line 327: Should the `load()` be part of the previous `if` to run at most once?

rsmmr (dismissed):
I like this, looks more natural.

PR #539, opened 2020-10-08
Add support for new %print hook to Spicy units.

rsmmr:
This allows a unit to customize its appearance when printed (or more
generally: rendered into a string).


hilti/runtime/include/types/struct.h
```diff
@@ -27,8 +27,23 @@ inline auto& value_or_exception(const std::optional<T>& t, const char* location)
...
+
+template<typename T>
+constexpr auto has__str__helper(int) -> decltype(std::declval<T>().__str__(), std::true_type{});
+
+template<typename T>
+using has__str__ = decltype(has__str__helper<T>(0));
+
 template<typename T, typename std::enable_if_t<std::is_base_of<trait::isStruct, T>::value>* = nullptr>
 inline std::string to_string(const T& x, adl::tag /*unused*/) {
+    if constexpr ( has__str__<T>() ) {
```
bbannier, line 42: Can we add a test for this to the [to_string testsuite](hilti/runtime/src/tests/to_string.cc)? This should not be hard.

hilti/runtime/include/types/struct.h
```diff
@@ -27,8 +27,23 @@ inline auto& value_or_exception(const std::optional<T>& t, const char* location)
...
+constexpr std::false_type has__str__helper(long);
+
+template<typename T>
+constexpr auto has__str__helper(int) -> decltype(std::declval<T>().__str__(), std::true_type{});
+
+template<typename T>
+using has__str__ = decltype(has__str__helper<T>(0));
+
 template<typename T, typename std::enable_if_t<std::is_base_of<trait::isStruct, T>::value>* = nullptr>
 inline std::string to_string(const T& x, adl::tag /*unused*/) {
```
bbannier, line 41: I am wondering if we add similar code to `to_string_for_print` which could forward to the the same hook (I don't think there's value in users customizing that one separately).

spicy/toolchain/src/compiler/codegen/codegen.cc
```diff
@@ -428,13 +428,25 @@ std::optional<hilti::declaration::Function> CodeGen::compileHook(
     else if ( item_type )
         params.push_back({ID("__dd"), *item_type, hilti::type::function::parameter::Kind::In, {}});
 
-    auto rt = hilti::type::function::Result(hilti::type::Void());
-    auto ft = hilti::type::Function(std::move(rt), params, hilti::type::function::Flavor::Hook, meta);
-    auto hid = fmt("__on_%s%s", id.local(), (foreach ? "_foreach" : ""));
+    std::string hid;
+    Type result;
+
+    if ( id.local().str() == "0x25_print" ) {
```
bbannier, line 434: While we are at it, we could make `IDBase::str` from
```.cpp
auto str() const { return _id; }
```
to
```.cpp
const auto& str() const { return _id; }
```

spicy/toolchain/src/compiler/codegen/unit-builder.cc
```diff
@@ -163,6 +163,7 @@ Type CodeGen::compileUnit(const type::Unit& unit, bool declare_only) {
     add_hook("0x25_init", {});
     add_hook("0x25_done", {});
     add_hook("0x25_error", {});
+    add_hook("0x25_print", {});
```
bbannier, line 166: We'll need a doc update for this as well eventually.

tests/spicy/types/unit/print-hook.spicy
```diff
@@ -0,0 +1,52 @@
...
+    local f2 = "-";
+
+    if ( self?.f1 )
+        f1 = "%s" % self.f1;
+
+    if ( self?.f2 )
+        f2 = "%s" % self.f2;
+
+    return "%s|%s" % (f1, f2);
+}
```
bbannier, line 52: Can we add another test case for nested units? Given two units `A`, `B` and an `A` containing a `B`, test that if only `A`, only `B`, or both customizing their string repr, we get the expected result for an `A` defining
```.r
on A::%print {
  return "b: %s" % b;
}
```

bbannier (changes requested):

hilti/runtime/include/types/struct.h
rsmmr: Added to ``struct`` tests, that's where the current "to_string" test resides.

hilti/runtime/include/types/struct.h
rsmmr: `struct`` doesn't have a custom `to_string_for_print` currently, so seems we don't need that?

spicy/toolchain/src/compiler/codegen/unit-builder.cc
rsmmr: I added docs for `%print`, do you mean something else?

bbannier: That one is perfect, missed it.

hilti/runtime/include/types/struct.h
bbannier: Makes perfect sense.

tests/spicy/types/unit/print-hook.spicy
rsmmr: Good point, will do. Not quite sure actually what would be best for a subunit has %print with a parent that does not. Right now, that will use the sub's %print when the field is rendered inside the parent, which I think is ok, but could also see if differently (which would be harder though.)

bbannier: Naively I would also expect that if a subunit has a custom `%print` if would be use everywhere (e.g., also when the unit is wrapped; might want to add another test that this it also is used when `print` is invoked from outside any unit scope e.g., from a free, non-hook function).

The only issue I could see with that behavior is that there is no way to "remove" that print hook again from the outside to get back to the default behavior. But that is an issue of hooks in general which is why they generally shouldn't be part of grammars intended for reuse.

rsmmr: Added that tests as well and hopefully fixed CI with a missing baseline update.

PR #543, opened 2020-10-12
Add support for packet analyzers to Zeek plugin

### LINKED ISSUE #497: Add support for creating Zeek packet analyzers.
Zeek 4.0 is adding pluggable L2 analyzers, per https://github.com/zeek/zeek/pull/1046. Once that's landed, we can extend the Spicy plugin to support these.

rsmmr:
Zeek master got new support for adding low-level "packet analyzers"
through plugins. This extends the Spicy plugin to support that. We now
allow a new section "packet analyzer" in *.evt files that defines the
unit to use for parsing a packet:

    packet analyzer spicy::RawLayer:
        parse with RawLayer::Packet;

This is all needed to make the new analyzer available to Zeek. Usually
you will needs add two more pieces to make it useful:

1. If the analyzer just strips away a header to then forward the
   remaining data to another analyzer, you need to call
   ``zeek::forward(protocol)`` at some point during the unit's
   parsing. `protocol` is the identifier of the next protocol (per
   Zeek's dispatcher tables).

2. You need to configure the Zeek-side dispatcher tables, similar to
   any other packet analyzers. 

This needs Zeek master.

Full example in tests/zeek/evt/packet-analyzer.zeek.

No CI integration so far as we aren't testing Zeek master (4.0-to-be).

Closes #497


rsmmr:
Just as a status update, this is waiting for https://github.com/zeek/zeek/issues/1215

rsmmr:
This works with current master now.

It doesn't compile against older Zeeks yet, still needs some #ifdefs.

spicy/toolchain/bin/spicy-config.cc
```diff
@@ -222,6 +223,15 @@ int main(int argc, char** argv) {
 #endif
         }
 
+        if ( opt == "--zeek-version" ) {
+#ifdef HAVE_ZEEK
+            result.emplace_back(ZEEK_VERSION_NUMBER);
+#else
+            result.emplace_back("0");
```
bbannier, line 230: Can we make this an error instead of making up a value which looks valid?

zeek/plugin/src/plugin.cc
```diff
@@ -325,6 +354,28 @@ void plugin::Zeek_Spicy::Plugin::InitPostScript() {
         }
     }
 
+#ifdef HAVE_PACKET_ANALYZERS
+    for ( auto& p : _packet_analyzers_by_subtype ) {
+        ZEEK_DEBUG(hilti::rt::fmt("Registering packet analyzer %s with Zeek", p.name_analyzer.c_str()));
+
+        p.parser = find_parser(p.name_analyzer, p.name_parser);
+
+        auto instantiate = [&]() -> ::zeek::packet_analysis::AnalyzerPtr { return ::spicy::zeek::rt::PacketAnalyzer::Instantiate(p.name_analyzer); };
```
bbannier, line 363: It looks like `packet_analysis::Component` will store this longer than this function is alive. Could we copy in `name_analyzer` instead of keeping a ref to eliminate e.g., potential for use-after-free issues?

zeek/plugin/src/packet-analyzer.cc
```diff
@@ -0,0 +1,64 @@
...
+static auto create_packet_state(PacketAnalyzer* analyzer) {
+    cookie::PacketAnalyzer cookie;
+    cookie.analyzer = analyzer;
+    return PacketState(cookie);
+}
+
+PacketAnalyzer::PacketAnalyzer(std::string name)
+    : ::zeek::packet_analysis::Analyzer(std::move(name)), _state(create_packet_state(this)) {}
+
+PacketAnalyzer::~PacketAnalyzer() {}
```
bbannier, line 27: Can this be defaulted?

zeek/plugin/include/runtime-support.h
```diff
@@ -76,6 +85,13 @@ void register_protocol_analyzer(const std::string& name, hilti::rt::Protocol pro
 void register_file_analyzer(const std::string& name, const hilti::rt::Vector<std::string>& mime_types,
                             const std::string& parser);
 
+/**
+ * Registers an Spicy packet analyzer with its EVT meta information the
+ * plugin's runtime.
```
bbannier, line 90: Probably missing an `in` near near the end of the line,
```.cpp
 * Registers an Spicy packet analyzer with its EVT meta information in the
 * plugin's runtime.
```

zeek/plugin/include/plugin.h
```diff
@@ -106,6 +130,19 @@ class Plugin : public zeek::plugin::Plugin {
      */
     ::zeek::analyzer::Tag tagForFileAnalyzer(const ::zeek::analyzer::Tag& tag);
 
+#ifdef HAVE_PACKET_ANALYZERS
+    /**
+     * Runtime method to retrieve the analyzer tag that should be passed to
+     * script-land when talking about a packet analyzer. This is normally the
+     * analyzer's standard tag, but may be replaced with somethign else if
```
bbannier, line 137: `s/somethign/something/`

zeek/plugin/include/packet-analyzer.h
```diff
@@ -0,0 +1,77 @@
...
+
+#ifdef HAVE_PACKET_ANALYZERS
+
+#include <memory>
+#include <optional>
+#include <string>
+#include <utility>
+
+#include <hilti/base/util.h>
+#include <hilti/rt/types/stream.h>
```
bbannier, line 15: This doesn't seem sorted, could you check?

I get 
```.cpp
#include <hilti/rt/types/stream.h>

#include <spicy/rt/driver.h>
#include <spicy/rt/parser.h>

#include <hilti/base/util.h>

```

bbannier (approved):

zeek/compiler/CMakeLists.txt
```diff
@@ -10,6 +10,10 @@ target_include_directories(zeek-compiler BEFORE PUBLIC "${CMAKE_CURRENT_SOURCE_D
 target_compile_options(zeek-compiler PRIVATE "-fPIC")
 target_link_libraries(zeek-compiler PRIVATE spicy)
 
+if ( ${ZEEK_VERSION_NUMBER} GREATER_EQUAL 30300 ) # Zeek >= 3.3 (aka 4.0)
```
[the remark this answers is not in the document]
rsmmr: there are branch for 3.x maintenance, and we won't release 3.{3,4}, so not much of a problem.

doc/zeek.rst
```diff
@@ -191,6 +191,49 @@ As a full example, here's what a new HTTP analyzer could look like::
...
+As a full example, here's what a new analyzer could look like:
+
+    packet analyzer spicy::RawLayer:
+        parse with Raw Layer::Packet;
+
+In addition to the Spicy-side configuration, packet analyzers also need to be
+registered with Zeek inside a ``zeek_init`` event handler; see the
+`Zeek documentation <https://docs.zeek.org/en/master/frameworks/packet-analysis.html>`_
+for more. You will need to use the
+`PacketAnalyzer::try_register_packet_analyzer_by_name <TODO>`_ for registering
```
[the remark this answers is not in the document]
rsmmr: Docs weren't there yet, but they are now, updating.

spicy/toolchain/bin/spicy-config.cc
rsmmr: It's convenient this way because it allows to check against an expected Zeek version number even if Zeek isn't available at all (with the right result).

zeek/plugin/src/plugin.cc
rsmmr: ack

zeek/plugin/src/plugin.cc
```diff
@@ -120,6 +136,12 @@ const spicy::rt::Parser* plugin::Zeek_Spicy::Plugin::parserForFileAnalyzer(const
     return _file_analyzers_by_subtype[tag.Subtype()].parser;
 }
 
+#ifdef HAVE_PACKET_ANALYZERS
+const spicy::rt::Parser* plugin::Zeek_Spicy::Plugin::parserForPacketAnalyzer(const ::zeek::packet_analysis::Tag& tag) {
+    return _packet_analyzers_by_subtype[tag.Subtype()].parser;
```
bbannier, line 141: Like for the other analyzer stores we don't check or assert that the element `tag` is actually present in the vector. We could just return a `nullptr` for such cases.

bbannier (approved):

PR #547, opened 2020-10-20
Zeek plugin: Respect %port and %mime-type properties in units.

### LINKED ISSUE #95: Spicy plugin doesn't pick up on %port and %mime-type
We don't make use yet of any meta-data properties defined in a unit to specify well-known ports and MIME types.

rsmmr:
Unit-side %port and %mime-type may properties may be used instead of
specifying them inside the EVT file.

Closes #95.

### DIFF
--- a/tests/zeek/evt/import-from.zeek
+++ b/tests/zeek/evt/import-from.zeek
@@ -15,6 +15,8 @@ event ssh::test(x: string, y: string)
 module SSH;
 
 public type Banner = unit {
+    %port = 22/tcp;
+
     magic   : /SSH-/;
     version : /[^-]*/;
     dash    : /-/;
@@ -47,8 +49,7 @@ public function y()  : string {
 
 # @TEST-START-FILE ssh.evt
 protocol analyzer spicy::SSH over TCP:
-    parse with SSH::Banner,
-    port 22/tcp;
+    parse with SSH::Banner;
 
 import X;
 import Y from a.b.c;
--- a/zeek/plugin/src/plugin.cc
+++ b/zeek/plugin/src/plugin.cc
@@ -266,6 +266,16 @@ void plugin::Zeek_Spicy::Plugin::InitPostScript() {
             ZEEK_DEBUG(hilti::rt::fmt("  Scheduling analyzer for port %s", port));
             ::zeek::analyzer_mgr->RegisterAnalyzerForPort(tag, transport_protocol(port), port.port());
         }
+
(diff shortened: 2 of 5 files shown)


duffy-ocraven:
Are port-ranges rare, or is my experience unusual? Both the protocols where I have ever dealt with ports: BACnet and VoIP, use port-ranges rather than single ports.

rsmmr:
> Are port-ranges rare, or is my experience unusual? Both the protocols where I have ever dealt with ports: BACnet and VoIP, use port-ranges rather than single ports.

I'd say they are rare, yes, but I take your comment as that it would be useful to add support. If you think so, please file a separate feature request ticket.

zeek/plugin/src/plugin.cc
```diff
@@ -303,6 +313,14 @@ void plugin::Zeek_Spicy::Plugin::InitPostScript() {
...
                 ::spicy::zeek::compat::event_register_Register("spicy_analyzer_for_mime_type");
             ::spicy::zeek::compat::event_mgr_Enqueue(handler, vals);
+        };
+
+        for ( auto mt : p.mime_types )
+            register_analyzer_for_mime_type(tag, mt);
+
+        if ( p.parser ) {
+            for ( auto mt : p.parser->mime_types )
+                register_analyzer_for_mime_type(tag, mt);
```
bbannier, line 323: Do we have anything which ensures that we register a mime type exactly once? If not it might make sense to e.g., here track in a set which mime types are already registered.

tests/zeek/evt/import-from.zeek
```diff
@@ -47,8 +49,7 @@ public function y()  : string {
 
 # @TEST-START-FILE ssh.evt
 protocol analyzer spicy::SSH over TCP:
-    parse with SSH::Banner,
-    port 22/tcp;
+    parse with SSH::Banner;
```
bbannier, line 52: Could you add an additional test which ensures that a port in an evt file overrides whatever was already set in the parser?

zeek/plugin/src/plugin.cc
rsmmr: it's actually ok to register the same MIME type multiple times, Zeek is fine with that and triggers them all.

tests/zeek/evt/import-from.zeek
rsmmr: Same as above: this would register them multiple times, which is ok.

Issue #546, opened 2020-10-20
Speed up compilation of generated C++ code


rsmmr:
I ran some generated C++ code through `-ftime-trace` + [ClangBuildAnalyzer](https://github.com/aras-p/ClangBuildAnalyzer/blob/master/ClangBuildAnalyzer.ini). Full output below, I think there's some low-hanging fruit that could make quite some difference:

- `std::function` seems particular expensive
     - ~I believe we can just remove use of that from `type-info.h` and go with standard function pointers instead for the accessors.~ (#548)
     - ~In `fiber.h`, we can swap out `std::function` for something simpler that just stores a lambda for us. I did a quick POC with [this code](https://www.codeproject.com/Articles/313312/Cplusplus11-Lambda-Storage-Without-libcplusplus) and saw a noticeable speed-up.~ (#566)

- ~`std::variant` is expensive in `type-info.h`. We should be able to switch to a classic `union`, it's all static information anyways.~ (#566)

- ~`libhilti.h` is an expensive header that's included by all generated C++ files. I believe we should be able to speed that up by precompiling it into PCH. If we generally produce a PCH during builds and install it along with the our other headers, we can add `--include hilti/libhilti.h` to `hilti::Configuration::runtime_cxx_flags` and have it picked up. (Not sure if we need to treat Clang and GCC differently here).~ (#573)

Full output:

```
Analyzing build trace from 'profile'...
**** Time summary:
Compilation (6 times):
  Parsing (frontend):            8.5 s
  Codegen & opts (backend):      7.7 s

**** Files that took longest to parse (compiler frontend):
  2676 ms: a.out.73467.tmp.d/a.out.0.tmp.cc.o
  2644 ms: a.out.73467.tmp.d/a.out.1.tmp.cc.o
  1528 ms: a.out.73467.tmp.d/a.out.2.tmp.cc.o
   906 ms: a.out.73467.tmp.d/a.out.0.tmp.cc.o
   784 ms: a.out.73467.tmp.d/a.out.1.tmp.cc.o

**** Files that took longest to codegen (compiler backend):
  4064 ms: a.out.73467.tmp.d/a.out.0.tmp.cc.o
  3576 ms: a.out.73467.tmp.d/a.out.1.tmp.cc.o
    29 ms: a.out.73467.tmp.d/a.out.2.tmp.cc.o

**** Templates that took longest to instantiate:
[...]

PR #548, opened 2020-10-20
Use plain function pointers for type info accessors.

bbannier:
It turns out that std::function not only adds runtime overhead, but also
that it is significantly slower to compiler than a function pointer.

This commit partially addresses #546.

### DIFF
--- a/doc/host-applications.rst
+++ b/doc/host-applications.rst
@@ -315,5 +315,5 @@ Our examples always passed the full input at once. You don't need to
 do that, Spicy's parsers can process input incrementally as it comes
 in, and return back to the caller to retrieve more. See the source of
 `spicy::Driver::processInput()
-<https://github.com/zeek/spicy/blob/master/spicy/src/rt/driver.cc#L121>`_
+<https://github.com/zeek/spicy/blob/master/spicy/src/rt/driver.cc>`_
 for an example of how to implement that.
--- a/hilti/runtime/include/type-info.h
+++ b/hilti/runtime/include/type-info.h
@@ -158,7 +158,7 @@ public:
      * Type of a function that, given the outer value, returns a pointer to
      * the contained element.
      */
-    using Accessor = std::function<const void*(const Value& v)>;
+    using Accessor = const void* (*)(const Value& v);
 
     /**
      * Constructor.
@@ -291,9 +291,9 @@ public:
      * refers to.
      *
      */
-    using Accessor = std::tuple<std::function<std::optional<std::any>(const Value&)>,    // begin()
-                                std::function<std::optional<std::any>(const std::any&)>, // next()
-                                std::function<const void*(const std::any&)>>;            // deref()
+    using Accessor = std::tuple<std::optional<std::any> (*)(const Value&),    // begin()
+                                std::optional<std::any> (*)(const std::any&), // next()
+                                const void* (*)(const std::any&)>;            // deref()
 
     /**
      * Constructor.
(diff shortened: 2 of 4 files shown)


doc/host-applications.rst
```diff
@@ -315,5 +315,5 @@ Our examples always passed the full input at once. You don't need to
 do that, Spicy's parsers can process input incrementally as it comes
 in, and return back to the caller to retrieve more. See the source of
 `spicy::Driver::processInput()
-<https://github.com/zeek/spicy/blob/master/spicy/src/rt/driver.cc#L121>`_
+<https://github.com/zeek/spicy/blob/4e3fcb80d19aaa2b9b205134b8e2453348dbaf50/spicy/runtime/src/driver.cc#L144>`_
```
rsmmr, line 318: This risks showing outdated code at some point in the future. Not sure what's better: that, or showing wrong line numbers. Hard to catch either way.

hilti/runtime/include/type-info.h
```diff
@@ -180,7 +180,7 @@ class DereferencableType {
 
 private:
     const TypeInfo* _vtype;
-    const Accessor _accessor;
+    Accessor* _accessor;
```
rsmmr, line 183: Could this remains `const`?

Also, how about turning the `Accessor` typedef into a `pointer to <function>` (i.e., `using Accessor = const void* (*) (const Value& v);`, then we can pass it around as just `Accessor` just like we do further below for `Accessor` types that are tuples.

hilti/runtime/include/type-info.h
```diff
@@ -605,7 +605,7 @@ class Map {
 
     const TypeInfo* _ktype;
     const TypeInfo* _vtype;
-    const Accessor _accessor;
+    Accessor _accessor;
```
rsmmr, line 608: Same `const` question (and more below).

hilti/runtime/include/type-info.h
```diff
@@ -691,7 +691,7 @@ class MapIterator {
 private:
     const TypeInfo* _ktype;
     const TypeInfo* _vtype;
-    const Accessor _accessor;
+    Accessor* _accessor;
```
rsmmr, line 694: Should this really be a pointer? `Accessor` is a pair here, no?

rsmmr (changes requested):
Are you seeing a different in compile times?

doc/host-applications.rst
bbannier: I changed this to now link to only the file, not the line number.

hilti/runtime/include/type-info.h
bbannier: Accessor is a function `const Value&` -> `std::pair<const void*, const void*` so this is fine. I changed this to a function pointer type as well like you suggested above.

bbannier:
> Are you seeing a difference in compile times?

I tested with by compiling `hilti/runtime/src/tests/type-info.cc` which includes moderate use of this header and see a significant, but only moderate speedup:

- before: 3.617 s ±  0.018 s
- after: 3.505 s ±  0.033 s

Looking at the time profile a bigger win be possible from getting rid of the `std::variant` use in this header. I'll look into that next.

rsmmr:
Try comparing it for one of the btests, say `spicy/types/unit/sub-unit.spicy`

bbannier:
> Try comparing it for one of the btests, say `spicy/types/unit/sub-unit.spicy`

For that one the difference is bigger:

- before: (11.979 s ± 0.610) s
- after: (9.793 s ± 0.242) s

FTR, I first safed the output to a file, and then benchmarked generating an object file from that source file
```.sh
spicyc -c spicy/types/unit/sub-unit.spicy -o foo.cc
$(spicy-config --cxx --cxxflags --ldflags) -c foo.cc -o foo.o
```

rsmmr:
Great, thanks, that's noticeable.

rsmmr (approved):

PR #550, opened 2020-10-21
Series of time-related fixes.

rsmmr:


hilti/runtime/include/types/interval.h
```diff
@@ -36,8 +37,18 @@ class Interval {
...
      *
      * @param secs interval in seconds.
+     * @throws OutOfRange if *secs* cannot be represented with the internal resolution
      */
-    explicit Interval(double secs, SecondTag /*unused*/) : _nsecs(static_cast<int64_t>(secs * 1e9)) {}
+    explicit Interval(double secs, SecondTag /*unused*/)
+        : _nsecs([&]() {
+              auto x = secs * 1'000'000'000;
+
+              auto limits = std::numeric_limits<int64_t>();
```
bbannier, line 46: Might as well declare this as `constexpr` so we have a chance of generating less work at runtime.
```.cpp
               constexpr auto limits = std::numeric_limits<int64_t>(); 
```

hilti/runtime/include/types/time.h
```diff
@@ -36,14 +36,15 @@ class Time {
      * Constructs a time from a double value.
      *
      * @param secs seconds since the UNIX epoch.
+     * @throws OutOfRange if *secs* cannot be represented with the internal resolution
```
bbannier, line 39: The formatting of `secs` will be inconsistent with other fields. Let's either leave it plain (`secs` instead of `*secs*`) like in the line above, or format them all as code inside backticks.

spicy/toolchain/src/compiler/parser/parser.yy
```diff
@@ -888,12 +907,11 @@ ctor          : CADDRESS                         { $$ = hilti::ctor::Address(hil
 
               | const_real                       { $$ = hilti::ctor::Real($1, __loc__); }
               | CUINTEGER                        { $$ = hilti::ctor::UnsignedInteger($1, 64, __loc__); }
```
bbannier, line 909: I am confused why we check some uses of `CUINTEGER`, but not others (e.g., this one here). Shouldn't we range-check _all uses_ of `CUINTEGER`?

Here and elsewhere, similarly in the hilti parser.

hilti/runtime/include/types/time.h
rsmmr: actually it's not: `*...*` is used more often than backticks. Not saying that that's the better way of doing it, but currently we're inconsistent either way.

spicy/toolchain/src/compiler/parser/parser.yy
rsmmr: We only need to range check cases where the `CUINTEGER` then turns into signed value. If it's used as an actual unsigned going forward, it'll always be in range (because we're already storing it in an uint64 anyways when we arrive in parser.yy)

hilti/runtime/include/types/interval.h
rsmmr: ack

bbannier (approved):

PR #551, opened 2020-10-21
Provide error message when attempting to have spicy-driver compile  when JIT is not available

### LINKED ISSUE #490: spicy-driver should complain if it's not compiled with JIT
Currently it silently accepts `.spicy` files even if JIT is not available, and then just reports it has no parser to execute.

### LINKED ISSUE #507: Clarify error messaging when attempting JIT without support compiled in
Latest 0.4.0-884.dirty (master 6b77b11) spicy-driver immediately prints `[error] spicy-driver: no parsers available` Whereas running spicy-build on the same .spicy file correctly compiles and runs giving the correct output, but then that is followed by a print of `[error] spicy-driver: no parsers available`

rsmmr:
Closes #490.
Closes #507.

### DIFF
--- a/hilti/toolchain/src/compiler/driver.cc
+++ b/hilti/toolchain/src/compiler/driver.cc
@@ -430,6 +430,7 @@ void Driver::_addUnit(Unit unit) {
 
     hookNewASTPreCompilation(unit.id(), unit.path(), unit.module());
     _pending_units.emplace_back(std::move(unit));
+    _need_jit = true;
 }
 
 Result<void*> Driver::_symbol(const std::string& symbol) {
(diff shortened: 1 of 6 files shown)


hilti/toolchain/src/compiler/driver.cc
```diff
@@ -480,12 +480,14 @@ Result<Nothing> Driver::addInput(const std::filesystem::path& path) {
             _addUnit(std::move(*unit));
         }
 
+        _need_jit = true;
```
bbannier, line 483: I feel this mutation would really fit better in `_addUnit`, but on the other hand we don't have `_external_cxxs` not as well encapsulated, so that would introduce an inconsistency.

bbannier (approved):

hilti/toolchain/src/compiler/driver.cc
rsmmr: Good point, moving it there (while leaving it with _external_cxxs)

bbannier (dismissed):

PR #552, opened 2020-10-21
Various fixes/clarifications to diagnostic messages

rsmmr:


bbannier (approved):
Can you add another patch sorting the usage messages?

rsmmr:
@bbannier I added one more commit at the end.

spicy/toolchain/src/ast/types/unit-items/field.cc
```diff
@@ -37,12 +37,12 @@ static Type _adaptType(const type::unit::item::Field& field, const Type& t, bool
 
 Type spicy::type::unit::item::Field::parseType() const {
     if ( isContainer() ) {
-        Type etype = _originalType().as<type::Vector>().elementType();
+        Type etype = originalType().as<type::Vector>().elementType();
```
bbannier, line 40: While it looks like we cannot change `originalType` to return a `const T&` instead of a value (AFAICT it is ineligible for const lifetime extension), let's at least cache its return value here so we don't have to call it twice.

bbannier (approved):

spicy/toolchain/src/ast/types/unit-items/field.cc
rsmmr: Ack

doc/examples/_usage-spicy-config.output
```diff
@@ -6,20 +6,26 @@ Available options:
 
     --build                 Prints "debug" or "release", depending on the build configuration.
     --bindir                Prints the path to the directory where binaries are installed.
+    --cmake-path            Prints the path to Spicy-provided CMake modules
     --cxx                   Print the path to the C++ compiler used to build Spicy
     --cxxflags              Print flags for C++ compiler. (These are addition to any that HILTI needs.)
     --debug                 Output flags for working with debugging versions.
     --distbase              Print path of the Spicy source distribution.
+    --dynamic-loading       Adjust --ldflags for host applications that dynamically load precompiled modules
```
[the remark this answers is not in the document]
rsmmr: This just pulls the previous changes to spicy-config into the docs, not going to make any changes in this PR.

rsmmr:
> Can you add another patch sorting the usage messages?

They are sorted by short-option name.

PR #555, opened 2020-10-21
Remove --include-linker option from driver.

### LINKED ISSUE #500: Linker glue can introduce incorrect enum redeclarations
Given a module
```
module Foo;

type E = enum {
    A = 0x01
};

type P = unit {
    x: uint16 &convert=E($$);
};
```

the code generated from `spicyc -cK` will include a declaration for `E`
```.cpp
namespace __hlt::Foo {
    struct P;
    enum class E : int64_t { A = 1, Undef = -1 };
}
```
which the linker glue will attempt to forward-declare the enum as
```.cpp
using E = enum class E : int64_t { A = 1, Undef = -1 };
```

Attempting to compile that code will fail since `E` is redeclared.

Note that this redeclaration is only emitted if `E` is used like in above unit.

rsmmr:
This option used to print the linker glue code along with the main,
generated C++ code. The problem is that the linker code was produced
as if it were a standalone C++ unit. However, by concatening it with
the main code, that didn't hold and we'd run into problems like
duplicaste type definitions. This wasn't salvagable easily, and
conceptually it looks kind of wrong anyways, so removing it.

The better way to do all this is using "-l" to generate *just* the
linker code, then compile the two together as two compilation units.
If one insists, one can also still "cat" the two and will be back to
what "-cK" used to produce, with the same issues obviously.

Closes #500.

### DIFF
--- a/doc/examples/_usage-spicyc.output
+++ b/doc/examples/_usage-spicyc.output
@@ -9,7 +9,6 @@ Options controlling code generation:
   -D | --compiler-debug <streams> Activate compile-time debugging output for given debug streams (comma-separated; 'help' for list).
   -e | --output-all-dependencies  Output list of dependencies for all compiled modules.
   -E | --output-code-dependencies Output list of dependencies for all compiled modules that require separate compilation of their own.
-  -K | --include-linker           With --output-c++, include HILTI linker glue code.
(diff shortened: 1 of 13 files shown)


bbannier (approved):
Makes sense to me.

PR #557, opened 2020-10-22
Provide has_label() method for enum values.

### LINKED ISSUE #499: Validating domain of enum fields seems not very streamlined
When parsing units with enum-type fields it seems users should always write validation for a `&convert` function to produce a non-`Undef` value, i.e., that the encountered value is indeed in the domain of the enumerator.

Given a module
```
module Foo;

type P = unit {
  foo: uint16 convert=E($$);
  switch(self.foo) {
    E::A -> bar: uint16;
    # Seems impossible to put a case for `E::Undef` which triggers an error here.
  };
};

type E = enum { A = 0x01 };
```

The workaround seems to be to declare `foo` similar to this,
```
foo: uint16 convert=E($$) { if (self.foo==E::Undef) assert False; }
```

Looking at our example parsers we almost never perform such validation, but instead let a `ParseError` from the unhandled `Undef` case propagate up.

This seems weird as parsers almost everywhere else automatically enforce that parsed values are in the domain of the field being parsed. Naively one would expect a parse error to be triggered when parsing `foo` in above example, but this seems not possible since users cannot raise this exception.

I feel the most declarative approach would be to declare `foo: E` which could hook in automatic validation of the parsed data for `Undef`; alternatively we could allow users to raise parse errors themselves.

rsmmr:
Returns true if the enum maps to a known type value other than Undef.

Closes #499.

(This uses our new type information for the check. We should move enum
printing to that style as well, instead of generating dedicated
functions for each enum type.)

### DIFF
--- a/hilti/runtime/include/types/enum.h
+++ b/hilti/runtime/include/types/enum.h
@@ -1,0 +1,34 @@
+#pragma once
+
+#include <string>
+#include <variant>
+
+#include <hilti/rt/logging.h>
+#include <hilti/rt/type-info.h>
+
+namespace hilti::rt {
+
+namespace enum_ {
+
+/**
+ * Returns true if an enum value maps to a known label.
+ *
+ * @param t enum value
+ * @param ti type information corresponding to enum's type
+ * @tparam T enum type
+ */
+template<typename T>
+bool has_label(const T& t, const TypeInfo* ti) {
+    auto et = std::get_if<type_info::Enum>(&ti->aux_type_info);
+    if ( ! et )
+        internalError("unexpected type info in enum_::has_label");
+
+    for ( const auto& l : et->labels() ) {
+        if ( l.value != -1 && static_cast<int64_t>(t) == l.value )
+            return true;
+    }
+
+    return false;
+}
+} // namespace enum_
+} // namespace hilti::rt
--- a/hilti/toolchain/include/ast/operators/enum.h
+++ b/hilti/toolchain/include/ast/operators/enum.h
@@ -35,4 +35,17 @@ Instantiates an enum instance initialized from an integer value. The value does
     }
 END_CTOR
 
+BEGIN_METHOD(enum_, HasLabel)
+    auto signature() const {
+        return Signature{.self = type::constant(type::Enum(type::Wildcard())),
+                         .result = type::Bool(),
+                         .id = "has_label",
+                         .args = {},
+                         .doc = R"(
+Returns *true* if the value of *op1* corresponds to a known enum label (other
+than ``Undef``), as defined by it's type.
+)"};
+    }
+END_METHOD
+
 } // namespace hilti::operator_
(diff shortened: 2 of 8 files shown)


hilti/runtime/include/types/enum.h
```diff
@@ -0,0 +1,34 @@
...
+
+namespace hilti::rt {
+
+namespace enum_ {
+
+/**
+ * Returns true if an enum value maps to a known label.
+ *
+ * @param t enum value
+ * @param ti tyoe information corresponding to enum's type
```
bbannier, line 17: `s/tyoe/type/`

hilti/toolchain/include/ast/operators/enum.h
```diff
@@ -35,4 +35,17 @@ Instantiates an enum instance initialized from an integer value. The value does
...
 END_CTOR
 
+BEGIN_METHOD(enum_, HasLabel)
+    auto signature() const {
+        return Signature{.self = type::constant(type::Enum(type::Wildcard())),
+                         .result = type::Bool(),
+                         .id = "has_label",
+                         .args = {},
+                         .doc = R"(
+Returns *true* if the value of *op1* corresponds to a known enum label (other
```
bbannier, line 45: At least to me "label" in the context of enums doesn't immediately ring a bell, and I could imagine that this could be hard to discover for users. What do you think about instead using something like `is_know`? That should capture both the cases of uninitialized enum values and enum values filled with values outside of the expected range.

duffy-ocraven: On https://docs.zeek.org/projects/spicy/en/latest/programming/language/types.html
in 5.2.5.5, the use of the word label is widespread and consistent for this concept, i.e.

Enum types associate labels with a numerical values.
(I have just reported that grammar error, over in the docs pinned Issue)

The only other uses of label to mean something different are in 5.2.5.2
Each RANGE has one of the forms LABEL: A or LABEL: A..B where A and B are a bit numbers.
(should be
Each RANGE has one of the forms LABEL: A or LABEL: A..B where A and B are bit numbers.
I reported that grammar error too, over in the docs pinned Issue)

and one use of the word label in 5.2.5.21
`...you can initialize unit instances through coercion from a list expression: my_unit = [$FIELD_1 = X_1, $FIELD_N = X_N, ...] where FIELD_I is the label of a corresponding field in my_unit’s type.`

The uses of the word label in 5.2.5.2 and 5.2.5.21 can probably change (perhaps to: name of the member).

rsmmr: Yeah, I had been looking at 5.2.5.5 when picking the name, so it is consistent with the docs. We also use "label" at other locations with enums in the docs. Google also shows "label" being used with enums in a various contexts. May not be the most common terminology, but there's precedent. :-) So, I'll leave as is.

PR #560, opened 2020-10-23
Fix stack overflow during validation

rsmmr:


bbannier (approved):

hilti/toolchain/src/compiler/visitors/printer.cc
```diff
@@ -984,6 +984,15 @@ void hilti::detail::printAST(const Node& root, std::ostream& out, bool compact)
 void hilti::detail::printAST(const Node& root, printer::Stream& stream) {
     util::timing::Collector _("hilti/printer");
 
+    if ( auto t = root.tryAs<Type>() ) {
+        if ( ! stream.isExpandSubsequentType() ) {
+            if ( auto id = t->typeID() ) {
+                stream << *id;
+                return;
```
duffy-ocraven, line 991: Did you perhaps here mean:
```
        if ( ! stream.isExpandSubsequentType() ) {
            if ( auto id = t->typeID() ) {
                stream << *id;
            }
            return;
        }
```
in case the ->typeID() call returns something that coerces to false?

rsmmr: No, it's correct as written: a type may or may not have a type ID. If it has one, we want to print it (unless explicitly asked not to). If it doesn't, we fall back to the remainder of the method, which will print the full type. With your suggested change, we wouldn't print anything at all in that case, which would leave a gap in the output.

But valid question, thanks for watching!

tests/Baseline/hilti.ast.imported-id/output
```diff
@@ -267,12 +267,12 @@
...
 [debug/ast-scopes]           - declaration::Parameter %3 <is_struct_param="false" kind="in"> (bar.hlt:9:20)
 [debug/ast-scopes]             - ID <name="bar"> (bar.hlt:9:25)
-[debug/ast-scopes]             - type::ResolvedID <resolved="%10"> (type: string) (bar.hlt:9:20) (non-const)
+[debug/ast-scopes]             - type::ResolvedID <resolved="%10"> (type: Bar::Bar1) (bar.hlt:9:20) (non-const)
 [debug/ast-scopes]               - ID <name="Bar::Bar1"> (bar.hlt:9:20)
 [debug/ast-scopes]             - node::None (foo.hlt:9:20)
 [debug/ast-scopes]           - declaration::Parameter %4 <is_struct_param="false" kind="in"> (bar.hlt:9:29)
 [debug/ast-scopes]             - ID <name="foo"> (bar.hlt:9:40)
-[debug/ast-scopes]             - type::ResolvedID <resolved="%14"> (type: bool) (bar.hlt:9:29) (non-const)
+[debug/ast-scopes]             - type::ResolvedID <resolved="%14"> (type: Foo::Foo1) (bar.hlt:9:29) (non-const)
```
bbannier, line 275: Interesting side effect of the recursion fix, but I think exactly what we want.

bbannier (approved):

PR #562, opened 2020-10-23
Make generated grammar symbols globally unique during Spicy codegen.

### LINKED ISSUE #531: Look aheads which contain another look ahead fail to compile
This script...
```
module Test;

import spicy;

public type Line = unit {
  switch {
     -> a  : A;
     -> b  : B;
  };
};

type A = unit {
  switch {
    -> x : X;
    -> y : Y;
  };
};

type B = unit {
  val: /abc/;
};

type X = unit {
  val: /xyz/;
};

type Y = unit {
  val: /XYZ/;
};
```

Fails to compile with this error...
```
[error] /x/broken-la-in-la.spicy:2:1-30:3: redefinition of '__re_switch_lha_2' defined in /x/broken-la-in-la.spicy:2:1-30:3
[error] spicy-dump: aborting after errors
```

rsmmr:
With nested productions, we could end up generating colissions.

Closes #531.

### DIFF
--- a/spicy/toolchain/include/compiler/detail/codegen/codegen.h
+++ b/spicy/toolchain/include/compiler/detail/codegen/codegen.h
@@ -3,6 +3,7 @@
 #pragma once
 
 #include <memory>
+#include <string>
 #include <unordered_set>
 #include <utility>
 #include <vector>
@@ -10,6 +11,7 @@
 #include <hilti/ast/declarations/function.h>
 #include <hilti/ast/declarations/property.h>
 #include <hilti/ast/node.h>
+#include <hilti/base/uniquer.h>
 #include <hilti/compiler/unit.h>
 
 #include <spicy/ast/types/unit-items/field.h>
@@ -52,6 +54,7 @@ public:
     NodeRef preserveNode(Expression x);
     NodeRef preserveNode(Statement x);
     NodeRef preserveNode(Type x);
+    auto uniquer() { return &_uniquer; }
 
     const auto& moduleProperties() const { return _properties; }
     void recordModuleProperty(hilti::declaration::Property p) { _properties.emplace_back(std::move(p)); }
@@ -75,6 +78,7 @@ private:
     hilti::Node* _root = nullptr;
     std::vector<Declaration> _new_decls;
     std::unordered_set<ID> _decls_added;
+    hilti::util::Uniquer<std::string> _uniquer;
 };
 
 } // namespace spicy::detail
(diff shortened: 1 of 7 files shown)


spicy/toolchain/include/compiler/detail/codegen/codegen.h
```diff
@@ -52,6 +54,7 @@ class CodeGen {
     NodeRef preserveNode(Expression x);
     NodeRef preserveNode(Statement x);
     NodeRef preserveNode(Type x);
+    auto uniquer() { return &_uniquer; }
```
bbannier, line 57: Not a real issue here, but it would be nice if this method could return a `const` pointer instead to be more in line with the overall spirit of this class. That would probably require some changes to `Cache` so it can provide a `const` interface over internally `mutable` state.

Feel free to ignore.

spicy/toolchain/src/compiler/codegen/grammar-builder.cc
```diff
@@ -19,14 +19,14 @@ using hilti::util::fmt;
...
-    Visitor(codegen::GrammarBuilder* gb, Grammar* g) : gb(gb), grammar(g) {}
+    Visitor(CodeGen* cg, codegen::GrammarBuilder* gb, Grammar* g) : cg(cg), gb(gb), grammar(g) {}
+    CodeGen* cg;
     codegen::GrammarBuilder* gb;
     Grammar* grammar;
 
     using CurrentField = std::pair<const spicy::type::unit::item::Field&, std::reference_wrapper<hilti::Node>&>;
     std::vector<CurrentField> fields;
     hilti::util::Cache<std::string, Production> cache;
-    hilti::util::Uniquer<std::string> uniquer;
```
bbannier, line 29: We have another (`static`!) `Uniquer` in `type::unit::item::Field`. Do you think we could clean that one up as well as part of this work so we have uniquing always as a part of cg?

spicy/toolchain/include/compiler/detail/codegen/codegen.h
```diff
@@ -4,12 +4,14 @@
 
 #include <memory>
 #include <unordered_set>
+#include <string>
```
bbannier, line 7: Should be added before the previous line.

spicy/toolchain/src/compiler/codegen/grammar-builder.cc
rsmmr: Don't see an easy an easy way to unify those as field doesn't have access to a codegen. That's ok: if it only needs to be unique across fields, it's ok to leave there.

bbannier (approved):

PR #566, opened 2020-10-26
Model `TypeInfo` with a union instead of a `std::variant`.

bbannier:


hilti/runtime/include/type-info.h
```diff
@@ -1133,66 +1133,300 @@ class WeakReference : public detail::DereferencableType {
...
+        Stream,
+        StreamIterator,
+        StreamView,
+        String,
+        StrongReference,
+        Struct,
+        Time,
+        Tuple,
+        Union,
+        UnsignedInteger_uint8_t,
```
rsmmr, line 1186: let's just skip the `_t` postfix.

hilti/runtime/include/type-info.h
```diff
@@ -1133,66 +1133,300 @@ class WeakReference : public detail::DereferencableType {
...
+        type_info::ValueReference* value_reference;
+        type_info::Vector* vector;
+        type_info::VectorIterator* vector_iterator;
+        type_info::Void* void_;
+        type_info::WeakReference* weak_reference;
+    };
+
+    TypeInfo() = default;
+
+    template<typename Type>
```
rsmmr, line 1249: Would this template, along with the constexpr type checks, be adding compile-time overhead again? Could just provide a bunch of overloaded constructors instead?

spicy/toolchain/bin/spicy-dump/printer-json.cc
```diff
@@ -13,130 +15,139 @@ using json = nlohmann::json;
...
+        }
+        case TypeInfo::VectorIterator: return convert(type.vector_iterator->value(v));
+        case TypeInfo::Void: return "<void>";
+        case TypeInfo::WeakReference: {
+            auto y = type.weak_reference->value(v);
+            return y ? convert(y) : json();
+        }
+    }
+
+    throw RuntimeError("unhandled type");
```
rsmmr, line 152: move into `default` case? the compiler will probably not let this get executed anyways, but instead complain about a case not handled if added more type tag.

hilti/runtime/include/fiber.h
```diff
@@ -1,8 +1,5 @@
 // Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
 //
-// TODO(robin): This may not be the most efficient solution for implementing fibers
-// yet (due to using std::function, std::any)
-//
 // There are also at least two proposals for upcoming version of C++ that could implement this:
 //      - Coroutunes: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4680.pdf
```
rsmmr, line 4: We can remove this, too (co-routines won't work for us, and fibers aren't on the horizon).

hilti/runtime/include/fiber.h
```diff
@@ -48,6 +45,140 @@ using Handle = detail::Fiber;
 
 namespace detail {
 
+namespace lambda {
```
rsmmr, line 48: It'd be worth moving this out of fiber into a general utility class. I can see using this at other places, too, where we currently use std::function to capture a lambda.

rsmmr (approved):

hilti/runtime/include/type-info.h
bbannier: Looking at the compiler traces this doesn't seem to be costly (around 6ms total, place ~170).

spicy/toolchain/bin/spicy-dump/printer-json.cc
bbannier: Ideally I wouldn't have added this potentially dead code, but have relied on the compiler checking that the cases are exhaustive. I added it here since I believe GCC could warn that such a function wouldn't return a value.

I'd suggest don't use a `default` since it would mask the warning about an unhandled case.

hilti/runtime/include/fiber.h
bbannier: I'll move it into a separate file in the hilti runtime library.

rsmmr (approved):

PR #567, opened 2020-10-27
Add CI job for centos7

### LINKED ISSUE #451: Get Spicy to build on CentOS 7

bbannier:
This closes #451.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -98,53 +98,6 @@ clang10_zeek_lts_ubuntu_release_task:
       type: text/xml
       format: junit
 
-clang11_zeek_lts_ubuntu_release_task:
-  container:
-    dockerfile: ci/Dockerfile
-    docker_arguments:
-      - zeek_packages: zeek=3.2.2-* zeek-core-dev=3.2.2-*
-    cpu: 4
-    memory: 8G
-
-  timeout_in: 120m
-
-  always:
-    ccache_cache:
-      folder: /tmp/ccache
-      fingerprint_script: echo $CIRRUS_TASK_NAME-$CIRRUS_OS
-
-  env:
-    CCACHE_DIR: /tmp/ccache
-    CCACHE_COMPRESS: 1
-
-  # Pull tags as well since by default Cirrus CI does not fetch them, but they
-  # are needed for `git describe` used in `scripts/autogen-version`. We also
-  # pull submodules here.
-  update_git_script:
-    - git fetch --tags
-    - git submodule update --recursive --init
-
-  configure_script:      ./ci/run-ci -b build configure release --cxx-compiler clang++-11 --with-zeek /opt/zeek --clang-format `which clang-format-11` --clang-tidy `which clang-tidy-11`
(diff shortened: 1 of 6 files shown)


.cirrus.yml
```diff
@@ -379,6 +332,45 @@ docker_alpine_3_12_task:
     path: build/spicy-*.tar.gz
     type: application/gzip
 
+docker_centos_7_task:
```
bbannier, line 335: We should add the artifacts of this job to the docs.

bbannier:
FTR, if we do not link libstdc++ statically we run into issues with number formatting, even inside a devtoolset environment.

[log.txt](https://github.com/zeek/spicy/files/5492523/log.txt)

rsmmr:
> FTR, if we do not link libstdc++ statically we run into issues with number formatting, even inside a devtoolset environment.
> 
> [log.txt](https://github.com/zeek/spicy/files/5492523/log.txt)

Looks like hex floats are part of C99, I'm wondering if we could use libc for formatting them instead of libc++.

(I don't really like the hex floats to begin with, but they do have better precision semantics; we could also (in addition) add some logic to use the hex syntax only if we'd loose precision otherwise).

bbannier:
> > FTR, if we do not link libstdc++ statically we run into issues with number formatting, even inside a devtoolset environment.
> > [log.txt](https://github.com/zeek/spicy/files/5492523/log.txt)
> 
> Looks like hex floats are part of C99, I'm wondering if we could use libc for formatting them instead of libc++.
> 
> (I don't really like the hex floats to begin with, but they do have better precision semantics; we could also (in addition) add some logic to use the hex syntax only if we'd loose precision otherwise).

I tested this and hexfloat support via libc seems to work even on centos7. Added a patch to that effect.

hilti/toolchain/src/compiler/codegen/ctors.cc
```diff
@@ -92,8 +94,10 @@ struct Visitor : hilti::visitor::PreOrder<std::string, Visitor> {
     result_t operator()(const ctor::Port& n) { return fmt("hilti::rt::Port(\"%s\")", n.value()); }
 
     result_t operator()(const ctor::Real& n) {
-        // hexfloat format for lossless output, at most 13 fraction hexits
-        return fmt("%a", n.value());
+        constexpr size_t size = 256;
```
rsmmr, line 97: Worth a comment why we aren't using `fmt()` here.

hilti/toolchain/src/compiler/visitors/printer.cc
```diff
@@ -236,7 +238,12 @@ struct Visitor : visitor::PreOrder<void, Visitor> {
 
     void operator()(const ctor::Port& n) { out << n.value(); }
 
-    void operator()(const ctor::Real& n) { out << fmt("%a", n.value()); }
+    void operator()(const ctor::Real& n) {
```
rsmmr, line 241: same

rsmmr (approved):
Cool, thanks!

PR #572, opened 2020-11-02
 Use stable names for hook impls.

rsmmr:


hilti/toolchain/src/compiler/codegen/codegen.cc
```diff
@@ -565,3 +567,14 @@ cxx::Expression CodeGen::addTmp(const std::string& prefix, const cxx::Type& t) {
...
     return std::string(tmp.id);
 }
+
+cxx::ID CodeGen::uniqueID(const std::string& prefix, const Node& n) {
+    std::string x;
+
+    if ( auto l = n.location() )
+        x = fmt("%x", util::hash(l) % 0xffff);
+    else
+        x = fmt("%p", &n);
```
bbannier, line 577: This ID will change across across runs which will make the lead to e.g., ASTs which are not reproducible.

Can we instead use something guaranteed to be stable? The combined hashes of the parent nodes up to the root node seem like something which could work, but I am not sure whether that would always be unique.

bbannier (approved):

hilti/toolchain/src/compiler/codegen/codegen.cc
rsmmr: That's the problem: I don't think we something that's always unique, which is why I left the pointer in as a fall-back. Best would be if we could identify those places the fall-back hits, and make sure we have location objects for those instead.

rsmmr:
Turns out CI is still failing. I wasn't able to reproduce this anymore with the change in place, whereas I could easily without. Not sure what's going on right now.

hilti/toolchain/src/compiler/codegen/codegen.cc
bbannier: We discussed making nodes without location a hard error. Let's either do that, or at least add a comment here that we would like to do that eventually.

rsmmr: Will do.

bbannier (approved):

PR #585, opened 2020-11-13
Redo parsing of look-ahead vectors

rsmmr:
No test for the problem we used to have as I'm not sure how to test for that reliably (we'd need to trigger stack overflow)

### DIFF
--- a/spicy/toolchain/include/compiler/detail/codegen/productions/while.h
+++ b/spicy/toolchain/include/compiler/detail/codegen/productions/while.h
@@ -8,33 +8,70 @@
 
 #include <spicy/ast/types/unit.h>
 #include <spicy/compiler/detail/codegen/production.h>
+#include <spicy/compiler/detail/codegen/productions/look-ahead.h>
 
 namespace spicy::detail::codegen::production {
 
 /**
- * A production executing as long as a given boolean expression evaluates
- * to true.
+ * A production executing as long as either a given boolean expression evaluates
+ * to true, or, if no expression is provided, as determined by look-ahead symbols.
  */
 class While : public ProductionBase, public spicy::trait::isNonTerminal {
 public:
+    /**
+     * Constructor for a while-loop using an expression as the condition for termination.
+     */
     While(const std::string& symbol, Expression e, Production body, const Location& l = location::None)
-        : ProductionBase(symbol, l), _expression(std::move(e)), _body(std::move(body)) {}
+        : ProductionBase(symbol, l), _body(std::move(body)), _expression(std::move(e)) {}
 
-    const Expression& expression() const { return _expression; }
-    const Production& body() const { return _body; }
+    /**
+     * Constructor for a while-loop using look-ahead as the condition for
+     * termination. When using this constructor, `preprocessLookAhead()` must
+     * later be called with the grammar that the production has been inserted
+     * into.
+     */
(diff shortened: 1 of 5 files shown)


spicy/toolchain/include/compiler/detail/codegen/productions/while.h
```diff
@@ -8,33 +8,70 @@
...
 #include <spicy/compiler/detail/codegen/production.h>
+#include <spicy/compiler/detail/codegen/productions/look-ahead.h>
 
 namespace spicy::detail::codegen::production {
 
 /**
- * A production executing as long as a given boolean expression evaluates
- * to true.
+ * A production executing as long as either a given boolean expression evaluates
+ * to true, or, if no expression is provided, as determined by look-ahead symbols
```
bbannier, line 17:
```suggestion
 * to true, or, if no expression is provided, as determined by look-ahead symbols.
```

spicy/toolchain/include/compiler/detail/codegen/productions/while.h
```diff
@@ -8,33 +8,70 @@
...
     bool eodOk() const { return nullable(); }
     bool atomic() const { return false; }
     bool supportsSynchronize() const { return hasSize() || _body.supportsSynchronize(); }
-    std::string render() const { return hilti::util::fmt("while(%s): %s", _expression, _body.symbol()); }
+    std::string render() const;
 
 private:
-    Expression _expression;
     Production _body;
+    std::optional<Expression> _expression;
```
bbannier, line 73: Let's add a comment here that only one of these should be set.

spicy/toolchain/include/compiler/detail/codegen/productions/while.h
```diff
@@ -8,33 +8,70 @@
...
+     */
     While(const std::string& symbol, Expression e, Production body, const Location& l = location::None)
-        : ProductionBase(symbol, l), _expression(std::move(e)), _body(std::move(body)) {}
+        : ProductionBase(symbol, l), _body(std::move(body)), _expression(std::move(e)) {}
 
-    const Expression& expression() const { return _expression; }
-    const Production& body() const { return _body; }
+    /**
+     * Constructor for a while-loop using look-ahead as the condition for
+     * termination. When using this constructor, `preprocessLookAhead()` must
```
bbannier, line 29: Could we instead pass the grammar into this constructor to make it easier to guarantee that exactly on of `_expression` or `_body_for_grammar` is set?

spicy/toolchain/include/compiler/detail/codegen/productions/while.h
rsmmr: Actually `_body` is always set.

spicy/toolchain/include/compiler/detail/codegen/productions/while.h
rsmmr: AST classes should be instantiable without having a grammar (yet), that's why this is a little bit awkward.

spicy/toolchain/include/compiler/detail/codegen/productions/while.h
bbannier: I meant that only one of `_expression` or `_body_for_grammar` should be set making this some sort of union type.

PR #588, opened 2020-11-13
Finish support for hook priorities

rsmmr:


tests/hilti/types/function/hook-priorities.hlt
```diff
@@ -0,0 +1,42 @@
...
+
+module Bar {
+
+import hilti;
+import Foo;
+
+public function hook void Foo::f() &priority=-5 { hilti::print("f with priority -5"); }
+public function hook void Foo::f() &priority=15 { hilti::print("f with priority 15"); }
+
+
```
bbannier, line 28: nit: dead line

tests/hilti/types/function/hook-priorities.hlt
```diff
@@ -0,0 +1,42 @@
...
+@TEST-START-FILE foo.hlt
+
+module Foo {
+
+import hilti;
+
+declare public hook void f();
+
+public function hook void f() { hilti::print("f with priority 0"); }
```
bbannier, line 12: Could make this `f with default priority` to decouple from whatever we use there.

tests/spicy/types/unit/hooks-priorities.spicy
```diff
@@ -0,0 +1,36 @@
+# @TEST-EXEC:  echo 12345| spicy-driver -p A::tA %INPUT b.spicy >output
+# @TEST-EXEC:  btest-diff output
+
+module A;
+
+public type tA = unit {
+       a: bytes &size=5 {
+            print "A: prio 0", self.a;
```
bbannier, line 8: similarly: `default prio`

doc/programming/parsing.rst
```diff
@@ -215,7 +215,7 @@ position. Example:
     :show-with: foo.spicy
 
 As you can see, ``x`` receives 6 bytes of input, which it then turns
-into three 16-bit integers.
+into three a-bit integers.
```
bbannier, line 218: Why this change?

doc/programming/parsing.rst
```diff
@@ -143,7 +143,7 @@ chapter.
 Unit fields always have a type. However, in some cases a field's type
 is not explicitly declared, but derived from what's being parsed. The
 main example of this is parsing a constant value: Instead of a type, a
-field can specify a constant of a parseable type. The field's type
+field can specify a constant of a parse able type. The field's type
```
bbannier, line 146: I feel the old form was already correct.

doc/programming/parsing.rst
```diff
@@ -314,7 +314,7 @@ constraints on their values. ``EXPR`` must a be boolean expression
 that will be evaluated after the parsing for the field has finished,
 but before any hooks execute. If ``EXPR`` returns ``False``, the
 parsing process will abort with an error, just as if the field had
-been unparseable in the first place (incl. executing any :ref:`%error
+been untraceable in the first place (incl. executing any :ref:`%error
```
bbannier, line 317: Was this change intentional? I think I understood _unparseable, but _untraceable_ would appear here the only time in the user-searchable docs, and at least for me means something different.

doc/programming/parsing.rst
```diff
@@ -469,6 +469,18 @@ There are three locations where hooks can be implemented:
...
+If multiple implementations are provided for the same hook, by default
+it remains undefined in which order they will execute. If a particular
+order is desired, you can specify priorities for your hook
+implementations::
+
+      on Foo::v priority=-5 { ... }
+      on Foo::v priority=5 { ... }
+
+Implementations then execute in order of their priorities: The lower a
+priority value, the earlier it will execute. If not specified, a
```
bbannier, line 481: Ah I see, maybe disregard my comment from above on possibly leaving them unset by default is not the behavior we want. Do you have a specific use case for hooks with negative priority in mind? -- maybe something worth adding an example for.

hilti/toolchain/include/ast/attribute.h
```diff
@@ -72,19 +73,40 @@ class Attribute : public NodeBase {
...
+            if ( ! hasValue() )
+                return result::Error(hilti::util::fmt("attribute '%s' requires a string", _tag));
+
             if ( auto e = value().tryAs<expression::Ctor>() )
                 if ( auto s = e->ctor().tryAs<ctor::String>() )
                     return s->value();
             return result::Error(hilti::util::fmt("value for attribute '%s' must be a string", _tag));
         }
 
+        if constexpr ( std::is_same<T, int64_t>::value ) {
```
bbannier, line 95: We could use `std::is_integral_v<T>` here.

Originally I was about to suggest this so this function also works with `integer::safe<T>`, but it looks like it currently doesn't implement that trait. Might be worth being more general here nevertheless.

bbannier (changes requested):

hilti/toolchain/src/compiler/codegen/codegen.cc
```diff
@@ -244,6 +244,17 @@ struct Visitor : hilti::visitor::PreOrder<void, Visitor> {
             return;
         }
 
+        int64_t priority = 0;
```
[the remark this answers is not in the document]
rsmmr: 0 is a deliberate choice. We're mimicking semantics of Zeek's event priorities here: zero is default, and users can chose to lower or higher.

tests/hilti/types/function/hook-priorities.hlt
rsmmr: See above.

doc/programming/parsing.rst
rsmmr: Oops, some accidental doc changes slipped in, will revert.

hilti/toolchain/include/ast/attribute.h
rsmmr: Not sure, it's meant to get an `int64_t` ... :)

rsmmr:
Turns out I had the sorting wrong: Mimicking Zeek's event priorities means higher priority value executes first. Fixed.

bbannier (dismissed):

bbannier (approved):
I suspect that hooks with same (e.g., default) priority are executed in undetermined order will bite us in the future when writing unit tests when we forget to normalize the output with `sort`.

Do you think it would make sense to use `std::stable_sort` instead of `std::sort` to order hooks? Not sure that would promise users more than we want, but at least it should make these sorting changes unnecessary and introduce some determinism.

rsmmr:
> I suspect that hooks with same (e.g., default) priority are executed in undetermined order will bite us in the future when writing unit tests when we forget to normalize the output with `sort`.

Yeah, unfortunately. 

> Do you think it would make sense to use `std::stable_sort` instead of `std::sort` to order hooks?

Won't help unfortunately. I don't think we can (or at least should not) rely on the internal order that the implementations are going to come into the linker.

bbannier (approved):
LGTM but let's wait for CI.

PR #592, opened 2020-11-17
Let Zeek plugin search for HLTO files at startup.

rsmmr:


zeek/plugin/src/plugin.cc
```diff
@@ -389,21 +391,26 @@ void plugin::Zeek_Spicy::Plugin::Done() {
...
+void plugin::Zeek_Spicy::Plugin::loadModule(const hilti::rt::filesystem::path& path) {
+    try {
+        ZEEK_DEBUG(hilti::rt::fmt("Loading %s", path.native()));
+
+        if ( ! _libraries.count(path) ) {
+            _libraries.insert({path, hilti::rt::Library(path)});
+            if ( auto load = _libraries.at(path).open(); ! load )
+                hilti::rt::fatalError(hilti::rt::fmt("could not open library path %s: %s", path, load.error()));
+        }
+    } catch ( const hilti::rt::EnvironmentError& e ) {
```
bbannier, line 403: Why are we catching this specific exception here? Could we catch `Exception` instead?

bbannier (approved):

zeek/plugin/src/plugin.cc
rsmmr: Not sure, that's existing code that I just moved. I assumed it was due to the library loading in some way?

zeek/plugin/src/plugin-jit.cc
```diff
@@ -31,6 +32,7 @@ plugin::Zeek_Spicy::Plugin* ::plugin::Zeek_Spicy::OurPlugin = &SpicyPlugin;
 using namespace spicy::zeek;
 
 void ::spicy::zeek::debug::do_log(const std::string_view& msg) {
+    PLUGIN_DBG_LOG(*plugin::Zeek_Spicy::OurPlugin, "%s", std::string(msg).c_str());
```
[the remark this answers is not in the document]
rsmmr: `PLUGIN_DBG_LOG` comes from Zeek, and all Zeek headers come in through `zeek-compat.h`.

zeek/plugin/src/plugin.cc
```diff
@@ -32,7 +32,10 @@ plugin::Zeek_Spicy::Plugin* plugin::Zeek_Spicy::OurPlugin = &SpicyPlugin;
 using namespace spicy::zeek;
 
 #ifndef ZEEK_HAVE_JIT
-void spicy::zeek::debug::do_log(const std::string_view& msg) { HILTI_RT_DEBUG("zeek", msg); }
+void spicy::zeek::debug::do_log(const std::string_view& msg) {
+    PLUGIN_DBG_LOG(plugin::Zeek_Spicy::OurPlugin, "%s", msg);
```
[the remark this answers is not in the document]
rsmmr: same

zeek/plugin/src/plugin.cc
bbannier: I checked and it looks like nothing in the loading throws this exception today. This pattern was probably introduced because of this review comment: PRhttps://github.com/zeek/spicy/pull/317#pullrequestreview-414229770-body-html-edit-form (the comment on `library.cc`); previously we caught `std::runtime_error`. 

The same pattern is copied in `hilti/toolchain/src/compiler/driver.cc`.

Issue #561, opened 2020-10-23
Swap out fiber implementation


rsmmr:
The current fiber/co-routine implementation borrows low-level code from libtask for context switching that uses a fixed stack size per fiber. Two problems there:

1. We can't catch if a fiber runs out of stack space. We'll just crash in ways that probably open doors widely for all kinds of stack-smashing attacks.

2. Not being able to grow the stack dynamically means we need to estimate upfront how much stack fibers need max, and then always allocate the worst-case, which is very wasteful.

I've been looking around for alternatives to libtask's code. It's difficult to find something that's both (1) addressing those issues, (2) well maintained, (3) cross-platform (for what we need), and  (4) narrow enough to not pull in a bunch of additional infrastructure we don't need (e.g., a full-featured event scheduler).

The best candidate I've found so is https://github.com/hnes/libaco. I'd like to give that a try. It can use a shared stack for all fibers. That comes with a performance impact (because it copies stack frames around) but I can't tell how much impact that would have for us. It also protects its stack with guard pages against overflow to at least crash reliably.

bbannier:
It looks like libaco [does not support macos at this point](https://github.com/hnes/libaco/blob/fe379c1a5964783aef26685e1167312aa40e146b/acosw.S#L3-L4). Using different fiber implementations for different host platforms seems like something we wouldn't want to do, so I do not think that libaco works for us.

bbannier:
> It looks like libaco [does not support macos at this point](https://github.com/hnes/libaco/blob/fe379c1a5964783aef26685e1167312aa40e146b/acosw.S#L3-L4). Using different fiber implementations for different host platforms seems like something we wouldn't want to do, so I do not think that libaco works for us.

I looked more carefully and it when build correctly libaco actually does support macos. Since the including build setup broke with the `grep` on my macos, I checked this with the following cmake setup:
```.cmake
project(aco C ASM)

add_library(aco OBJECT acosw.S aco.c)

set(THREADS_PREFER_PTHREAD_FLAG ON)
find_package(Threads REQUIRED)

enable_testing()

set(tests "test_aco_tutorial_0;test_aco_tutorial_1;test_aco_tutorial_2;test_aco_tutorial_3;test_aco_tutorial_4;test_aco_tutorial_5;test_aco_tutorial_6;test_aco_synopsis;test_aco_benchmark")
foreach (test ${tests})
    add_executable(${test} ${test}.c)
[...]

PR #593, opened 2020-11-17
Move fiber runtime to libaco.

bbannier:


bbannier:
Closes #561.

bbannier:
I added a small benchmark which in addition to timing `hilti::rt::init` and `hilti::rt::done` exercises the following scenarios:

- `execute_one`: run a single fiber which immediately returns. This should be a baseline measurement.
- `execute_one_yield`: run a single fiber which first yields and returns on a subsequent run. This should measure the cost of resuming a fiber.
- `execute_yield_to_other`: construct two fibers, one like in `execute_one_yield` and a second one which resumes the first fiber. This should measure the cost of yielding to another fiber.

I ran this benchmark for three versions:

1) libtask-based baseline
2) libaco with private, per-fiber stacks
3) libaco with a single shared stack

The timing differences were:

    Comparing fiber_bench_0_baseline.json to fiber_bench_1_aco_private_stack.json
    Benchmark                                       Time             CPU      Time Old      Time New       CPU Old       CPU New
    ----------------------------------------------------------------------------------------------------------------------------
    init_mean                                    -0.0226         -0.0237          2423          2369          2415          2357
    init_median                                  -0.0209         -0.0207          2434          2383          2429          2379
    init_stddev                                  -0.2206         -0.2764           146           114           144           104
    done_mean                                    -0.0541         -0.0575          1584          1498          1589          1498
    done_median                                  -0.0355         -0.0426          1559          1504          1567          1500
    done_stddev                                  +0.2784         +0.2276            66            84            66            80
    execute_one_mean                             -0.0284         -0.0284          1127          1095          1125          1093
    execute_one_median                           -0.0132         -0.0143          1123          1108          1123          1107
    execute_one_stddev                           +2.3864         +2.2937            10            34            10            33
    execute_one_yield_mean                       -0.0869         -0.0856          1374          1255          1372          1255
    execute_one_yield_median                     -0.0869         -0.0860          1369          1251          1368          1251
    execute_one_yield_stddev                     -0.8039         -0.7997            49            10            49            10
    execute_yield_to_other_mean                  -0.0343         -0.0351          1556          1502          1556          1501
    execute_yield_to_other_median                -0.0419         -0.0429          1536          1471          1535          1469
    execute_yield_to_other_stddev                +0.3769         +0.3468            39            54            40            54

    ============================================================================================================================

    Comparing fiber_bench_0_baseline.json to fiber_bench_2_aco_shared_stack.json
    Benchmark                                       Time             CPU      Time Old      Time New       CPU Old       CPU New
    ----------------------------------------------------------------------------------------------------------------------------
    init_mean                                    +1.3162         +1.3031          2423          5613          2415          5561
    init_median                                  +1.3374         +1.3229          2434          5690          2429          5643
    init_stddev                                  +0.6884         +0.6865           146           246           144           243
    done_mean                                    +1.5900         +1.5787          1584          4103          1589          4099
    done_median                                  +1.6028         +1.5933          1559          4059          1567          4063
    done_stddev                                  +0.8200         +0.6643            66           120            66           109
    execute_one_mean                             +0.0152         +0.0129          1127          1144          1125          1139
    execute_one_median                           +0.0225         +0.0204          1123          1148          1123          1146
    execute_one_stddev                           +5.4060         +5.0395            10            64            10            61
    execute_one_yield_mean                       -0.0083         -0.0075          1374          1362          1372          1362
    execute_one_yield_median                     -0.0100         -0.0109          1369          1356          1368          1353
    execute_one_yield_stddev                     +0.4740         +0.4444            49            72            49            71
    execute_yield_to_other_mean                  -0.0706         -0.0706          1556          1446          1556          1446
    execute_yield_to_other_median                -0.0665         -0.0655          1536          1433          1535          1435
    execute_yield_to_other_stddev                +0.0411         +0.0178            39            41            40            41

Overall the timing differences seem small while with approach (3) we would still get the memory benefits of libaco.

rsmmr:
Couple quick suggestions on the benchmarks:

1. is this measuring just single fiber invocation each time? Would suggest a loop of a few 1000s or so, so that it dominates the baseline overhead of doing init/done.

2. try adding one version where the fiber uses a substantial stack. Could be just a local array of like 4kb.

bbannier:
> 1. is this measuring just single fiber invocation each time? Would suggest a loop of a few 1000s or so, so that it dominates the baseline overhead of doing init/done.

Yes, each of the benchmarks I showed results from above ran a single fiber to completion. With the benchmarking library I used only the part inside the benchmark loop `for (auto _: state) ...` was timed. Additionally, I also removed per-loop fiber setup time from the loop by manually stopping and resuming the clock (the benchmarks used `fiber::execute` instead of explicitly constructing and running `Resumable`s, but the results are not changed a lot by that). The library then takes care of running the loop repeatedly to obtain averaged results. That's the numbers I looked at and I think that should be fine.

> 2. try adding one version where the fiber uses a substantial stack. Could be just a local array of like 4kb.

I have done that now, but it looks like results where not much affected by that above the noise. This was surprising to me so I also repeated the benchmark without optimizations to make sure the stack variables were not removed and got the same result (I also checked with coverage information that the lines were hit).

What has a way stronger effect on the timings is to construct many fibers and measuring the per fiber time to resume (checked for fibers which immediately return, and for fibers which have already yielded once -- this could have an effect for especially the current implementation which does different work on the initial run and resumes). For larger N I get roughly constant time here with a shared libaco stack, while versions with a per-fiber stack seem to become slower the more fibers are active. This seems to tell me that the time to resume a fiber might be more influenced by cache failures than by the raw cost of the instructions. 

(In the first plot below the different points with same color and (x, y) were measured with different-sized `char` arrays on the stack, `size={1, 4096, 163840}`).
 
returning
yielding

bbannier:
Below the individual results for above microbenchmarks.

- [slightly cleaned up libtask version](https://github.com/zeek/spicy/files/5567670/0.log)
- [libaco with per-fiber, private stack](https://github.com/zeek/spicy/files/5567672/1.log)
- [libaco with global shared stack](https://github.com/zeek/spicy/files/5567673/2.log)

bbannier:
> > 1. is this measuring just single fiber invocation each time? Would suggest a loop of a few 1000s or so, so that it dominates the baseline overhead of doing init/done.

As we discussed offline, I had misunderstood what you meant here; you asked me to make sure we can actually measure these short times.

I looked into how the benchmark library measures these times, and it seems to try the benchmark loop with a number of fixed iterations, first to identify a good iteration count by trying larger and larger iteration counts, and then performing the actual measurement with the previously determined target iteration count. Verbose logs from such a run are:

```
-- LOG(2): Running execute_one<1> for 1
-- LOG(2): Ran in 1.1e-05/2.70751e-05
-- LOG(2): Running execute_one<1> for 10
-- LOG(2): Ran in 1.7e-05/1.81079e-05
-- LOG(2): Running execute_one<1> for 100
-- LOG(2): Ran in 0.000146/0.000148276
-- LOG(2): Running execute_one<1> for 1000
-- LOG(2): Ran in 0.00145/0.00144662
-- LOG(2): Running execute_one<1> for 10000
-- LOG(2): Ran in 0.013921/0.0174766
-- LOG(2): Running execute_one<1> for 100000
-- LOG(2): Ran in 0.122368/0.122543
-- LOG(2): Running execute_one<1> for 572045
-- LOG(2): Ran in 0.70823/0.712106
-----------------------------------------------------------------------------------------------------
Benchmark                                                           Time             CPU   Iterations
-----------------------------------------------------------------------------------------------------
execute_one<1>                                                   1245 ns         1238 ns       572045
```

Here it tried execution of the benchmarking loop with 1, 10, 100, 1.000, 10.000, and 100.000 iterations and ran the actual benchmark with 572.045 iterations determined from the last try loop. The benchmark report references the target iteration count, so it does not look like timings from different runs are mixed in one sample.

bbannier:
I looked into why we see no dependence of the time it takes to `Fiber::run` on the size of stack variables created in the lambda (e.g., see the points with same color in the first plot above).

The code executing on a `Fiber` is stored in a [member variable](https://github.com/zeek/spicy/blob/2cbe6323c8fee0d69aa7eb670c71ab2371cf4de7/hilti/runtime/include/fiber.h#L123) which effectively stores the equivalent of a `std::function`. With that approach the stack of the code executing on a `Fiber` is always stored on the heap; the "stack" of the `Fiber` as far as e.g., libaco is concerned is the stack of whatever lives in the trampoline loop which is independent and always of constant size.

rsmmr:
> With that approach the stack of the code executing on a `Fiber` is always stored on the heap; the "stack" of the `Fiber` as far as e.g., libaco is concerned is the stack of whatever lives in the trampoline loop which is independent and always of constant size.

I don't quite follow this. The trampoline runs with the same stack as the "payload" of the fiber, it's the trampoline that does the call into the lambda the function.

bbannier:
Proof-of-concept implementation using the fiber library is https://github.com/zeek/spicy/tree/topic/bbannier/fiber. This allows to run other coroutines directly from a coroutine without having to leave the coroutine like with libaco.

This library currently doesn't provide a shared stack, but this could be added, even from the outside.

In the benchmarks this shows similar performance as libaco with a shared stack.

bbannier:
I have opened #601 for using the fiber library instead.

bbannier:
Superseded by #601.

Issue #535, opened 2020-10-05
Add vector::resize() method


rsmmr:
In contrast to `reserve`, this would actually resize the vector to the given size, default-construcingt any new slots that remain empty. 

Once we have that, update  the `spicy.track_offsets` code in `parser-builder.cc` to use it.

rsmmr:
This has landed.

PR #595, opened 2020-11-18
Add new vector methods resize/sub/at.

### LINKED ISSUE #584: give vector an at() method
bytes, stream, and view  datatypes have .at() operation, but vector does not. We should be give vector an at() method which would return an iterator.

Also bytes and view datatypes have .sub() operations, but vector does not. We could provide sub(), but it could be expensive to copy the vector if it’s large (which however is true for bytes/stream as well).

rsmmr:
Closes #584 and #535.

(Also uses new resize() method at one place in parser builder.)

### DIFF
--- a/doc/autogen/types/vector.rst
+++ b/doc/autogen/types/vector.rst
@@ -6,6 +6,10 @@
     less than *i* elements a sufficient number of default-initialized
     elements is added to carry out the assignment.
 
+.. spicy:method:: vector::at vector at False <iterator> (i: uint<64>)
+
+    Returns an iterator referring to the element at vector index *i*.
+
 .. spicy:method:: vector::back vector back False <type~of~element> ()
 
     Returns the last element of the vector. It throws an exception if the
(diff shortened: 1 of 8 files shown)


doc/autogen/types/vector.rst
```diff
@@ -26,6 +30,22 @@
     change the vector in any observable way but provides a hint about the
     size that will be needed.
 
+.. spicy:method:: vector::resize vector resize False void (n: uint<64>)
+
+    Resizes the vector to hold exactly *n* elements. If *n* is larger than
+    the current size, the new slots are filled with default values. If *n*
+    is smaller than the current size, the excessive elements are removed.
+
+.. spicy:method:: vector::sub vector sub False vector (begin: uint<64>, end: uint<64>)
```
duffy-ocraven, line 39: Since you are here documenting the sub() and at() methods, would this be a good time to also there document the begin() and end() methods, which have long-been overlooked, and omitted.

hilti/runtime/include/types/vector.h
```diff
@@ -296,6 +296,53 @@ class Vector : protected std::vector<T, Allocator> {
...
+    }
+
+    /**
+     * Extracts a subsequence from the vector.
+     *
+     * @param from start index
+     * @param end end index (not including)
+     * @returns new vector with a copy of the range's elements
+     */
+    Vector<T> sub(uint64_t start, uint64_t end) {
```
bbannier, line 319: This could be `const`.

hilti/runtime/include/types/vector.h
```diff
@@ -296,6 +296,53 @@ class Vector : protected std::vector<T, Allocator> {
...
+        return v;
+    }
+
+    /**
+     * Extraces a subsequence from the vector.
+     *
+     * @param end end index (not including)
+     * @returns new vector with a copy of the elements from the beginning to *end*
+     */
+    Vector<T> sub(uint64_t end) {
```
bbannier, line 337: Could be `const`.

hilti/runtime/src/tests/vector.cc
```diff
@@ -88,6 +88,33 @@ TEST_CASE("assignment") {
     }
 }
 
+TEST_CASE("at") {
+        Vector<int> xs;
+        xs = Vector<int>({1, 2, 3, 4, 5});
```
bbannier, line 93: Could be simplified to

```.cpp
auto xs = Vector<int>({1, 2, 3, 4, 5});
```

Same in cases below.

bbannier (approved):

doc/autogen/types/vector.rst
rsmmr: Yeah, recorded in the documentation ticket already. I need to figure out how to get that into the autogenerated documentation (because it's a generic operator applied to different data types).

PR #598, opened 2020-11-24
Speed up the compiler

rsmmr:
This branch provides a major speed boost to the Spicy-side of the compiler: generating C++ code is much improved for more complex code. Most of the speed-up comes from a single change (the caching of operator results). The branch contains a series of other smaller improvements as well, including a couple to help with compile time of the toolchain itself.

Best reviewed by individual commits.

### DIFF
--- a/hilti/runtime/include/json-fwd.h
+++ b/hilti/runtime/include/json-fwd.h
@@ -1,0 +1,19 @@
+// Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
+
+#pragma once
+
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wtautological-overlap-compare"
+#elif defined(__GNUC__)
+// Note that clang #defines __GNUC__ as well.
+#pragma GCC diagnostic push
+#endif
+
+#include <hilti/rt/3rdparty/nlohmann/json_fwd.hpp>
+
+#if defined(__clang__)
+#pragma clang diagnostic pop
+#elif defined(__GNUC__)
+#pragma GCC diagnostic pop
+#endif
--- a/hilti/toolchain/include/base/intrusive-ptr.h
+++ b/hilti/toolchain/include/base/intrusive-ptr.h
@@ -1,0 +1,31 @@
+// Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
+
+#pragma once
+
+#include <utility>
+
+#include <hilti/rt/intrusive-ptr.h>
+
+namespace hilti {
+
+template<class T>
+using IntrusivePtr = hilti::rt::IntrusivePtr<T>;
+
+namespace intrusive_ptr {
+using AdoptRef = ::hilti::rt::intrusive_ptr::AdoptRef;
+using NewRef = ::hilti::rt::intrusive_ptr::NewRef;
+using ManagedObject = ::hilti::rt::intrusive_ptr::ManagedObject;
+} // namespace intrusive_ptr
+
+template<class T, class... Ts>
+IntrusivePtr<T> make_intrusive(Ts&&... args) {
+    // Assumes that objects start with a reference count of 1!
+    return {intrusive_ptr::AdoptRef{}, new T(std::forward<Ts>(args)...)};
+}
+
+template<class T, class U>
+IntrusivePtr<T> cast_intrusive(IntrusivePtr<U> p) noexcept {
+    return {intrusive_ptr::AdoptRef{}, static_cast<T*>(p.release())};
+}
+
+} // namespace hilti
--- a/hilti/toolchain/include/base/optional-ref.h
+++ b/hilti/toolchain/include/base/optional-ref.h
@@ -1,0 +1,85 @@
+// Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
+//
+// This was originally inspired by the version in
+// https://github.com/Chlorie/clu, but it turned into pretty much a rewrite. (That
+// code comes with an MIT license, FWIW).
+
+#pragma once
+
+#include <memory>
+#include <optional>
+#include <stdexcept>
+#include <utility>
+
+namespace hilti {
+
+/**
+ * Similar to `std::optional<>T` but storing a reference to the wrapped
+ * instance instead of a full copy. The caller must ensure that the
+ * underlying instance remains valid as long as necessary.
+ **/
+template<typename T>
+class optional_ref {
+public:
+    using nonConstT = typename std::remove_const<T>::type;
+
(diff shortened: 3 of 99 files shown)


hilti/toolchain/include/compiler/detail/cxx/unit.h
```diff
@@ -36,7 +37,7 @@ class Linker;
 
 namespace linker {
 
-using MetaData = nlohmann::json;
+using MetaData = hilti::rt::ValueReference<nlohmann::json>;
```
bbannier, line 40: We don't seem to need the shared behavior of `ValueReference`, so maybe we should use something like `std::unique_ptr` to make this work with fwd decls instead (makes construction slightly more ugly in an isolated number of places).

hilti/toolchain/include/base/intrusive-ptr.h
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
```
bbannier, line 1: Why do we need to duplicate this type into this namespace?

hilti/toolchain/include/base/optional-ref.h
```diff
@@ -0,0 +1,83 @@
+// From https://github.com/Chlorie/clu/tree/4d8bc5cfa780b9ba1d19d2e911c5bb80246a6fa4. Extended.
```
bbannier, line 1: Let's update `3rdparty/LICENSE.3rdparty`.

It would be nice if we could point out our modification, but that is probably easier said than done. If the modifications we might even be able to pull this in as is in `3rdparty/` as a full submodule (and then only use that single header). All of that should make updating this easier in the medium and long term.

hilti/toolchain/include/base/optional-ref.h
```diff
@@ -0,0 +1,83 @@
...
+class optional_ref final {
+private:
+    T* ptr_ = nullptr;
+
+public:
+    using nonConstT = typename std::remove_const<T>::type;
+    constexpr optional_ref() noexcept = default;
+    constexpr optional_ref(std::nullopt_t) noexcept {}
+    constexpr optional_ref(std::nullptr_t) noexcept {}
+    constexpr optional_ref(T& reference) noexcept : ptr_(std::addressof(reference)) {}
```
bbannier, line 20: Not sure on my rule of X, but does this allow constructing from r-value refs? If it does it would be great to delete that constructor to make it harder to construct dangling optional refs.

bbannier (dismissed):
This looks great, only left a couple nits.

While this patchset does indeed add caching, this acts more like filling in derived data/attributes. I wonder whether we should rename e.g., `clearCache` to `invalidateDerived` or similar instead.

I feel that ideally all this data should not live in ad hoc caches, but be embedded in the AST itself, e.g., in a `DerivedData` node under `Node`, or even as derived fields in `Node` itself. Resolving the AST would then include filling in all derived fields where an AST with no unset derived fields is fully resolved. Not using the name "cache" now could make that a little clearer.

hilti/runtime/include/json-fwd.h
```diff
@@ -0,0 +1,19 @@
...
+
+#pragma once
+
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wtautological-overlap-compare"
+#elif defined(__GNUC__)
+// Note that clang #defines __GNUC__ as well.
+#pragma GCC diagnostic push
+#endif
```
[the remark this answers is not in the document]
rsmmr: Hmm, not sure how I would do that? The one in `json.h` wraps including `nlohmann/json.hpp` whereas this one wraps around `nlohmann/json.hpp`.

hilti/toolchain/include/compiler/detail/cxx/unit.h
rsmmr: I wanted to keep the value semantics, not immediately sure if changing that would break anything. Sounds like you're saying it wouldn't?

hilti/toolchain/include/base/intrusive-ptr.h
rsmmr: For the most part, the toolchain code doesn't use the `hilti::rt:*` namespace directly, but imports what it needs. Not a strong reason for doing so, just that it doesn't feel quite right to use "runtime" inside the toolchain. We are not fully consistent on that, and I wouldn't fight to hard to abandon the policy, but that's why I did it here as well (admittedly, this one one particular ugly because of the complex declarations.)

hilti/toolchain/include/base/optional-ref.h
rsmmr: Ah, I meant to make a clean up pass over `optional-ref.h` to clean it up and see if there's  reasonable way to do the 3rd party thing. Forgot to do that, will look into it.

hilti/toolchain/include/base/optional-ref.h
rsmmr: ack, will do

hilti/runtime/include/json-fwd.h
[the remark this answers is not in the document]
bbannier: You are right, we not only push diagnostics here, but also pop them after the include, so including this header does nothing. 🤦

hilti/toolchain/include/compiler/detail/cxx/unit.h
bbannier: I didn't check, but I would expect that should mostly work as we don't seem to store `Metadata`. We'll probably would need fix function returns to use `std::move` though.

rsmmr:
> Resolving the AST would then include filling in all derived fields where an AST with no unset derived fields is fully resolved. Not using the name "cache" now could make that a little clearer.

That sounds like a nice model, but would require a bit of changes though in terms of assumptions being made on how to the AST can be used. Something to think about.

hilti/toolchain/include/compiler/detail/cxx/unit.h
rsmmr: So I started to make that change, but turns out it's completely trivial at least due to current ownership semantics (or lack thereof).Will just leave as is for now.

hilti/toolchain/include/base/optional-ref.h
rsmmr: As cleaned it up further, the move moved away more from the original version. Also, it feels like overkill to pull in the whole repository as a submodule just for this one pretty straight-forward file. So I went the opposite route and rewrote the header file as I'd normally written it for this code base. I think it's safe to declare this our own at this point (I still left a note for the source of inspiration in).

PR #601, opened 2020-11-29
Move fiber runtime to fiber library.

bbannier:


bbannier:
This is an alternative implementation for #561 (we already have #593 to move the fiber runtime to libaco). The benefit of using the fiber library is that in principle we can directly switch from one fiber to another. The downside is that the fiber library does not allow to use a shared stack by default (instead each fiber gets its own statically allocated stack).

hilti/runtime/src/fiber.cc
```diff
@@ -1,95 +1,206 @@
...
+            std::memcpy(static_cast<char*>(to->stack) - detail::fiber_stack_size(to) + to->stack_size,
+                        stored_stack->second._stack.get(), stored_stack->second._size);
+        else
+            HILTI_RT_DEBUG("fibers", fmt("stack for to=%p not found", to));
+    }
+    else
+        HILTI_RT_DEBUG("fibers", fmt("not copying stack for to=%p since it is the main_co", to));
+
+
+    return ::fiber_switch(fiber_switch_trampoline, to);
```
bbannier, line 111: This needs additional handling for ASAN. We probably need to move `Fiber`'s `asan` parameters into a subclass encapsulating both the `::Fiber` and the `asan` parameters.

Similarly for `fiber_switch` before it dispatches to this function.

bbannier: Currently the following tests fail:

```
spicy.types.sink.reassembler.gap
spicy.types.sink.reassembler.basic
spicy.types.sink.reassembler.skip
spicy.types.sink.reassembler.init-seq
spicy.types.sink.reassembler.overlap
spicy.types.unit.filter-chained
spicy.types.unit.params-forward
```

```
RESULT FILE .tmp/spicy.lib.protocols.http.gc-release/.stderr:
/bin/sh: line 1: 37803 Broken pipe: 13         awk -f generate.awk
     37804 Bus error: 10           | HILTI_DEBUG=spicy-driver spicy-driver http.hlto -i 83 -p HTTP::Replies > output 2> debug.log

RESULT FILE .tmp/spicy.types.sink.reassembler.basic/.stderr:
/bin/sh: line 1: 38973 Segmentation fault: 11  spicy-driver -p Mini::Main /Users/bbannier/src/spicy/tests/.tmp/spicy.types.sink.reassembler.basic/basic.spicy > output < /dev/null

RESULT FILE .tmp/spicy.types.sink.reassembler.gap/.stderr:
[libhilti] Internal error: code is executing that should not be reachable

--- Aborting in libhilti
/bin/sh: line 1: 58349 Abort trap: 6           (core dumped) spicy-driver -d -p Mini::Main /Users/bbannier/src/spicy/tests/.tmp/spicy.types.sink.reassembler.gap/gap.spicy > output < /dev/null

RESULT FILE .tmp/spicy.types.sink.reassembler.init-seq/.stderr:
/bin/sh: line 1: 39084 Segmentation fault: 11  spicy-driver -p Mini::Main /Users/bbannier/src/spicy/tests/.tmp/spicy.types.sink.reassembler.init-seq/init-seq.spicy > output < /dev/null

RESULT FILE .tmp/spicy.types.sink.reassembler.overlap/.stderr:
[libhilti] Internal error: code is executing that should not be reachable

--- Aborting in libhilti
/bin/sh: line 1: 39160 Abort trap: 6           spicy-driver -p Mini::Main /Users/bbannier/src/spicy/tests/.tmp/spicy.types.sink.reassembler.overlap/overlap.spicy > output < /dev/null

RESULT FILE .tmp/spicy.types.sink.reassembler.skip/.stderr:
[libhilti] Internal error: code is executing that should not be reachable

--- Aborting in libhilti
/bin/sh: line 1: 39290 Abort trap: 6           spicy-driver -p Mini::Main /Users/bbannier/src/spicy/tests/.tmp/spicy.types.sink.reassembler.skip/skip.spicy > output < /dev/null

RESULT FILE .tmp/spicy.types.unit.filter-chained/.stderr:
[libhilti] Internal error: code is executing that should not be reachable

--- Aborting in libhilti
/bin/sh: line 1: 44341 Done                    echo "H4sIAOVzEV0CA/NIzcnJ11EILshMrlQEACp6Q+YNAAAA"
     44342 Abort trap: 6           | spicy-driver -p Test::X /Users/bbannier/src/spicy/tests/.tmp/spicy.types.unit.filter-chained/filter-chained.spicy.hlto > output

RESULT FILE .tmp/spicy.types.unit.params-forward/.stderr:
[libhilti] Internal error: code is executing that should not be reachable

--- Aborting in libhilti
/bin/sh: line 1: 43385 Done                    printf xxx
     43386 Abort trap: 6           | spicy-driver -d -p Mini::Test /Users/bbannier/src/spicy/tests/.tmp/spicy.types.unit.params-forward/params-forward.spicy > output
```

bbannier:
Currently the following tests fail:

```
spicy.types.sink.reassembler.gap
spicy.types.sink.reassembler.basic
spicy.types.sink.reassembler.skip
spicy.types.sink.reassembler.init-seq
spicy.types.sink.reassembler.overlap
spicy.types.unit.filter-chained
spicy.types.unit.params-forward
```

```
RESULT FILE .tmp/spicy.lib.protocols.http.gc-release/.stderr:
/bin/sh: line 1: 37803 Broken pipe: 13         awk -f generate.awk
     37804 Bus error: 10           | HILTI_DEBUG=spicy-driver spicy-driver http.hlto -i 83 -p HTTP::Replies > output 2> debug.log

RESULT FILE .tmp/spicy.types.sink.reassembler.basic/.stderr:
/bin/sh: line 1: 38973 Segmentation fault: 11  spicy-driver -p Mini::Main /Users/bbannier/src/spicy/tests/.tmp/spicy.types.sink.reassembler.basic/basic.spicy > output < /dev/null

RESULT FILE .tmp/spicy.types.sink.reassembler.gap/.stderr:
[libhilti] Internal error: code is executing that should not be reachable

--- Aborting in libhilti
/bin/sh: line 1: 58349 Abort trap: 6           (core dumped) spicy-driver -d -p Mini::Main /Users/bbannier/src/spicy/tests/.tmp/spicy.types.sink.reassembler.gap/gap.spicy > output < /dev/null

RESULT FILE .tmp/spicy.types.sink.reassembler.init-seq/.stderr:
/bin/sh: line 1: 39084 Segmentation fault: 11  spicy-driver -p Mini::Main /Users/bbannier/src/spicy/tests/.tmp/spicy.types.sink.reassembler.init-seq/init-seq.spicy > output < /dev/null

RESULT FILE .tmp/spicy.types.sink.reassembler.overlap/.stderr:
[libhilti] Internal error: code is executing that should not be reachable

--- Aborting in libhilti
/bin/sh: line 1: 39160 Abort trap: 6           spicy-driver -p Mini::Main /Users/bbannier/src/spicy/tests/.tmp/spicy.types.sink.reassembler.overlap/overlap.spicy > output < /dev/null

RESULT FILE .tmp/spicy.types.sink.reassembler.skip/.stderr:
[libhilti] Internal error: code is executing that should not be reachable

--- Aborting in libhilti
/bin/sh: line 1: 39290 Abort trap: 6           spicy-driver -p Mini::Main /Users/bbannier/src/spicy/tests/.tmp/spicy.types.sink.reassembler.skip/skip.spicy > output < /dev/null

RESULT FILE .tmp/spicy.types.unit.filter-chained/.stderr:
[libhilti] Internal error: code is executing that should not be reachable

--- Aborting in libhilti
/bin/sh: line 1: 44341 Done                    echo "H4sIAOVzEV0CA/NIzcnJ11EILshMrlQEACp6Q+YNAAAA"
     44342 Abort trap: 6           | spicy-driver -p Test::X /Users/bbannier/src/spicy/tests/.tmp/spicy.types.unit.filter-chained/filter-chained.spicy.hlto > output

RESULT FILE .tmp/spicy.types.unit.params-forward/.stderr:
[libhilti] Internal error: code is executing that should not be reachable

--- Aborting in libhilti
/bin/sh: line 1: 43385 Done                    printf xxx
     43386 Abort trap: 6           | spicy-driver -d -p Mini::Test /Users/bbannier/src/spicy/tests/.tmp/spicy.types.unit.params-forward/params-forward.spicy > output
```

bbannier:
I have reset the changes here to only include the refactoring from libtask to the fiber library so we can tackle implementation of the shared stack in a follow-up.

bbannier:
This should in general be ready for review @rsmmr while I look into the sanitizer build failure.

hilti/runtime/include/global-state.h
```diff
@@ -69,6 +70,12 @@ struct GlobalState {
      * still need this information.
      */
     std::vector<hilti::rt::detail::HiltiModule> hilti_modules;
+
+    /** The main coroutine for fiber execution. */
+    std::optional<::Fiber> main_co;
+
+    /** The currently executing fiber .*/
+    ::Fiber* cur_co = nullptr;
```
rsmmr, line 78: I''m not quite seeing yet what we need `cur_co` for? Don't we always know the two fibers that we need to switch between (`main` && current fiber inside our ``Fiber`` class.

If we need it, it should also move into `Context`.

hilti/runtime/include/global-state.h
```diff
@@ -69,6 +70,12 @@ struct GlobalState {
      * still need this information.
      */
     std::vector<hilti::rt::detail::HiltiModule> hilti_modules;
+
+    /** The main coroutine for fiber execution. */
+    std::optional<::Fiber> main_co;
```
rsmmr, line 75: We should move this into `Context` instead of global. Right now it won't make much of a difference (there's just one global context) but in the future, if/when we introduce threads, there'd be one context per thread, and we'd want to tie fibers to that.

hilti/runtime/include/fiber.h
```diff
@@ -124,10 +115,16 @@ class Fiber {
...
+    /** The underlying coroutine of this fiber. */
+    ::Fiber _fiber;
+
+    /**
+     * The coroutine this fiber yields to.
+     *
+     * This is typically the coroutine of the fiber which invoked `run` on this
+     * coroutine.
+     */
+    ::Fiber* _parent = nullptr;
```
rsmmr, line 127: Is the parent for a fiber always the main fiber? If so, do we need store it given that have it globally available now?

bbannier: No, `_parent` is the fiber we yield to and is the fiber that triggered us.

hilti/runtime/src/fiber.cc
```diff
@@ -18,114 +18,91 @@
...
 
 using namespace hilti::rt;
-using namespace hilti::rt::detail;
 
 const void* _main_thread_bottom = nullptr;
 std::size_t _main_thread_size = 0;
 
-extern "C" {
+// A dummy function which will be put on the bottom of each fiber's call stack. This function should never execute.
+[[noreturn]] static void fiber_bottom(Fiber* fiber, void* args) { abort(); }
```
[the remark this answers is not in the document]
bbannier: Since functions executing on these fibers always have the same signature I think we do not need this to have C linkage (and it seems to pass CI), but other than that I am not certain.

hilti/runtime/CMakeLists.txt
```diff
@@ -94,6 +93,32 @@ target_compile_definitions(hilti-rt-tests-library-dummy1 PUBLIC RETURN_VALUE=1)
 add_library(hilti-rt-tests-library-dummy2 SHARED src/tests/library-dummy.cc)
 target_compile_definitions(hilti-rt-tests-library-dummy2 PUBLIC RETURN_VALUE=2)
 
+##### Link directories
```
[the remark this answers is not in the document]
bbannier: I think this might be the first externally visible library we add which is more than a single header. That means the library needs to be able to find its own headers and dependencies under its hardcoded include paths (e.g., `fiber.h` needs to be able to include `<fiber/fiber_mach.h>` or `<hu/lang.h>`). We can adjust paths of includes for our own code, but not this library (w/o forking). For that reason I put it into a "global" location.

An alternative to what I did here would be to add the 3rdparty include directory to the global include directories (e.g., reported by `hilti-config --cxxflags`). That would be messy in a different way. I am not sure there are other options.

hilti/runtime/CMakeLists.txt
```diff
@@ -111,7 +136,8 @@ install(CODE "file(REMOVE_RECURSE ${CMAKE_INSTALL_FULL_INCLUDEDIR}/hilti/rt/hilt
 
 # Install the 3rdparty headers that we need individually.
 install_headers(${CMAKE_SOURCE_DIR}/3rdparty/SafeInt hilti/rt/3rdparty/SafeInt)
-install_headers(${CMAKE_SOURCE_DIR}/3rdparty/libtask hilti/rt/3rdparty/libtask)
+install_headers(${CMAKE_SOURCE_DIR}/3rdparty/fiber/include/fiber fiber)
```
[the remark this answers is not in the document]
bbannier: See above.

hilti/runtime/include/fiber.h
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
 
 #pragma once
+#include <fiber/fiber.h>
```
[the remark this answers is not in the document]
bbannier: See above.

hilti/runtime/src/fiber.cc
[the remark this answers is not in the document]
rsmmr: I'm not quite sure who'd be doing the actual call into this function, if that's coming from the `fiber` C code (if it ever were called), I think it would need the "C" linkage.

hilti/runtime/include/global-state.h
rsmmr: Ah, of course! Nevermind.

hilti/runtime/src/fiber.cc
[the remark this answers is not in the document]
bbannier: I changed the linkage.

hilti/runtime/CMakeLists.txt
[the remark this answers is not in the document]
bbannier: By following the approach for justrx and by using a pimpl we are able to hide the dependency and I posted patches for that. With my approach `libhilti-rt.a` is still underlinked though:

```.sh
$ nm lib64/libhilti-rt.a | grep fiber_
                 U fiber_init_toplevel
                 U fiber_alloc
                 U fiber_destroy
                 U fiber_reserve_return
                 U fiber_switch
```

I built fiber with a `SHARED` configuration so `-fPIC` gets injected; I then consume the object file like in the case of justrx. With that e.g., `hilti-rt-tests` links and executes successfully since it directly links against the hilti object library, but `libhilti-rt.a` is still underlinked which shows in the btests which cannot find the symbol.

Do you have any idea what I am missing here?

hilti/runtime/include/fiber.h
```diff
@@ -124,10 +115,16 @@ class Fiber {
     std::optional<std::any> _result;
     std::exception_ptr _exception;
 
-    ucontext_t _uctx{};
-    jmp_buf _fiber{};
-    jmp_buf _trampoline{};
-    jmp_buf _parent{};
+    /** The underlying coroutine of this fiber. */
+    ::Fiber _fiber;
```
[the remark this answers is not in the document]
bbannier: Behind a pimpl now.

hilti/runtime/include/fiber.h
```diff
@@ -18,12 +17,18 @@
...
 namespace hilti::rt {
 
 namespace detail {
+extern "C" {
 void _Trampoline(void* argsp);
+}
 
 class Fiber;
+
+using FiberImpl = std::unique_ptr<::Fiber>;
```
rsmmr, line 31: I'd prefer to skip this typedef as it hides that `Fiber*` and `FiberImpl` are the same thing, just with different ownership semantics.

hilti/runtime/CMakeLists.txt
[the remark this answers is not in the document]
bbannier: Closing this as this should hopefully be addressed by the recently added patches.

.gitmodules
```diff
@@ -34,3 +31,6 @@
 [submodule "3rdparty/filesystem"]
 	path = 3rdparty/filesystem
 	url = https://github.com/gulrak/filesystem.git
+[submodule "3rdparty/fiber"]
+	path = 3rdparty/fiber
+	url = https://github.com/bbannier/fiber.git
```
bbannier, line 36: Change this to upstream if https://github.com/simonfxr/fiber/pull/1 gets merged, else create and use zeek fork.

bbannier:
@rsmmr, could you have a look? It looks like the minimal changes for global cleanups we discussed let this pass in CI.

rsmmr:
Nice, looks good. Let me take one more look at the full diff later, then I'll approve for merge.

rsmmr:
Oh, and it might need a rebase, the diff isn't clean.

hilti/runtime/include/context.h
```diff
@@ -45,6 +45,12 @@ struct Context {
      */
     resumable::Handle* resumable = nullptr;
 
+    /** The main fiber. */
+    std::unique_ptr<::Fiber> main_fiber;
+
+    /** The currently executing fiber .*/
+    std::unique_ptr<::Fiber>* current_fiber = nullptr;
```
rsmmr, line 52: Do we want leave this at being raw pointer? The `*` is easy to miss there with the `unique_ptr<T>` part of the same type. (C++ lacks a `std::` thing for a observing, non-managed pointer; that would make it clear here).

rsmmr (dismissed):
Just one more tiny comment, otherwise good to merge. This turned out nicely!

hilti/runtime/include/context.h
bbannier: I missed to clean this up what I removed the typedef for `unique_ptr<::Fiber>` and we don't gain anything by exposing how the `::Fiber` is managed. I'll switch to a raw ptr here and for `Fiber::_caller`.

PR #604, opened 2020-12-02
Add field declarations to Spicy.

rsmmr:
The following creates a field declaration:

    field Foo = bytes &size=5;

Then `Foo` can be used in place of a type inside unit fields:

    type MyUnit = {
        x: Foo;
    };

Now `x` will be parsed as a bytes field of 5 characters. Non-type field
specifications work as well (e.g. `field Foo = /Fo+Ba+r/;`), and hooks
may be included, too. See the documentation for more.


spicy/toolchain/src/compiler/parser/parser.yy
```diff
@@ -383,6 +385,16 @@ hook_decl     : ON unit_hook_id unit_hook        { ID unit = $2.namespace_();
                                                  }
               ;
 
+/* field_decl follows a similar set of rules as unit_field itself, see below. */
```
bbannier, line 388: It would be great if we could move the RHS-parts of `field_decl` and `unit_field` into its own rule to avoid duplication.

tests/spicy/types/unit/field-declaration.spicy
```diff
@@ -0,0 +1,27 @@
...
+field X5 = uint8 { print "in decl", $$; }
+
+public type Test = unit {
+    x1: X1;
+    x2: X2;
+    x3: X3;
+    x4: X4;
+    x5: X5 { print "in unit", $$; }
+    on %done   { print self; }
+};
```
bbannier, line 27: I could see users trying to pass a field decl as a function parameter, so I had a look at the following test:

```.ruby
function foo(x: X1) : uint64 {
    return 0;
}
```

This fails correctly but with a slightly ugly error which one can probably already trigger with the current code:
```
[error] ../tests/spicy/types/unit/field-declaration.spicy:29:14: ID 'X1' does not resolve to a N5hilti11declaration4TypeE (but to unit field)
```

* * *

I also checked whether one can import field decls from other modules, but it looks like this does not yet work:

```.ruby
// a.spicy
module a;

import b;

field I = int8;

type X = unit {
    i1 : I;
    i2 : b::I;
};
```

```.ruby
// b.spicy
module b;

field I = string; // `pub` not recognized here.
```

```
$ spicyc -c a.spicy
[error] a.spicy:9:5: unknown ID 'b::I'
```

spicy/toolchain/src/compiler/parser/parser.yy
rsmmr: they have different structure, don't think that would work without making it more messy

tests/spicy/types/unit/field-declaration.spicy
rsmmr: I'll look into the error message.

Forgot to add a test, but was expecting `public` to work, will give it a try.

rsmmr: Confirming that this is a bug: the `public` linkage isn't propagated (and the error message is misleading). I fixed that, however there's now another problem: ID lookups inside the imported module aren't resolved (because we copy the sub-AST over into the importing module). Need to look into how to fix that.

rsmmr:
Thanks for the feedback, I'm doing another iteration of this one, will integrate.

bbannier:
> Thanks for the feedback, I'm doing another iteration of this one, will integrate.

Great. Not sure why I put my comments on the test since they require extending this feature and its impl to work, sorry for that.

rsmmr:
Closting this for now. I have branch with some more progress, but it's proving difficult and will need more work.

PR #607, opened 2020-12-02
Add unit tests for IntrusivePtr.

### LINKED ISSUE #492: Add unit tests for IntrusivePtr
With  #491 came a new IntrusivePtr class (adapted from Zeek). We need more test coverage for that.

bbannier:
Closes #492.

### DIFF
--- a/hilti/runtime/src/tests/intrusive-ptr.cc
+++ b/hilti/runtime/src/tests/intrusive-ptr.cc
@@ -1,5 +1,7 @@
 // Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
 
+#include <utility>
+
 #include <hilti/rt/doctest.h>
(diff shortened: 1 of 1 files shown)


hilti/runtime/src/tests/intrusive-ptr.cc
```diff
@@ -35,4 +37,243 @@ TEST_CASE("managed objects") {
     CHECK_EQ(Managed::instances, 0);
 }
 
+struct TestObject : intrusive_ptr::ManagedObject {
+    static uint64_t references;
```
rsmmr, line 41: `s/references/instances/` ? I got confused with terminology for a second when looking at the tests (e.g., `TestObject::references == 1` when there are two intrusive ptrs)

rsmmr (approved):

PR #614, opened 2020-12-09
Implement a shared stack for fiber library

bbannier:
This PR bases on #601.

### DIFF
--- a/hilti/runtime/src/fiber.cc
+++ b/hilti/runtime/src/fiber.cc
@@ -35,21 +35,6 @@ void fiber_destroy(::Fiber* fbr) {
     return ::fiber_destroy(fbr);
 }
 
-size_t fiber_stack_size(::Fiber* fbr) {
-    assert(fbr);
-    assert(fbr->stack);
-
-    // FIXME(bbannier): this function should only return the actually used
-    // stack size. Confusingly it seem as if `::fiber_stack_size(fbr) -
-    // ::fiber_stack_free_size(fbr)` returns a way to big number. For now we
-    // return the full stack size here which below leads to us copying way to
-    // much data and not saving any memory.
-    return ::fiber_stack_size(fbr);
-    // return ::fiber_stack_free_size(fbr);
-    // return ::fiber_stack_size(fbr);
-    // return ::fiber_stack_size(fbr) - ::fiber_stack_free_size(fbr);
-}
-
 struct PrepareSwitchArgs {
     ::Fiber* fiber_switch_trampoline = nullptr;
     ::Fiber* from = nullptr;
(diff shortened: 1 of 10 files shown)


hilti/runtime/src/fiber.cc
```diff
@@ -1,131 +1,248 @@
...
+size_t fiber_stack_size(::Fiber* fbr) {
+    assert(fbr);
+    assert(fbr->stack);
+
+    // FIXME(bbannier): this function should only return the actually used
+    // stack size. Confusingly it seem as if `::fiber_stack_size(fbr) -
+    // ::fiber_stack_free_size(fbr)` returns a way to big number. For now we
+    // return the full stack size here which below leads to us copying way to
+    // much data and not saving any memory.
+    return ::fiber_stack_size(fbr);
```
simonfxr, line 47: @bbannier Out of interest I took a look at what you are doing with my fiber lib and it seems I introduced a bug in fiber_stack_free_size() when I changed the meaning of the Fiber.stack field, it should be fixed in the latest commit!

bbannier: Hi @simonfxr, great to see you here. Thanks for that fix, I was still wondering whether I was "holding it wrong".

bbannier: And since I have you here @simonfxr, what I am trying to do here is to implement a shared fiber stack. That stack could be huge, but by inspecting the actually used stack size one could get away with storing away only small amounts for inactive fibers.

We previously evaluated https://github.com/hnes/libaco which already supports a shared stack, but doesn't allow jumping between arbitrary fibers (fiber supports this); we felt that implementing this on top of fiber was more straightforward. Even with fiber, we probably still need to implement some trampoline fiber for `::fiber_switch` with its own, not shared stack which can take care of storing away the stack of the `from` `Fiber` and restoring the stack of the `to` `Fiber`. The good thing is that that could be completely opaque to users of the fibers if we provide a modified `::fiber_switch` implementation.

We currently plan to implement this here in C++ outside of the fiber library, but if this is something interesting for fiber itself or already planned we'd be more than interested.

simonfxr: spicy looks like a really cool project, I'm glad that fiber is useful to you!

Regarding the shared stack: Do you plan to have a dedicated shared execution stack for fibers only or do you want to reuse the system thread stack (openjdk project loom does this afaik)? You are going to need a fixed execution stack base address, its not clear to me how that should look like when using the thread stack, but maybe any value is good enough, e.g. defined implicityly by the first switch away from the toplevel fiber. 

I'm not sure if this shared stack feature should be in fiber itself, because it will need some way of managing off stack storage, but I like to be as less prescriptive as possible. But at least fiber could provide a new primitive to mount / unmount the fiber stack from the execution stack without the need for a trampoline / double switch.

bbannier:
Superseded by #624.

Issue #602, opened 2020-11-30
Switch Dockerfiles to use new stable download links for Zeek package


rsmmr:
See https://github.com/zeek/zeek/issues/1207#issuecomment-734942590

PR #620, opened 2020-12-10
Add paths to Spicy and Zeek to ubuntu Docker image PATHs.

bbannier:


rsmmr (dismissed):

bbannier:
I'll wait for #602 to land to make sure we can actually build the images.

PR #628, opened 2020-12-15
Updates for debian-10 Docker image

mmguero:
* updated Bison and CMake to latest versions
* removed "contrib non-free" from sources.list as it's not necessary for Zeek/Spicy to build and run

Thanks for adding the debian-10 image to your CI list. I saw the comments on those commits and made a few updates. I tested that the image builds and that basic Zeek stuff runs (zeek -N, etc.).

I'm in slack if you need to find me there. Cheers. -SG


bbannier:
Thanks for the updates @mmguero, I've included them in the branch.

Issue #582, opened 2020-11-12
Align CI platforms with Zeek


rsmmr:
Zeek's policy is documented [here](https://github.com/zeek/zeek/wiki/Platform-Support-Policy), we should do the same for our CI.

rsmmr:
The two (now closed) tickets referenced above already have starting points for adding Debian 10 and FreeBSD.

bbannier:
We require additional builds for the following platforms:

- debian-9
- debian-10
- fedora-32
- fedora-33
- freebsd-12.2
- freebsd-11.4
- ubuntu-20.04
- ubuntu-16.04

bbannier:
With #627 merged we still need to add the following platforms:

- macos Big Sur
- freebsd-12.2
- freebsd-11.4

bbannier:
Closing this as CI now covers macOS Big Sure, freebsd-11 and freebsd-12.

PR #632, opened 2020-12-17
Add FreeBSD to the CI

bbannier:
This work is related to #582.

### DIFF
--- a/ci/prepare_freebsd.sh
+++ b/ci/prepare_freebsd.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+echo "Preparing FreeBSD environment"
+sysctl hw.model hw.machine hw.ncpu
+set -e
+set -x
+
+env ASSUME_ALWAYS_YES=YES pkg bootstrap
+pkg install -y bash git cmake flex bison python3 ninja llvm11 zeek base64 ccache
+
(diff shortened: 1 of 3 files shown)


bbannier:
I filed #633 for the failure on freebsd-11.

rsmmr (dismissed):

ci/prepare_freebsd.sh
```diff
@@ -0,0 +1,18 @@
...
+pkg install -y bash git cmake flex bison python3 ninja llvm11 zeek base64 ccache
+
+pyver=$(python3 -c 'import sys; print(f"py{sys.version_info[0]}{sys.version_info[1]}")')
+pkg install -y "$pyver"-pip
+pip install btest
+
+# The Zeek port is currently incomplete, get the missing files from a checkout.
+# The branch we pick here roughly matches the version of zeek packaged in above
+# port.
+git clone --branch release/zeek-3.0 https://github.com/zeek/cmake /usr/local/share/zeek/cmake
```
bbannier, line 18: This shouldn't be needed anymore as the distribution package now includes these files

bbannier: Looks like the package is still not built. I'd suggest we wait a couple more days for it to become available instead of building it ourself.

bbannier: The package is now available.

rsmmr (approved):

rsmmr (approved):

PR #635, opened 2020-12-20
Collection of cleanups from consuming Spicy in external project

bbannier:


rsmmr (commented):
Looks good for ArticleEnumClass-v2 (I'm wondering if that dependency is worth keeping, given that we use it only at one place, but that's for some other time).

For `nlohman/json`, I'd prefer to consider that an implementation detail and avoid installing the headers (also because we may want to switch to some other library eventually). I'm wondering if we could limit use of JSON completely to our *.cc files. It's only shows up in a couple of headers right now, and I believe we could switch those to just passing a `std::string` around (or some strong typedef aliasing to `std::string` to keep the typing). Up for giving that a try?

bbannier:
> For `nlohman/json`, I'd prefer to consider that an implementation detail and avoid installing the headers (also because we may want to switch to some other library eventually). I'm wondering if we could limit use of JSON completely to our *.cc files. It's only shows up in a couple of headers right now, and I believe we could switch those to just passing a `std::string` around (or some strong typedef aliasing to `std::string` to keep the typing). Up for giving that a try?

I had a look at this. Changing the top-level functions `to_json` and `from_json` to work with `string` instead of `json` seemed straightforward.

The issue this probably introduces is for this to somewhat to work we need to serialize any object as JSON strings.

This approach breaks down when objects are composed into larger objects and so on since even in *.cc files we loose access to the automation nlohmann::json provides; while we can e.g., deserialize top-level objects into dictionaries of their fields, to deserialize them we would now need to 1) parse the serialized object (a JSON string) as a C++ string to remove any escaping the library might have added (needed for nesting) and 2) parse that C++ string again as an object. From what I have tried this composes badly, even for our own types, and also interacts poorly with nlohmann::json's support for C++ containers.

It seems that if we go down this route we loose a lot of the benefits using a library brings, and would almost be better of the write all the boilerplate code directly.

This seems pretty drastic as I feel this library is useful, so I think we should either hide the library behind a JSON interface we can own (unclear how much work that is), or maybe just leak the library in the interfaces (i.e., leave things as is for now). Not hiding this library seems acceptable to me, given that knowledge of it is not required to consume the Spicy API (we just provide hooks for nlohmann::json which in C++ are hard to hide).

rsmmr:
I don't fully follow (would just need to look more closely I'm sure), but I hear you that if this would turn into a larger changes, it's not worth it; better to then just go and install the headers. I'm fine with that then.

rsmmr (approved):

PR #636, opened 2020-12-21
Overhauling regexp matching semantics.

### LINKED ISSUE #631: Bug in regexp matching
Spotted by ASAN:

```
diff --git a/hilti/toolchain/include/ast/ctors/regexp.h b/hilti/toolchain/include/ast/ctors/regexp.h
index e2d9a0ee..ab29a56c 100644
--- a/hilti/toolchain/include/ast/ctors/regexp.h
+++ b/hilti/toolchain/include/ast/ctors/regexp.h
@@ -33,7 +33,7 @@ public:
             return true;

         for ( const auto& p : _patterns ) {
-            if ( auto i = p.find('('); i >= 0 && (i == 0 || p[i-1] != '\\') )
+            if ( auto i = p.find('('); i != std::string::npos && (i == 0 || p[i-1] != '\\') )
                 return false;
         }
```

This is regression introduced by the recent regex changes.

However, that now triggers test failures because it leads to using the minimal DFA matcher in some cases where it's not supported with `&nosub`. Need to look into how to fix that.

rsmmr:
[Probably challenging to review once more, but at least the runtime regexp.cc got simpler overall.]

We were inconsistent in how regular expressions (1) handled being
anchored vs not, and (2) allowed to retrieve the matching data. The main
underlying problem was that we determined at regexp compile time which
of the JRX matchers to use: the standard matcher (which allows to
retrieve the matching part as well as all capture groups), or the
minimal matcher which matches more efficiently but allows to retrieve
only the final position of a match (and no capture groups). Likewise, we
so far had to decide at compile time whether to consider an expression
anchored not not. These decisions upfront then meant that the actual
matching methods had to work with what they were handed, leading to
unintuitive restrictions and/or lack of efficiency.

This commits changes this as follows:

- We always compile regexps for the standard matcher (which the minimal
  matcher can use as well).

- We always compile regexps as anchored, no implicit `.*` at compile
  time anymore.

- We let the matching methods decide which type of matcher to choose
  (this required an update to justrx to extend its API accordingly; this
  commit includes the corresponding submodule update).

- We rename the regexp methods for more consistency: `match` now means
  "matching is anchored at the beginning of the data", and `find` means
  we find matches anywhere in the data. This leads to the following
  renames:

    * <regexp>.find -> <regexp>.match
    * <regexp>.find_groups -> <regexp>match_groups
    * <regexp>.find_span -> <regexp>.find

  `<bytes>.match` adapts to the new semantics as well.

  Same renaming happens at the C++ level.

- We remove the `&anchor` attribute because that's the default now.

- `&no_sub` remains available to override the default choice of the
  matcher type: it'll use the minimal matcher and disable capture group
  support.

- The one downside of all this is that the new `find` is a bit
  inefficient: because all our regexps are compiled anchored, it needs
  to try matching at all possible begin positions, meaning quadratic
  runtime. Doesn't seem like a big issue right now, but we could improve
  this in the future by compiling two DFAs for each regexp: anchored and
  non-anchored, and then using the standard matcher instead of the
  minimal matcher with `find`.

Closes #631.


hilti/runtime/src/types/regexp.cc
```diff
@@ -309,256 +311,110 @@ Vector<Bytes> RegExp::findGroups(const Bytes& data) const {
...
+        jrx_match_state ms;
+        auto rc = _search_pattern(&ms, cur, endp - cur, &so, &eo);
 
         if ( rc > 0 ) {
-            // Match.
-            acc = rc;
+            auto nlen = (eo - so);
+            auto olen = (cur_eo - cur_so);
+            so += (cur - startp);
+            eo += (cur - startp);
```
bbannier, line 334: Maybe add a comment (and maybe assertions) here mentioning that even if `so` and `eo` are in uninitialized state this is never off-by-one since `_search_pattern is guaranteed to set them to values >= 0 if there was a match,

```suggestion
           
            // If a match was found `_search_patterns` always initializes `so` and `eo` to valid values.
            assert(so != -1);
            assert(eo != -1);
            so += (cur - startp);
            eo += (cur - startp);
```

bbannier (approved):
Thanks for the detailed commit message which really help me see what is happening here.

I left only small stylistic nits below.

doc/autogen/types/regexp-match-state.rst
```diff
@@ -6,9 +6,10 @@
     where it left off last time. If *final* is true, this is assumed to be
     the final piece of data; any further advancing will then lead to an
     exception. Returns a 2-tuple with (1) a integer match indicator with
```
duffy-ocraven, line 8: `a integer` 
should be
`an integer`

doc/autogen/types/bytes.rst
```diff
@@ -36,7 +36,8 @@
 
     Matches the ``bytes`` object against the regular expression *regex*.
     Returns the matching part or, if *group* is given, then the
-    corresponding subgroup.
+    corresponding subgroup. The expression is considered anchored to the
```
duffy-ocraven, line 39: I am sure that every individual works out their mental model for comprehending the difference between match() and find(), in a manner such that there is no one way of expressing it perfectly for all audiences. But the term `anchored` that you use, could be combined with mention of the `^` regexp operator, then to likely  achieve wider correct comprehension for people whose conception in built in terms of that `^` regexp operator. So I suggest something like: `Returns the matching part or, if *group* is given, then the corresponding subgroup. The expression is considered anchored, as though it starts with an implicit ^ regexp operator, to the beginning of the data.`

hilti/toolchain/src/compiler/codegen/ctors.cc
```diff
@@ -119,9 +119,6 @@ struct Visitor : hilti::visitor::PreOrder<std::string, Visitor> {
     result_t operator()(const ctor::RegExp& n) {
         std::vector<std::string> flags;
 
-        if ( n.isAnchor() )
-            flags.emplace_back(".anchor = 1");
-
         if ( n.isNoSub() )
             flags.emplace_back(".no_sub = 1");
```
duffy-ocraven, line 123: .no_sub was revised from a :1 bitfield to a `bool` in the changes. Assigning it as `= 1` is unlikely to still be correct.
BTW, "`&no_sub` remains available to override the default choice" in the preamble is inaccurate, as the attribute is everywhere `&nosub`. (Maybe to reduce confusion, the struct member could be better similarly as .nosub).

hilti/runtime/src/types/regexp.cc
```diff
@@ -309,256 +311,110 @@ Vector<Bytes> RegExp::findGroups(const Bytes& data) const {
...
 
-#ifdef _DEBUG_MATCHING
-        std::cerr << fmt("-> rc=%d ms->offset=%d\n", rc, ms->offset);
-#endif
-        if ( use_stdmatcher && rc == 0 )
-            // No further match.
-            return acc;
+    for ( auto cur = data.data(); cur < endp; cur++ ) {
+        jrx_offset so = -1;
+        jrx_offset eo = -1;
```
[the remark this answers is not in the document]
rsmmr: let me add a comment instead, the `-1` are really just to not leave them uninitialized at all, they should always be set by the functional if they become relevant.

hilti/toolchain/src/compiler/codegen/ctors.cc
rsmmr: `= 1` still works, but I'll update (the change, btw, was due to bitfields not supporting initializing to default values, that's a C++20 feature).

rsmmr: I leave the spelling as is as well, prefer it this way, even if not consistent between attribute and C++.

Issue #441, opened 2020-07-17
Documentation meta issue (just add to it when you notice something in the docs)


rsmmr:

rsmmr:
Add end-to-end file analyzer example.

rsmmr:
Document properties that can be set at unit or module level (e.g., `%byte-order`)

rsmmr:
Explain how to use Spicy enums from Zeek. note that they need to be `public`.

duffy-corelight:
`write` isn't in the 5.2.8.1 list https://docs.zeek.org/projects/spicy/en/latest/programming/language/appendix.html#reserved-keywords

duffy-corelight:
`filter` `forward` and `sink` also aren't in the 5.2.8.1 list

duffy-corelight:
`self` isn't in the 5.2.8.1 list

duffy-corelight:
`connect` isn't in the 5.2.8.1 list

duffy-corelight:
`bitfield` isn't in the 5.2.8.1 list

duffy-corelight:
In 5.1.6.8 Vector
standard field specification definining how to parse the vector elements.
should be
standard field specification defining how to parse the vector elements.

rsmmr:
Ah, it's generating the reserved list from the wrong source, will fix (some of that are not reserved words though, write/forward/etc are method names)

duffy-ocraven:
On https://docs.zeek.org/projects/spicy/en/latest/programming/language/types.html
in 5.2.5.2
Each RANGE has one of the forms LABEL: A or LABEL: A..B where A and B are a bit numbers.
should be
Each RANGE has one of the forms LABEL: A or LABEL: A..B where A and B are bit numbers.

and
in 5.2.5.5
Enum types associate labels with a numerical values.
should be
Enum types associate labels with numerical values.

duffy-ocraven:
On https://docs.zeek.org/projects/spicy/en/latest/programming/parsing.html#unit-parameters
```
Fields may use an attribute &requires=EXPR to enforce additional constraints on their values. EXPR must a be boolean expression that will be evaluated after the parsing for the field has finished,
```
should be
```
Fields may use an attribute &requires=EXPR to enforce additional constraints on their values. EXPR must be a boolean expression that will be evaluated after the parsing for the field has finished,
```

rsmmr:
The `begin()`/`end()` operators aren't documented (more generic operators may be missing too?)

[...]

PR #641, opened 2020-12-23
Documentation cleanup

bbannier:
This fixes the remaining issues reported in #441.

### DIFF
--- a/doc/programming/language/statements.rst
+++ b/doc/programming/language/statements.rst
@@ -269,9 +269,9 @@ exception.
    [else
         BLOCK]
 
-A while-loop that executed ``BLOCK`` for as long as the boolean
-``COND`` evaluates to true. The second form initializes a new local
-variable ``IDENT`` with ``EXPR``, and makes it available inside both
-``COND`` and the body. If an ``else`` block is specified, that will
-executes once the ``COND`` evaluates to false at the loop beginning.
+``while`` introduces a loop that executes ``BLOCK`` for as long as the
+boolean ``COND`` evaluates to true. The second form initializes a new
+local variable ``IDENT`` with ``EXPR``, and makes it available inside
+both ``COND`` and ``BLOCK``. If an ``else`` block is specified, it will
+execute one time once ``COND`` evaluates to false at the loop beginning.
 It will not execute if the loop aborts with a :ref:`statement_break`.
--- a/doc/programming/parsing.rst
+++ b/doc/programming/parsing.rst
@@ -228,9 +228,8 @@ them all.
 
 .. todo::
 
-    Parsing a regular expression would make a nice exmaple for
-    ``&size`` as well, but we currently have a bug there
-    (:issue:`286`).
+    Parsing a regular expression would make a nice example for
+    ``&size`` as well.
 
 .. rubric:: Anonymous Fields
 
--- a/doc/programming/parsing.rst
+++ b/doc/programming/parsing.rst
@@ -916,7 +916,7 @@ elements into the final vector.
 
 Spicy uses a specific syntax to define fields of type vector::
 
-    NAME : ELEM_TYPE[SIZE].
+    NAME : ELEM_TYPE[SIZE]
 
 ``NAME`` is the field name as usual. ``ELEM_TYPE`` is type of the
 vector's elements, i.e., the type that will be repeatedly parsed.
@@ -982,9 +982,9 @@ add the parameters to it (e.g., ``x: (MyUnit("arg1"))[]``).
 
 .. note::
 
-    The ``x: (<T>)[]`` syntax is quite flexbile. In fact, ``<T>`` is
+    The ``x: (<T>)[]`` syntax is quite flexible. In fact, ``<T>`` is
     not limited to subunits, but allows for any standard field
-    specification definining how to parse the vector elements. For
+    specification defining how to parse the vector elements. For
     example, ``x: (bytes &size=5)[];`` parses a vector of 5-character
     ``bytes`` instances.
 
--- a/doc/programming/parsing.rst
+++ b/doc/programming/parsing.rst
@@ -439,7 +439,7 @@ There are three locations where hooks can be implemented:
 
     type Foo = unit {
         x: uint32;
-        v: unint8[];
+        v: uint8[];
 
         on %init { ... }
         on x { ... }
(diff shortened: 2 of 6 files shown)


doc/programming/language/statements.rst
```diff
@@ -269,9 +269,9 @@ exception.
    [else
         BLOCK]
 
-A while-loop that executed ``BLOCK`` for as long as the boolean
-``COND`` evaluates to true. The second form initializes a new local
-variable ``IDENT`` with ``EXPR``, and makes it available inside both
-``COND`` and the body. If an ``else`` block is specified, that will
-executes once the ``COND`` evaluates to false at the loop beginning.
+``while`` introduces a loop that executed ``BLOCK`` for as long as the
```
duffy-ocraven, line 272: introduces a loop that executed
should be
introduces a loop that executes

doc/programming/language/statements.rst
```diff
@@ -269,9 +269,9 @@ exception.
...
+boolean ``COND`` evaluates to true. The second form initializes a new
+local variable ``IDENT`` with ``EXPR``, and makes it available inside
+both ``COND`` and ``BLOCK``. If an ``else`` block is specified, it will
+execute one time once ``COND`` evaluates to false at the loop beginning.
```
duffy-ocraven, line 276: it will execute one time once
should be
it will execute one time if

doc/programming/parsing.rst
```diff
@@ -982,9 +982,9 @@ add the parameters to it (e.g., ``x: (MyUnit("arg1"))[]``).
 
 .. note::
 
-    The ``x: (<T>)[]`` syntax is quite flexbile. In fact, ``<T>`` is
+    The ``x: (<T>)[]`` syntax is quite flexibile. In fact, ``<T>`` is
```
duffy-ocraven, line 993: syntax is quite flexibile
should be
syntax is quite flexible

doc/programming/language/statements.rst
bbannier: I am not a native speaker, but I parse 

>  it will execute one time once ``COND`` evaluates to false at the loop beginning.

as _at the point it time where `! COND` execute it_, but

>  it will execute one time if ``COND`` evaluates to false at the loop beginning.

at least has a ring of _if at the beginning `! COND` execute it_, i.e., either execute the loop at least once or execute the `else` block.

At least from looking at the tests the implementation does the first of these.

duffy-ocraven: The piece that we are revising is the trailing half of this sentence: If an ``else`` block is specified, it will execute one time once ``COND`` evaluates to false at the loop beginning.
So the more accurate statement is:
If an ``else`` block is specified, it will execute one time if ``COND`` evaluates to false at the loop beginning.

I.e. It will never execute at all,  if ``COND`` evaluates to true at the loop beginning.

"...once ``COND`` evaluates to ..." is linguistically a usage for a situation which is sure to eventually happen.

bbannier: > "...once COND evaluates to ..." is linguistically a usage for a situation which is sure to eventually happen.

And that is exactly what is happening here. The `else` block will always be executed unless the loop terminates via `break`. As an example, [this loop](https://github.com/zeek/spicy/blob/3bfb6cf3a8790cb154eac35e445aa733c77a619a/tests/spicy/statements/while.spicy#L30-L36) is executed like [this](https://github.com/zeek/spicy/blob/3bfb6cf3a8790cb154eac35e445aa733c77a619a/tests/Baseline/spicy.statements.while/output#L15-L21). Note that the next loop in that test is terminated via `break` so the `else` block is not run.

duffy-ocraven: If I have misunderstood whether the else is exclusively ever executed *instead of* ever doing the body, then my suggestion for clarification is clarity but in the wrong direction.

if the else is nearly always executed, one time, after zero-or-more loops through the body, then we should wholesale rewrite the sentence to capture that. Since if the else is not executed at all, when a break occurs.

duffy-ocraven: Yeah, looking at https://github.com/zeek/spicy/blob/3bfb6cf3a8790cb154eac35e445aa733c77a619a/tests/Baseline/spicy.statements.while/output#L15-L21 I had formed an inaccurate perception from the pre-existing text: ... it will
execute one time once ``COND`` evaluates to false at the loop beginning.

To my ear, "at the loop beginning" is a one-time moment, just prior to the first evaluation of COND.

bbannier: Okidoki, I think that resolves this.

duffy-ocraven: If we are implementing a construct which is: "zero-or-more loops through the body, then perform a statement" we should term that statement as `then` rather than as `else`. Python's statement `finally` is another good choice, though Python's `finally` does execute even when break is used to prematurely drop out from the loop.

duffy-ocraven: I doubt that there is good justification for having a construct with semantics:
This statement will not be executed after "zero-or-more loops through the body, if one of those is a conditional to drop out from the loop based upon a conditional other than the COND". 

The use of two distinct conditionals is usually for clarity. It will not be common that an author needs to execute the "finally" kind of statement if the COND happened, and doesn't need to execute that, if execution fell out of the loop at the within-loop conditional. You could just always have that statement, with no else, and it always executes subsequent to the loop.

```
while (idx--) {
   if (isRepresentable(charInSomeEncoding) {
       print charInSomeEncoding.representation();
   else
       break;
}
print "\n";
```

bbannier: I agree and filed #644.

bbannier:
> > Okidoki, I think that resolves this.
> 
> I suppose best to proceed with _once_ until #644 so as not to muddy things before we decide whether to even have this construct. Once is still the wrong word (_if and when_ is more appropriate for something that might not happen) Once connotes something which definitely eventually happens.

Are you suggesting we reopen the issue https://github.com/zeek/spicy/pull/641#discussion_r548050838? Blocking this PR until #644 is resolved is a recipe for staling a simple doc cleanup for something which has a less clearer timeline.

Note that the sentence we are discussing is directly followed by

> It will not execute if the loop aborts with a :ref:`statement_break`.

At least to me this seems to sufficiently qualify the preceding _once_.

duffy-ocraven:
> Are you suggesting we reopen the issue [#641 (comment)](https://github.com/zeek/spicy/pull/641#discussion_r548050838)? Blocking this PR

The opposite. Don't block this PR. The text documenting this construct is fine, as long as the construct exists. The text documenting this construct will disappear when the construct disappears, if #644 proceeds in that direction.

rsmmr (approved):

PR #642, opened 2020-12-23
Add optional CI task for Zeek nightly

bbannier:


bbannier:
Since I have only enabled this task for `master`, I ran a previous test where it was not skipped. The mechanics appear to work there, but the [task still failed](https://cirrus-ci.com/task/6024776714027008) since the Zeek package installed something we did not expect. I am not sure who needs to fix this, but since this task is optional I think this PR is ready.

.cirrus.yml
```diff
@@ -665,3 +665,65 @@ docker_fedora33_task:
   packages_artifacts:
     path: build/spicy-*.tar.gz
     type: application/gzip
+
+clang11_zeek_nightly_task:
+  # This task is only executed on master.
+  # Failing this task does not fail the build.
+  only_if: $CIRRUS_BRANCH == 'master'
+  allow_failures: $CIRRUS_BRANCH == 'master'
```
rsmmr, line 673: Could this just be`true` given that `only_if` limits already where it runs?

rsmmr (approved):

.cirrus.yml
bbannier: Given [the docs](https://cirrus-ci.org/guide/writing-tasks/#failure-toleration) I wasn't sure what syntax this expected. If this is just a shell expression (`true` command) or YAML (`true` literal), using `true` here should work. I adjusted the patch, let's see if it is accepted by Cirrus (it passes `cirrus validate` for me locally).

rsmmr (approved):

PR #651, opened 2021-01-04
Fix for macOS 10.16.

rsmmr:
Started to need this after upgrading to Big Sur. Should be fine
elsewhere too, but CI will tell us ...


bbannier:
I opened PR #643 to add Big Sur to our CI, but that fails relatively early on since the installation of the compiler seems not working. I am not sure what was missing and do not currently run Big Sur myself, so any info you could share over there on what to do to even let the build start would be welcome.

bbannier (approved):

Issue #644, opened 2020-12-23
Consider dropping while-else construct


bbannier:
A discussion on #641 brought up the issue of the usefulness of the while-else construct.

> I doubt that there is good justification for having a construct with semantics:
This statement will not be executed after "zero-or-more loops through the body, if one of those is a conditional to drop out from the loop based upon a conditional other than the COND". 
>
> The use of two distinct conditionals is usually for clarity. It will not be common that an author needs to execute the "finally" kind of statement if the COND happened, and doesn't need to execute that, if execution fell out of the loop at the within-loop conditional. You could just always have that statement, with no else, and it always executes subsequent to the loop.
>
> ```
> while (idx--) {
>   if (isRepresentable(charInSomeEncoding) {
>       print charInSomeEncoding.representation();
>   else
>       break;
>}
>print "\n";
>```

_Originally posted by @duffy-ocraven in https://github.com/zeek/spicy/pull/641#discussion_r548134296_

I agree that this construct appears rather unusual, easy to misunderstand, and probably easy to replace with more easily approachable explicit logic.

We should consider removing it without replacement.

duffy-ocraven:
Notice that when an esoteric situation arises that does not want the trailing statement to be executed because the loop prematurely terminated, that same conditional can be repeated in the trailing statement, to there suppress it .

rsmmr:
This construct exists because it's being reused from the HILTI-side, and there it comes in handy for auto-generated code (makes scoping easier in some cases).  I agree that for Spicy it's not very useful to have. I suggest we just remove support from the Spicy parser.

PR #653, opened 2021-01-04
Remove while-else construct from Spicy

bbannier:
This PR removes the while-else construct from Spicy. While we continue to support it in HILTI, we do not expose it in Spicy.


rsmmr (approved):
Assuming this is ready for review, looks good.

bbannier:
Closes #644.

PR #654, opened 2021-01-04
Add script to create release tarball.

### LINKED ISSUE #650: Automate upload of release tar balls to GitHub
When preparing release versions, it'd be good to have release tarballs for download on GitHub. I believe we can't use GitHub's default tarballs because they wouldn't include the submodules (iirc, GitHub just does `git archive`). However, we should be able to deploy a GitHub action to make &. upload our own, see:

- https://github.com/actions/create-release#example-workflow---create-a-release
- https://github.com/actions/upload-release-asset

(Ideally, we'd do a quick build & test run of that tar ball too).

bbannier:
This PR adds a script to create a release tarball and a GH action
attaching which creates and attaches the tarball to any release tags
pushed to the repo. We assume that that release tags are of the form
`v*`.

Closes #650.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -1,0 +1,39 @@
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
+      - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+
+name: Create Release
+
+jobs:
+  build:
+    name: Upload Release Asset
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Prepare archive
+        run: |
+          ./scripts/make-tarball.sh
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
+      - name: Get release version
+        id: get_version
+        uses: battila7/get-version-action@v2
+      - name: Upload Release Asset
+        id: upload-release-asset
+        uses: actions/upload-release-asset@v1
+        env:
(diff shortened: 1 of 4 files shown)


rsmmr (dismissed):
Looks good. Let's add VERSION to `.gitignore` as well.

rsmmr (dismissed):

bbannier:
I added a GH action to upload the tarball created by the script. For that I changed the script to now always generate a file `spicy.tar.bz` which means we can hardcode its path in the action.

rsmmr:
> I added a GH action to upload the tarball created by the script. For that I changed the script to now always generate a file `spicy.tar.bz` which means we can hardcode its path in the action.

Can we still set a name when uploading? Would be nice if people downloading the tarball get a versioned file name.

bbannier:
> > I added a GH action to upload the tarball created by the script. For that I changed the script to now always generate a file `spicy.tar.bz` which means we can hardcode its path in the action.
> 
> Can we still set a name when uploading? Would be nice if people downloading the tarball get a versioned file name.

That's exactly what I did [here](https://github.com/zeek/spicy/blob/992873886ddd24ab3fdbefa889c1d6494bf95757/.github/workflows/create-release.yml#L38).

rsmmr:
> That's exactly what I did [here]

Excellent!

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  build:
+    name: Upload Release Asset
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Prepare archive
+        run: |
+          ./foo.sh
```
rsmmr, line 18: This correct?

rsmmr (approved):

.github/workflows/create-release.yml
bbannier: Most definitely not, but a leftover from testing. Fixed now.

PR #668, opened 2021-01-18
On macOS, pass CMAKE_OSX_SYSROOT through to JIT linker.

rsmmr:


bbannier (approved):
I tested this patch in #643 and it makes CI pass :shipit:

PR #675, opened 2021-01-20
Zeek plugin build fixes

rsmmr:


zeek/plugin/src/plugin.cc
```diff
@@ -27,9 +24,11 @@
...
 #include <zeek-spicy/zeek-reporter.h>
 
+#ifdef HAVE_TOOLCHAIN
 namespace spicy::zeek::debug {
 const hilti::logging::DebugStream ZeekPlugin("zeek");
 }
+#endif
 
 plugin::Zeek_Spicy::Plugin SpicyPlugin;
 plugin::Zeek_Spicy::Plugin* ::plugin::Zeek_Spicy::OurPlugin = &SpicyPlugin;
```
bbannier, line 34: (commenting here since I cannot leave a comment on l.42)

When building with `BUILD_TOOLCHAIN=OFF` we do not link against hilti and `hilti::util::fmt` will be unknown. Let's use `hilti::rt::fmt` instead and include `<hilti/rt/fmt.h>`.

bbannier (dismissed):

zeek/plugin/src/plugin.cc
rsmmr: Good catch!

bbannier (approved):

PR #680, opened 2021-01-21
Collection of smaller cleanups in individual commits

rsmmr:


scripts/precompile-headers.sh
```diff
@@ -40,16 +40,16 @@ VERSION=$(${HILTI_CONFIG} --version | cut -d ' ' -f1)
 # if set; else a patch under the user's home directory is used.
 CACHE=${SPICY_CACHE:-$HOME/.cache/spicy/${VERSION}}
 
-echo "Clearing cache directory $CACHE"
+# echo "Clearing cache directory $CACHE"
```
bbannier, line 43: Let's just get rid of these lines instead of commenting them out.

Here and below.

hilti/runtime/include/libhilti.h
```diff
@@ -1,6 +1,9 @@
-// Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
+// Copyright (c) 2020-2021 by the Zeek Project. See LICENSE for details.
 
-#pragma once
+// Do not use "#pragma once here". We use this file as top-level for header
+// pre-compilation, and GCC doesn't like that there.
```
bbannier, line 4: Let's add a link to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64117 here.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2020 by the Zeek Project. See LICENSE for details.
+# Copyright (c) 2020-2021 by the Zeek Project. See LICENSE for details.
```
bbannier, line 1: Let's also bump the date in `LICENSE:1`.

bbannier (approved):

PR #693, opened 2021-01-22
Allow to pass mime type to data_in when sending data to Zeek.

0xxon:
This change adds zeek::file_data_in_with_mime_type - which allows to
pass an initial mime type to Zeek when sending file data.

I also added the timestamp removal to the Zeek log canonification. This
is needed for the testcase - and it also seems probable that this will
be needed for future tests. It also is used for nearly all btests in the
Zeek source tree.


rsmmr:
Excellent, thanks! 

One question: you probably didn't see #693, I was wondering if we should instead pass the MIME type to `file_begin()` as an optional parameter. That doesn't match the Zeek API, but it seems more intuitive to me. Internally we'd have to buffer the MIME type inside the cookie because we'd still need to pass it to `DataIn()` eventually.

0xxon:
That might indeed be the better idea.

rsmmr:
Ok, I'll change that on merge.

0xxon:
Heh, I saw the last comment too late and kind of am done with this. I will push this in a minute.

0xxon:
I have not quite figured out the sed newline problem on alpine yet, though.

rsmmr:
This looks all good, we just need to figure out Alpine issued. Have you already looked into that?

rsmmr:
Found the Alpine problem: BusyBox sed!

have a patch, just testing it a bit more

PR #702, opened 2021-01-26
Support `&size` attributes at the unit level.

rsmmr:


spicy/toolchain/include/ast/types/unit.h
```diff
@@ -114,7 +114,7 @@ class Unit : detail::AssignIndices,
     }
 
     /**
-     * Returns all properties of a given name.
+      Returns all properties of a given name.
```
bbannier, line 117: This is now formatted inconsistently (no leading `* ` on lines in block). Let's either keep the old format or make this a single-line comment.

spicy/toolchain/src/compiler/parser/parser.yy
```diff
@@ -332,8 +332,12 @@ global_scope_decl
               | property_decl                    { $$ = std::move($1); }
               | hook_decl                        { $$ = std::move($1); }
 
-type_decl     : opt_linkage TYPE scoped_id '=' type opt_attributes ';'
-                                                 { $$ = hilti::declaration::Type(std::move($3), std::move($5), std::move($6), std::move($1), __loc__); }
+type_decl     : opt_linkage TYPE scoped_id '=' type opt_attributes ';' {
+                                                   if ( auto x = $5.tryAs<type::Unit>(); x && $6 && *$6 )
+                                                       $5 = Type(type::Unit::addAttributes(*x, *$6));
```
bbannier, line 337: Could you add an validator which checks passed attributes against a safelist of allowed attributes (right now: contains only `&size`).

bbannier (approved):

PR #703, opened 2021-01-26
Allow to define a minimum Spicy version that a module requires.

### LINKED ISSUE #685: Support `%minimum-spicy-version` module property
This would let Spicy enforce a minimum version required to compile the module.

rsmmr:
Adding `%spicy-version = "1.0";` to a module now lets it require to
compile with a Spicy of at least that version. The version can be
specified as either `x.y` or `x.y.z`.

(The commit uses the opportunity to document the module-level
`%byte-order` property as well.)

Closes #685.


doc/programming/language/modules.rst
```diff
@@ -66,3 +69,18 @@ implicitly public.
...
+
+    ``%byte-order = ORDER;``
+        Defaults the byte order for any parsing inside this module to
+        `<expr>`, where ``ORDER`` must be of typei is type
+        :ref:`spicy_ByteOrder`.
+
+    ``%spicy-version = "VERSION";``
+        Specifies that the module requires a given minimum version of
+        Spicy, where ``VERSION`` must be a string of the form ``X.Y``
+        or ``X.Y.Z``.
```
bbannier, line 86: Do you think it would make sense to also support just `X`? This would become interesting especially once we release v1, v2 and so on.

doc/programming/language/modules.rst
```diff
@@ -66,3 +69,18 @@ implicitly public.
...
+-----------------
+
+A module may define the following global properties:
+
+    ``%byte-order = ORDER;``
+        Defaults the byte order for any parsing inside this module to
+        `<expr>`, where ``ORDER`` must be of typei is type
+        :ref:`spicy_ByteOrder`.
+
+    ``%spicy-version = "VERSION";``
```
bbannier, line 83: I think it would be better to call this `spicy-min-version` even now to leave the door open to some form of maximum version down the line to work around breaking API changes.

spicy/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -154,6 +154,47 @@ struct PreTransformVisitor : public hilti::visitor::PreOrder<void, PreTransformV
...
+                        int minor = parse_number(v[1]);
+                        int patch = 0;
+
+                        if ( v.size() == 3 )
+                            patch = parse_number(v[2]);
+
+                        // This must match the computation in our autogen-version script.
+                        int version = major * 10000 + minor * 100 + patch;
+                        if ( hilti::configuration().version_number < version )
+                            error(fmt("module %s requires at least Spicy version %s (have %s)", m.id(), s->value(),
```
bbannier, line 185: We already output these condense version numbers at least in `library::Version::checkCompatibility`, but I think it would help users if we would instead format this as a user-level version string (e.g., 0.4.0` instead of `400`). Could you introduce a formatting helper for that with this patch?

doc/programming/language/modules.rst
rsmmr: `%spicy-version = "1"` looks at bit odd to me, semantics aren't clear immediately. And one can express the same through `%spicy-version = "1.0"`

spicy/toolchain/src/compiler/visitors/validator.cc
rsmmr: it is showing the string version, not the numerical integer, for both: `s->value()` is what the user typed, and `version_string` is our normal full version string.

doc/programming/language/modules.rst
rsmmr: Hmm, I don't see a max version going to be helpful (as we chatted about) because one won't be able to anticipate that number in advance. If we wanted to do support better control, we'd need some form of prerprocessors-tyle `if/else`. Given that, I prefer the shorter version here.

doc/programming/language/modules.rst
bbannier: As discussed offline, `"1"` would have the same implicit trailing zero semantics as e.g., `"1.1"`; `"1"` would require `>=1.0.0` like `"1.1" would require `>=1.1.0`. Let either allow single digit versions as well or always require all three digits.

duffy-ocraven: I favor with the whole specificity. New syntactic features are introduced at a specific version, best-specified with the whole specificity, not "this specific number is not good enough, but you're good-to-go as long as you are later than this". It would be a boon if what a user will see when their local Spicy is out-of-date with respect to that unsupported syntax, is an explicit message about the required minimum version.

duffy-ocraven: OTOH, if two different syntactic supports are needed, the value has to be the larger of the two. Probably conventional usage over time in this case will shift to overly conservatively putting a number that is not so precise, 
perhaps the next-higher minor, to capture something the author will think is some kind of "margin of safety".

rsmmr: Folks, I'll leave this as it is for now, don't think it's worth iterating further at the moment.  If we want some other semantics, we can discuss, but for now "1.0" is just the same as "1.0.0", seems clear enough to me, and fits with semantic versioning as well where the patch level would usually not matter (i.e., we wouldn't introduce new features at that level). Duffy, the error messages includes what's required.

duffy-ocraven:
%spicy-version should get added into
https://github.com/zeek/spicy/pull/728/commits/99c8e4e296995e995320669b503cb5f080658f45

jsiwek:
> %spicy-version should get added into
> [99c8e4e](https://github.com/zeek/spicy/commit/99c8e4e296995e995320669b503cb5f080658f45)

The Pygments lexer doesn't use a pre-defined set of properties, just looks for identifiers preceded by `%`: https://github.com/zeek/spicy/commit/99c8e4e296995e995320669b503cb5f080658f45#diff-47520806e7d2ad8954dab2a68957c442e4aa34f192630a3781d310bab2ffe460R70-R72

(no strong idea on whether it *should*: I didn't understand yet why Spicy's own lexer didn't pre-define properties, unlike how it does for `&`-attributes, but for purpose of syntax highlighting, likely don't have to be that conservative in what it accepts)

rsmmr:
>  I didn't understand yet why Spicy's own lexer didn't pre-define properties, unlike how it does for `&`-attributes,

Context here is that `&`-attributes started out as a non-predefining pattern as well, until it turned out that the parser got into ambiguities distinguishing them from some expressions. Predefining fixed that. We should probably do the same for properties now, yes.

PR #704, opened 2021-01-27
Add strptime Spicy library function.

### LINKED ISSUE #701: Consider adding strptime Spicy library function
We currently provide a `strftime` Spicy library function, so it seems to make sense to also provider the inverse `strptime`. It has similar usability issues around e.g., locales and extensions.

bbannier:
Closes #701.

### DIFF
--- a/spicy/lib/spicy.spicy
+++ b/spicy/lib/spicy.spicy
@@ -82,7 +82,7 @@ public function bytes_to_hexstring(value: bytes) : string &cxxname="spicy::rt::b
 ## Returns the value of an environment variable, if set.
 public function getenv(name: string) : optional<string> &cxxname="hilti::rt::getenv" &have_prototype;
 
-##  Formats a time according to user-specified format string.
+## Formats a time according to user-specified format string.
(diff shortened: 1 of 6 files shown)


spicy/lib/spicy.spicy
```diff
@@ -93,3 +93,15 @@ public function getenv(name: string) : optional<string> &cxxname="hilti::rt::get
...
+
+## Parse time from string.
+##
+## This function uses the currently active locale and timezone to parse values.
+##
+## The format string can contain format specifiers supported by POSIX strptime, see
+## https://pubs.opengroup.org/onlinepubs/009695399/functions/strptime.html.
+##
+## This function can raise InvalidArgument if the string could not be parsed
+## with the given format string, or OutOfRange if the parsed time value cannot
```
rsmmr, line 105:
```suggestion
## This function raises InvalidArgument if the string could not be parsed
## with the given format string, or OutOfRange if the parsed time value cannot
```

rsmmr (approved):

PR #705, opened 2021-01-28
Zeek plugin: Add functions to access UIDs and FUIDs.

### LINKED ISSUE #688: Zeek plugin: provide access to current UID and FUID

rsmmr:
zeek::uid() and zeek::fuid() return the current connection UID and file
FUID, respectively.

Closes #688.

### DIFF
--- a/zeek/plugin/src/runtime-support.cc
+++ b/zeek/plugin/src/runtime-support.cc
@@ -165,6 +165,15 @@ hilti::rt::Bool rt::is_orig() {
     else
         throw ValueUnavailable("is_orig() not available in current context");
 }
+std::string rt::uid() {
+    auto cookie = static_cast<Cookie*>(hilti::rt::context::cookie());
+    assert(cookie);
+
+    if ( auto c = std::get_if<cookie::ProtocolAnalyzer>(cookie) )
+        return c->analyzer->Conn()->GetUID().Base62("C");
+    else
+        throw ValueUnavailable("uid() not available in current context");
+}
 
 void rt::flip_roles() {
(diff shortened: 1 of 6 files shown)


zeek/plugin/src/runtime-support.cc
```diff
@@ -165,6 +165,15 @@ hilti::rt::Bool rt::is_orig() {
     else
         throw ValueUnavailable("is_orig() not available in current context");
 }
+std::string rt::uid() {
+    auto cookie = static_cast<Cookie*>(hilti::rt::context::cookie());
+    assert(cookie);
```
bbannier, line 170: This seems pretty drastic and hard to work around. Could we instead either return an `optional<string>` here, or reuse below's `throw ValueUnavailable("uid() not available in current context");`?

zeek/plugin/src/runtime-support.cc
```diff
@@ -220,6 +229,18 @@ void rt::file_begin() {
     // Nothing todo.
 }
 
+std::string rt::fuid() {
+    auto cookie = static_cast<Cookie*>(hilti::rt::context::cookie());
+    assert(cookie);
```
bbannier, line 234: See above.

zeek/plugin/src/runtime-support.cc
rsmmr: cookie is always available, it's set by the plugin (otherwise it's a bug, hence the assert)

zeek/plugin/src/runtime-support.cc
rsmmr: See above.

PR #706, opened 2021-01-28
Make Docker images standalone.

bbannier:
This changeset updates the included Docker images so that the included
Zeek installation is usable with Spicy without further configuration.


bbannier:
This PR is should fail as is as a test since I have not yet fixed the installation of the ubuntu-20 image.

rsmmr (approved):

PR #714, opened 2021-01-29
Canonify baseline for test `spicy.attributes.spicy-version-fail`.

bbannier:
This patch canonifies the test baseline for
`spicy.attributes.spicy-version-fail` so that it does not contain the
exact version of the current checkout. This is a cleanup commit for
1fd0e80f3f9c916fff94888c724997d6051d00fb.


rsmmr (approved):

rsmmr:
Thanks!

PR #715, opened 2021-01-29
Add script to store and upload release artifacts.

### LINKED ISSUE #699: Publish stable binary artifacts releases
This issue was broken out of #683.

We should provide stable binary artifacts for releases. Currently artifacts are built in Cirrus and directly linked in the docs. Instead the release artifacts should be uploaded to the corresponding release page; documentation should reference these artifacts instead.

bbannier:
This patch adds a script to download packaging artifacts of a Cirrus
build and optionally upload them to a release.

Closes #699.

### DIFF
--- a/scripts/store-release-artifacts
+++ b/scripts/store-release-artifacts
@@ -1,0 +1,79 @@
+#!/usr/bin/env bash
+
+set -eu
+set -o pipefail
+
+usage() {
+    cat << EOF
+Downloads CI packaging artifacts and optionally uploads them to a release.
+
+Usage: $0 --commit COMMIT --upload
+
+This script downloads Cirrus CI packaging artifacts for the given COMMIT from
+all completed tasks and stores them in a temporary directory. The packaging
+artifacts will be renamed to include te name of the job which produced them.
+The output directory will be reported.
+
+With '--upload' artifacts are uploaded to the release page and we expect a valid
+Github token with 'repo' scope passed on stdin. In this case COMMIT should
+refer to a tag for which a release already exists.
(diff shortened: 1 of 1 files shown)


scripts/store_release_artifacts
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
+
+set -eu
+set -o pipefail
+
+usage() {
+    cat << EOF
+Downloads CI packaging artifacts and optionally uploads them to a release.
+
+Usage: $0 COMMIT <GITHUB_TOKEN>
```
rsmmr, line 10: Couple requests on usage:

1. Let's read the token from stdin, that way one can avoid getting it into the command line history
2. Add an option to explicitly trigger the upload, so that it's easier to anticipate if the script will be doing that

scripts/store_release_artifacts
```diff
@@ -0,0 +1,79 @@
...
+Downloads CI packaging artifacts and optionally uploads them to a release.
+
+Usage: $0 --commit COMMIT --upload
+
+This script downloads Cirrus CI packaging artifacts for the given COMMIT from
+all completed tasks and stores them in a temporary directory. The packaging
+artifacts will be renamed to include te name of the job which produced them.
+The output directory will be reported.
+
+If '--upload' artifacts are uploaded to the release page and we expect a valid
```
rsmmr, line 17:
```suggestion
With '--upload' artifacts are uploaded to the release page and we expect a valid
```

rsmmr (dismissed):

PR #716, opened 2021-01-29
Make offsets usable from unit field attributes.

bbannier:


rsmmr (approved):
Looks good. It would be nice if we could avoid saving the offset twice, but from a quick look I also don't see a straight-forward way to do that; seems fine.

PR #725, opened 2021-02-03
Add support for shared unit contexts with lifetime determined by host application.  

### LINKED ISSUE #696: Zeek plugin: Provide access to parser for other side of the connection
Some Zeek analyzers need a way to correlate information across a connection's
two sides. To facilitate that, the unit that is parsing the originator side needs to
gain access to the unit parsing the responder side, and vice versa.

From a Spicy perspective, that's a bit tricky unfortunately because a Spicy
unit doesn't know anything about "the other side", especially not its unit type.
Because everything in Spicy is statically typed, we need to provide that
information somehow upfront somehow.

Here's one idea to implement this:

We could extend units with the concept of an external "context" object. A
host application (here: Zeek) would instantiate a context before kicking off
parsing and fill it with any information needed (e.g., the state for the
"other side"). Inside the unit needing access, we declare the type of the context 
object, and make it accessible through a new magic unit attribute, `__context__`. 

Possible syntax: A unit property `%context = T` will let `self.__context__` have a 
yield type of `optional<T&>`, with the optional remaining unset if the host application 
doesn't make a context object available.

Here's an example that gives to units access to each other, with each flagging for the
other when it's done:

```
public type Foo_Request = unit {
    %context = Foo_Request; 

    ...

    on %done {
        (*self.__context__).request_done = True;
    }

    var reply_done: bool;
};

public type Foo_Reply = unit {}
    %context = Foo_Request; 

    ...

    on %done {
        (*self.__context__).reply_done = True;
    }

    var request_done: bool;
};
```

Downside of all of this: Such a parser won't work with `spicy-driver` anymore because it wouldn't make the context available (maybe we could make it work in batch mode at least where we should be able to create the same context as the Zeek plugin).

rsmmr:
This introduces a new unit property for public unit types:

    %context = <type>;

The unit's new `context()` method will then return a value of type
`<type>&` (i.e, a reference to an instance of the given type). This will
be initialized with a reference to a default-constructed instance of the
type.

What makes this special is that the host application can determine how a
context instance is being shared across different parsers: it can (1)
let multiples units link their context to the same instance, and (2)
retain a context instance beyond any specific unit's lifetime.

We use this new mechanism to provide contexts with connection-level
semantics: the Zeek plugin, and also spicy-driver's batch mode, will
create a single context instance that the two originator-side and
responder-side units will share (they must declare matching contexts
for that to work). The object will live until the connections gets
torn down (also meaning it'll be reused across the per-packet parsers
for UDP).

An example how to use this for associating replies with requests in
a pipelining protocol:

    type Connection = tuple<pending: vector<bytes>>;

    public type Requests = unit {
        %context = Connection;
        : (Request(self))[];
    };

    public type Replies = unit {
        %context = Connection;
        : (Reply(self))[];
    };

    public type Request = unit(inout reqs: Requests) {
        cmd: /[A-Za-z]+/;
        : b"\n";

        on %done {
            reqs.context().pending.push_back(self.cmd);
            }
    };

    public type Reply = unit(inout replies: Replies) {
        response: /[0-9]+/;
        : b"\n";

        on %done {
            if ( |replies.context().pending| ) {
                print "%s -> %s" % (replies.context().pending.back(), self.response);
                replies.context().pending.pop_back();
                }
            else
                print "<missing request> -> %s", self.response;
        }
    };

Feeding spicy-driver with a litte batch file containing a couple
connections, each having a series of request/reply pairs:

    msg -> 100
    put -> 200
    CAT -> 555
    end -> 300
    get -> 400
    LST -> 666

Closes #696.


rsmmr:
(Not ready for reviewing yet)

rsmmr:
This supports explicit access to the unit of the other side of a unit as well, though one has to make that happen manually.

Example:

```
module Test;

type Connection = tuple<orig: Requests&, resp: Replies&>;

public type Requests = unit {
    %context = Connection;

    on %init { self.__context__.orig = self; }
    : (Request(self))[];
    on %done { print "replies", self.replies_seen; }

    var replies_seen: uint64;
};

public type Replies = unit {
    %context = Connection;

    on %init { self.__context__.resp = self; }
    : (Reply(self))[];
    on %done { print "requests", self.requests_seen; }

    var requests_seen: uint64;
};

public type Request = unit(inout reqs: Requests) {
    cmd: /[A-Za-z]+/;
    : b"\n";

    on %done {
        ++reqs.__context__.resp.requests_seen;
        }
};

public type Reply = unit(inout replies: Replies) {
    response: /[0-9]+/;
    : b"\n";

    on %done {
        ++replies.__context__.orig.replies_seen;
    }
};
```

(There's some bug currently that let's Spicy reject the `__context__`` assignments here, need to look into that).

rsmmr:
suggestion from @bbannier: change syntax from self.__context__ to self.context().

0xxon:
Just two leave my two cents here - this looks pretty neat to me. I agree with the suggestion of preferring ``self.context().``

duffy-ocraven:
%context should get added into
https://github.com/zeek/spicy/pull/728/commits/99c8e4e296995e995320669b503cb5f080658f45

hilti/toolchain/include/ast/types/reference.h
```diff
@@ -22,8 +22,8 @@ class StrongReference : public TypeBase,
     StrongReference(Wildcard /*unused*/, Meta m = Meta()) : TypeBase({node::none}, std::move(m)), _wildcard(true) {}
     StrongReference(Type ct, Meta m = Meta()) : TypeBase({std::move(ct)}, std::move(m)) {}
     StrongReference(Type ct, bool treat_as_non_constant, Meta m = Meta()) : TypeBase({std::move(ct)}, std::move(m)) {
-        if ( treat_as_non_constant )
-            _state().flags -= type::Flag::Constant;
+        // if ( treat_as_non_constant )
+        //  _state().flags -= type::Flag::Constant;
```
bbannier, line 26: Let's just remove this.

spicy/runtime/src/driver.cc
```diff
@@ -224,6 +227,15 @@ driver::ParsingState::State driver::ParsingState::_process(size_t size, const ch
         return Done;
     }
 
+#ifdef DEBUG
```
bbannier, line 230: Let's either use a proper way to figure out whether a debug logger is enabled, or execute this code unconditionally (the logging would still figure out whether debug logging is enabled).

spicy/runtime/src/driver.cc
```diff
@@ -163,7 +166,7 @@ Result<spicy::rt::ParsedUnit> Driver::processInput(const spicy::rt::Parser& pars
 
         if ( ! r ) {
             DRIVER_DEBUG(fmt("beginning parsing input (eod=%s)", data->isFrozen()));
-            r = parser.parse3(unit, data, {});
+            r = parser.parse3(unit, data, {}, {});
```
bbannier, line 169: I am probably missing something, but shouldn't we inject a context from the outside here as well? I would have expected this function to take an `optional<UnitContext>` and pass it on here.

spicy/runtime/include/sink.h
```diff
@@ -62,8 +62,8 @@ auto _connectUnit(UnitRef<U>& unit) {
     auto self = hilti::rt::ValueReference<U>::self(&*unit);
 
     auto& state = unit->__sink;
-    state = new sink::detail::State();                   // NOLINT
-    state->resumable = (*parse2)(self, state->data, {}); // Kick-off parsing with empty data.
+    state = new sink::detail::State();                       // NOLINT
+    state->resumable = (*parse2)(self, state->data, {}, {}); // Kick-off parsing with empty data.
```
bbannier, line 66: Should this pass on `unit`'s context?

spicy/runtime/include/unit-context.h
```diff
@@ -0,0 +1,102 @@
...
+     *
+     * @tparam T type of the a unit's `%context`.
+     * @param obj reference to a concrete context's instance
+     * @param ti type information matching *obj*
+     */
+    template<typename T>
+    UnitContext(hilti::rt::StrongReference<T> obj, const hilti::rt::TypeInfo* ti) {
+        _object = obj;
+        _type_info = ti;
+    }
```
bbannier, line 40: Can we use an initializer list here and move from `obj`?

doc/programming/parsing.rst
```diff
@@ -1654,3 +1654,112 @@ control Spicy's gap handling, including when to stop buffering data
...
+
+.. spicy-code::
+
+    public type Foo = unit {
+        %property = bytes;
+        [...]
+    };
+
+By default, each instance of such a unit will then receive a unique
+context value of that type. The context value be accessed through the
```
bbannier, line 1688:
```suggestion
context value of that type. The context value can be accessed through the
```

doc/programming/parsing.rst
```diff
@@ -1654,3 +1654,112 @@ control Spicy's gap handling, including when to stop buffering data
...
+
+    module Test;
+
+    # We wrap the state into a tuple to make it easy to add more attributes if needed later.
+    type Pending = tuple<pending: vector<bytes>>;
+
+    public type Requests = unit {
+        %context = Pending;
+
+        : Request[] foreach { self.context().pending.push_back($$.cmd); }
```
bbannier, line 1729: We should probably add a note somewhere in here that production code should include some check on the size of `pending` so parsers do not die from out-of-memory errors.

doc/programming/parsing.rst
```diff
@@ -1654,3 +1654,112 @@ control Spicy's gap handling, including when to stop buffering data
...
+
+Parsing may need to retain state beyond any specific unit's lifetime.
+For example, a UDP protocol may want to remember information across
+individual packets (and hence units), or a bi-directional protocol may
+need to correlate the request side with the response side. One option
+for implementing this is maintaining the state in :ref:`global
+variables <variables>`. However, doing so can be cumbersome because a
+global will often not match semantics well. As an alternative, a unit
+can make use of a dedicated *context* value, which is an instance of a
+custom type that has its lifetime determined suitably by the host
```
bbannier, line 1672: We talked before how this allows creating new reference cycles. Looking at this patch, it seems that this is still possibility for our users of contexts. If I read that right, let's capture it in a improvement ticket.

tests/hilti/types/tuple/assign-element.hlt
```diff
@@ -0,0 +1,26 @@
...
+
+module Foo {
+
+import hilti;
+
+global auto t1 = ("Hello!", True);
+
+assert t1 == ("Hello!", True);
+t1[0] = "Tach!";
+t1[1] = False;
```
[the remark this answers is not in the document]
rsmmr: I'll leave that for the other ticket to fix.

spicy/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -269,6 +270,17 @@ struct PreTransformVisitor : public hilti::visitor::PreOrder<void, PreTransformV
...
 
+        else if ( i.id().str() == "%context" ) {
+            if ( auto e = i.expression(); ! e )
+                error("%context requires an argument", p);
+            else if ( ! e->isA<hilti::expression::Type_>() )
+                error("%context requires a type", p);
+
+            auto decl = p.findParent<hilti::declaration::Type>();
+            if ( decl && decl->get().linkage() != hilti::declaration::Linkage::Public )
+                error("only public units can have %context", p);
```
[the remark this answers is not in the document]
rsmmr: It would need extra code to create and set those contexts. And conceptually I prefer to restrict them to public units to emphasize that that's what their purpose is.

spicy/runtime/src/driver.cc
rsmmr: yeah, the latter is fine

spicy/runtime/src/driver.cc
rsmmr: `processInput()` doesn't have a notion of a connection, so default behavior is fine (which means: the parser will automatically create a conetxt instance if we don't pass one in).

spicy/runtime/include/sink.h
rsmmr: I'm not quite sure yet what to do about sink's and contexts (if anything), but just passing on the context isn't unlikely to work because the types would probably not match.

doc/programming/parsing.rst
rsmmr: I'll leave that out, seem orthogonal and not different than similar issues elsewhere.

doc/programming/parsing.rst
rsmmr: as we chatted about, it's a general problem we have without a cycle detector. As this point, context at least don't make it worst I believe. :)

doc/programming/parsing.rst
bbannier: The danger is that people copy such snippets from the docs and easily run into issues. We should probably call this out whenever we add samples which are not production-ready.

bbannier (approved):

hilti/toolchain/src/compiler/codegen/operators.cc
```diff
@@ -925,6 +925,11 @@ struct Visitor : hilti::visitor::PreOrder<std::string, Visitor> {
         return fmt("%s.emplace_back(%s)", self, args[0]);
     }
 
+    result_t operator()(const operator_::vector::PopBack& n) {
+        auto [self, args] = methodArguments(n);
+        return fmt("%s.pop_back()", self);
```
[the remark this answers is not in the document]
rsmmr: Good catch, I'm changing it to throw when empty, similar to `back()`.

PR #727, opened 2021-02-03
Fix documentation of `set_input`.

bbannier:
The following construct is expected to work, but according to the previous
documentation it was malformed,

    public type Foo = unit {
        %random-access;
        : uint8;
        a: uint8 {
            # Skip forward beyond current position.
            self.set_input(self.input() + 4);
        }
        b: uint8;
    };


rsmmr:
This might not just be a documentation issue. I'd need to try to be sure but I believe behavior of moving the input pointer forward isn't well-defined if the data that's now been pointed to hasn't arrived yet. If that's right, then I think the right fix would be adding that `throw` to actually flag the issue as the documentation says. That would be a bit unfortunate for cases where one knows that data is there already (e.g., with a UDP packet where we always get the full payload); but I don't see at the moment how to make this generally well-defined.

rsmmr (approved):
I tried a couple more things and they were all working as expected. So cool: an unintended feature! :-)

PR #741, opened 2021-02-09
Swapping in an external "any" class.

rsmmr:


hilti/runtime/include/any.h
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2020-2021 by the Zeek Project. See LICENSE for details.
+
+#pragma once
+
+#include <hilti/rt/3rdparty/any/any.hpp>
+
+namespace hilti::rt {
+
+// We ran into some trouble with older versions of hilti::rt::any::any on some
```
bbannier, line 9: `s/hilti::rt::any::any/std::any/`

hilti/runtime/include/any.h
```diff
@@ -0,0 +1,20 @@
...
+// platforms, so we're bringing in an external implementation for the time
+// being. Specifically, we observed the issue in
+// https://stackoverflow.com/a/52414724 on Debian 10, with no obvious way to
+// work around it. See https://github.com/zeek/spicy/issues/629 for the full
+// discussion.
+//
+// Note that this implementation pulls in the *experimental* any interface,
+// which is slightly different from the standardized C++17 API.
+namespace any = linb;
```
bbannier, line 18: Adding an additional namespace makes this pretty verbose. What do you think about instead pulling the names we need into `hilti::rt`, i.e., something like ```
using linb::any;
using linb::any_cast;
using linb::bad_any_cast;
```

**Review by bbannier — COMMENTED** [review]
I tried to run the test suite for the project on my platform but ran into build failures. Included into Spicy the code seems to build fine though.

File: hilti/runtime/include/any.h (hunk 2 above)
Comment (rsmmr) [reply hilti/runtime/include/any.h:18]: Is there any easy way to pull in templates? I went this route because `any_cast` is a series of templated functions that would all need to be pulled in individually, assuming I'm not missing anything. We don't use `any` that often, so I figured this is ok too.

Comment (bbannier) [reply hilti/runtime/include/any.h:18]: Not sure I am missing something, but I believe this should look just like that, even for templated types or functions. The `using` just pulls a name into a scope.

E.g., the following just works,

```.cpp
#include <memory>

using std::unique_ptr;
using std::make_unique;

int main() {
  unique_ptr<int> x = make_unique<int>(1);
}
```

rsmmr: doesn't work for me, I'm getting a bunch of errors like ` error: no template named 'any_cast' in namespace 'hilti::rt'; did you mean 'linb::any_cast'?`

bbannier: Very weird, I had actually checked this against Apple clang (reports 12.0.0) and gcc-9.3.0 on ubuntu-16 before suggesting it. 😓 

As a workaround you could define a template `any_cast` in `hilti::rt` which has the same template and function parameters as `linb::any_cast` and make it call that impl (probably forwarding any passed values).

bbannier: Pushed a fixup commit doing what I suggested to this branch. If that passes CI we should squash it, or else roll it back.

rsmmr: Ok, this works for me now too, I just needed to wipe out my build directory once first!

bbannier (approved):

PR #742, opened 2021-02-09
Fix crash in spicy-driver when used with units that require parameters.

### LINKED ISSUE #738: Segfault when using spicy-driver on unit with parameter
The following unit segfaults when called with:

```
$ echo "abcde" | spicy-driver test4.spicy
Segmentation fault: 11
```

Unit:
```
module test;

public type Top = unit(length: uint32) {
	a: bytes &eod;
};
```

This is a bit mean, since I am not passing the required parameter - but it probably should raise an error message instead.

rsmmr:
These units can't be used but spicy-driver missed checking if that's
the case (as flagged by the corresponding parse functions remaining
unset in the parser description.)

Closes #738.


spicy/runtime/src/driver.cc
```diff
@@ -334,12 +338,12 @@ Result<hilti::rt::Nothing> Driver::processPreBatchedInput(std::istream& in) {
             else
                 return hilti::rt::result::Error(hilti::rt::fmt("unknown session type '%s'", m[2]));
 
-            if ( auto parser = lookupParser(parser_name) ) {
+            if ( auto parser = lookupParser(parser_name); parser && (*parser)->parse1 ) {
```
bbannier, line 341: I see that `_process` has similar uses of `parse1` which according to the updated logic can also be unset. We should add code there to check it before use as we do for `parse3` here now.

rsmmr: Ack, missed protecting the  actual parse function. Just pushed a better fix.

spicy/runtime/src/driver.cc
```diff
@@ -14,6 +14,8 @@
 
 #include <spicy/rt/driver.h>
 
+#include "exception.h"
```
bbannier, line 17: Should be `hilti/rt/exception.h`.

spicy/runtime/src/driver.cc
```diff
@@ -232,6 +240,11 @@ driver::ParsingState::State driver::ParsingState::_process(size_t size, const ch
                 auto input = hilti::rt::reference::make_value<hilti::rt::Stream>(data, size);
                 input->freeze();
 
+                if ( ! _parser->parse1 )
+                    throw InvalidUnitType(
```
bbannier, line 244: Since this is an internal function I am not sure where exactly we should document that this exception can be thrown, but we should document it _somewhere_ since this our external API or very close to it.

bbannier (dismissed):

bbannier (approved):

PR #745, opened 2021-02-12
    Fix non-converging AST in case of ambiguous ID usage.

rsmmr:


bbannier (approved):
The commit message references an incorrect issue. I manually linked the correct issue, but it would probably be a good idea to also fix the commit message to make this less confusing for anybody looking at the history later on.

PR #746, opened 2021-02-12
Fix a couple of vector parsing issues.

### LINKED ISSUE #726: Until-including does not seem to work with vectors
This file:

```
module Test;
public type Testing = unit {
	la: Sub[] &until-including=($$.second_el == b"e");
};
type Sub = unit() {
	second_el: bytes &size=1;
};
```

Yields:
```
$ echo "ananananeee" | spicy-driver -d test.spicy
[error] terminating with uncaught exception of type hilti::rt::NullReference: attempt to access null reference (test.spicy:4:2)
```

I see no obvious reason why this should not work.

Workaround is a construct like this (with an external bool):

```
ifd: IFD(self)[] &while=(self.continue_parsing) foreach { if ( $$.next_ifd == 0 ) { self.continue_parsing = False; } }
```

### LINKED ISSUE #740: Interaction between foreach and &while is undefined
I currently use the following pattern in a number of places:

```
type test = unit {
  var continue_parsing = True;
  element: subtype(self)[] &while(continue_parsing) foreach {
    if ( evaluate_something($$) ) {
      self.continue_parsing = False
    }
  }
};
```

It turns out that this works... most of the time. However, I just encountered the first case where, apparently, the content of foreach was executed after while was evaluated. This manifested in the vector trying to consume another element.

One can work around this by having a subtype like this:

```
type subtype = unit(top: test) {
  sub: actual_subtype if (top.continue_parsing);
}
```

which aborts when continue_parsing was actually already set to false. This is, however, actually not optimal.

To me this feels like a bug - I think foreach should always be executed before &while is evaluated.

I sadly don't have a small testcase. I have one case where this happens consistently for me - however that is with a rather long script and a test-file that is several megabytes large.

rsmmr:
1. "foreach" were run only after evaluating a loop's end condition,
   preventing some use cases. Closes #740.

2. $$ wasn't accessible from inside an `&until-including` expression.
  Closes #726.

Note about 1: I think there are arguments either way here, and in fact
we had a test checking for the existing behaviour.

(Changing both in a single commit because it's rewriting the whole chunk
of code there.)


tests/spicy/types/vector/parse-while-foreach.spicy
```diff
@@ -0,0 +1,23 @@
...
+    x: bytes &size=5;
+    y: int8;
+};
+
+public type test = unit {
+    element: subtype()[] &while=self.continue_parsing foreach {
+        if ( $$.y == 'C' ) {
+            self.continue_parsing = False ;
+        }
+    }
```
bbannier, line 18: TBH, I find the behavior of this confusing and I would have expected `foreach` to operate on the values ending up in the unit, not everything seen. Since `&while` is an attribute which will always appear before a `foreach` and its statement/block, reading the field declaration seems to suggest that we'd examine `continue_parsing` before including the element. Do you think it would make sense change the behavior to have that order instead?

rsmmr: I wouldn't change the parsing order (because that would be problematic for other parts), but yeah, I go back and forth on whether this is a good change, too. Your comment makes me feel inclined to change it back. @0xxon  thoughts?

rsmmr: Changing it back after all.

0xxon: I actually think I expressed myself wrong in the bug report - sorry. I actually want exactly the behavior that @bbannier sketches.

Thinking about it again, the buggy behavior that I wanted to describe (...but did not actually describe) is that the following happend:

&while executed, seeing continue_parsing=True, adding the element.
foreach execute, setting continue_parsing to False, because this was to be the lase element.
...and after that, another parsing step was started, and another element was added, even though continue_parsing was already set to false.

This is also why I said that the behavior typically is correct and that I found an edge-case where it does not seem to work correctly. I think I can reliably trigger this - but it is part of a larger example.

PR #747, opened 2021-02-15
Provide Spicy-land Port constructor

### LINKED ISSUE #723: Provide Spicy-land Port constructor
Currently `Port` values can only be constructed from literals in Spicy. We should provide a constructor so users can construct them from an `uint16` and a `hilti::Protocol`.

This issue was first reported by user Tom Carroll on [Zeek Slack](https://zeekorg.slack.com/archives/C012CQV1K29/p1611768550080300).

bbannier:
Closes #723.


hilti/toolchain/include/ast/expressions/builtin-function.h
```diff
@@ -0,0 +1,83 @@
...
+#include <hilti/ast/node.h>
+#include <hilti/ast/type.h>
+
+namespace hilti {
+namespace expression {
+
+/** AST node representing a builtin function. */
+class BuiltinFunction : public NodeBase, public trait::isExpression {
+public:
+    BuiltinFunction(std::string name, std::string cxxname, hilti::Type type,
```
rsmmr, line 18: This could use a comment string explaining the semantics of the parameters, not immediately clear without further context.

hilti/toolchain/include/ast/expressions/builtin-function.h
```diff
@@ -0,0 +1,83 @@
...
+                    std::vector<declaration::Parameter> parameters, std::vector<Expression> arguments, Meta m = Meta())
+        : NodeBase(nodes(arguments), std::move(m)),
+          _name(std::move(name)),
+          _cxxname(std::move(cxxname)),
+          _type(std::move(type)),
```
rsmmr, line 23: Type should normally be child node as well so that it participates in AST traversal (and hence type resolution). Any particular reason for not making it one?

hilti/toolchain/include/ast/expressions/builtin-function.h
```diff
@@ -0,0 +1,83 @@
...
+          _parameters(std::move(parameters)) {}
```
rsmmr, line 24: Same: should be child nodes.

hilti/toolchain/src/compiler/visitors/apply-coercions.cc
```diff
@@ -401,6 +401,13 @@ struct Visitor : public visitor::PreOrder<void, Visitor> {
         }
     }
 
+    void operator()(const expression::BuiltinFunction& n, position_t p) {
+        auto coerced = coerceCallArguments(&p.node, n.arguments(), n.parameters());
```
rsmmr, line 405: Not needed, left-over?

hilti/toolchain/src/compiler/visitors/apply-coercions.cc
```diff
@@ -401,6 +401,13 @@ struct Visitor : public visitor::PreOrder<void, Visitor> {
...
+        if ( auto coerced = coerceCallArguments(&p.node, n.arguments(), n.parameters()); coerced && *coerced ) {
```
rsmmr, line 406: Formatting nit: skip braces for single-element blocks

hilti/toolchain/src/compiler/codegen/expressions.cc
```diff
@@ -32,6 +32,32 @@ struct Visitor : hilti::visitor::PreOrder<std::string, Visitor> {
         return fmt("%s = %s", cg->compile(n.target(), true), cg->compile(n.source()));
     }
 
+    result_t operator()(const expression::BuiltinFunction& n) {
+        // We use a statement expression (this is an extension supported by
+        // both GCC and Clang) in order for this to be callable in global
+        // contexts.
+        // This "obvious" approach runs into issues there since temporaries
+        // potentially created via `CodeGen::compile` require a block which is
+        // not present for certain globals:
```
rsmmr, line 41: I'm still wondering why this isn't working and/or whether we'll run into the same problem elsewhere when initializing globals. But we can see, this will work for now.

spicy/toolchain/src/compiler/parser/parser.yy
```diff
@@ -904,6 +904,9 @@ expr_g        : '(' expr ')'                     { $$ = hilti::expression::Group
                                                    auto index = hilti::expression::Ctor(hilti::ctor::UnsignedInteger($1, 64, __loc__), __loc__);
                                                    $$ = hilti::expression::UnresolvedOperator(hilti::operator_::Kind::Index, {std::move(captures), std::move(index)}, __loc__);
                                                  }
+              | expr_h
```
rsmmr, line 907: Hmm ... I think this should have a priority that's either like normal function calls (add to `expr_d`) or other ctors (add to `expr_f`); not quite sure which one, but I'd say the latter.

hilti/toolchain/src/compiler/parser/parser.yy
```diff
@@ -753,11 +753,14 @@ expr_f        : ctor                             { $$ = hilti::expression::Ctor(
 
 expr_g        : '(' expr ')'                     { $$ = hilti::expression::Grouping(std::move($2)); }
               | scoped_id                        { $$ = hilti::expression::UnresolvedID(std::move($1), __loc__); }
+              | expr_h
 
 
 member_expr   : local_id                         { $$ = hilti::expression::Member(std::move($1), __loc__); }
               | ERROR                            { $$ = hilti::expression::Member(ID("error", __loc__), __loc__); } // allow methods of that name even though reserved keyword
 
+expr_h        : PORT '(' expr ',' expr ')'       { $$ = hilti::builder::port(std::move($3), std::move($5), __loc__); }
```
rsmmr, line 762: (Same comment about priority here as below in the Spicy version.)

spicy/toolchain/src/compiler/parser/parser.yy
```diff
@@ -904,6 +904,9 @@ expr_g        : '(' expr ')'                     { $$ = hilti::expression::Group
...
+
+expr_h        : PORT '(' expr ',' expr ')'       { $$ = hilti::builder::port(std::move($3), std::move($5), __loc__); }
```
rsmmr, line 909: Please add this to the port type's documentation as well.

hilti/toolchain/src/compiler/codegen/expressions.cc
bbannier: I have filed #748 for this issue (putting more data of `BuiltinFunction` nodes in the AST did not help this issue).

rsmmr (approved):

PR #749, opened 2021-02-16
Add unit `position` method.

### LINKED ISSUE #731: Feature request: stream iterator to current position
It would be neat if there was a way to get a stream iterator to the current position. I know this is basically equivalent to just calling input() + offset(). However, often you want to add something to the current position, and then save that - and the syntax for that is slightly ugly at the moment:

```
self.input() + (cast<uint32>(self.offset()) + 4);
```

Additionally - it would be neat to have variants for "current position" (as in - current element is not parsed) and "following position" (so - just behind the element that currently is being parsed).

bbannier:
This method returns an iterator to the current position in the unit's
input stream.

Closes #731.


spicy/toolchain/include/ast/operators/unit.h
```diff
@@ -36,6 +36,26 @@ property.
...
+        return hilti::operator_::Signature{.self = hilti::type::constant(spicy::type::Unit(type::Wildcard())),
+                                           .result = hilti::type::stream::Iterator(),
+                                           .id = "position",
+                                           .args = {},
+                                           .doc = R"(
+Returns an iterator to the current position in the unit's input stream. If
+executed from inside a field hook, the position will represent the first byte
+that the field has been parsed from. If this method is called before the unit's
+parsing has begun, it will throw a runtime exception. Once parsing has started,
+the offset will remain available for the unit's entire life time.
```
rsmmr, line 50: This sentence doesn't seem to make much sense for this method, can just skip I think.

rsmmr (approved):

PR #751, opened 2021-02-17
 Fix codegen's result type of unit.offset().

rsmmr:


spicy/toolchain/src/compiler/codegen/codegen.cc
```diff
@@ -23,6 +23,8 @@
 #include <spicy/compiler/detail/codegen/grammar-builder.h>
 #include <spicy/compiler/detail/codegen/grammar.h>
 
+#include "ast/types/integer.h"
```
bbannier, line 26:
```suggestion
#include <hilti/ast/types/integer.h>
```

bbannier (approved):

duffy-ocraven:
I would suggest also adding a test (or one each to the left and right) that would have failed because of the difference being codegen'ed internally not having higher precedence than a high-precedence adjacent operator.

rsmmr:
> I would suggest also adding a test (or one each to the left and right) that would have failed because of the difference being codegen'ed internally not having higher precedence than a high-precedence adjacent operator.

That probably was actually tied to the codegen for `offset()`, so the test covers it.

That said, we might have more instances of similar issues, probably need to revisit at some point how to codegen expressions so that we ensure precedence is always correct.

PR #752, opened 2021-02-17
Convert unit properties into attributes.

### LINKED ISSUE #713: Convert unit properties into attributes
#702 introduced support for unit-level attributes. That seems the better mechanism for some functionality that is currently available through unit properties. For example,  `%byte-order` can turn into a unit-level `&byte-order`. That's  more consistent with field attributes. Need to look at what properties make sense to convert.

bbannier:
Closes #713.


doc/programming/parsing.rst
```diff
@@ -352,12 +352,12 @@ Example:
     :expect-failure:
 
 One can also enforce conditions globally at the unit level through a
-property ``%requires = EXPR``. ``EXPR`` works similar as with the
+attribute ``&requires = EXPR``. ``EXPR`` works similar as with the
 attribute: It will be evaluated once the unit has been full parsed,
```
rsmmr, line 356:
```suggestion
attribute ``&requires = EXPR``. ``EXPR`` 
will be evaluated once the unit has been full parsed,
```

rsmmr (approved):
I don't see a documentation change for `%byte-order` -> `&byte-order`. If we didn't have this documented, please add it.

doc/programming/parsing.rst
```diff
@@ -650,6 +648,34 @@ Unit Attributes
...
+
+    .. spicy-code::
+
+        type Foo = unit {
+            version: uint32;
+        } &byte-order=spicy::ByteOrder::Little;
+
+``&requires=EXPR``
+   Enforces post-conditions on the parsed unit. ``EXPR`` must be a boolean
+   expression that will be evaluated after the parsing for the field has
```
rsmmr, line 665:
```suggestion
   expression that will be evaluated after the parsing for the unit has
```

rsmmr (approved):

PR #754, opened 2021-02-17
Add description of enum use to Zeek plugin documentatio (and fix a bug)

rsmmr:


tests/hilti/codegen/init.hlt
```diff
@@ -0,0 +1,20 @@
...
+
+init function void test_init() {
+    hilti::print("init");
+}
+
+preinit function void test_preinit() {
+    # Calling runtime functionality isn't well-defined from
+    # inside a preinit function but this happens to work.
+    hilti::print("preinit");
+}
```
bbannier, line 17: We support multiple `init` and `preinit` functions and should maybe test that here (if hooks of same linkage emit identical strings this shouldn't mess up the btest-diff test).

zeek/compiler/glue-compiler.cc
```diff
@@ -760,6 +760,8 @@ bool GlueCompiler::compile() {
     auto import_ = hilti::declaration::ImportedModule(ID("zeek_rt"), std::string(".hlt"));
     init_module.add(std::move(import_));
 
+    auto preinit_body = hilti::builder::Builder(_driver->context());
```
bbannier, line 763: We could move this down before the loop using it.

bbannier (approved):

zeek/compiler/glue-compiler.cc
rsmmr: prefer to leave where it is as other part could later be adding to it to.

PR #755, opened 2021-02-17
Clarify documentation around unit parameters of basic types.

### LINKED ISSUE #674: inout does not work with basic types
According to the documentation at https://docs.zeek.org/projects/spicy/en/latest/programming/parsing.html?highlight=inout#unit-parameters, a string parameter should be able to be declared as inout.

Trying this:
```
module TESTING;
public type Handshake = unit(inout version: string) {
  handshakes: bytes &size=0;
};
```

Yields the following error message:
```
[error] test3.spicy:3:1-5:3: only parameters of reference type can be 'inout' for struct parameters
```

Expected behavior: no error and parameter is modifiable in the unit.

bbannier:
Closes #674.


rsmmr (commented):
Hmm ... I'm unsure if we should document it like this; it's a bug, not a feature. :)

How about this: put it into a `.. note`, say it's a current problem and link to the issue, then show this as a work-around?

rsmmr (approved):

PR #776, opened 2021-02-22
Support nested fields in parser generator

rsmmr:


spicy/toolchain/include/compiler/detail/codegen/production.h
```diff
@@ -82,7 +87,7 @@ class Meta {
 
 private:
     template<class T>
-    std::optional<T> _value(const NodeRef& n) const {
+    hilti::optional_ref<const T> _value(const NodeRef& n) const {
```
bbannier, line 90: Is this needed? The callers of this function usually return `auto` so any external caller likely gets an `optional_ref`. These external callers sometimes store these values, unaware of the lifetime constraints of `optional_ref`.

spicy/toolchain/src/compiler/codegen/codegen.cc
```diff
@@ -104,7 +104,7 @@ struct VisitorPassInit : public hilti::visitor::PreOrder<void, VisitorPassInit>
         }
 
         auto ns = cg->compileUnit(nu, false);
-        auto attrs = AttributeSet::add(t.attributes(), Attribute("&on-heap"));
+        auto attrs = AttributeSet({Attribute("&on-heap")});
```
bbannier, line 107: I checked whether this would ever throw away a existing, non-empty list of attributes and it seems that does indeed happen. Is that intentional?

spicy/toolchain/src/compiler/codegen/codegen.cc
```diff
@@ -409,58 +409,39 @@ std::optional<hilti::declaration::Function> CodeGen::compileHook(
...
+        params.push_back({ID("__dd"), hilti::type::Auto(), hilti::type::function::parameter::Kind::In, {}});
         params.push_back({ID("__stop"), type::Bool(), hilti::type::function::parameter::Kind::InOut, {}});
     }
-    else if ( item_type ) {
-        params.push_back({ID("__dd"), *item_type, hilti::type::function::parameter::Kind::In, {}});
+    else if ( original_field_type ) {
+        params.push_back({ID("__dd"), hilti::type::Auto(), hilti::type::function::parameter::Kind::In, {}});
 
-        if ( original_item_type && original_item_type->isA<type::RegExp>() )
+        if ( original_field_type->isA<type::RegExp>() && ! is_container )
```
bbannier, line 439: Could you add a comment here why we exclude the code below for fields with containers of regexps?

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -73,20 +78,33 @@ struct ProductionVisitor
...
+
     auto popDestination() {
         auto back = _destinations.back();
         _destinations.pop_back();
+
+        if ( _destinations.size() ) {
+            HILTI_DEBUG(spicy::logging::debug::ParserBuilder, fmt("- pop destination, now: %s", destination()));
+        }
+        else
+            HILTI_DEBUG(spicy::logging::debug::ParserBuilder, fmt("- push destination, now: none"));
```
bbannier, line 94:
```suggestion
            HILTI_DEBUG(spicy::logging::debug::ParserBuilder, fmt("- pop destination, now: none"));
```

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -73,20 +78,33 @@ struct ProductionVisitor
     auto popBuilder() { return pb->popBuilder(); }
 
     auto destination() { return _destinations.back(); }
-    auto pushDestination(const std::optional<Expression>& e) { _destinations.push_back(e); }
+    auto pushDestination(Expression e) {
```
bbannier, line 81:
```suggestion

    auto pushDestination(Expression e) {
```

spicy/toolchain/src/compiler/codegen/unit-builder.cc
```diff
@@ -221,6 +226,21 @@ Type CodeGen::compileUnit(const type::Unit& unit, bool declare_only) {
     auto ft = _pb.parseMethodFunctionType({}, unit.meta());
     v.addField(type::struct_::Field(type::struct_::Field("__parse_stage1", std::move(ft))));
 
+    if ( auto convert = AttributeSet::find(unit.attributes(), "&convert") ) {
+        auto expression = *convert->valueAs<Expression>();
```
bbannier, line 230: I checked whether this works with the following example modified from the new docs,

```.ruby
    module Test;

    function foo(d: Data): int64 {
        return d.data.to_int();
    }

    type Data = unit {
        data: bytes &size=2;
    } &convert=foo(self);

    public type Foo = unit {
        numbers: Data[];

        on %done { print self.numbers; }
    };
```

This fails with the following error after codegen,

```
executing printf 12345678 | spicy-driver programming/examples/_parse-convert-unit.spicy

WARNING: execution failed: /var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/hilti-rt-test-0iOmjx/Test.cc:36:20: error: use of undeclared identifier 'foo'; did you mean 'Foo'?
            return foo(__self);
                   ^
/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/hilti-rt-test-0iOmjx/Test.cc:11:12: note: 'Foo' declared here
    struct Foo;
           ^
1 error generated.
[error] JIT compilation failed
[error] (no error output)
```

bbannier (commented):
I have made a pass over the code, but especially in the patch adding support for nested fields it wasn't always clear to me with certain changes were made I was only able to review this superficially.

spicy/toolchain/src/compiler/codegen/codegen.cc
rsmmr: Yeah, I was trying to remember if there was a reason for transferring all existing attributes over from the original unit to the new struct type, and couldn't remember any. So changed it to only add the one attribute we want.

hilti/toolchain/include/ast/types/auto.h
```diff
@@ -0,0 +1,62 @@
...
+#include <vector>
+
+#include <hilti/ast/type.h>
+#include <hilti/ast/types/unknown.h>
+
+namespace hilti {
+namespace type {
+
+/** AST node for an "auto" type. */
+class Auto : public TypeBase,
```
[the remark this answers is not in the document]
rsmmr: Well, it's complicated. :) The reason for adding `auto` is purely technical: it helps getting around some cyclic type dependencies. One of the later commits uses it for the `$$` hook parameters. I wasn't able to get them to reliably resolve without this new `auto` mechanism--which is nicer anyways because it removes some duplication too.

I don't think replacing this with `type::Unknown` would work right now because generally, type resolution doesn't replace such nodes explicitly. `type::Unknown` is more of a temporary stand-in until we have means to resolve the type otherwise (e.g., some `type()` methods that are dynamically calculating a type return `Unknown` until they can do so). `Auto` on the other hand gives us a specific node to search and replace/update.

Honestly, our type resolution isn't great right now: it has diverged into a mix of multiple approaches over time, and `auto` is certainly a bit of hack here to make things work. I do like the idea of `auto` conceptuallly, but it would be better to use it consistently everywhere where we the language infers types by itself (e.g., `local x = 42` would use it too). But overall, the real task is to clean up type resolution across the AST infrastructure, and probably more along the lines of what you have in mind: always store types in explicit nodes (which could then be `type::Unknown`) and have dynamic computation replace those.

In the meantime I think this will do (and it's HILTI-only, not exposed to Spicy users).

hilti/toolchain/src/compiler/visitors/printer.cc
```diff
@@ -877,7 +881,26 @@ struct Visitor : visitor::PreOrder<void, Visitor> {
...
+            if ( block && block->statements().empty() ) {
+                out << " {}";
+                out.endLine();
+            }
+            else if ( block && block->statements().size() == 1 ) {
+                auto old_compact = out.setCompact(true);
+                out << " { " << block->statements().front() << " }";
+                out.setCompact(old_compact);
+                out.endLine();
+            }
```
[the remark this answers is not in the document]
rsmmr: I found it helpful when looking at the generated code, and the compact flag was already there.

spicy/toolchain/include/compiler/detail/codegen/production.h
rsmmr: Yeah, the point is to avoid the copy of T, which would mean aren't talking about the same node anymore. Semantics match other uses of `optional_ref`: the callers return references to child nodes, and the assumption is that they stay around for as along as the parent node.

spicy/toolchain/src/compiler/codegen/codegen.cc
rsmmr: Will do: only regex fields provide captures (i.e., $1, $2, $3); for vector of regex, it wouldn't be clear what to bind too.

spicy/toolchain/src/compiler/codegen/unit-builder.cc
rsmmr: Nice one, will fix.

bbannier (dismissed):
LGTM :shipit:

PR #778, opened 2021-02-23
Minor documentation fixes

jsiwek:


rsmmr (approved):
Looks good. Time/interval indeed seem  to be working, but we aren't testing them. I'll add a test during merge.

duffy-ocraven:
Is this now the documentation meta compendium, taking over from #441?
I just found a new one in 5.1.11.2. Leveraging MIME Types:
`(if that’s multiple, all will connected and all will receive the same data).`
should be
`(if that’s multiple, all will be connected and all will receive the same data).`

duffy-ocraven:
In 5.1.12 Contexts, is the appearance at 
```public unit may declare such a context through a unit-level property called %context, which takes an arbitrary type as its argument. For example:
public type Foo = unit {
    %property = bytes;
```
of `%property` a typo? That whole example should just be thrown out if so. The example with `%context = int64;` which immediately follows, is a good example.

bbannier:
This issue is closed. Could you open separate, new issues for your above comments @duffy-ocraven so they can be tracked more easily? You can tag them `Documentation`.

duffy-ocraven:
Do you want separate issues, or a compendium approach like #441 ? I mean, since I already have two, those couple can go together, but then it is better for the third to just be a comment there, even if it is days from now? Or will it be better to have each one start a fresh ticket?

bbannier:
With mega issue like #441 it becomes a lot of work to see what is fixed and what work is still open. Since we have tags and issues are cheap, avoid expanding on the issue in comments and feel free to open multiple ones.

PR #783, opened 2021-02-23
Build with -Werror in CI.

bbannier:
We already have this on for CI setups using `run-ci`, but need to
specify it explicitly for platforms using the document build method via
`./configure`.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -171,7 +171,7 @@ macos_catalina_task:
     - pip3 install "btest>=0.66" sphinx_rtd_theme
 
   configure_script:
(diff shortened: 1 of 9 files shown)


rsmmr (commented):
Don't we want this in `ci/run-ci too`?

bbannier:
> Don't we want this in `ci/run-ci` too?

There we already have it, https://github.com/zeek/spicy/blob/759eeadef8c17914e7fb8dba651da3d91c6e5444/ci/run-ci#L97.

rsmmr:
Ah, perfect, good to go then!

rsmmr (approved):

rsmmr (approved):

PR #785, opened 2021-02-24
Add &max-size attribute.

### LINKED ISSUE #767: No way to constrain bytes/string parsing by both total length and null-termination
In context to what I described in https://github.com/zeek/spicy/discussions/765, I was looking for a way to constrain the parsing of null-terminated string by whether it reaches an arbitrary N-byte limit before seeing the null-terminator and this is one of the first ways I thought to try expressing that:

```
module Test;

public type NullTerminatedString = unit {
    s: bytes &size=4 &until=b"\x00";
};
```

```
$ printf 'testing\000' | spicy-dump test.spicy
[fatal error] terminating with uncaught exception of type spicy::rt::ParseError: parse error: end-of-data reached before &until expression found (test.spicy:4:8)
```

Not sure that's the expected/intuitive way for that to have failed, but is there ultimately some way to accomplish  "parse until a either a given byte-sequence is seen *or* an N-byte limit is reached", or can the combination of `&size` and `&until` be defined to have that meaning?

bbannier:
This patch adds a &max-size field and unit attribute. The attribute
allows setting up a maximum number of bytes a parse step should consume.

It is conceptually similar to the existing &size attribute but differs
from it in that an error is raised only if more than the specified bytes
were extracted (&size enforces that exactly the specified number of
bytes is extracted).

@max-size cannot be used together with &size.

Closes #767.


spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -877,15 +911,32 @@ struct ProductionVisitor
...
+            // the last byte acting as sentinel. Once parsing has finished we can
+            // then validate that parsing hasn't reached the sentinel byte.
+            maxSize = builder::incrementPrefix(builder::coerceTo(*a->valueAs<Expression>(), type::UnsignedInteger(64)));
+
+        if ( size && maxSize )
+            hilti::logger().internalError("&size and &max-size cannot be combined");
+        else if ( maxSize )
+            length = maxSize;
+        else if ( size )
+            length = size;
```
bbannier, line 930: This or the finalizer code currently triggers an HILTI error with the following unit decl,

```.ruby
type Foo = unit {
    xs: bytes &eod;
} &max-size=1;
```

We also should add tests for this (probably similar to the field-level tests).

spicy/toolchain/src/compiler/parser/scanner.ll
```diff
@@ -48,7 +48,7 @@ static std::string expandEscapes(Driver* driver, std::string s, spicy::detail::p
 address4  ({digits}"."){3}{digits}
 address6  ("["({hexs}:){7}{hexs}"]")|("["0x{hexs}({hexs}|:)*"::"({hexs}|:)*"]")|("["({hexs}|:)*"::"({hexs}|:)*"]")|("["({hexs}|:)*"::"({hexs}|:)*({digits}"."){3}{digits}"]")
 
-attribute \&(bit-order|byte-order|chunked|convert|count|cxxname|default|eod|internal|ipv4|ipv6|length|no-emit|nosub|on-heap|optional|originator|parse-at|parse-from|priority|requires|responder|size|static|synchronize|transient|try|type|until|until-including|while|have_prototype)
+attribute \&(bit-order|byte-order|chunked|convert|count|cxxname|default|eod|internal|ipv4|ipv6|length|max-size|no-emit|nosub|on-heap|optional|originator|parse-at|parse-from|priority|requires|responder|size|static|synchronize|transient|try|type|until|until-including|while|have_prototype)
```
bbannier, line 51: Still need to add the attribute to the docs.

tests/spicy/types/unit/unit-max-size-fail.spicy
```diff
@@ -0,0 +1,12 @@
+# @TEST-EXEC-FAIL: spicyc -j %INPUT >>output 2>&1
+# @TEST-EXEC: btest-diff output
+#
+# @TEST-DOC: Test invalid uses of unit `&max-size`
+
+# FIXME(bbannier): This currently triggers a bus error, at leats on macos-10.15.7.
```
bbannier, line 6: This needs to be fixed.

tests/spicy/types/unit/field-max-size-fail.spicy
```diff
@@ -0,0 +1,10 @@
+# @TEST-EXEC-FAIL: spicyc -j %INPUT >>output 2>&1
```
bbannier, line 1: With this branch rebased onto `master` no diagnostic is triggered here as of 28733d55e388050bffd2d702f257f5812c1db20b.

bbannier: Still needs a rebase on the fix for #791.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -933,6 +949,13 @@ struct ProductionVisitor
 
         pb->finalizeUnit(true, p.location());
 
+        if ( auto a = AttributeSet::find(p.unitType().attributes(), "&max-size") ) {
+            // Restore parser state.
+            auto ncur = state().ncur;
+            popState();
+            builder()->addAssign(state().cur, *ncur);
+        }
+
```
rsmmr, line 958: As we just talked about, this isn't reliably triggering an error in all cases, let's trigger it here directly.

rsmmr (changes requested):

doc/programming/parsing.rst
```diff
@@ -200,34 +200,41 @@ Limiting Input Size
 ^^^^^^^^^^^^^^^^^^^
 
 You can limit the input that a field receives by attaching a
-``&size=EXPR`` attribute that specifies the number of raw bytes to
-make available. This works on top of any other attributes that control
-the field's parsing. From the field's perspective, such a size limit
-acts just like reaching the end of the input stream at the specified
-position. Example:
+``&size=EXPR`` or ``&max-size=EXPR`` attribute that specifies the
```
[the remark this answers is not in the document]
bbannier: I split this into two sections now, one for parsing fields with known size and one for `&max-size` as a defensive mechanism.

doc/programming/parsing.rst
```diff
@@ -237,6 +237,40 @@ them all.
     Parsing a regular expression would make a nice example for
     ``&size`` as well.
 
+Defensively Limiting Input Size
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+On its own, parser place no intrinsic upper limit on the size of variable-size
```
rsmmr, line 243:
```suggestion
On its own, parsers place no intrinsic upper limit on the size of variable-size
```

doc/programming/parsing.rst
```diff
@@ -237,6 +237,40 @@ them all.
...
+
+.. spicy-output:: max-size.spicy
+    :exec: printf '\001\002\003\004\005\000' | spicy-driver %INPUT
+    :show-with: foo.spicy
+
+Here ``x`` will parse a ``NULL``-terminated byte sequence (excluding the
+terminating ``NULL``), but never more than 1024 bytes.
+
+If more than ``&max-size`` bytes are consumed during parsing, an error will be
+triggered.
```
rsmmr, line 270: I'd move this to above just before "This attributes works on top ...".

doc/programming/parsing.rst
```diff
@@ -237,6 +237,40 @@ them all.
...
+
+On its own, parser place no intrinsic upper limit on the size of variable-size
+fields or units. This can have negative effects like out-of-memory errors,
+e.g., when available memory is constrained, or for malformed input.
+
+As a defensive mechanism you can put an upper limit on the data a field or unit
+receives by attaching a ``&max-size=EXPR`` attribute where ``EXPR`` is an
+unsigned integer specifying the upper limit of number of raw bytes a field or
+unit should receive. This attribute works on top of any other attributes that
+control parsing.
```
rsmmr, line 251:
```suggestion
control parsing. Example:
```

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -645,7 +656,23 @@ struct ProductionVisitor
...
+                                              builder::memberCall(*ncur, "offset", {}));
+            auto exceeded = builder()->addIf(std::move(cond));
+            pushBuilder(exceeded, [&]() {
+                // We didn't finish parsing the data, which is an error.
+                if ( ! destination().type().isA<type::Void>() && ! field->isTransient() )
+                    // Clear the field in case the type parsing has started
+                    // to fill it.
+                    builder()->addExpression(builder::unset(state().self, field->id()));
+
+                pb->parseError("parsing not done with &max-size bytes", a->meta());
```
rsmmr, line 671:
```suggestion
                pb->parseError("parsing not done within &max-size bytes", a->meta());
```

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -933,7 +973,20 @@ struct ProductionVisitor
 
         pb->finalizeUnit(true, p.location());
 
-        if ( auto a = AttributeSet::find(p.unitType().attributes(), "&size") ) {
+        if ( auto a = AttributeSet::find(p.unitType().attributes(), "&max-size") ) {
+            // Check that we did not read into the sentinel byte.
+            auto cond = builder::greaterEqual(builder::memberCall(state().cur, "offset", {}),
+                                              builder::memberCall(*state().ncur, "offset", {}));
+            auto exceeded = builder()->addIf(std::move(cond));
+            pushBuilder(exceeded, [&]() { pb->parseError("parsing not done with &max-size bytes", a->meta()); });
```
rsmmr, line 981:
```suggestion
            pushBuilder(exceeded, [&]() { pb->parseError("parsing not done within &max-size bytes", a->meta()); });
```

rsmmr (dismissed):

PR #787, opened 2021-02-24
Integrate spicy-analyzers submodule

rsmmr:


rsmmr:
@bbannier you could take a look at this at this point, not cast in stone yet, open for feedback on the submodule integration

tests/Makefile
```diff
@@ -0,0 +1,24 @@
+
```
bbannier, line 1: We already have two ways to trigger tests (BTest and via CMake where the former also executes the latter), and introducing another way to rum them will it easier to miss one (especially since the files in this directory makes it look like just running `btest` should be sufficient).

I would much prefer if we'd just add BTest which 
- checks that we build with the analyzers with a `@TEST-REQUIRES: have-zeek-plugin` (I think that's missing below),
- creates a build directory for the analyzer build,
- configures the analyzer build in a way that it detects an build-directory or installed Spicy to support both installation tests as well,
- builds the analyzers in their build directory,
- executes the analyzer tests from that BTest with `btest -c <ANALYZER_PATH>/tests/btest.cfg -j 5` (we don't have a way to access the current BTest instance `-j` value, so just use something smallish for CI; currently we just have 3 analyzers).

This should also avoid a lot of the curn this introduces in e.g., the setup.

We'll probably still need something in `.cirrus.yml` to archive the package BTest artifacts.

.gitmodules
```diff
@@ -43,3 +43,6 @@
 [submodule "3rdparty/any"]
 	path = 3rdparty/any
 	url = https://github.com/thelink2012/any.git
+[submodule "zeek/analyzers"]
+	path = zeek/spicy-analyzers
+	url = https://github.com/zeek/spicy-analyzers
```
bbannier, line 48: Could we add these to `3rdparty/spicy-analyzers` instead? Putting them alongside our sources opens up the possibility of us depending on code in them as well by accident (where the package should instead be a pure consumer).

hilti/toolchain/src/base/util.cc
```diff
@@ -19,6 +19,14 @@
 
 // We include pathfind directly here so we do not have to work
 // around it being installed by its default install target.
+
+#ifdef __linux__
+#ifndef __gnu_linux__
+// pathfind checks for __gnu_linux__, which isn't set on Alpine.
```
bbannier, line 25: Let's link https://github.com/bkloppenborg/pathfind/issues/5 here.

rsmmr: Upstream is fixed, so updating the submodule instead.

rsmmr:
@bbannier take another look

tests/Makefile
rsmmr: As discussed, recursive btest calls don't work very well.

.gitmodules
rsmmr: I prefer to keep 3rdparty to software that we don't control.

bbannier:
I like this much better since the default build path is not slower now, and one can run the Spicy BTests and if interested also explicitly trigger just the analyzer tests.

It would be nice to also tests building and running analyzers against an installed Spicy (some `test-analyzer-install` target in `test/Makefile`), but we can safe that for some later point since that file already triggers the build of the external analyzers so it should be possible to tweak it for that in the future.

Issue #771, opened 2021-02-19
Undocumented vector &until boolean parsing logic


jsiwek:
Contrived example, but maybe works like I thought it would:

```
module Test;

type IntegerPairs = unit {
    i0: uint8;
    i1: uint8;
};

public type MyUnit = unit {
    a: IntegerPairs[] &until=(|self.a| == 2);
};
```
```
$ printf 'testtest' | spicy-dump test.spicy
Test::MyUnit {
  a: [
    Test::IntegerPairs {
      i0: 116
      i1: 101
    }
    Test::IntegerPairs {
      i0: 115
      i1: 116
    }
  ]
}
```

Except boolean conditions seems contrary/missing to existing docs for `vector &until` (or maybe it's actually not intended to work):

https://github.com/zeek/spicy/blob/b3fc756045704cf61427c0060d6ae9fd57643d86/doc/programming/parsing.rst

> &until=EXPR
>
> Parses elements until one with the value EXPR is encountered. **EXPR must be of the same type as the vector's elements**. Once the specified element is encountered, vector parsing stops without including the matching one into the field's vector value.

Also may be worth stressing that using a boolean `&until` does push the last-encountered element into the array while a value-equality `&until` does not, or is that a bug ?

jsiwek:
> Also may be worth stressing that using a boolean `&until` does push the last-encountered element into the array while a value-equality `&until` does not, or is that a bug ?

Think that's my mistake/confusion with the contrived test.  Extending it a bit further think it makes sense and shows the expected element that triggered the first successful `&until` evaluation did actually get thrown out:

```
module Test;

type IntegerPairs = unit {
    i0: uint8;
    i1: uint8;
};

public type MyUnit = unit {
    a: IntegerPairs[] &until=(|self.a| == 2);
    b: IntegerPairs[];
};
```
$ printf 'testtest' | spicy-dump test.spicy
Test::MyUnit {
  a: [
    Test::IntegerPairs {
      i0: 116
      i1: 101
    }
    Test::IntegerPairs {
      i0: 115
      i1: 116
    }
  ]
  b: [
    Test::IntegerPairs {
      i0: 115
      i1: 116
    }
  ]
}
```

bbannier:
I think you are correct @jsiwek. Is there anything that needs to be done here @rsmmr?

rsmmr:
Yeah, I think we're good here.

PR #789, opened 2021-02-24
Improve docs for vector &until parsing

jsiwek:
Explicitly mention that boolean expressions are allowed.

(i.e. what would help in relation to #771)


rsmmr:
Oh, I get the confusion now. The current documentation is simply wrong: it always needs to be a boolean expression.

PR #793, opened 2021-02-25
Bring back validation of field parseability.

### LINKED ISSUE #791: Add back validation of field parseability
28733d55e388050bffd2d702f257f5812c1db20b removed a baseline validating that types are checked for parseabiltiy and in fact the corresponding test does not pass anymore. We should fix the implementation so the tests removed in 28733d55e388050bffd2d702f257f5812c1db20b pass again and can be added back.

rsmmr:
Closes #791.


bbannier (approved):
LGTM.

PR #795, opened 2021-02-25
Use same filenames for TFTP doc example snippets.

bbannier:


rsmmr (dismissed):
I believe this must have been just a glitch, the `show-with` should take care of that. I just tried regenerating the files and it comes out right (i.e., `rm tutorial/examples/_rrq.spicy.output*`)

bbannier:
> I believe this must have been just a glitch, the `show-with` should take care of that. I just tried regenerating the files and it comes out right (i.e., `rm tutorial/examples/_rrq.spicy.output*`)

Regenerating the file again also fixed it for me so I rolled back the change to the input.

rsmmr (approved):

PR #797, opened 2021-02-26
Add language support to explicitly trigger a parse error.

### LINKED ISSUE #768: Add language support to explicitly trigger a parse error
Wasn't sure if it's intended for `&requires` to be evaluated before field hooks:

```
module Test;

public type MyUnit = unit {
    s: bytes &size=4 &requires=(self.flag);

    on s { self.flag = True; }

    var flag: bool;
};
```
```
$ printf 'test' | spicy-dump test.spicy
[fatal error] terminating with uncaught exception of type spicy::rt::ParseError: parse error: &requires failed ($$ == test) (test.spicy:4:22)
```

Initially thought that field hooks might get evaluated first in real example here and found otherwise:

https://github.com/zeek/spicy/compare/topic/jsiwek/pe#diff-36ceb1dc5c1f8f65e1fb3553b2d552ef12f58b3651f1df59aafa9e6c3fafc63eR64-R67

* Generally seems to be lack of documentation on evaluation order and adding that could help.

* For what I was trying to do, would be convenient if the order was different, but realize it's kind of arbitrary/implicit, so might generally be better to have an explicit way to induce parsing errors from within arbitrary hook-code itself.  E.g. either with a `throw` or special `return` value.  Kind of like mentioned in previous error-handling discussions https://github.com/zeek/spicy/issues/89#issuecomment-648637582 and https://github.com/zeek/spicy/discussions/616#discussioncomment-185009

bbannier:
This patch adds a Spicy `throw` statement of the form

    throw EXPR;

which when hit triggers a parse error with the message `EXPR`.

Closes #768.


doc/programming/language/statements.rst
```diff
@@ -207,22 +207,14 @@ there's no matching case.
...
-Throws an exception, letting execution stop at the current location
-and propagating ``EXPR`` up the call stack to the nearest matching
-:ref:`statement_try`. If not match is found, the exception eventually
-bubbles up to the host application. If that doesn't catch it either,
-the runtime will abort execution. ``EXPR`` must be of type
-:ref:`type_exception`. ``EXPR`` can be skipped to rethrow the current
-exception inside a ``catch`` block.
+Triggers a parse error exception with the message indicated by ``EXPR``. ``EXPR`` needs
+to be a :ref:`type_string`. ``throw`` aborts parsing and if not handled in the
+calling process the runtime.
```
rsmmr, line 217:
```suggestion
to be a :ref:`type_string`. ``throw`` aborts parsing.
```

spicy/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -210,6 +210,12 @@ struct PreTransformVisitor : public hilti::visitor::PreOrder<void, PreTransformV
             error("'stop' can only be used inside a 'foreach' hook", p);
     }
 
+    void operator()(const hilti::ctor::Exception& e, position_t p) {
+        if ( auto x = e.value().tryAs<hilti::expression::Ctor>() )
+            if ( ! x->type().isA<type::String>() )
+                error("exceptions need to be a string", p);
+    }
+
```
rsmmr, line 218: This should probably move into the HILTI validator: currently all exception arguments must be a single strings anyways.

rsmmr (approved):

PR #799, opened 2021-03-01
Enable Spicy scripts to compile code conditionally based on Spicy/Zeek version.

rsmmr:
This generalizes the preprocessor logic we already had for EVT files and makes it available to Spicy scripts as well.


hilti/runtime/include/util.h
```diff
@@ -6,6 +6,7 @@
 
 #include <algorithm>
 #include <list>
+#include <map>
```
bbannier, line 9: Unused in this file?

hilti/toolchain/include/base/preprocessor.h
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Process one preprocessor directive of the form `@<id> [expression]`.
+     *
+     * This evaluates the directive and updates internal state accordingly.
+     *
+     * @param directive a supported directive ID, which must start with `@`.
+     * @param expression string with expression that the directive is taking
+     * @result the new inclusion state for subsequent input, or an error if the directive could not be processed
+     */
+    Result<State> processLine(const std::string_view& directive, const std::string_view& expression = "");
```
bbannier, line 49:
```suggestion
    Result<State> processLine(std::string_view directive, std::string_view expression = "");
```

hilti/toolchain/src/base/preprocessor.cc
```diff
@@ -0,0 +1,108 @@
...
+    auto id = std::string(m[0]);
+
+    std::string op;
+    int want = 0;
+
+    if ( m.size() == 3 ) {
+        // "<id> <operator> <expr>"
+        op = std::string(m[1]);
+        want = 0;
+        hilti::rt::atoi_n(m[2].begin(), m[2].end(), 10, &want);
```
bbannier, line 77: Not a new issue, but we probably should check the return value here to catch parse errors.

zeek/compiler/driver.cc
```diff
@@ -80,6 +80,11 @@ Driver::Driver(const std::string& argv0, int zeek_version) : spicy::Driver("<Spi
         hilti::configuration().initLocation(argv0);
 
     spicy::Configuration::extendHiltiConfiguration();
+
+    auto& config = spicy::configuration();
+    config.preprocessor_constants["HAVE_ZEEK"] = 1;
```
bbannier, line 85: Not really needed right now, but we could prefix this name `SPICY_HAVE_ZEEK`. This would only become relevant should we allow users to set their own preprocessing directives, but might be good to do now that we make these constants available to users.

tests/spicy/tools/preprocessor.spicy
```diff
@@ -0,0 +1,60 @@
+# @TEST-EXEC: echo == spicyc >>output
```
bbannier, line 1: Since we support stray spaces in some places,  let's maybe make this test less "pretty", e.g.,

```.patch
diff --git a/tests/spicy/tools/preprocessor.spicy b/tests/spicy/tools/preprocessor.spicy
index 9df8b5d7..e69c043b 100644
--- a/tests/spicy/tools/preprocessor.spicy
+++ b/tests/spicy/tools/preprocessor.spicy
@@ -11,7 +11,7 @@

 module Foo;

-@if HAVE_ZEEK
+@if  HAVE_ZEEK
 print "have zeek";
 @else
 print "not have zeek";
@@ -25,11 +25,11 @@ print "not have zeek";
 print "have spicy version";
 @endif

-@if SPICY_VERSION >= 400
+@if SPICY_VERSION  >= 400
 print "have spicy version >= 0.4";
 @endif

-@if SPICY_VERSION >= 40000
+@if SPICY_VERSION >=  40000
 print "have spicy version >= 4";
 @endif

@@ -42,7 +42,7 @@ no valid Spicy syntax here.
 @endif

 @if HAVE_ZEEK
-    @if ZEEK_VERSION > 10000
+    @if ZEEK_VERSION  >  10000
     print "have zeek and zeek version > 1.0";
     @else
     print "have zeek but zeek version < 1.0";
```

doc/scripts/spicy-pygments.py
```diff
@@ -48,6 +49,8 @@ class SpicyLexer(RegexLexer):
             (r'#.*$', Comment),
         ],
 
+        'directives': [(r'(@(if|else|endif))\b', Comment.Preproc)],
```
bbannier, line 52: Nice.

bbannier (approved):

zeek/compiler/driver.cc
rsmmr: hmm, there's `ZEEK_VERSION` as well, and `SPICY_ZEEK_VERSION` isn't too great. I could see adding an explicit namespace (`Spicy::ZEEK_VERSION`), but it seems overkill right now, and I'm also not very worried about conflicts. Let's just leave as is.

PR #816, opened 2021-03-04
Reorg splitting the spicy-plugin out of the main repository

rsmmr:


bbannier (commented):
There's a lot going on here, so I wasn't able to go very deep, but it looks like most of the changes here are shuffling stuff around and adjusting configurations in a lot of places.

PR #818, opened 2021-03-04
Distribute Linux binary artifacts as deb or rpm files

### LINKED ISSUE #815: Distribute Linux binary artifacts as deb or rpm files
At least on Linux, we should distribute binary artifacts as distribution packages to make it easier to install and uninstall them.

bbannier:
Closes #815.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -330,15 +330,16 @@ docker_centos_7_task:
   configure_script:
     - source scl_source enable devtoolset-9 && LDFLAGS="-static-libstdc++ -static-libgcc" ./configure --generator=Ninja --enable-ccache --prefix=/opt/spicy --with-zeek=/opt/zeek --enable-werror
   build_script:
-    - source scl_source enable devtoolset-9 && ninja -j4 -C build install install-spicy-analyzers
+    - source scl_source enable devtoolset-9 && ninja -j4 -C build check all package
+    - cp build/spicy.rpm .
+    - rpm -i ./spicy.rpm
+    - rm -rf build
   test_install_script:
     - source scl_source enable devtoolset-9 && SPICY_INSTALLATION_DIRECTORY=/opt/spicy make -C tests test-all-install
     - zeek -N | grep Spicy
-  packaging_script:
-    - source scl_source enable devtoolset-9 && ninja -C build package
 
   packages_artifacts:
-    path: build/spicy-*.tar.gz
+    path: ./spicy.{deb,rpm}
     type: application/gzip
 
 docker_centos_8_task:
@@ -370,15 +371,16 @@ docker_centos_8_task:
     # flex-2.6.1 which generates code which triggers sign-compare diagnostics.
     - ./configure --generator=Ninja --enable-ccache --prefix=/opt/spicy --with-zeek=/opt/zeek
   build_script:
-    - ninja -j4 -C build install install-spicy-analyzers
+    - ninja -j4 -C build check all package
+    - cp build/spicy.rpm .
+    - rpm -i ./spicy.rpm
+    - rm -rf build
   test_install_script:
     - SPICY_INSTALLATION_DIRECTORY=/opt/spicy make -C tests test-all-install
     - zeek -N | grep Spicy
-  packaging_script:
-    - ninja -C build package
 
   packages_artifacts:
-    path: build/spicy-*.tar.gz
+    path: ./spicy.{deb,rpm}
     type: application/gzip
 
 validate_release_tarball_task:
@@ -470,15 +472,16 @@ docker_debian9_task:
   configure_script:
     - ./configure --generator=Ninja --enable-ccache --prefix=/opt/spicy --with-zeek=/opt/zeek --enable-werror
   build_script:
-    - ninja -j4 -C build install install-spicy-analyzers
+    - ninja -j4 -C build check all package
+    - cp build/spicy.deb .
+    - dpkg --install ./spicy.deb
+    - rm -rf build
   test_install_script:
     - SPICY_INSTALLATION_DIRECTORY=/opt/spicy make -C tests test-all-install
     - zeek -N | grep Spicy
-  packaging_script:
-    - ninja -C build package
 
   packages_artifacts:
-    path: build/spicy-*.tar.gz
+    path: ./spicy.{deb,rpm}
     type: application/gzip
 
 docker_debian10_task:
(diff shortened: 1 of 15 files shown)


CMakeLists.txt
```diff
@@ -263,5 +263,27 @@ message(
 # unmodified links to the latest version.
 string(TOLOWER ${CMAKE_SYSTEM_NAME} SPICY_SYSTEM_NAME)
 set(CPACK_PACKAGE_FILE_NAME "${CMAKE_PROJECT_NAME}-${SPICY_SYSTEM_NAME}")
+set(CPACK_BINARY_TZ OFF)
```
bbannier, line 266: Need to add the generated distro packages to the cirrus build artifacts.

CMakeLists.txt
```diff
@@ -263,5 +263,27 @@ message(
...
+set(CPACK_BINARY_TGZ ON)
```
bbannier, line 267: Think about adding a step checking whether the archives can be installed.

Especially the DEB packager seems to have issues with the soft links for 3rdparty stuff; let's see whether the CMake-related patch in this branch helps already.

bbannier:
This is close to ready and most tests seem to pass when running from a Spicy installed from a distribution package. It still needs a rebase on the merge of #816 though since the BTest suite doesn't really deal correctly with disabling the plugin. Once that is merged and this one rebased, removing the `DO NOT COMMIT` patches should lead to something close to mergable.

.cirrus.yml
```diff
@@ -167,15 +167,15 @@ macos_catalina_task:
...
 
   configure_script:
-    - ./configure --generator=Ninja --with-bison=/usr/local/opt/bison --with-flex=/usr/local/opt/flex --enable-ccache --enable-werror
+    - ./configure --generator=Ninja --with-bison=/usr/local/opt/bison --with-flex=/usr/local/opt/flex --enable-ccache --enable-werror --build-zeek-plugin=no
   build_script:
-    - ninja -C build all build-spicy-analyzers check
+    - ninja -C build all check
   test_build_script:
-    - make -C tests test-all-build
+    - make -C tests test-spicy-build
```
rsmmr, line 178: `make test` is equivalent to `make test-spicy-build`. I suggest we use `test` as (1) it's more intuitive (in the sense of "don't care what specifically we're testing"), and (2) easy to change what it binds to in the Makefile if we wanted to later.

.cirrus.yml
```diff
@@ -296,7 +296,7 @@ docker_alpine_3_12_task:
   build_script:
     - ninja -j4 -C build install install-spicy-analyzers
   test_install_script:
-    - SPICY_INSTALLATION_DIRECTORY=/opt/spicy make -C tests test-all-install
+    - SPICY_INSTALLATION_DIRECTORY=/opt/spicy make -C tests test-spicy-install
```
rsmmr, line 299: Similar to above, use `test-install` -- except that that doesn't exist yet, so let's add that as a shortcut for `test-spicy-install`.

CMakeLists.txt
```diff
@@ -195,6 +195,42 @@ execute_process(COMMAND ${PROJECT_SOURCE_DIR}/scripts/autogen-version
...
                 OUTPUT_STRIP_TRAILING_WHITESPACE)
 
+# Packaging.
+# Hide version number in package filename so that we can e.g., provide
+# unmodified links to the latest version.
+string(TOLOWER ${CMAKE_SYSTEM_NAME} SPICY_SYSTEM_NAME)
+set(CPACK_PACKAGE_FILE_NAME "${CMAKE_PROJECT_NAME}")
+
+set(CPACK_PACKAGE_NAME "spicy")
+set(CPACK_PACKAGE_CONTACT "spicy@lists.zeek.org")
```
rsmmr, line 205: Let's use `info@zeek.org`

doc/installation.rst
```diff
@@ -18,56 +18,62 @@ Linux
...
 We provide pre-built binaries for some platforms from the Git ``master``
-branch. Artifacts are distributed as TAR archives which can be unpacked to any
-location::
+branch. Artifacts are distributed as DEB or RPM distribution packages, or TAR archives; TAR archives need to be unpacked to `/opt/spicy`::
 
-    # Unpack archive to e.g., /opt/spicy overwriting any previous installation.
-    # Writing to /opt/spicy likely requires superuser priviledges.
-    # The path /opt/spicy can be replaced with a custom path.
+    # Using DEB distribution packages.
+    dpkg --install spicy.deb
```
rsmmr, line 24: I don't recall what we said about the names: can we get version numbers into them? So for releases it would be `spicy-1.0.0.{tar.gz,deb,rpm}`, for master/nightly a date would be nice.

doc/installation.rst
```diff
@@ -18,56 +18,62 @@ Linux
...
+    # Unpack archive to /opt/spicy (required) replacing any previous
+    # installation. Writing to /opt/spicy likely requires superuser privileges.
     rm -rf /opt/spicy && mkdir /opt/spicy
-    tar xf spicy-linux.tar.gz -C /opt/spicy --strip-components=1
+    tar xf spicy.tar.gz -C /opt/spicy --strip-components=1
 
 The binaries might require installation of additional dependencies;
 see the ``Dockerfile`` for the respective platform.
 
 alpine-3.12
```
rsmmr, line 39: I keep thinking that it would be nice to turn all the download links into a real table. And then maybe have two columns "release" and "development" with corresponding links ("release" filled once we cut 1.0.0).

docker/Dockerfile.debian-9
```diff
@@ -10,7 +10,7 @@ ENV CCACHE_DIR "/var/spool/ccache"
 ENV CCACHE_COMPRESS 1
 
 ENV CMAKE_DIR "/opt/cmake"
-ENV CMAKE_VERSION "3.17.2"
+ENV CMAKE_VERSION "3.18.0"
```
rsmmr, line 13: Do we have at least one platform left that's testing with CMake 3.15? Because that's our current mininum requirement, so we should make sure it works with that still (outside of packaging, that's fine).

doc/installation.rst
bbannier: We want to link to packages in the docs, and I am unsure how to do that without too much complications if we include version numbers. If we'd only link to artifacts for releases we could use the existing release automation for that, but for builds from master it seems we'd need to some form of templating in the documentation. Do you know if there's something in Sphinx we could use?

rsmmr: Then lets do it just for releases. Leaving master builds at some generic name (say `spicy-dev.tbz`) is fine, less important for that.

docker/Dockerfile.debian-9
bbannier: I explicily set a single Docker platform (centos-7) to use the minimum CMake version.

PR #826, opened 2021-03-15
Do not include Spicy in Docker images.

bbannier:


ci/Dockerfile.dockerhub
```diff
@@ -0,0 +1,46 @@
+# This Dockerfile is used to create images published to DockerHub. It needs to
+# be kept in sync with the corresponding Dockerfile we use and test in CI
+# (docker/Dockerfile.ubuntu-20).
+
+FROM ubuntu:focal
+
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
+ARG ZEEK_LTS=1
+ARG ZEEK_VERSION=3.0.11-0
```
rsmmr, line 10: Let's switch this to 4.0.

ci/Dockerfile.dockerhub
```diff
@@ -0,0 +1,46 @@
...
+ && rm -rf /var/lib/apt/lists/* \
+ # Install a recent CMake.
+ && mkdir -p /opt/cmake \
+ && curl -L https://github.com/Kitware/CMake/releases/download/v3.18.0/cmake-3.18.0-Linux-x86_64.tar.gz | tar xzvf - -C /opt/cmake --strip-components 1
+
+ENV PATH="/opt/zeek/bin:/opt/cmake/bin:${PATH}"
+
+# Install Spicy.
+ADD . /opt/spicy/src
+RUN cd /opt/spicy/src && ./configure --generator=Ninja --prefix=/opt/spicy --with-zeek=/opt/zeek && ninja -C build install && rm -rf build
```
rsmmr, line 45: Are we going to keep installing the Zeek plugin and analyzers through this as well? If so, I'd add an explicit `--build-zeek-plugin=yes` so that it won't break when we switch the default. And currently it won't install the analyzers.

Or we go the zkg route here as well, either is way fine I think.

rsmmr (approved):

ci/Dockerfile.dockerhub
bbannier: The packages will be ready later today or tomorrow. I'd suggest that we do a more general sweep of setups once that happens, see #827.

ci/Dockerfile.dockerhub
bbannier: I went with Spicy w/o plugin and installation of additional Zeek components via zkg.

rsmmr (approved):

PR #838, opened 2021-03-17
Fix vector parsing potentially missing EOD.

### LINKED ISSUE #817: File analyzer tries to parse beyond end of file in some circumstances
When using a file analyzer in zeek, it tried to parse beyond the end of the file in some circumstances.

This behavior is currently exhibited for one file in the test-case of the PNG file analyzer. To reproduce - check out the ``topic/johanna/png`` branch of spicy-analyzers (or wait until https://github.com/zeek/spicy-analyzers/pull/5 is merged).

Then execute ``btest -t file/png.zeek`` in that repository. This will yield a weird.log containing:

```
#separator \x09
#set_separator	,
#empty_field	(empty)
#unset_field	-
#path	weird
#open	2021-03-04-14-57-51
#fields	ts	uid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	name	addl	notice	peer	source
#types	time	string	addr	port	addr	port	string	string	bool	string	string
1258577885.355204	-	-	-	-	-	parse error: expecting 4 bytes for unpacking value (/Users/johanna/spicy/spicy-analyzers/analyzer/file/png/PNG.spicy:22:10)	FYdCSR32lxKYM7mSmk	F	zeek	-
#close	2021-03-04-14-57-51
```

When one extracts this file and runs it against spicy-driver manually, this error does not occur. Inspection of the file shows that there is no trailing data - parsing should just end.

The file is parsed using this construct:

```
: Chunk(self)[] &eod;
```

And I assume that, for some reason, parsing of the next chunk already starts, even though the file ends. The parsing error occurs while trying to parse the first element of Chunk.

rsmmr:
When parsing a vector bounded by &eod, the parser could enter the
code for reading the next element too early before it could reliably
tell that EOD hadn't been reached yet. The fix is to change `atEod()` so
that if stream is still unfrozen, it waits for at least one available
byte before making the decision.

This is hard to test for because it requires yielding & resuming the
input in the right way, but the PNG can act as a test case: it was
producing an unexpected weird, which is now gone.

Closes #817.

Merge with https://github.com/zeek/spicy-analyzers/pull/19

### DIFF
--- a/spicy/runtime/src/tests/parser.cc
+++ b/spicy/runtime/src/tests/parser.cc
@@ -57,16 +57,18 @@ TEST_CASE("ParserPort") {
 
 TEST_CASE("atEod") {
     auto stream = hilti::rt::ValueReference<hilti::rt::Stream>();
+    auto filters = hilti::rt::StrongReference<filter::detail::Filters>();
 
     SUBCASE("empty") {
         bool expanding = false;
         SUBCASE("expanding view") { expanding = true; }
         SUBCASE("not expanding view") { expanding = false; }
 
-        CHECK_EQ(detail::atEod(stream, stream->view(expanding)), ! expanding);
+        // This would now yield to wait for input before deciding whether EOD has been reached.
+        // CHECK_EQ(detail::atEod(stream, stream->view(expanding), filters), ! expanding);
 
         stream->freeze();
-        CHECK(detail::atEod(stream, stream->view(expanding)));
+        CHECK(detail::atEod(stream, stream->view(expanding), filters));
     }
 
     SUBCASE("not empty") {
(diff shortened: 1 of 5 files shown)


spicy/runtime/src/tests/parser.cc
```diff
@@ -57,16 +57,18 @@ TEST_CASE("ParserPort") {
...
+    auto filters = hilti::rt::StrongReference<filter::detail::Filters>();
 
     SUBCASE("empty") {
         bool expanding = false;
         SUBCASE("expanding view") { expanding = true; }
         SUBCASE("not expanding view") { expanding = false; }
 
-        CHECK_EQ(detail::atEod(stream, stream->view(expanding)), ! expanding);
+        // This would now yield to wait for input before deciding whether EOD has been reached.
+        // CHECK_EQ(detail::atEod(stream, stream->view(expanding), filters), ! expanding);
```
bbannier, line 68: Can we just remove this commented out part? "Now" will not be useful in the future.

bbannier (approved):

PR #841, opened 2021-03-17
Bump Zeek LTS version to 4.0.0 in some CI and Docker images.

### LINKED ISSUE #827: Update setups using zeek-lts to 4.0.0
Once packages for zeek-lts switch to 4.0.0 later today or tomorrow we should update all setups using zeek to reflect that.

bbannier:
We still keep at least a single setup with zeek-3.0.0 so we also have
some coverage there.

Closes #827.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -9,7 +9,7 @@ clang9_zeek_ubuntu_debug_task:
     dockerfile: ci/Dockerfile
     docker_arguments:
       - ZEEK_LTS:
-      - ZEEK_VERSION: 3.2.2-0
+      - ZEEK_VERSION: 4.0.0-0
     cpu: 4
     memory: 12G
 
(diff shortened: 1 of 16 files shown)


rsmmr (dismissed):
We sometimes set LTS=1 for 4.0 and sometimes not. Is that on purpose?

Looks like it's the static build that remains at 3.0 to keep that coverage, right? Suggest to leave a comment in there saying so.

bbannier:
> We sometimes set LTS=1 for 4.0 and sometimes not. Is that on purpose?

The way the Dockerfiles are written, the Docker build variable `ZEEK_LTS` has an effect only if it contains a value. Leaving it defined but unset is equivalent to choosing a non-LTS release.

Since today the latest Zeek release is also an LTS release the different settings for `ZEEK_LTS` for zeek-4.0.0 have no effect, but when e.g., zeek-4.1.0 gets released we'll bump the versions only for non-LTS setups.

> Looks like it's the static build that remains at 3.0 to keep that coverage, right? Suggest to leave a comment in there saying so.

✅

PR #846, opened 2021-03-22
Add `!in` to HILTI and Spicy parser.

### LINKED ISSUE #450: provide the  not in  operator
`!(a in set)` is the current supported syntax. But: `a not in set` both reads more intuitively and avoids the not being applied across more than was intended.

spicy operator precedence does a pretty good job so far of not applying that leading ! to an expression, as long as there is no operator which can be on the right of the parentheses and bind tighter than ! does. But ! is only third level and spicy attributes such as &size bind quite tightly.

Over in #zq on brim Slack, on 14-Jul Phil Rzewski expressed: There's not a not in operator. You were on the right path with !(a in set) (or you could have done not (a in set)). We can think about making a change there as well. In prior discussions we've often shied away from making multiple ways to achieve the same thing the language, and not is already there to invert other logic. However, it doesn't seem to be helping here since it exposed the other usability problems with the left to right eval, such that you'd have to be double-wrapping to (not (a in set)) in order to put it as the first search term, which is, indeed, pretty ugly compared to a not in set.

bbannier:
Closes #450.

### DIFF
--- a/spicy/toolchain/src/compiler/parser/parser.yy
+++ b/spicy/toolchain/src/compiler/parser/parser.yy
@@ -29,7 +29,7 @@ namespace spicy { namespace detail { class Parser; } }
 %verbose
 
 %glr-parser
-%expect 121
+%expect 131
 %expect-rr 157
 
 %union {}
@@ -182,6 +182,7 @@ static int _field_width = 0;
 %token         NETWORK
 %token         NEW
 %token         NONE
+%token         NOT_IN
 %token         OBJECT
 %token         ON
 %token         OPTIONAL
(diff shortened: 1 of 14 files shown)


spicy/toolchain/src/compiler/parser/parser.yy
```diff
@@ -29,7 +29,7 @@ namespace spicy { namespace detail { class Parser; } }
 %verbose
 
 %glr-parser
-%expect 121
+%expect 131
```
rsmmr, line 32: (Weird that's it's jumping like that for such a little change.)

rsmmr (commented):
The code change looks good, but I'm wondering about the docs part. You were arguing that we shouldn't need to touch every type that implements `in` to also support `!in`. That's right for the code, but now every doc string for `in` needs to mention `!in`. How about instead changing spicy-doc to create an entry for `!in` automatically whenever it sees an `in`, with a generic description like along the lines of "returns the inverted result of `in`". Alternatively, do that in `spicy-doc-to-rst`: just create a second entry for `!in` along with each `in`. Does that make sense?

bbannier:
> Alternatively, do that in `spicy-doc-to-rst`: just create a second entry for `!in` along with each `in`.

I did that in the latest update. Since `in` can be overloaded we might generate multiple doc entries for `!in` as well (this is expected and correct), so I went with the following docstring:

```
Performs the inverse of the corresponding ``in`` operation.
```

rsmmr (approved):

PR #863, opened 2021-03-25
Prepare for master->main branch renaming.

rsmmr:


.github/workflows/docker-latest.yml
```diff
@@ -4,13 +4,13 @@ name: Publish Docker
...
-      # Since this action pushes a `latest` image it is only active for the `master` branch.
-      - master
+      # Since this action pushes a `latest` image it is only active for the `main` branch.
+      - main
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@main
```
bbannier, line 13: This should stay as is as this is their repo.

.github/workflows/docker-tags.yml
```diff
@@ -4,15 +4,15 @@ name: Publish Docker
...
+      - main
     tags:
       # Since this action pushes just a tagged image it is only active on tags.
       - v*
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@main
```
bbannier, line 15: This should stay as is as this is their repo.

scripts/autogen-version
```diff
@@ -81,7 +81,7 @@ dirty=$(echo "${git_version}" | awk -F - '{print $4}')
 test -n "${commit}" -a -n "${dirty}" && commit="${commit}.${dirty}"
 
 case "${branch}" in
-    master)
+    main|master)
```
bbannier, line 84: We should be able to get rid of `master` here altogether, right?
```suggestion
    main)
```

bbannier (changes requested):

scripts/autogen-version
rsmmr: I left it so that it would continue to work with existing checkouts of master. I believe that if somebody just does `git pull` on that, it'll start to pull from main, but locally keep looking like its still `master` (this is just guessing though).

.github/workflows/docker-latest.yml
rsmmr: Oops, of course.

scripts/autogen-version
bbannier: Hmm, existing checkouts of `master` should have a version of this script which just looks for `master`, so I think they should be fine if we remove `master`, unless they plan to copy this file out-of-tree.

rsmmr: GitHub will redirect master pulls to main, so I believe what will happen is that people will stay on the master branch locally, but still get the new files from main (unless they explicitly switch to main of course).

bbannier: Makes sense.

bbannier (approved):

PR #866, opened 2021-03-26
Fix doc links to package artifacts.

bbannier:
We recently renamed packages from e.g., `spicy.deb` to `spicy-dev.deb`
which wasn't reflected in the docs.

Additionally, due to the CI setup `.tar.gz` artifacts live in
`packages/build` while `.deb` and `.rpm` artifacts live in `packages/`;
we fix the links for that as well.


bbannier:
I validated this manually by commenting out link to Cirrus CI artifacts in `linkcheck_ignore` in `doc/conf.py`. Before this patch this reported the broken links, but now passes.

rsmmr (approved):

PR #867, opened 2021-03-26
Fix access to map elements of composite type.

bbannier:


hilti/runtime/include/types/map.h
```diff
@@ -245,14 +245,37 @@ class Map : protected std::map<K, V> {
      * @return the value
      * @throw `IndexError` if `k` is not set in the map
      */
-    const V& get(const K& k) const {
+    const V& get(const K& k) const& {
```
rsmmr, line 200: Oh, I don't know this, what's the difference when  overloading by `&`?

hilti/runtime/include/types/map.h
```diff
@@ -305,6 +246,13 @@ class Map : protected std::map<K, V> {
      */
     auto operator[](const K& k) && { return this->get(k); }
 
+    auto index_assign(const K& key, V value) {
```
rsmmr, line 249: result can be `void`

hilti/runtime/src/tests/map.cc
```diff
@@ -33,40 +33,12 @@ TEST_CASE("subscript") {
...
-        (void)p;
-        CHECK(! m.contains(2));
-
-        // Proxy objects stringify to the referenced value if it exists, and throw otherwise.
-        CHECK_EQ(to_string(m[1]), "11");
-        CHECK_THROWS_WITH_AS(to_string(m[2]), "key is unset", const IndexError&);
-
-        CHECK_EQ(fmt("%s", m[1]), "11");
-
-        // Proxy objects only invalidate iterators if an element was actually inserted.
```
rsmmr, line 54: Should we keep this? Looks like we still have the same iterator invalidation semantics?

hilti/toolchain/src/compiler/visitors/operator-resolver.cc
```diff
@@ -152,6 +153,17 @@ struct Normalizer : public hilti::visitor::PostOrder<void, Normalizer> {
...
         }
     }
+
+    void operator()(const expression::Assign& assign, position_t p) {
+        auto& lhs = assign.childs().front();
+        if ( const auto& index_non_const = lhs.tryAs<operator_::map::IndexNonConst>() ) {
+            const auto map = index_non_const->op0();
+            const auto key = index_non_const->op1();
+
+            auto index_assign = builder::memberCall(std::move(map), "index_assign", {std::move(key), assign.source()});
```
rsmmr, line 163: I think I'd prefer to transform this into a new operator `IndexAssign` that's overloaded by `map`, instead of a `map` method. That would be more natural to extend to other types needing a custom index-assign operation (e.g., vector). But this is also related to the question if need the explicit method `index assign`, see below.

tests/Baseline/spicy.types.map.ops/output
```diff
@@ -1,3 +1,6 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
 {1: b"AAA", 2: b"BBB", 3: b"CCC"}
-b"BBB"
+BBB
```
rsmmr, line 3: Nice, that looks like the better output actually.

rsmmr (commented):
One conceptual question:

> We still support updating existing elements through through the new
index & assign operation.

Does that mean we can't update existing elements through the index operator? Why's that?And if so, I suppose that's reason why we need the `index_assign` method that some of the tests are now using?

hilti/runtime/include/types/map.h
bbannier: This is an overload just for `const` lvalues as we wouldn't want to give out a ref to a value in a temporary map.

bbannier:
> One conceptual question:
> 
> > We still support updating existing elements through through the new
> > index & assign operation.
> 
> Does that mean we can't update existing elements through the index operator? Why's that?And if so, I suppose that's reason why we need the `index_assign` method that some of the tests are now using?

No, we still support updating existing elements.

hilti/toolchain/src/compiler/visitors/operator-resolver.cc
```diff
@@ -152,6 +154,36 @@ struct Normalizer : public hilti::visitor::PostOrder<void, Normalizer> {
...
+        auto& lhs = assign.childs().front();
+        if ( const auto& index_non_const = lhs.tryAs<operator_::map::IndexNonConst>() ) {
+            const auto map = index_non_const->op0();
+
+            const auto map_type = map.type().as<type::Map>();
+            const auto key_type = map_type.keyType();
+            auto value_type = map_type.elementType();
+
+            auto key = index_non_const->op1();
+            if ( key.type() != key_type ) {
```
rsmmr, line 169: Actually there's a better way than what I had said earlier: if you look into `c.nexpr` instead of `c.coerced`, you can skip the `if` check if the types match (because `coerceExpression` checks that already; I hadn't thought about that).

hilti/toolchain/src/compiler/visitors/operator-resolver.cc
```diff
@@ -152,6 +154,36 @@ struct Normalizer : public hilti::visitor::PostOrder<void, Normalizer> {
...
+                if ( auto c = hilti::coerceExpression(key, key_type); c.coerced )
+                    key = *c.coerced;
+            }
+
+            auto value = assign.source();
+            if ( value.type() != value_type ) {
+                if ( auto c = hilti::coerceExpression(value, value_type); c.coerced )
```
rsmmr, line 176: same here.

rsmmr (approved):
Coo, this is nice, good to see that it works as advertised. :-)

hilti/toolchain/src/compiler/visitors/operator-resolver.cc
bbannier: Ah, that's convenient.

PR #869, opened 2021-03-30
Fix regexp matching of multi-chunk inputs.

bbannier:


doc/autogen/types/sink.rst
```diff
@@ -50,8 +50,8 @@
 
 .. spicy:method:: sink::gap sink gap False void (seq: uint<64>, len: uint<64>)
 
-    Reports a gap in the input stream. *seq* is the sequence number of the
```
rsmmr, line 53: Have you checked what style is more common across all operators? I thought I had been meaning to use `*XX*` for  parameters, though obviously not consistently. Either way, if we make these changes, should ensure it's what we do elsewhere too.

hilti/runtime/src/tests/regexp.cc
```diff
@@ -12,10 +13,17 @@
 using namespace hilti::rt;
 using namespace hilti::rt::bytes::literals;
 
+namespace hilti::rt::detail::adl {
+template<typename A, typename B>
+std::string to_string(const std::tuple<A, B>& xs, tag /*unused*/) {
```
rsmmr, line 18: Hmmm .. I'm surprised we need this, tuples should have a to_string() already. Also should output them as `(..)`.

doc/autogen/types/sink.rst
rsmmr: note: the changes below from single-tick to double-tickets are ok, they are just typos accidentally switching to markdown.

rsmmr (dismissed):
Nice find. This looks good, except for the question on markup consistency.

doc/autogen/types/sink.rst
bbannier: I missed that, `*par*` seems indeed like the preferred way to refer to parameters. I rolled back the changes on that here.

PR #874, opened 2021-04-08
Add support for `%skip` properties

### LINKED ISSUE #26: Add pre/post skipping
- Spicy
    - Unit properties
        - [ ] %skip-pre, %skip-post

  Old tests in `*.spicy` in https://github.com/rsmmr/hilti/blob/master/tests/spicy/unit/skip-property-disable.spicy and https://github.com/rsmmr/hilti/blob/master/tests/spicy/unit/skip-property.spicy

bbannier:
Closes #26.


tests/spicy/types/unit/skip.spicy
```diff
@@ -0,0 +1,38 @@
...
+
+# @TEST-EXEC: echo "Parsing X w/ content like skip-post" >> output
+# @TEST-EXEC: printf '01' | spicy-driver -p Test::X test.hlto >> output
+# @TEST-EXEC: printf '0011' | spicy-driver -p Test::X test.hlto >> output
+
+# @TEST-EXEC: echo "Parsing X w/o any skipping since skip-pre and -post in the wrong place" >> output
+# @TEST-EXEC: printf '1AB0' | spicy-driver -p Test::X test.hlto >> output
+# @TEST-EXEC: echo "Parsing Y w/o any skipping since skip-pre and -post in the wrong place" >> output
+# FIXME(bbannier): This is currently broken.
+# @_TEST-EXEC: printf '1AB0' | spicy-driver -p Test::Y test.hlto >> output
```
bbannier, line 17: Support for this is currently broken since we extract `X` values `['1', 'A', 'B']`. After that we still haven't reached EOD so the parser for `Y` kicks off parsing of another `X` value; that parse then `%skip-pre` drops the `0` and then has no more data to extract `&size=1` bytes.

In order for this to work we would either need to communicate some stop information from the parser of `X::x` up into the parser of `Y::xs`, or alternatively drop support for `%skip-pre` and `%skip-post` and let it all be handled by `%skip` (this would work since we could extract the `0` in a post step and then then detect EOD as usual).

spicy/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -217,6 +217,33 @@ struct PreTransformVisitor : public hilti::visitor::PreOrder<void, PreTransformV
...
+            ; // Nothing; handled in validator for `hilti::Module`.
+
+        else if ( i.id().str() == "%byte-order" ) {
+            if ( auto e = i.expression(); ! e ) {
+                error("%byte-order requires an argument", p);
+                return;
+            }
+        }
+
+        else if ( const auto& prop = i.id().str(); prop == "%skip" || prop == "%skip-post" || prop == "%skip-pre" ) {
```
bbannier, line 231: This still needs to be documented.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -974,9 +1034,21 @@ struct ProductionVisitor
...
+        if ( const auto& skipPre = anyProperty(p.unitType(), {"%skip-pre", "%skip"}) ) {
+            auto ctor = skipPre->expression()->as<hilti::expression::Ctor>().ctor();
+            if ( auto c = ctor.tryAs<hilti::ctor::RegExp>() )
+                skipRegExp(*c);
+        }
+
         for ( const auto& i : p.fields() )
             parseProduction(i);
 
+        if ( const auto& skipPost = anyProperty(p.unitType(), {"%skip-post", "%skip"}) ) {
```
bbannier, line 1045: This still needs to be documented.

tests/spicy/types/unit/skip.spicy
rsmmr: As we chatted about, I think these semantics are ok, we just need to document them.

tests/spicy/types/unit/skip.spicy
```diff
@@ -0,0 +1,38 @@
...
+
+# @TEST-EXEC: echo "Parsing Y w/ skipping pre and post" >> output
+# @TEST-EXEC: printf '0AB1' | spicy-driver -p Test::Y test.hlto >> output
+
```
rsmmr, line 21: Please add a test with incremental input as well (`-i 1`)

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -974,9 +1033,21 @@ struct ProductionVisitor
...
```
rsmmr, line 1045: I think we have a misunderstanding about `%skip` here: it should't skip only at the beginning/end, but also in between fields. I suggest this logic:

```
if ( have %pre-skip )
  skip regex;

if ( have %skip )
  skip regex;

for ( const auto& i : p.fields() ) {
     parseProduction(i);

     if ( have %skip )
         skip regex;
}


if ( have %post-skip )
  skip regex;

```

\+ maybe some other way to resolve the case where we have both `%skip` *and*  `%pre/post-skip.`

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -61,6 +61,17 @@ void ParserState::printDebug(const std::shared_ptr<builder::Builder>& builder) c
 
 namespace spicy::detail::codegen {
 
+// Helper function to get any of the specified unit properties.
+std::optional<type::unit::item::Property> anyProperty(const type::Unit& unit,
```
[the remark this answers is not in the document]
rsmmr: (but see other comment about `%skip` semantics below)

spicy/toolchain/src/compiler/codegen/parser-builder.cc
rsmmr: Another thing that's missing: disabling any of the %skip properties by assigning `Null`, see ticket and old tests.

tests/spicy/types/unit/skip.spicy
rsmmr: Then also add tests for `%skip` between fields, and disabling skipping through `Null`.

rsmmr (changes requested):

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -726,6 +737,54 @@ struct ProductionVisitor
...
+        // No match found, leave `cur` unchanged.
+        auto no_match = switch_.addCase(builder::integer(0));
+        pushBuilder(no_match);
+        builder()->addBreak();
+        popBuilder();
+
+        // Match found, update `cur`.
+        auto match = switch_.addDefault();
+        pushBuilder(match);
+        builder()->addAssign(state().cur, builder::id("ncur"));
```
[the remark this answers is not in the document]
bbannier: For some reason the conditional in `advanceInput` doesn't detect the temporary `ncur` as a `stream::View` even though using it here triggers an error message during codegen which shows that it has the correct type (`cannot resolve operator: <view<stream>>.advance(<view<stream>>)`), so I stayed with the current, manual update. I did add an explicit trim, though.

doc/programming/language/modules.rst
```diff
@@ -79,10 +79,17 @@ A module may define the following global properties:
...
+        `<expr>`, where ``ORDER`` must be of type is type
         :ref:`spicy_ByteOrder`.
 
     ``%spicy-version = "VERSION";``
         Specifies that the module requires a given minimum version of
         Spicy, where ``VERSION`` must be a string of the form ``X.Y``
         or ``X.Y.Z``.
+
+    ``%skip = ( REGEXP | Null );``
+        Specifies a pattern which should be skipped when encountered in the input stream in between parsing of unit fields.
```
rsmmr, line 91:
```suggestion
        Specifies a pattern which should be skipped when encountered in the input stream in between parsing of unit fields (incl. before/after the first/last field).
```

rsmmr: We can leave out the `Null` here: at the module-level that's not going to be useful (it's just restating the default).

doc/programming/parsing.rst
```diff
@@ -825,6 +825,16 @@ properties:
     make use of the information to decide which unit type to use for
     parsing a connection's payload.
 
+``%skip = ( REGEXP | Null );``
+    Specifies a pattern which should be skipped when encountered in the input stream in between parsing of unit fields.
+    This overwrites a value set at the module level; use ``Null`` to reset the property.
```
rsmmr, line 830:
```suggestion
    This overwrites a value set at the module level; use ``Null`` to reset the property (i.e., not skip anything).
```

Suggest same change below.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -726,6 +726,53 @@ struct ProductionVisitor
     // Returns a boolean expression that's 'true' if a 'stop' was encountered.
     Expression parseProduction(const Production& p) { return _parseProduction(p, p.meta()); }
 
+    void skipRegExp(const hilti::ctor::RegExp& c) {
+        auto re = ID("__re");
+        int i = 0;
+        while ( pb->cg()->haveAddedDeclaration(re) )
+            re = ID(fmt("__re_%" PRId64, ++i));
```
rsmmr, line 733:
```suggestion
            re = ID(fmt("__re_%" PRId64, ++i));
            
```

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -974,9 +1021,34 @@ struct ProductionVisitor
             pushState(std::move(pstate));
         }
 
-        for ( const auto& i : p.fields() )
+        if ( const auto& skipPre = p.unitType().propertyItem("%skip-pre") ) {
+            auto ctor = skipPre->expression()->as<hilti::expression::Ctor>().ctor();
+            if ( auto c = ctor.tryAs<hilti::ctor::RegExp>() )
+                skipRegExp(*c);
+        }
```
rsmmr, line 1028: Could compress this (and below) if `skipRegExp` took an `Expression` and did the cast and null check there.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
rsmmr, line 733: Hmm ... Feels like we should have a codegen method here that (1) takes care of creating a unique ID, and (2) checks if the same constant already exists and reuses that instead to avoid duplication. it's ok to leave as is for now, but add comment at least to that regard I'd say.

rsmmr (approved):

spicy/toolchain/src/compiler/codegen/parser-builder.cc
bbannier: What is the change here? These lines seem identical to me.

rsmmr: Insert an empty line after the loop body. :-)

PR #876, opened 2021-04-09
Update docs for v1.0.0

bbannier:


README.md
```diff
@@ -16,10 +16,9 @@
 
 <td>
 <table>
-<tr><th></th>                 <th>Release</th> <th>Development</th> </tr>
-<tr><td> Documentation </td>  <td> (coming soon) </td> <td> <a href="https://docs.zeek.org/projects/spicy">Spicy Manual</a> </td></tr>
-<tr><td> Changelog </td>      <td> (coming soon) </td> <td> <a href="/CHANGES">CHANGES</a> </td></tr>
-<tr><td> Build status </td>   <td> (coming soon) </td> <td> <a href="https://cirrus-ci.com/github/zeek/spicy/main"><img src="https://api.cirrus-ci.com/github/zeek/spicy.svg" alt="Build status"> </a> </td></tr>
+<tr><td>Documentation</td><td><a href="https://docs.zeek.org/projects/spicy">Spicy Manual</a> </td></tr>
+<tr><td>Changelog</td>    <td><a href="/CHANGES">CHANGES</a> </td></tr>
+<tr><td>Build status</td> <td><a href="https://cirrus-ci.com/github/zeek/spicy/main"><img src="https://api.cirrus-ci.com/github/zeek/spicy.svg" alt="Build status"> </a> </td></tr>
```
rsmmr, line 21: I was thinking we'd leave both columns in here, so that people see the information for both current release and dev. Except I guess for the docs, where we don't have two versions, can just span that across the two columns.

.github/workflows/docker-tags.yml
```diff
@@ -38,4 +38,6 @@ jobs:
         file: ci/Dockerfile.dockerhub
         push: true
         tags: |
+          # TODO(bbannier): Automatically detect whether this is the latest
+          # release and in that case also push this to `spicy:latest`.
```
rsmmr, line 42: Can we always push to latest: the most recently pushed release will always be the latest, won't it?

rsmmr (dismissed):

.github/workflows/docker-tags.yml
bbannier: Right now that is true, but once we have two releases out and do patch releases. We would need to compare the commit of the current tag against `release` or `beta` or something similar.

I don't think this is too big an issue for now, and we have it documented in our release checklist.

README.md
bbannier: I cannot find a documented or undocumented way to get the Cirrus CI build status svg for a tag (branch would be possible, but that's not the same as the release), https://cirrus-ci.org/guide/writing-tasks/#embedded-badges.

I could add some text for a tag's status here (e.g., `Cirrus CI status`). Like I mentioned offline, I am not sure this is too useful as tags are immutable.

rsmmr: Let's use the release branch for the status then? Not exactly the same but close enough I'd say. 

The alternative is to do get rid of this table altogether and replace it with something else showing the "Docs / CHANGES / build status" information.

(Not a release blocker either way of course)

rsmmr: So this would be my suggestion 
```
<td>
<table>
<tr><th></th>              <th>Release</th> <th>Development</th> </tr>
<tr><td>Documentation</td> <td colspan="2" align="center"><a href="https://docs.zeek.org/projects/spicy">Spicy Manual</a></td></tr>
<tr><td>Changelog</td>     <td><a href="https://github.com/zeek/spicy/blob/v1.0.0/CHANGES">CHANGES</td> <td><a href="/CHANGES">CHANGES</a></td></tr>
<tr>
  <td>Build status</td>
  <td><a href="https://cirrus-ci.com/github/zeek/spicy/release"><img src="https://api.cirrus-ci.com/github/zeek/spicy.svg?branch=release" alt="Build status release"></a></td> 
  <td><a href="https://cirrus-ci.com/github/zeek/spicy/main">   <img src="https://api.cirrus-ci.com/github/zeek/spicy.svg" alt="Build status development"></a></td>
</tr>
</table>
</td>
```

rsmmr (dismissed):

PR #881, opened 2021-04-13
Fix integer to enum conversion.

rsmmr:


doc/autogen/types/integer.rst
```diff
@@ -12,10 +12,17 @@
...
+    Converts the value into an enum instance. The value does *not* need to
+    correspond to any of the target type's enumerator labels.
+
 .. spicy:operator:: integer::Cast enum cast<enum-type>(uint)
 
     Converts the value into an enum instance. The value does *not* need to
-    correspond to any of the target type's enumerator labels
+    correspond to any of the target type's enumerator labels. It must not
+    be larger that the maximum that a *signed* integer value can
+    represent.
```
duffy-ocraven, line 25: *signed* integer value of the same length
would be a better phrase. But is this constraint really going in?!? That an enum cast from uint8 can't have value 130.

duffy-ocraven (commented):
Sorry, I meant that as a comment, not a review, but clicked on the brightly colored button in the dialog

hilti/runtime/include/types/enum.h
```diff
@@ -29,5 +31,38 @@ bool has_label(const T& t, const TypeInfo* ti) {
...
+ * Converts a signed integer value into an enum value. The value does
+ * not need to correspond to a valid label. (Internally, this is a
+ * straight-forward cast.)
+ *
+ * @param t numerical value to convert
+ * @tparam T enum type
+ */
+template<typename T>
+T from_int(int64_t n) {
+    return static_cast<T>(n);
```
bbannier, line 45: Could we add a check whether the [underlying type of the enum](https://en.cppreference.com/w/cpp/types/underlying_type) can represent the value? We probably should also `static_assert` that `T` [is an enum](https://en.cppreference.com/w/cpp/types/is_enum).

hilti/runtime/include/types/enum.h
```diff
@@ -29,5 +31,38 @@ bool has_label(const T& t, const TypeInfo* ti) {
...
+ * larger than the maximum possible signed int64 value. (Internally,
+ * this is mostly a straight-forward cast, we just add the range
+ * check.)
+ *
+ * @param t numerical value to convert
+ * @tparam T enum type
+ * @throws InvalidValue if value exceeds range
+ */
+template<typename T>
+T from_uint(uint64_t n) {
```
bbannier, line 60: Let's add a range check of the underlying type and assert that `T` is an enum here as well.

bbannier (approved):

hilti/toolchain/include/ast/operators/enum.h
```diff
@@ -24,17 +24,30 @@ STANDARD_OPERATOR_2x(
...
 
+BEGIN_CTOR(enum_, CtorUnsigned)
+    auto ctorType() const { return type::Enum(type::Wildcard()); }
+
+    auto signature() const {
+        return Signature{.args = {{.id = "value", .type = type::UnsignedInteger(type::Wildcard())}}, .doc = R"(
+Instantiates an enum instance initialized from an unsigned integer
+value. The value does *not* need to correspond to any of the type's
+enumerator labels. It must not be larger that the maximum that a
+*signed* 64-bit integer value can represent.
```
duffy-ocraven, line 46: This has a typo, needs to fix: than for that
Oh, and this clarifies that the limitation is not at the same size, so nevermind about my concern regarding 130 in a unit8. But this:
the maximum that a *signed* 64-bit integer value can represent
is a much more explanatory phrasing than the existing proposed phrasing in the .rst

doc/autogen/types/integer.rst
rsmmr: Oh, there's an import bit missing, it's supposed to be: "It must not be larger that the maximum that a *signed* 64-bit integer value can represent." So that shouldn't really be much of a limitation at all.

hilti/runtime/include/types/enum.h
rsmmr: Our enums are always 64-bit signed underneath. Are you thinking in case this got used with something else that codegen'd enums?

hilti/toolchain/include/ast/operators/enum.h
rsmmr: Ack, will fix.

hilti/runtime/include/types/enum.h
bbannier: I think this would be a good sanity check to add while we are at it for future-proofing. I wouldn't expect this to produce an error right now.

rsmmr: ok, will do

rsmmr:
> Still has the typo: than for that

Autogenerated docs weren't recreated, fixed now, thanks.

hilti/runtime/src/tests/enum.cc
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2020-2021 by the Zeek Project. See LICENSE for details.
+
+#include <hilti/rt/doctest.h>
+#include <hilti/rt/types/enum.h>
+
+#include "exception.h"
```
bbannier, line 6: Let's include <hilti/rt/exception.h>` instead.

bbannier (approved):

PR #883, opened 2021-04-14
Add support for &size attribute for unit switch statement.

### LINKED ISSUE #844: Support &size with switch
Would be useful to be able to add `&size` to a switch so that it applies to all cases:

```
switch ( x ) {
        0 -> a: ..;
        1 -> b: ..;
    } &size=5;
```

Right now one can work-around the lack of this by adding the same `&size` to each case. The OpenVPN analyzer does this, but it's cumbersome.

Should check if there are other attributes that would make sense to support in this way.

bbannier:
Closes #844.


doc/programming/examples/_parse-switch-size.spicy
```diff
@@ -0,0 +1,12 @@
+# Automatically generated; edit in Sphinx source code, not here.
+module Test;
+
+public type Foo = unit {
+    tag: uint8;
+    switch ( self.tag ) {
+       1 -> b1: bytes &eod;
+       2 -> b2: bytes &eod &convert=$$.lower();
+    };
```
rsmmr, line 9: `&size` missing here?

doc/programming/parsing.rst
```diff
@@ -1404,6 +1404,31 @@ once you have subunits that are recognizable by how they start:
     :exec: printf 'A ' | spicy-driver %INPUT; printf '\377\377' | spicy-driver %INPUT
     :show-with: foo.spicy
 
+.. rubric:: Switching Over Fields With Common Size
+
+You can limit the input any field in a unit switch receives by attaching an
+optional ``&size=EXPR`` attribute that specifies the number of raw bytes to
+make available. This is analog to the `field size attribute <attribute_size>`_
+and especially useful to remove duplication.
```
rsmmr, line 1412:
```suggestion
and especially useful to remove duplication when each case is subject to the same constraint.
```

doc/programming/parsing.rst
```diff
@@ -1404,6 +1404,31 @@ once you have subunits that are recognizable by how they start:
...
+.. spicy-code:: parse-switch-size.spicy
+
+    module Test;
+
+    public type Foo = unit {
+        tag: uint8;
+        switch ( self.tag ) {
+           1 -> b1: bytes &eod;
+           2 -> b2: bytes &eod &convert=$$.lower();
+        };
```
rsmmr, line 1423: Same as above.

spicy/toolchain/include/ast/types/unit-items/switch.h
```diff
@@ -60,10 +60,12 @@ inline Node to_node(Case c) { return Node(std::move(c)); }
 class Switch : public hilti::NodeBase, public spicy::trait::isUnitItem {
 public:
     Switch(std::optional<Expression> expr, const std::vector<switch_::Case>& cases, Engine e,
-           std::optional<Expression> cond, std::vector<Hook> hooks, Meta m = Meta())
-        : NodeBase(nodes(std::move(expr), std::move(cond), cases, std::move(hooks)), std::move(m)),
+           std::optional<Expression> cond, std::vector<Hook> hooks, std::optional<AttributeSet> attributes,
```
rsmmr, line 63: I'm always torn if it should be `optional<AttributeSet>` or `AttributeSet`, with the latter just remaining empty if there are no attributes. We are currently inconsistent across AST nodes on this. Advantage of the latter is that access gets easier: one doesn't need to check first if there's actually a set available. Advantage of the former is that it avoids creating lots of empty `AttributeSet` instances (less of a problem in this case because there aren't many switches, but more for other node types of which we create plenty). 

It's ok to leave as is, just wanted to mention.

tests/spicy/types/unit/switch.spicy
```diff
@@ -4,6 +4,8 @@
 # @TEST-EXEC: echo 3234567890 | spicy-driver %INPUT.hlto >>output
 # @TEST-EXEC: echo 4234567890 | spicy-driver %INPUT.hlto >>output
 # @TEST-EXEC: btest-diff output
+#
+# @TEST-DOC: Checks unit switch statement.
```
rsmmr, line 8: Want to add a test case for the error case where not all of size is being consumed?

rsmmr (approved):

tests/spicy/types/unit/switch.spicy
bbannier: Did that and found an actual bug. 🤦

bbannier:
Thanks for the review @rsmmr. Could you make another pass? In order to properly detect cases where not the full `&size` amount is consumed, I needed to make another slight adjustment to `parser-builder.cc`.

rsmmr (approved):
Yep, that should work.

Issue #882, opened 2021-04-14
CI: Add OpenSUSE Leap 15.2


bbannier:
Zeek recently added opensuse-15.2 to their list of supported platforms, see e.g., zeek/zeek#1482. We should add it to our CI as well.

0xxon:
It actually turns out to not be this bad. This needs #885 merged, for bison-3.0 compatibility. Afterwards, you just need to install gcc9 and gcc9-c++, in addition to the typical packages you need for Zeek (see Zeek ci dockerfile).

I then was able to get it compiled with all tests passing issuing:
```
CC=/usr/bin/gcc-9 CXX=/usr/bin/g++-9 ./configure --prefix=$HOME/spicy/install --enable-debug --disable-broker-tests --generator=Ninja --disable-zkg --disable-zeekctl --disable-python --disable-archiver 
./configure --prefix=$HOME/spicy/install --with-c-compiler=/usr/bin/gcc-9 --with-cxx-compiler=/usr/bin/g++-9 --with-zeek=$HOME/spicy/install --enable-debug --generator=Ninja --build-zeek-plugin=yes
```

When you create tests for this, note that the Zeek packets for OpenSUSE Leap 15.2 are currently created with the default gcc (so 7.5.0).  I did not test if there are any compatibility problems with using a different compiler to compile Zeek vs spicy.

PR #885, opened 2021-04-14
Make spicy compile with bison 3.0.

0xxon:
This also allows compiling under OpenSUSE Leap 15.2, which only ships with bison 3.0.4.

Relates to GH-882

### DIFF
--- a/hilti/toolchain/src/compiler/parser/parser.yy
+++ b/hilti/toolchain/src/compiler/parser/parser.yy
@@ -1,7 +1,11 @@
 /* Copyright (c) 2020-2021 by the Zeek Project. See LICENSE for details. */
 
+/* This grammar is written against the bison-3.3 API. If an older Bison version
+ * was detected we perform preprocessing to support versions down to at least
+ * bison-3.0, see the CMake macro `BISON_TARGET_PP`. */
+%require "3.3"
+
 %skeleton "lalr1.cc"                          /*  -*- C++ -*- */
-%require "3.4"
 %defines
 
 %{
--- a/spicy/toolchain/src/compiler/parser/parser.yy
+++ b/spicy/toolchain/src/compiler/parser/parser.yy
@@ -1,7 +1,11 @@
 /* Copyright (c) 2020-2021 by the Zeek Project. See LICENSE for details. */
 
+/* This grammar is written against the bison-3.3 API. If an older Bison version
+ * was detected we perform preprocessing to support versions down to at least
+ * bison-3.0, see the CMake macro `BISON_TARGET_PP`. */
+%require "3.3"
+
 %skeleton "lalr1.cc"                          /*  -*- C++ -*- */
-%require "3.4"
 %defines
 
 %{
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@ execute_process(COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/scripts/autogen-version --cm
 project(spicy VERSION "${SPICY_VERSION}" LANGUAGES ASM C CXX)
 
 set(flex_minimum_version "2.6")
-set(bison_minimum_version "3.4")
+set(bison_minimum_version "3.0")
 set(python_minimum_version "2.4")
 set(macos_minimum_version "19.0.0") # macOS 10.15.0 (Catalina)
 
--- a/ci/Dockerfile.packages
+++ b/ci/Dockerfile.packages
@@ -13,14 +13,11 @@ SHELL [ "/usr/bin/scl", "enable", "gcc-toolset-9"]
 # Use ld.gold instead.
 RUN alternatives --set ld /usr/bin/ld.gold
 
-# Need a more recent Bison than available.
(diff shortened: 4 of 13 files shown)


hilti/toolchain/src/compiler/parser/parser.yy
```diff
@@ -21,7 +21,7 @@ namespace hilti { namespace detail { class Parser; } }
 %lex-param   {class Driver* driver}
 
 %define api.namespace {hilti::detail::parser}
-%define api.parser.class {Parser}
+%define parser_class_name {Parser} /* new name after switching to bison 3.3 -> api.parser.class */
```
bbannier, line 24: A newer Bison will emit almost the same information,
```
/spicy/hilti/toolchain/src/compiler/parser/parser.yy:24.1-34: warning: deprecated directive: '%define parser_class_name {Parser}', use '%define api.parser.class {Parser}' [-Wdeprecated]
   24 | %define parser_class_name {Parser} /* new name after switching to bis...
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      | %define api.parser.class {Parser}
```

Let's reword the comment so we make it very clear from the deprecation warning that this is intentional, e.g.,
```suggestion
%define parser_class_name {Parser} /* Using deprecated directive for bison-3.0 compatibility. */
```

spicy/toolchain/src/compiler/parser/parser.yy
```diff
@@ -22,7 +22,7 @@ namespace spicy { namespace detail { class Parser; } }
 %lex-param   {class Driver* driver}
 
 %define api.namespace {spicy::detail::parser}
-%define api.parser.class {Parser}
+%define parser_class_name {Parser} /* new name after switching to bison 3.3 -> api.parser.class */
```
bbannier, line 25: Ditto.

bbannier (approved):
Thanks for the patch, having users install one less out-of-band package is great.

I tested this patch by switching our Docker images to the distro-default Bison version and that seems to work great, see #886.

hilti/toolchain/src/compiler/parser/parser.yy
rsmmr: Oh, that means we'd now always be getting a warning with new Bisons? That's not so great. I like #886  very much, but I don't like unnecessary warnings ...

bbannier: FTR, this is a non-fatal warning from Bison which does not break the build, even with `--enable-werror`. I think suppressing the warning would not be great either, since that could hide genuinely interesting warnings.

rsmmr: I'd like to keep the build generally free of warnings so would prefer some way of conditionally compiling this based on Bison version. It looks like there's no direct way to do that unfortunately, so we might need to do some manual replacement through `sed` or such prior to compilation. Not great, but better than always being warned IMO.

bbannier: @0xxon, I updated the build setup to adjust the sources if needed. I also appended the container update #886 to this PR. PTAL.

hilti/toolchain/src/compiler/parser/parser.yy
```diff
@@ -1,7 +1,7 @@
 /* Copyright (c) 2020-2021 by the Zeek Project. See LICENSE for details. */
 
 %skeleton "lalr1.cc"                          /*  -*- C++ -*- */
-%require "3.4"
+%require "3.3"
```
rsmmr, line 4: I'm wondering if we should set this to 3.0 as that's "mostly true", and avoids confusion by somebody looking just at this file. And we can add comment to the `api.parser.class` class line that it'll be replaced for < 3.3.

Alternatively, I'd at least a comment here saying that this 3.3 actually is a 3.0 through PP.

spicy/toolchain/src/compiler/parser/parser.yy
```diff
@@ -1,7 +1,7 @@
 /* Copyright (c) 2020-2021 by the Zeek Project. See LICENSE for details. */
 
 %skeleton "lalr1.cc"                          /*  -*- C++ -*- */
-%require "3.4"
+%require "3.3"
```
rsmmr, line 4: Same as above.

rsmmr (dismissed):

hilti/toolchain/src/compiler/parser/parser.yy
bbannier: I wanted to avoid the grammar being inconsistent (e.g., incompatibility between version requirement and used API), so the choice here was to either declare a 3.0 requirement and preprocess to use a 3.3 API if detected, or the conditonal downgrade done here. I went with the downgrade since I expect platforms to upgrade their Bison packages so that we can remove the downgrade support without code changes.

I'll go with a comment explaining that we might downgrade the file (here and in the other grammar),

```
/* This grammar is written against the bison-3.3 API. If an older Bison version
 * was detected we perform preprocessing to support versions down to at least
 * bison-3.0, see the CMake macro `BISON_TARGET_PP`. */
%require "3.3"
```

spicy/toolchain/src/compiler/parser/parser.yy
bbannier: See above.

PR #887, opened 2021-04-15
Port Spicy to Apple silicon.

### LINKED ISSUE #766: Build fails on Apple M1 (arm64)
```
$ ./configure --with-flex=/opt/homebrew/opt/flex --with-bison=/opt/homebrew/opt/bison && make
...

Scanning dependencies of target fiber
[  5%] Building ASM object 3rdparty/fiber/CMakeFiles/fiber.dir/src/fiber_asm_aarch64_apcs.S.o
/Users/jon/pro/zeek/spicy/3rdparty/fiber/src/fiber_asm_aarch64_apcs.S:58:1: error: unknown directive
.size fiber_asm_switch, .-fiber_asm_switch
^
/Users/jon/pro/zeek/spicy/3rdparty/fiber/src/fiber_asm_aarch64_apcs.S:68:1: error: unknown directive
.size fiber_asm_invoke, .-fiber_asm_invoke
^
/Users/jon/pro/zeek/spicy/3rdparty/fiber/src/fiber_asm_aarch64_apcs.S:79:1: error: unknown directive
.size fiber_asm_exec_on_stack, .-fiber_asm_exec_on_stack
^
make[3]: *** [3rdparty/fiber/CMakeFiles/fiber.dir/src/fiber_asm_aarch64_apcs.S.o] Error 1
```

rsmmr:
We won't have CI for a while, so this is best effort.

Closes #766.


hilti/runtime/src/fiber.cc
```diff
@@ -249,11 +249,11 @@ std::pair<char*, char*> detail::StackBuffer::activeRegion() const {
     // The direction in which the stack grows is platform-specific. It's
     // probably gong to be growing downwards pretty much everywhere, but to be
     // safe we whitelist platforms that we have confirmed to do so.
-#if __x86_64__
+#if __x86_64__ || __arm64__
```
bbannier, line 252: Looking at e.g., https://stackoverflow.com/a/664779/176922, it seems that ARM has no fixed stack growth direction. I think we should either detect the growth direction during the configure phase, or use a more restricted selection of platforms (e.g., if you know Apple M1 grows down, explicitly select for that here instead of `__arm64__`).

hilti/runtime/src/fiber.cc
```diff
@@ -268,7 +268,7 @@ size_t detail::StackBuffer::liveRemainingSize() const {
     assert(::fiber_is_executing(_fiber)); // must be live
 
     // Whitelist architectures where we know how to do this.
-#if __x86_64__
+#if __x86_64__ || __arm64__
```
bbannier, line 271: Ditto.

hilti/runtime/src/fiber.cc
rsmmr: Yeah, pretty sure it's that way here, but you're right on ARM in general. Need to figure out how to test for M1 specifically then.

rsmmr: Done

tests/spicy/types/real/coercion-fail.spicy
```diff
@@ -6,7 +6,7 @@ module Foo;
 global untrue: bool = 1.01;
 global falser: bool = 0.0;
 
-global u64max: real = 0xFFFFFFFFFFFFFFFF;
+# global u64max: real = 0xFFFFFFFFFFFFFFFF; # seems that this may work depending on platform
```
bbannier, line 9: Let's add a ticket to add validation to reject this (either on the Spicy or on the HILTI level).

bbannier (approved):

PR #888, opened 2021-04-16
Fix versioning for development.

rsmmr:
update-changes and autogen-version didn't agree on how to number
development versions.


bbannier (approved):
Like you mentioned offline, for this to work users need to delete any existing local `v1.1.0-dev` tag and refetch the updated upstream tag. With that this patch LGTM.

rsmmr:
Yeah, not much we can do about that unfortunately, but shouldn't affect many people in the end.

PR #891, opened 2021-04-19
Add support for skip fields.

### LINKED ISSUE #37: Add parsing support for skipping fields
This actually never got implemented for the prototype Spicy, but would be a nice feature to have:

```
module Mini;

export type test = unit {
       a: bytes &length=4;
        : skip &length=7;
       b: bytes &length=5;
        : skip &length=2;
       c: bytes &length=2;

       on %done {
         print self;
       }
};
```

However, we might want to call this something other than `skip` because that term already has two different meanings in unit parsing ...

That's from a non-functional test in https://github.com/rsmmr/hilti/blob/master/tests/spicy/unit/skip.spicy

bbannier:
Closes #37.


bbannier:
Could you have a look at this @rsmmr? This patch takes the easiest possible approach to implement the feature, but I would like to make sure I didn't miss a crucial aspect between the lines of the original issue.

rsmmr:
Using an anonymous field makes sense. I'm torn if reusing `bytes` is the way to go. The advantage is of course that we can reuse the existing machinery and attributes, but that's also the disadvantage: one thing `skip` allows to do is optimizing the parsing to simply skip over some data, vs. `bytes` parsing needing to at least temporarily store the value and make it available for hooks and other stuff (even if we don't allow this here, we'd either still generate that code or we'd need to special case the `bytes` parsing to not emit certain pieces when `&skip` is present). 

An alternative would be giving this field a type of `void` and then handling this separately in the parser generator, generating only minimal code. We'd then limit `skip` to a smaller set of attributes (just `&size` and `eod`?).

We could also treat this as an implementation detail and go with wrapping `bytes` first, and optimize later, maybe as part of https://github.com/zeek/spicy/wiki/Project:-Improve-efficiency-of-the-generated-C---parsing-code.

Yeah, I can see using that terser syntax; would take care of disallowing hooks as well.

tests/spicy/types/unit/skip-field.spicy
```diff
@@ -0,0 +1,17 @@
...
+module Test;
+
+public type X = unit {
+    : skip &size=2;
+    x: uint8;
+    : skip &size=0;
+};
+
+public type Y = unit {
+    : skip &size=1;
```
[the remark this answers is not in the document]
bbannier: Not sure we need to explicitly document the now final form of an anon `void` field with an explicit `&size` attribute.

spicy/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -335,6 +335,38 @@ struct PreTransformVisitor : public hilti::visitor::PreOrder<void, PreTransformV
...
+                // direct child). Below we instead check that a `$$` has a hook
+                // and a field as parent where the field has a `void` type
+                // which is not the same, but seems close enough for now.
+                //
+                // TODO(bbannier): Consider performing this test as part of the
+                // `Hook` validation. This would require a mechanism to filter
+                // all descendants of a node by type.
+
+                auto hook = p.findParent<spicy::Hook>();
+                auto field = p.findParent<type::unit::item::Field>();
```
rsmmr, line 352: This looks like it'll work only for inline hooks, not ones added externally. Given the test below, it wouldn't catch this one:

```
hook void Test::X() { print $$; }
```

This is tricky to solve because the external hook won't be tied to the Unit anymore (but to the HILTI-side struct that's been created from the unit), so there's no field. What we could do is check if the hook's parameter list includes a parameter called `__dd`, as that's what ``$$`` will be mapped to. Not exactly nice, but another artifact of the current resolver process where we mix HILTI and Spicy AST processing into the just a single loop. 

We could also just punt and skip the check, unlikely that users will ever run into this.

spicy/toolchain/src/compiler/codegen/parsers/types.cc
```diff
@@ -167,7 +167,12 @@ struct Visitor : public hilti::visitor::PreOrder<Expression, Visitor> {
         return performUnpack(destination(t), t, t.width() / 8, {state().cur, fieldByteOrder()}, t.meta(), is_try);
     }
 
-    result_t operator()(const hilti::type::Void& /* t */) { return hilti::expression::Void(); }
+    result_t operator()(const hilti::type::Void& /* t */) {
+        // Even though we we do not store parsed data, we still need to consume
+        // data in the input stream so that e.g., `&size` checks can work.
+        pb->advanceInput(builder::size(state().cur));
```
rsmmr, line 173: This comment feels a bit misleading because I think it's really *only* with `&size` where skipping ahead over the all current input is the right thing to do. Are we already rejecting `void` fields that don't have any attribute? If we don't then this would act like `&eod` for them, right? If we want that, we should make it explicit and require `: void &eod`. 

In other words, let's make sure we validate that only supported cases can arrive here, and update the comment to make that assumption clear.

rsmmr (commented):
Also need to document what we support for parsing `void` (there's already a section on that that we can extend).

spicy/toolchain/src/compiler/codegen/parsers/types.cc
bbannier: Thanks for the careful reading, this indeed introduced a bug.

The current behavior is for `void` fields w/o explicit `&size` is to consume nothing which is required behavior if used in unit switch statements. I added a new test checking that unit-level `void` fields do indeed consume nothing, and also introduced a validator to reject any attributes other than `&size` for `void` fields (e.g., `&max-size` would introduce off-by-one errors on `void` fields).

While I was at it I also added code to reject `void` fields with names.

spicy/toolchain/src/compiler/visitors/validator.cc
bbannier: I added a patch to this PR which rejects `void` fields with names so that it would become impossible to declared external field hooks on `void` fields.

We still have an issue with `$$` being accepted in hooks which do not provide it, e.g., `%done` hooks. I took a stab at adding validation for that as well along the lines you suggested above, but it looks like I cannot see the parameters of such hooks, e.g., slightly modifying the lines changed here to also include

```.cpp
                const auto& parameters = hook->get().type().parameters();
                if ( parameters.end() ==
                     std::find_if(parameters.begin(), parameters.end(), [](auto&& p) { return p.id() == ID("__dd"); }) )
                    error("$$ is undefined", p);
```

Here I _always_ see an empty parameter list for any of the following hook impls,

```.ruby
module Test;

public type X = unit {
    a: uint8 { print $$; } # Okay.
    on %done { print $$; } # Should fail: `$$` not available.
};

on Test::X::%done { print $$; } # Should fail: `$$` not available.
```

We probably need a broader fix which properly resolves `$$` in hooks instead of magically pasting in code without checking whether the value is defined; this should work for different kind of hooks and e.g., also for hooks on `void` fields which have a different signature. If we did that we wouldn't even need the fix here, but instead could use normal name resolution to reject such code. Should I drop this validation patch here, and instead file a new ticket?

spicy/toolchain/src/compiler/codegen/parsers/types.cc
```diff
@@ -167,7 +167,13 @@ struct Visitor : public hilti::visitor::PreOrder<Expression, Visitor> {
         return performUnpack(destination(t), t, t.width() / 8, {state().cur, fieldByteOrder()}, t.meta(), is_try);
     }
 
-    result_t operator()(const hilti::type::Void& /* t */) { return hilti::expression::Void(); }
+    result_t operator()(const hilti::type::Void& /* t */) {
+        if ( auto size = AttributeSet::find(meta.field()->attributes(), "&size") )
+            // Even though we we do not store parsed data, we still need to consume
+            // data in the input stream so that `&size` checks can work.
+            pb->advanceInput(*size->valueAs<Expression>());
```
rsmmr, line 174: Insert an empty line after block.

rsmmr (approved):

tests/spicy/types/unit/skip-field.spicy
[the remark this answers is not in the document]
rsmmr: I think we need to. Currently we're saying: "The Void type can be used as a place-holder for not parsing anything.". Not exactly intuitive that by adding `&size`, this changes.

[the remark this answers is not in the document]
bbannier: Sorry, I missed this. I opened a follow-up PR to also update the docs, #893.

PR #892, opened 2021-04-20
Fix hook handling for anonymous void fields.

### LINKED ISSUE #889: Empty default branch inside switch triggers error
```
module Test;

public type VarInt = unit {
    switch ( 1 ) {
        *    -> : void;
    };
};
```

```
# spicyc -j ./x.spicy
[error] ./x.spicy:6:16: struct type does not have a method `__on_anon`
[error] ./x.spicy:6:16: cannot resolve operator: <Test::VarInt>.__on_anon()
```

Adding an empty hook to the default brach makes it work:

```
        *    -> : void {}
```

bbannier:
We previously would unconditionally call field hooks for anonymous void
fields, even though they might not have been declared by the user. We
now only invoke the hooks if they are actually declared.

Closes #889.


spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -1505,7 +1505,10 @@ void ParserBuilder::newValueForField(const production::Meta& meta, const Express
     if ( value.type().isA<type::Void>() ) {
         // Special-case: No value parsed, but still run hook.
         beforeHook();
-        builder()->addMemberCall(state().self, ID(fmt("__on_%s", field->id().local())), {}, field->meta());
+
+        if ( ! field->hooks().empty() )
```
rsmmr, line 1509: This won't work: we do always need to call hooks unconditionally because they could be implemented in a different translation unit. The bug must be somewhere else.

rsmmr (approved):

PR #894, opened 2021-04-23
Support passing window bits into zlib when decompressing from Spicy.

rsmmr:
This adds a new Spicy runtime function `spicy::zlib_init(window_bits)`
that initializes a `ZlibStream` for decompression, passing the window
bits argument into zlib's `inflateInit2()` function.

The existing API remains: if one creates a ZlibStream directly without
calling `zlib_init()`, we continue to use 47 as the default value, which
means "check for, and require, a gzip file". (We could enforce going
through `zlib_init()`, but that would break the current API; seems
unnecessary).

zlib's window_bits parameter is messy. For details, see the
documentation for inflateInit2() at https://www.zlib.net/manual.html.


tests/spicy/rt/zlib.spicy
```diff
@@ -18,3 +21,11 @@ public type X = unit {
         print(out);
     }
 };
+
+# Expect raw deflate data (like inside a zip archive).
+public type X2 = unit {
+    : bytes &eod {
+        local z = spicy::zlib_init(-15); # maximum negative value
```
bbannier, line 28: Let's instead capture the semantics of the value `-15`, e.g.,

```suggestion
        local z = spicy::zlib_init(-15); # Negative for deflating raw data, 15 maximum window size. 
```

spicy/runtime/include/zlib_.h
```diff
@@ -61,6 +68,12 @@ class Stream {
     std::shared_ptr<detail::State> _state;
 };
 
+/** Instantiates a new `Stream` objcect, forwarding arguments to its constructor. */
```
bbannier, line 71:
```suggestion
/** Instantiates a new `Stream` object, forwarding arguments to its constructor. */
```

spicy/lib/spicy.spicy
```diff
@@ -58,6 +58,12 @@ public type Side = enum {
 ## Captures the state of gzip decompression for the corresponding library functions.
 public type ZlibStream = __library_type("spicy::rt::zlib::Stream");
 
+## Initializes a zlib stream for decompression.
+##
+## ``window_bits``: Same as the corresponding parameter for zlib's `inflateInit2`
+## (see https://www.zlib.net/manual.html).
```
bbannier, line 64: Let's also document that this function can throw a `ZlibError`.

spicy/runtime/src/tests/zlib.cc
```diff
@@ -28,6 +28,12 @@ TEST_CASE("decompress") {
             CHECK_EQ(zlib::finish(stream), ""_b);
         }
 
+        SUBCASE("raw deflate") {
+            stream = zlib::Stream(-15);
+            CHECK_EQ(zlib::decompress(stream, "\x33\x34\x84\x01\x2e\x00"_b), "1111111111\n"_b);
+            CHECK_EQ(zlib::finish(stream), ""_b);
```
bbannier, line 34: This mutates `stream` from the shared fixture. This isn't a bug, but still potentially confusing. I'd suggest we use a dedicated stream instead, e.g.,

```suggestion
            auto raw_stream = zlib::Stream(-15);
            CHECK_EQ(zlib::decompress(raw_stream, "\x33\x34\x84\x01\x2e\x00"_b), "1111111111\n"_b);
            CHECK_EQ(zlib::finish(raw_stream), ""_b);
```

bbannier (dismissed):

bbannier (approved):

PR #895, opened 2021-04-26
Fix crash on macOS ARM when writing over an existing HLTO file.

rsmmr:
On macOS ARM, there are odd crashes when loading an HLTO if we created it
by writing over an already existing HLTO file when. Fixing by always removing
existing any files first.

### DIFF
--- a/hilti/runtime/src/library.cc
+++ b/hilti/runtime/src/library.cc
@@ -119,6 +119,15 @@ hilti::rt::Result<Nothing> hilti::rt::Library::remove() const {
 
 hilti::rt::Result<hilti::rt::Nothing> hilti::rt::Library::save(const hilti::rt::filesystem::path& path) const {
     std::error_code ec;
+
+    // On macOS ARM, there are weird crashes during execution if we don't remove an existing file first. Note that
+    // `is_regular_file(`) fails if we cannot access the file due to permissions, but we let it fail during removal.
+    if ( hilti::rt::filesystem::is_regular_file(path, ec) ) {
+        hilti::rt::filesystem::remove(path, ec);
+        if ( ec )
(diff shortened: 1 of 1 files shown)


hilti/runtime/src/library.cc
```diff
@@ -119,6 +119,17 @@ hilti::rt::Result<Nothing> hilti::rt::Library::remove() const {
...
+
+    try {
+        // On macOS ARM, there'll be weird crashes during execution if we don't
+        // remove an existing file first. Note that is_regular_file() throws if
+        // we can't access the file due to permissions.
+        if ( hilti::rt::filesystem::is_regular_file(path) )
+            hilti::rt::filesystem::remove(path);
+    } catch ( const hilti::rt::filesystem::filesystem_error& e ) {
+        return result::Error(fmt("could not remove existing library when saving to %s: %s", path, e.what()));
+    }
```
bbannier, line 131: Alternative style to make the control flow slightly more explicit:

```suggestion
    // On macOS ARM, there are weird crashes during execution if we don't remove an existing file first. Note that
    // `is_regular_file(`) fails if we cannot access the file due to permissions, but we let it fail during removal.
    if ( hilti::rt::filesystem::is_regular_file(path, ec) ) {
        hilti::rt::filesystem::remove(path, ec);
        if ( ec )
            return result::Error(fmt("could not remove existing library when saving to %s: %s", path, ec.message()));
    }
```

bbannier (approved):

hilti/runtime/src/library.cc
rsmmr: Ah, I had missed that there's an `ec` version of `is_regular_file()` as well.

bbannier (approved):

PR #896, opened 2021-04-26
Force type information to come last in generated C++ code.

rsmmr:
We could run into some dependencies where the types being referred to in
the type information weren't declared yet.

No test included, as it's hard to recreate the situation; will be
tested through an upcoming change to the Zeek plugin.

### DIFF
--- a/hilti/toolchain/src/compiler/cxx/unit.cc
+++ b/hilti/toolchain/src/compiler/cxx/unit.cc
@@ -246,6 +246,11 @@ void Unit::_generateCode(Formatter& f, bool prototypes_only) {
         }
 
         if ( ! prototypes_only || ! util::endsWith(ns, "::") ) { // skip anonymous namespace
+            if ( ns == "__hlt::type_info::" )
+                // We force this to come last later below because creating the type information needs access to all
+                // other types.
+                continue;
+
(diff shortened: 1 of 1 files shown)


hilti/toolchain/src/compiler/cxx/unit.cc
```diff
@@ -273,6 +278,14 @@ void Unit::_generateCode(Formatter& f, bool prototypes_only) {
         }
     }
 
+    // Add the contents of the type information namespace. We know that there are only constants in there.
+    if ( ! prototypes_only ) {
+        for ( const auto& i : _constants ) {
+            if ( i.second.id.namespace_() == cxx::ID("__hlt::type_info::") )
+                f << i.second;
```
bbannier, line 285: Small prettification:

```suggestion
        for ( const auto& [id, constant] : _constants ) {
            if ( constant.id.namespace_() == cxx::ID("__hlt::type_info::") )
                f << constant;
```

bbannier (approved):

hilti/toolchain/src/compiler/cxx/unit.cc
rsmmr: I'll leave as is for now as the surrounding code uses the same style.

PR #898, opened 2021-04-26
Provide mktime() library function.

rsmmr:
This allows creating a `time` value from individual compenents, similar
to mktime(3). Providing to both HILTI and Spicy code.

### DIFF
--- a/hilti/runtime/src/types/time.cc
+++ b/hilti/runtime/src/types/time.cc
@@ -18,6 +18,27 @@ Time time::current_time() {
     return Time(t, Time::SecondTag());
 }
 
+Time time::mktime(uint64_t y, uint64_t m, uint64_t d, uint64_t H, uint64_t M, uint64_t S) {
+    if ( y < 1970 || (m < 1 || m > 12) || (d < 1 || d > 31) || H > 23 || M > 59 || S > 59 )
+        throw InvalidValue("value out of range");
+
+    struct tm t;
+    t.tm_sec = S;
+    t.tm_min = M;
+    t.tm_hour = H;
+    t.tm_mday = d;
+    t.tm_mon = m - 1;
+    t.tm_year = y - 1900;
+    t.tm_isdst = -1;
+
+    time_t teatime = mktime(&t);
+
+    if ( teatime < 0 )
+        throw InvalidValue("cannot create time value");
+
+    return Time(teatime, Time::SecondTag());
+}
+
 Time::operator std::string() const {
     if ( _nsecs == 0 )
         return "<not set>";
--- a/doc/autogen/spicy-functions.spicy
+++ b/doc/autogen/spicy-functions.spicy
@@ -45,6 +45,19 @@ Finalizes a base64 stream used for decoding or encoding.
 
 Returns the current wall clock time.
 
+.. _spicy_mktime:
+
+.. rubric:: ``function spicy::mktime(y: uint64, m: uint64, d: uint64, H: uint64, M: uint64, S: uint64) : time``
+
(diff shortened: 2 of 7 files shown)


hilti/runtime/src/types/time.cc
```diff
@@ -18,6 +18,24 @@ Time time::current_time() {
...
 }
 
+Time time::mktime(uint64_t y, uint64_t m, uint64_t d, uint64_t H, uint64_t M, uint64_t S) {
+    struct tm t;
+    t.tm_sec = S;
+    t.tm_min = M;
+    t.tm_hour = H;
+    t.tm_mday = d;
+    t.tm_mon = m - 1;
+    t.tm_year = y - 1900;
```
bbannier, line 28: More an observation: The epoch will depend on whether we use POSIX or not. Currently that seems to be the case, but might become an issue one we need to support e.g., win platforms. I think above unit test might catch when the epoch becomes a problem.

hilti/runtime/src/types/time.cc
```diff
@@ -18,6 +18,24 @@ Time time::current_time() {
...
+    t.tm_isdst = -1;
```
bbannier, line 29: Should we also allow users to override DST and `tm_zone`? Without giving them this option this function seem to be hard to use to convert general time stamps which differ from the setting of the particular machine this was running on.

Same question for the function exposed in the Spicy API.

hilti/runtime/src/types/time.cc
bbannier, line 28: Let's add validation that we do not underflow here. This probably requires throwing an exception (and a matching test case).

hilti/runtime/src/types/time.cc
rsmmr: mktime(3) always seems to use local time, with `dst` and `tm_zone` being non-portable extensions if considered at all. We could use `timegm()` instead, which takes UTC, but that would lead to unexpected values too I think. I don't see an easy alternative, guess we could consider finding a different implementation of `mktime` that provides a TZ parameter.

hilti/runtime/src/types/time.cc
rsmmr: What's not portable in this part?

bbannier: You are right, `mktime` is POSIX and uses the UNIX epoch as well, https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/mktime-mktime32-mktime64?view=msvc-160.

hilti/runtime/src/types/time.cc
bbannier: I still stand by my objection that this is hard to use in general, but I didn't mean to suggest that we need to provide a full set of time functions. Looking at [Python](https://docs.python.org/3/library/time.html) they seem to (at least at that level) mirror POSIX; [Golang](https://golang.org/pkg/time/) reinvents this, but I am unsure what shape their wheel has.

As an idea, we could accept optional DST and TZ parameters here and perform the conversion inside the function if they are given, or we follow whatever Zeek choose to follow (since they currently are our biggest user and maybe already needed to solve for this).

rsmmr: Yeah, I think we'd need to do what you suggest and at least accept a `TZ` parameter. But I'm not sure that doing that conversion is straight-forward. If you have an idea how to implement that easily, happy to add; otherwise I'd go with this version for now and we can add a defaulted `TZ` parameter later when need comes up (`dst` I'd just skip).

rsmmr: Going ahead with this, as discussed.

PR #903, opened 2021-04-29
Add fuzzing for generated parsers.

bbannier:


3rdparty/CMakeLists.txt
```diff
@@ -11,11 +11,6 @@ set(FIBER_SHARED OFF)
 set(FIBER_OBJECT ON)
 add_subdirectory(fiber)
 
-set(BENCHMARK_ENABLE_GTEST_TESTS OFF)
-set(BENCHMARK_ENABLE_TESTING OFF)
-set(BENCHMARK_ENABLE_INSTALL OFF)
-add_subdirectory(benchmark)
-
```
rsmmr, line 18: I assume this is because the fiber benchmark is being removed, not sure if that's on purpose  (see question below)

hilti/runtime/CMakeLists.txt
```diff
@@ -161,8 +161,3 @@ target_compile_options(hilti-rt-configuration-tests PRIVATE $<$<CONFIG:Debug>:-O
 target_link_options(hilti-rt-configuration-tests PRIVATE $<$<CONFIG:Debug>:-O0>)
 target_link_libraries(hilti-rt-configuration-tests PRIVATE $<IF:$<CONFIG:Debug>,hilti-rt-debug,hilti-rt> doctest)
 add_test(NAME hilti-rt-configuration-tests COMMAND ${CMAKE_BINARY_DIR}/bin/hilti-rt-configuration-tests)
-
```
rsmmr, line 164: Why's this going out?

ci/fuzz/fuzz.cc
```diff
@@ -0,0 +1,62 @@
...
+            if ( name && p->name == *name )
+                break;
+        }
+    }
+
+    assert(parser);
+    assert(parser->parse1);
+
+    hilti::rt::ValueReference<hilti::rt::Stream> stream;
+    stream->append(reinterpret_cast<const char*>(Data), Size);
```
rsmmr, line 36: Extension for later: we could also test parsing the data incrementally, one byte at a time. But let's get this going for a while first to see if anything shows up this way.

ci/fuzz/fuzz.cc
```diff
@@ -0,0 +1,62 @@
...
+
+    hilti::rt::ValueReference<spicy::rt::ParsedUnit> pu;
+
+    try {
+        assert(parser->parse1 || parser->parse3);
+        if ( parser->parse1 )
```
rsmmr, line 42: parse1 should always be available, or did you see something different?

ci/fuzz/fuzz.cc
```diff
@@ -0,0 +1,62 @@
...
+            parser->parse1(stream, {}, {});
+        else if ( parser->parse3 )
+            parser->parse3(pu, stream, {}, {});
+    } catch ( const spicy::rt::ParseError& ) {
+    } catch ( const hilti::rt::StackSizeExceeded& ) { // FIXME(bbannier): should we trigger this on small inputs?
+    } catch ( ... ) {
+    }
```
rsmmr, line 49: How about print out the exception here so that when running, one sees what's going on (I assume one can see output).

ci/fuzz/fuzz.cc
```diff
@@ -0,0 +1,62 @@
...
+
+    try {
+        assert(parser->parse1 || parser->parse3);
+        if ( parser->parse1 )
+            parser->parse1(stream, {}, {});
+        else if ( parser->parse3 )
+            parser->parse3(pu, stream, {}, {});
+    } catch ( const spicy::rt::ParseError& ) {
+    } catch ( const hilti::rt::StackSizeExceeded& ) { // FIXME(bbannier): should we trigger this on small inputs?
+    } catch ( ... ) {
```
rsmmr, line 48: I'm wondering if we can limit the catch-all here to HILTI/Spicy exceptions. Conceptually that would be nice, however our current runtime code can probably through `std`  exceptions legitimate as well, so maybe that's not viable.

ci/fuzz/run.sh
```diff
@@ -0,0 +1,100 @@
...
+
+# shellcheck disable=SC2043
+for parser in Message; do
+mkdir -p spicy-fuzz/corpus-fuzz-dhcp-"$parser"
+docker run -it --rm \
+    -v "${PWD}"/spicy-fuzz:/work \
+    -e SPICY_FUZZ_PARSER="$parser" \
+    spicy-fuzz \
+    /work/fuzz-dhcp -max_total_time=$MAX_TOTAL_TIME -jobs="$(nproc)" /work/corpus-fuzz-dhcp-"$parser"
+done
```
rsmmr, line 32: Any reason not to factor this out into a function instead of repeating it below?

rsmmr (approved):

hilti/runtime/CMakeLists.txt
bbannier: This shouldn't have ended up in this PR, I'll fix it.

ci/fuzz/fuzz.cc
bbannier: I think that's not practical. Fuzzing will invoke parsing between hundreds or thousands of times per second, so we would generate a huge amount of logs with no clear idea on how to analyze them.

I think if we want to look into specific exceptions we need to patch this code to whitelist expected exceptions.

ci/fuzz/fuzz.cc
bbannier: That's an interesting idea. I don't recall whether we made the requirement to only ever throw our runtime exceptions from parsers, but if that is so we could certainly only catch our errors here to detect leaking incorrect exceptions.

ci/fuzz/run.sh
bbannier: The way this is currently written we assume that fuzzers would be run sequentially which can take a long time. I wasn't sure whether that was the best way forward.

I'll look into how to create multidimensional arrays in here to capture data of the form `[GRAMMAR, [PARSER..]` we can iterate over.

ci/fuzz/fuzz.cc
rsmmr: Fair enough.

ci/fuzz/fuzz.cc
rsmmr: yeah, we didn't make that exploit, and may indeed have just been accepted throwing `std` exceptions. Something for future cleanup, it would be nice to have exception semantics that trigger only our own ones.

ci/fuzz/run.sh
rsmmr: I meant just the bodies of the for loops; that's always the same, isn't it?

ci/fuzz/fuzz.cc
bbannier: I filed zeek/spicy#907 to capture this.

rsmmr (approved):

ci/fuzz/fuzz.cc
bbannier: It doesn't seem to be available for units which take a parameter, e.g., `Content` from spicy-analyzer's HTTP parser.

rsmmr: Oh, good point, I forgot that those can now make into the parsers table.

rsmmr (approved):

PR #905, opened 2021-05-03
Fix unit-level convert using `self`

rsmmr:
Depending on processing order, `self` could be resolved to the wrong unit type. (This is another result of our current messy type resolution scheme, will get better once that's overhauled.)

The first commit is independent of the fix, just something I noticed while looking into what's going on.


hilti/toolchain/src/compiler/visitors/printer.cc
```diff
@@ -982,7 +982,22 @@ struct Visitor : visitor::PreOrder<void, Visitor> {
...
+                if ( ! first )
+                    out << ", ";
+
+                if ( id )
+                    out << id << ": " << ty;
+                else
+                    out << ty;
+
+                first = false;
+            }
```
bbannier, line 998: This looks equivalent to a `transform` and a `join`,

```suggestion
            out << util::join(util::transform(n.elements(),
                                              [](auto&& element) {
                                                  auto&& [id, ty] = element;

                                                  if ( id )
                                                      return fmt("%s: %s", id, ty);
                                                  else
                                                      return fmt("%s", ty);
                                              }),
                              ",");

```

This requires an `#include <hilti/base/util.h>`.

tests/spicy/types/unit/unit-convert-two.spicy
```diff
@@ -0,0 +1,21 @@
+# @TEST-EXEC: printf '\776\777' | spicy-dump -d %INPUT >output
```
bbannier, line 1:
```suggestion
# @TEST-EXEC: ${SCRIPTS}/printf '\776\777' | spicy-dump -d %INPUT >output
```

spicy/toolchain/include/ast/types/unit-items/field.h
```diff
@@ -101,7 +101,8 @@ class Field : public hilti::NodeBase, public spicy::trait::isUnitItem {
     Node& itemNode() { return childs()[2]; }
     Node& attributesNode() { return childs()[4]; }
 
-    std::optional<std::pair<Expression, bool>> convertExpression() const;
+    // For unit-level converts, returns the unit type as well.
```
bbannier, line 104: Let's add document what this function does, and only then the behavior for unit-level converts, e.g.,

```suggestion
    // Get the `&convert` expression, if any.
    //
    // For unit-level converts, returns the unit type as well.
```

spicy/toolchain/src/ast/types/unit-items/field.cc
```diff
@@ -59,7 +59,45 @@ Type spicy::type::unit::item::Field::itemType() const {
...
+
+                auto field = st->field("__convert");
+                assert(field);
+
+                auto f = field->type().as<type::Function>();
+                return adapt_type(f.result().type());
+            });
+        }
+        else {
+            return type::Computed(convert->first, [adapt_type](Node& n) -> Type {
```
bbannier, line 96:
```suggestion
            return type::Computed(convert->first, [&adapt_type](Node& n) -> Type {
```

spicy/toolchain/src/ast/types/unit-items/field.cc
```diff
@@ -59,7 +59,45 @@ Type spicy::type::unit::item::Field::itemType() const {
...
+
+                auto st = t.tryAs<hilti::type::Struct>();
+                if ( ! st )
+                    return type::unknown;
+
+                auto field = st->field("__convert");
+                assert(field);
+
+                auto f = field->type().as<type::Function>();
+                return adapt_type(f.result().type());
```
bbannier, line 92: Could you add a comment explaining why we go via the `__convert` field for unit-level converts, but directly via the expression for field-level converts? I naively would have expected these to take a similar route with the unit-level path possibly taking the unit type into account during resolution or similar.

rsmmr: There is no __convert for field-level attributes, we just added that to get around typing issues for unit-level. I'll add a note.

bbannier (dismissed):

PR #906, opened 2021-05-03
Port autogen-version to python3

sykhro:
As part of the Windows port effort.
Output checked against the old bash version with several configurations.

- ~~Bump the minimum python3 version to 3.7. 
While asking for python3.7 in 2021 is probably reasonable I realize that this version bump might be a bit extreme, considering that the other python scripts in the repo apparently run on 3.2. With some readability sacrifices we can target something lower than 3.7 if that's a requirement.~~
After a minor change the now required minimum is 3.6, which is what centos7 ships,
- Explicitly invoke `Python3_EXECUTABLE` in the cmake commands that need them (workaround for shebangs not working on Windows). For now this is limited to autogen-version calls.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,11 +2,7 @@
 
 cmake_minimum_required(VERSION 3.15.0)
 
-execute_process(COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/scripts/autogen-version --cmake --git-root ${CMAKE_CURRENT_SOURCE_DIR} --store ${CMAKE_SOURCE_DIR}/VERSION
-    OUTPUT_VARIABLE SPICY_VERSION
-    OUTPUT_STRIP_TRAILING_WHITESPACE)
-
-project(spicy VERSION "${SPICY_VERSION}" LANGUAGES ASM C CXX)
+project(spicy LANGUAGES ASM C CXX)
 
 set(flex_minimum_version "2.6")
 set(bison_minimum_version "3.0")
@@ -183,6 +179,12 @@ endif ()
 
 add_subdirectory(3rdparty)
 
+# Compute project version.
+execute_process(COMMAND ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/scripts/autogen-version --cmake --git-root ${CMAKE_CURRENT_SOURCE_DIR} --store ${CMAKE_SOURCE_DIR}/VERSION
+    OUTPUT_VARIABLE SPICY_VERSION
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+set(CMAKE_PROJECT_VERSION ${SPICY_VERSION})
+
 ## Print build summary
 string(TOUPPER ${CMAKE_BUILD_TYPE} BuildType)
 
@@ -193,7 +195,7 @@ string(STRIP "${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_${BuildType}}" cxxflags)
 add_custom_target(check COMMAND ctest --output-on-failure -C $<CONFIG> DEPENDS tests)
 add_custom_target(tests DEPENDS hilti-tests spicy-tests)
 
-execute_process(COMMAND ${PROJECT_SOURCE_DIR}/scripts/autogen-version
+execute_process(COMMAND ${Python3_EXECUTABLE} ${PROJECT_SOURCE_DIR}/scripts/autogen-version
                 OUTPUT_VARIABLE VERSION
                 OUTPUT_STRIP_TRAILING_WHITESPACE)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@ project(spicy VERSION "${SPICY_VERSION}" LANGUAGES ASM C CXX)
 
 set(flex_minimum_version "2.6")
 set(bison_minimum_version "3.0")
-set(python_minimum_version "2.4")
+set(python_minimum_version "3.6")
 set(macos_minimum_version "19.0.0") # macOS 10.15.0 (Catalina)
 
 ## Initialize defaults & global options
--- a/scripts/autogen-version
+++ b/scripts/autogen-version
@@ -1,166 +1,178 @@
-#!/bin/sh
-#
+#! /usr/bin/env python3
 # Generates a readable representation of the current version number; with
 # --short a short one. Or alternatively with --header, a version.h file with
 # various constants defined accordingly.
 
-set -e
+import argparse
+import os
+import re
+import shutil
+import subprocess
+import tempfile
 
-usage() {
-    echo "usage $(basename "$0") [--short | --header <file> | --cmake | --store <file>] [--git-root <dir>] [<commit>]"
-}
 
-# Make the top-level git repository the current directory.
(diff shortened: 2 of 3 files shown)


sykhro:
I'll do a minor edit and bring it down to 3.6, then we should be fine with centos.

scripts/autogen-version
```diff
@@ -1,166 +1,147 @@
...
+        except:
+            pass
+
+        describe_arg = ref if ref != "HEAD" else "--dirty"
+        raw_git_version = subprocess.check_output(
+            ["git", "describe", "--always", "--tags", "--match", "v*", describe_arg],
+            universal_newlines=True,
+        )
+        git_version = re.sub(
+            r"-([0-9]*)-g[0-9a-z]*", r".\1", re.sub(r"^v", "", raw_git_version)
```
bbannier, line 41: This is complicated enough that it needs a comment documenting the intended effect.

That we match trailing `-g...` here seems to break version computation on release tags, e.g.,

```
# In clean checkout of this PR:
$ git tag v9.9.9
$ git checkout v9.9.9
$ ./scripts/autogen-version
Traceback (most recent call last):
  File "/Users/bbannier/src/spicy/build/./scripts/autogen-version", line 88, in <module>
    version, commit, *_ = git_version.split("-")
ValueError: not enough values to unpack (expected at least 2, got 1)
```

scripts/autogen-version
```diff
@@ -1,166 +1,147 @@
...
+        ).strip()
+
+        branch = hash
+        try:
+            branch = subprocess.check_output(
+                ["git", "symbolic-ref", "--short", ref],
+                universal_newlines=True,
+                stderr=subprocess.DEVNULL,
+            ).strip()
+        except:
```
bbannier, line 32:
```suggestion
        except subprocess.CalledProcessError:
```

scripts/autogen-version
```diff
@@ -1,166 +1,147 @@
...
+output_mode_args = parser.add_mutually_exclusive_group()
+output_mode_args.add_argument("--short", action="store_true")
+output_mode_args.add_argument("--cmake", action="store_true")
+parser.add_argument("--header", metavar="file", nargs="?", const="version.h")
+parser.add_argument("--store", metavar="file", nargs="?", const="VERSION")
+parser.add_argument("--git-root", metavar="directory", nargs=1)
+parser.add_argument("--commit", metavar="COMMIT_HASH", default="HEAD")
+
+options = parser.parse_args()
+# Move to the repository root
```
bbannier, line 70:
```suggestion
# Move to the repository root.
```

scripts/autogen-version
```diff
@@ -1,166 +1,147 @@
...
+parser = argparse.ArgumentParser(
+    description="Generates a readable representation of the current version number"
+)
+output_mode_args = parser.add_mutually_exclusive_group()
+output_mode_args.add_argument("--short", action="store_true")
+output_mode_args.add_argument("--cmake", action="store_true")
+parser.add_argument("--header", metavar="file", nargs="?", const="version.h")
+parser.add_argument("--store", metavar="file", nargs="?", const="VERSION")
+parser.add_argument("--git-root", metavar="directory", nargs=1)
+parser.add_argument("--commit", metavar="COMMIT_HASH", default="HEAD")
```
bbannier, line 67: While we are at it we could add some meaningful documentation on the semantics of these flags, e.g.,
```suggestion
output_mode_args.add_argument(
    "--short", action="store_true", help="output a short version string")
output_mode_args.add_argument(
    "--cmake", action="store_true", help="output version string for use with CMake")
parser.add_argument("--header", metavar="file", nargs="?",
                    const="version.h", help="generate a version.h file for compilation")
parser.add_argument("--store", metavar="file", nargs="?", const="VERSION",
                    help="generate VERSION file persisting Git information")
parser.add_argument("--git-root", metavar="directory", nargs=1,
                    help="path to directory containing .git folder")
parser.add_argument("--commit", metavar="COMMIT_HASH", default="HEAD",
                    help="commit to generate version information for")
```

bbannier (changes requested):
Thanks for this PR. This looks already really good.

Unfortunately we do not exercise all of the functionality of this script in PR CI, in particular not version computation on a release tag. I left a comment on that on a broken line.

CMakeLists.txt
```diff
@@ -183,6 +179,12 @@ endif ()
 
 add_subdirectory(3rdparty)
 
+# Compute project version
```
bbannier, line 182:
```suggestion
# Compute project version.
```

bbannier (changes requested):

scripts/autogen-version
sykhro: I'll have to investigate what's going on here, it doesn't seem to break on my end
```
$ git tag v1.33.7
$ git checkout v.1.33.7
$ ./scripts/autogen-version 
1.33.7 (02be372e)
```
Does your git output match mine?
```
$ git describe --always --tags --match "v*"
v1.33.7

without tagging:

$ git describe --always --tags --match "v*"
v1.1.0-dev-61-g02be372e
```
(regex was copied from 3f7dedd41490a36d1a445c7fa5c12bfd9c602e46, `regex.sub()` should behave just like `sed`)

bbannier: I do get similar output.

Are you sure you are invoking this in a clean checkout? The script would e.g., append `--dirty` if there are unstaged changes which where `git describe` would then produce e.g., `v1.33.7-dirty` which can make the splitting below work (around l.88).

sykhro: Whoops, I cloned the repo but issued the command inside the work one.
I guess we can either fix the regex or change line 116 like this, which is not pretty but I guess should work fine:
`version, commit, *_ = git_version.split("-") + [""]`

bbannier: Yes, I think fixing this after the fact is probably okay. You could also go with

```.python
version, commit, *_ = git_version.split('-') if '-' in git_version else (git_version, "")
```

In any case, let's document this edge case in a comment.

sykhro: Went with your version since it makes it easier to understand what's going on

bbannier (approved):
Thanks for the cleanup, will merge after CI passes.

bbannier:
@sykhro, just realized that we need to support python-3.5 for debian-9 and ubuntu-16. We do not run CIs for those on PRs by default, but [do on `main`](https://cirrus-ci.com/build/6318849554382848) (they can also be triggered manually from the Cirrus CI).

I cannot see anything obvious which would require 3.6, and wouldn't work on 3.5. Do you recall what would be required to support 3.5 instead of 3.6?

sykhro:
> @sykhro, just realized that we need to support python-3.5 for debian-9 and ubuntu-16. We do not run CIs for those on PRs by default, but [do on `main`](https://cirrus-ci.com/build/6318849554382848) (they can also be triggered manually from the Cirrus CI).
> 
> I cannot see anything obvious which would require 3.6, and wouldn't work on 3.5. Do you recall what would be required to support 3.5 instead of 3.6?

f-strings have been added in 3.6, so lines 156-166 would have to be rewritten to use `.format()` instead. Once that's done the script should work all the way down to 3.2

bbannier:
I went all the way to >=python-3.2 by also replacing `subprocess.run` with `subprocess.call` in #909.

PR #912, opened 2021-05-06
Add `encode()` method to strings for conversion to bytes.

### LINKED ISSUE #911: missing ability to do string-to-bytes conversion
In talking with @rsmmr in slack we noted that while it's possible to go from `bytes` to `string` via [`decode`](https://docs.zeek.org/projects/spicy/en/latest/programming/language/types.html#method-bytes::decode), there is no `string` to `bytes` conversion.

In my particular example, I'm trying to store a [print-f formatted](https://docs.zeek.org/projects/spicy/en/latest/programming/language/types.html#operator-string::Modulo) string into a bytes field and not finding a way to be able to do so.

rsmmr:
Closes #911.


bbannier (approved):

bbannier (dismissed):
LGTM. Can you break out the submodule changes into their own PR (is this zeek/spicy-plugin#36?)

rsmmr:
> LGTM. Can you break out the submodule changes into their own PR (is this [zeek/spicy-plugin#36](https://github.com/zeek/spicy-plugin/pull/36)?)

Ah, those shouldn't have been in there, will revert.

PR #916, opened 2021-05-10
Add optimizer pass to remove unimplemented member functions

bbannier:


hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -0,0 +1,235 @@
...
+        _stage = Stage::PRUNE;
+
+        bool modified = false;
+
+        for ( auto&& [hook_id, uses] : *_hooks ) {
+            // Linker joins are implemented via functions, so if we remove all
+            // functions data dependencies (e.g., needed for subunits) might
+            // get broken. Leave at least one function in unit so it gets emitted.
+            //
+            // TODO(bbannier): Explicitly express data dependencies in joins.
```
bbannier, line 67: Filed #918 for this.

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    result_t operator()(const type::struct_::Field& x, position_t p) {
+        if ( auto type_ = x.type().tryAs<type::Function>(); ! type_ )
+            return false;
+
+        auto fn = x.childsOfType<Function>();
+        assert(fn.size() <= 1);
+
+        bool is_keep = ! fn.empty() && AttributeSet::find(fn.front().attributes(), "&keep");
```
rsmmr, line 92: any idea for a more self-explanatory attribute name? `&no-optimize` perhaps?

spicy/runtime/include/parser.h
```diff
@@ -76,6 +77,57 @@ inline std::string to_string(const spicy::rt::ParserPort& x, adl::tag /*unused*/
 
 namespace spicy::rt {
 
+namespace detail {
+
+// Helper traits to detect whether a parser implements sink functions.
```
rsmmr, line 82:
```suggestion
// Helper traits to detect whether a parser implements sink hooks.
```

spicy/runtime/include/parser.h
```diff
@@ -76,6 +77,57 @@ inline std::string to_string(const spicy::rt::ParserPort& x, adl::tag /*unused*/
...
+
```
rsmmr, line 83: can we avoid the repitition here create and create a helper the hides the boilerplate common to all four?

rsmmr (commented):
So left comments on little stuff, but this looks good overall, along of the lines of what I had been hoping for in terms of complexity. 

Additional pieces: 

- I'm tempted to move the global-optimizer code into the `visitors/` directory, but I realize that it's not quite the same, so if you prefer leaving it where it is, that's fine.

- It would be nice if we had a better way of assembling those ID triples, it's pretty "manual" right now and makes some assumptions on the structure of our namespacing (e.g., that we don't have nested namespaces). That's ok for now, and we do it elsewhere as well already (multiple times even, iirc).  But I'm  wondering if we could provide additional infrastructure here to centralize this logic, so that the global optimizer (and other places) could just say, e.g., "give me the fully scoped ID for this hook". For example, the existing ID resolver could compute the information once and record it with the `Declaration`; and then other places would have that available and become much easier. (I'm fine leaving this as is right now, but wanted to put it out there).

- for how to treat external C++ code, how about this (similar to what we talked about):
   1. Generally leave optimization on by default, and have an option to turn it off.
   2. During linking, when optimization is on, reject any external C++ code that comes with HILTI linker information.

- the optimization doesn't seem to kick in for imported modules: try a Spicy module with `import filter`: it'll have all the hooks in there for types defined inside `filter.spicy`

- I don't see tests for the optimization yet?

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -0,0 +1,235 @@
...
+
+        auto field_id = x.id();
+
+        auto unit_type = typeID(p.parent().as<type::Struct>());
```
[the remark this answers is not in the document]
bbannier: I believe this is needed as we call `typeID` which can return nothing.

hilti/toolchain/src/compiler/driver.cc
```diff
@@ -609,6 +594,33 @@ Result<Nothing> Driver::compileUnits() {
     return Nothing();
 }
 
+Result<Nothing> Driver::codegenUnits() {
+    if ( _stage != Stage::COMPILED )
+        logger().internalError("unexpected driver stage in codegenUnits()");
+
+    if ( _driver_options.output_hilti && ! _driver_options.include_linker )
+        // No need to kick off code generation.
+        return Nothing();
```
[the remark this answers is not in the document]
bbannier: Hmm, this is called directly in `Driver::compile`. Currently functions there are called unconditionally and might skip internally.

I think keeping this branch here makes sense as it is an implementation detail on how we codegen (e.g., might skip altogether). WDYT?

hilti/toolchain/src/compiler/driver.cc
```diff
@@ -609,6 +594,33 @@ Result<Nothing> Driver::compileUnits() {
...
+
+    HILTI_DEBUG(logging::debug::Compiler, "compiling modules to C++");
```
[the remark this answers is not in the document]
bbannier: Same question as on the branch skipping codegen, moving this up on level seems leaky to me.

hilti/toolchain/include/compiler/global-optimizer.h
```diff
@@ -0,0 +1,33 @@
...
+    ~GlobalOptimizer() { _units = nullptr; }
+
+    void run();
+
+    std::vector<Unit>* _units = nullptr;
+    std::shared_ptr<Context> _ctx;
+
+    // Mapping of (module, unit, field) tuples to uses. Bit 0 is set for
+    // declared hooks, bit 1 for defined hooks and bit 2 for referenced hooks.
+    using Hooks = std::map<std::tuple<ID, ID, ID>, std::bitset<3>>;
```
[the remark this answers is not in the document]
bbannier: I introduced typedefs for the different `ID`s (also cross-checked with a strong typedef), and a dedicated `struct` for the uses.

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -0,0 +1,235 @@
...
+
+        for ( auto i : this->walk(&node) ) {
+            if ( auto x = dispatch(i) )
+                modified = modified && x;
+        }
+
+        if ( ! modified )
+            return;
+
+        prune(node);
```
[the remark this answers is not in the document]
bbannier: I went for the recursive version without much thinking, replaced it with a `do`/`while` loop.

hilti/toolchain/src/compiler/global-optimizer.cc
bbannier: This isn't really about _not optimizing_, but more about _keep a function which is part of some API_, which I think `&no-optimize` doesn't capture (and we might even end up optimizing the internals of such functions in the future).

WDYT about e.g., `&always_emit` or `&external_api`?

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -0,0 +1,235 @@
...
+                continue;
+
+            Visitor(&_hooks).collect(unit.imported(id));
+            collectedDependencies.insert(id);
+        }
+
+        Visitor(&_hooks).collect(unit.imported(unit.id()));
+    }
+
+    for ( auto& unit : *_units )
```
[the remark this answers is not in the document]
bbannier: Added logic to prune dependencies as well.

hilti/toolchain/src/compiler/driver.cc
```diff
@@ -479,6 +490,9 @@ Result<Nothing> Driver::addInput(const hilti::rt::filesystem::path& path) {
     }
 
     else if ( path.extension() == ".cc" || path.extension() == ".cxx" ) {
+        if ( _driver_options.global_optimizations )
+            return result::Error("Loading C++ files is not supported with transformations enabled, rerun with '-g'");
+
```
rsmmr, line 495: Iiirc, for this we arrived at allowing C++ to be added as long as they don't include HILTI linker information (because then they are probably just additional library code that won't have trouble with our optimizations).

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -0,0 +1,244 @@
...
+            // see https://github.com/zeek/spicy/issues/918.
+            if ( std::get<2>(hook_id) == std::string("__str__") ) {
+                uses.defined = true;
+                continue;
+            }
+        }
+
+        for ( auto i : this->walk(&node) ) {
+            if ( auto x = dispatch(i) )
+                modified = modified && x;
```
rsmmr, line 74:
```suggestion
                modified = modified || x;
```

No?

hilti/toolchain/src/compiler/global-optimizer.cc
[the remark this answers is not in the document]
rsmmr: I don't see the while loop?

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -0,0 +1,244 @@
...
+
+    result_t operator()(const type::struct_::Field& x, position_t p) {
+        if ( auto type_ = x.type().tryAs<type::Function>(); ! type_ )
+            return false;
+
+        auto field_id = x.id();
+
+        auto struct_type = typeID(p.parent().as<type::Struct>());
+        if ( ! struct_type )
+            return false;
```
rsmmr, line 91:
```suggestion
            return false;
            
```

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -0,0 +1,244 @@
...
+        auto&& [module_id, struct_id] = *struct_type;
+
+        auto type_ = p.findParent<declaration::Type>();
+        bool is_cxx = type_ && AttributeSet::find(type_->get().attributes(), "&cxxname");
+
+        auto ids = std::tuple<ModuleID, StructID, FieldID>{module_id, struct_id, field_id};
```
rsmmr, line 97: `make_tuple`?

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -0,0 +1,244 @@
...
+                    (*_hooks)[ids].defined = true;
+
+                break;
+            }
+
+            case Stage::PRUNE: {
+                // Remove hooks without implementation.
+                if ( auto hook = _hooks->at(ids); ! hook.defined ) {
+                    HILTI_DEBUG(logging::debug::GlobalOptimizer,
+                                util::fmt("removing field for unused default hook %s::%s::%s", module_id, struct_id,
```
rsmmr, line 129:
```suggestion
                                util::fmt("removing field for unused hook %s::%s::%s", module_id, struct_id,
```

(not sure what `default` would mean)

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -0,0 +1,244 @@
...
+        if ( ! struct_type )
+            return false;
+
+        auto&& [module_id, struct_id] = *struct_type;
+
+        auto&& member = x.op1().tryAs<expression::Member>();
+        if ( ! member )
+            return false;
+
+        auto hook_id = std::tuple<ModuleID, StructID, FieldID>{module_id, struct_id, member->id()};
```
rsmmr, line 188: `make_tuple`

rsmmr (commented):
- Quite a few of tests are adding `-g`. Do we need those? Would be preferable if most tests could go untouched.

- would be good to extend the tests further and add lower-level tests at the HILTI layer. We should have one test there for each basic transformation/situation. And what do you think about adding a separate tests hierarchy for optimizations (i.e., collect them into `test/{hilt, spicy}I/optimizations`)?

hilti/toolchain/include/compiler/global-optimizer.h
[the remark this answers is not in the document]
rsmmr: so personally I'd still go with single ID, but fine stay with the 3-typle typedefs too, that does make it more readable.

hilti/toolchain/src/compiler/global-optimizer.cc
rsmmr: I like `&always-emit` , we already `&no-emit` elsewhere, that matches.

hilti/toolchain/src/compiler/global-optimizer.cc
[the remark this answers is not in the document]
bbannier: This must have gotten lost during rebasing, adding it back in.

rsmmr (commented):
As discussed, let's add lower-level HILTI-side tests as well.

tests/hilti/optimization/unimplemented_hook.hlt
```diff
@@ -0,0 +1,28 @@
+# @TEST-EXEC: hiltic -d -j -D global-optimizer %INPUT > log 2>&1
+# @TEST-EXEC: btest-diff log
+
+# @TEST-EXEC: ${HILTIC} -d -j -g %INPUT -D ast-codegen >ast.0 2>&1
+# @TEST-EXEC: ${HILTIC} -d -j    %INPUT -D ast-codegen >ast.1 2>&1
+# @TEST-EXEC: /usr/bin/env diff -u ast.0 ast.1 | tail -n+3 >output.diff
```
rsmmr, line 6: This `diff` usage doesn't look very portable. I think it'd be better to determine the right `diff` in `btest.cfg`
and pass it in through an environment variable.

Also, let's document what diff we need (GNU or BSD?) and check that we got it.

tests/hilti/optimization/unimplemented_hook.hlt
```diff
@@ -0,0 +1,28 @@
+# @TEST-EXEC: hiltic -d -j -D global-optimizer %INPUT > log 2>&1
+# @TEST-EXEC: btest-diff log
+
+# @TEST-EXEC: ${HILTIC} -d -j -g %INPUT -D ast-codegen >ast.0 2>&1
+# @TEST-EXEC: ${HILTIC} -d -j    %INPUT -D ast-codegen >ast.1 2>&1
```
rsmmr, line 5:
```suggestion
# @TEST-EXEC: hiltic -d -j -g %INPUT -D ast-codegen >ast.0 2>&1
# @TEST-EXEC: hiltic -d -j    %INPUT -D ast-codegen >ast.1 2>&1
```

tests/hilti/optimization/unimplemented_hook.hlt
```diff
@@ -0,0 +1,28 @@
...
+# @TEST-EXEC: /usr/bin/env diff -u ast.0 ast.1 | tail -n+3 >output.diff
+# @TEST-EXEC: btest-diff output.diff
+
+# @TEST-DOC: Tests optimizations removing unimplemented hooks.
```
rsmmr, line 9: Is the optimization also handling global hooks (outside of struct) already? If so, add a test for that as well (and if not, something for the TODO list to support as well).

tests/spicy/optimization/default-parser-functions.spicy
```diff
@@ -0,0 +1,24 @@
+# @TEST-EXEC: ${SPICYC} -p -o /dev/null %INPUT -D global-optimizer >log 2>&1
+# @TEST-EXEC: btest-diff log
+
+# @TEST-EXEC: ${SPICYC} -d -j -g %INPUT -D ast-codegen >ast.0 2>&1
+# @TEST-EXEC: ${SPICYC} -d -j    %INPUT -D ast-codegen >ast.1 2>&1
```
rsmmr, line 5:
```suggestion
# @TEST-EXEC: spicyc -p -o /dev/null %INPUT -D global-optimizer >log 2>&1
# @TEST-EXEC: btest-diff log

# @TEST-EXEC: spicyc -d -j -g %INPUT -D ast-codegen >ast.0 2>&1
# @TEST-EXEC: spicyc -d -j    %INPUT -D ast-codegen >ast.1 2>&1
```

both works, but for consistency we can use the simpler version (I know the existing tests aren't consistent either)

doc/host-applications.rst
```diff
@@ -177,7 +177,7 @@ Finally, we compile it altogether:
 ::
 
     # spicyc -c my-http.spicy -o my-http.cc
-    # spicyc -l my-http.cc -o my-http-linker.cc
+    # spicyc -l my-http.spicy -o my-http-linker.cc
```
rsmmr, line 180: This is just to get around the spicyc error message, right? 

I think we should still add a note on optimization, at least something along the lines of "if you compile multiple Spicy modules independently into C++ code, you will currently need to turn off global optimizations".

tests/hilti/optimization/unimplemented_hook.hlt
bbannier: I replaced this with a custom diff util which I added to `tests/Scripts`. This also allows me to roll back the changes to the Alpine Docker image.

tests/spicy/optimization/default-parser-functions.spicy
```diff
@@ -0,0 +1,24 @@
...
+# @TEST-EXEC: /usr/bin/env diff -u ast.0 ast.1 | tail -n+3 >output.diff
```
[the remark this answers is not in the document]
bbannier: See above.

tests/hilti/optimization/unimplemented_hook.hlt
bbannier: This is currently not handled. I added a `TODO`.

rsmmr (approved):

PR #923, opened 2021-05-17
Draft 1.1 release notes.

rsmmr:


NEWS.md
```diff
@@ -0,0 +1,68 @@
...
+
+  * Port Spicy to Apple silicon.
+
+  * Add Dockerfile for OpenSUSE 15.2.
+
+Changed Functionality
+---------------------
+
+  * Reject `void` fields with names.
+  * Lower minimum required Python version to 3.2.
```
bbannier, line 50: I think this isn't new since 1.0.0, but the commit in question instead fixed a patch introduced since then which inadvertently raised the minimum required version.

README.md
```diff
@@ -6,7 +6,7 @@
 <table>
 <tr><th></th>              <th>Release</th> <th>Development</th> </tr>
 <tr><td>Documentation</td> <td colspan="2" align="center"><a href="https://docs.zeek.org/projects/spicy">Spicy Manual</a></td></tr>
-<tr><td>Changelog</td>     <td><a href="https://github.com/zeek/spicy/blob/v1.0.0/CHANGES">CHANGES</td> <td><a href="/CHANGES">CHANGES</a></td></tr>
+<tr><td>Release Notes</td> <td align="center"><a href="https://github.com/zeek/spicy/blob/v1.1.0/NEWS">NEWS</td> <td><a href="/CHANGES">CHANGES</a></td></tr>
```
bbannier, line 9: Let's also include the `NEWS` file literally in `doc/release-notes.rst` (probably before the `CHANGES`).

bbannier (approved):

README.md
bbannier: Thinking about that, that would require to use rst for `NEWS`, maybe that makes sense given that all our docs are in rst?

NEWS.md
rsmmr: Yeah, but compared to the 1.0, we still lowered it, it used to be 3.4.

README.md
rsmmr: Good point, moving it over into the docs sounds good. I made it MD (1) because the commit messages use MD at least partially, and (2) so that it'd be easier to copy into the GitHub releases.  But seems the docs should be our primary place for this, and that's reST.

NEWS.md
bbannier: Ah right.

rsmmr:
@bbannier I'll hold off on merging this, you can do that with the 1.1 release.

NEWS.rst
```diff
@@ -0,0 +1,67 @@
...
+  ``&until`` or ``&until_including``.
+
+- Add ``encode()`` method to ``string`` for conversion to ``bytes``.
+
+- Extend parsing of ``void`` fields:
+
+   - Add support for ``&eod`` to skip all data until the end of the
+     current input is encountered.
+
+   - Add support for ``&until`` to skip all data until a deliminator
```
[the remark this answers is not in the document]
duffy-ocraven: Or maybe it is gaining as a neologism. https://www.definitions.net/definition/deliminator does substantiate it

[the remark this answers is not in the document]
rsmmr: "originated among Perl hackers in connection with the Perl split() function; however, it has been sighted in live use among Java and even Visual Basic programmers." :-) 

agree that "delimiter" seems better/righter (but won't fix anymore, release is cut).

PR #927, opened 2021-05-19
Fix missing operator<< for enums when generating debug code.

rsmmr:


bbannier (approved):
LGTM. Let's wait for some CI to pass.

PR #930, opened 2021-05-25
Make only draft releases from automation [skip ci].

bbannier:
While this still requires us to create public tags, marking the releases
initially as drafts removes confusion for users who might end up using a
release we ultimately might not intend to publish.


rsmmr (approved):

bbannier:
Backported into 1.1 with 701833cd3f65748b277395494897efb2e9c34616.

PR #931, opened 2021-05-26
Update CI to clang-tidy-12

### LINKED ISSUE #915: Errors with recent clang-tidy
I'm getting a number of errors with clang-tidy 12 when running across the code base. It's pretty noisy, though could be just a a couple issues being repeated; haven't investigated yet.

bbannier:
Closes #915.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -32,10 +32,7 @@ clang9_zeek_ubuntu_debug_task:
     - git fetch --tags
     - git submodule update --recursive --init
 
-  # We currently run the build with sanitizers against clang-10 since clang-11 triggers unfixed diagnostics.
-  #
-  # TODO(bbannier): Switch this job to llvm-11.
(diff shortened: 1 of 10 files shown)


hilti/toolchain/include/ast/expression.h
```diff
@@ -46,6 +46,7 @@ using expression::detail::to_node;
 /** Constructs an AST node from any class implementing the `Expression` interface. */
 template<typename T, typename std::enable_if_t<std::is_base_of<trait::isExpression, T>::value>* = nullptr>
 inline Node to_node(T t) {
+    // NOLINTNEXTLINE(clang-analyzer-cplusplus.NewDeleteLeaks)
```
rsmmr, line 49: There are quite a few of these, looks like we should just disable this checker? Some quick Googling shows quite a few reports about false positives. We also have ASAN to catch such stuff.

bbannier: I disabled this check for now and removed the adjustments it triggered.

rsmmr (approved):

PR #932, opened 2021-05-28
Adjust processes used to build Docker image.

bbannier:
Github action runners only have limited resources (currently for Linux:
2 CPUs, 7GB RAM). In the build of the spicy-plugin and spicy-analyzers
Zeek packages we previously hardcoded a certain level of parallelism,
which caused us to run into especially RAM limits when building
spicy-analyzers. With recent versions of these packages we can now
optionally externally parameterize the level of parallelism to use.

This patch adjust the Github action setup so we build these packages
with a number of processes compatible with the runners we will be faced
with.

### DIFF
--- a/.github/workflows/docker-latest.yml
+++ b/.github/workflows/docker-latest.yml
@@ -35,6 +35,7 @@ jobs:
         context: .
         file: ci/Dockerfile.dockerhub
         push: true
+        build_args: "SPICY_ZKG_PROCESSES=1"
         tags: |
           ${{ secrets.DOCKER_USERNAME }}/spicy-dev:latest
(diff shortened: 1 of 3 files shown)


bbannier:
This still needs new package releases including zeek/spicy-plugin#41 and zeek/spicy-analyzers#60 to have visible effects (but should be safe to merge even before that happens).

We probably want to backport this to `release/1.1` as well.

rsmmr (approved):

bbannier:
Backported into 1.1 with 26e5c6b12c74d22b3ab3b26e946c7c11cbe36d12.

PR #933, opened 2021-05-28
Add support for unit switch &parse-at and &parse-from attributes.

### LINKED ISSUE #913: Add support for switch-level `&parse-from`/`&parse-at`
Two more attributes to support at the level of `switch` so that it applies to all cases.

bbannier:
Closes #913.


spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -981,8 +981,39 @@ struct ProductionVisitor
...
+            pstate.trim = builder::bool_(false);
+            pstate.lahead = builder()->addTmp("parse_lah", look_ahead::Type, look_ahead::None);
+            pstate.lahead_end = builder()->addTmp("parse_lahe", type::stream::Iterator());
+            auto expr = a->valueAs<Expression>();
+
+            auto cur = builder::memberCall(state().cur, "advance", {*expr});
+            pstate.cur = builder()->addTmp("parse_cur", cur);
+            pstate.ncur = {};
+            pushState(std::move(pstate));
+        }
```
rsmmr, line 1013: This looks like the code we already have? Let's factor it out to avoid the duplication then.

rsmmr (changes requested):

rsmmr (approved):

PR #935, opened 2021-06-03
Use forward declarations in linker metadata.

### LINKED ISSUE #918: Explicitly track type dependencies in linker joins
Currently we use linker joins to track what functions a unit exposes (e.g., to implement hook priorities). When optimizing away redundant default hook implementations (e.g., #916) we might remove hooks so that linker joins would not contain certain types in `aux_types` anymore. This can then cause compilation issues where types are not known anymore.

We probably should augment linker joins to explicitly contain exposed types (both public and private) to make linker joins work, even if hooks get removed.

bbannier:
Closes #918.


hilti/toolchain/src/compiler/cxx/linker.cc
```diff
@@ -63,8 +68,11 @@ void cxx::Linker::finalize() {
             if ( ! c.declare_only )
                 unit.add(c.callee);
 
-            for ( const auto& t : c.aux_types )
+            for ( auto t : c.aux_types ) {
+                // For the linker file forward declarations are sufficient.
+                t.forward_decl = true;
```
bbannier, line 73: The issue somewhere in this patch is still that types without hooks lead to a different linker metadata than files with hooks. While for the former we emit a forward declaration, for the latter (e.g., `Pointer` here) we emit a full declaration of the type, but do not include its own type dependencies.

rsmmr (approved):
Looks fine, but feels like a work-around around the problem in #949, right? Seems we could run into other types that still have problems (like an enum). Assuming that's right, I suggest putting a comment into the code to remind us about the issue and to remove the work-around later.

bbannier:
> Looks fine, but feels like a work-around around the problem in #949, right? Seems we could run into other types that still have problems (like an enum). Assuming that's right, I suggest putting a comment into the code to remind us about the issue and to remove the work-around later.

While this works around the incomplete list of dependent types I am not sure we'd want to remove the code touched here later. The current mechanics of `CodeGen::typeDeclaration` make no distinction between forward and full declarations which can lead to a lot of code being emitted with no way for a user to request e.g., a forward declaration (we don't even explicitly store forward declarations outside of units and `CodeGen` doesn't know about them). In the places touched here we know for sure that we just need forward declarations, so we can get away with emitting much less code (e.g., we never need declarations for all dependencies).

rsmmr:
> The current mechanics of `CodeGen::typeDeclaration` make no distinction between forward and full declarations 

Good point. Maybe we could add that later to typeDeclaration as a flag. Either way, looks good for now.

PR #936, opened 2021-06-08
Implement optimzing out unimplemented global hooks

bbannier:


hilti/toolchain/src/ast/scope.cc
```diff
@@ -91,8 +91,9 @@ void Scope::render(std::ostream& out, const std::string& prefix) const {
         for ( const auto& x : v ) {
             auto s = util::fmt("%s%s -> %s", prefix, k, x ? x->render(false) : "<invalid ref>");
 
-            if ( auto d = x->tryAs<declaration::Expression>() )
-                s += util::fmt(" (type: %s)", d->expression().type());
+            if ( x )
+                if ( auto d = x->tryAs<declaration::Expression>() )
+                    s += util::fmt(" (type: %s)", d->expression().type());
```
rsmmr, line 96:
```suggestion
            if ( x ) {
                if ( auto d = x->tryAs<declaration::Expression>() )
                    s += util::fmt(" (type: %s)", d->expression().type());
            }
```

hilti/toolchain/src/compiler/unit.cc
```diff
@@ -316,7 +316,8 @@ Result<Nothing> Unit::compile() {
 Result<Nothing> Unit::codegen() {
     auto& module = imported(_id);
 
-    _dumpASTs(logging::debug::AstCodegen, "AST for codegen");
+    // We do not include scopes for the final AST as we do not keep their references up to date.
+    _dumpASTs(logging::debug::AstCodegen, "AST for codegen", 0, false);
```
rsmmr, line 320: I'm wondering if instead we should clear out the scopes, then there's no risk of anybody accidentally trying to do something with that stale information (). Could make `Unit::_resetNodes` a public function for that and call it just before codegen. (Then we also don't need the changes to `dumpAST`).

hilti/toolchain/src/compiler/visitors/operator-resolver.cc
```diff
@@ -222,7 +222,7 @@ struct Visitor : public hilti::visitor::PostOrder<void, Visitor> {
                 return std::get<Type>(c.operands()[1].type).template as<type::Member>() == member;
             });
 
-            resolved = _resolve(candidates, u.operands(), u.meta());
+            resolved = _resolve(filtered, u.operands(), u.meta());
```
rsmmr, line 225: Oh, good catch.

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -142,31 +219,58 @@ struct Visitor : hilti::visitor::PreOrder<bool, Visitor> {
...
-                        module_id = module->get().id();
-                }
-
-                auto struct_id = x.id().sub(-2);
-                auto field_id = x.id().sub(-1);
+                // Record this hook as declared if it is not already known.
+                auto& function = (*_functions)[function_id];
+                function.declared = true;
+
+                for ( auto&& fn : x.childsOfType<Function>() ) {
```
rsmmr, line 234: There should be one function at most, and that one you get with `function()`, no?

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -212,12 +325,72 @@ struct Visitor : hilti::visitor::PreOrder<bool, Visitor> {
...
 
-    GlobalOptimizer::Hooks* _hooks = nullptr;
+    result_t operator()(const operator_::function::Call& call, position_t p) {
+        if ( ! call.hasOp0() )
+            return false;
+
+        auto id = call.op0().tryAs<expression::ResolvedID>();
+
+        if ( ! id )
+            return false;
```
rsmmr, line 344: This shouldn't ever happen right? let's trigger an internal error then to make sure we catch it if it does.

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -212,12 +325,72 @@ struct Visitor : hilti::visitor::PreOrder<bool, Visitor> {
...
+
+        auto module_id = id->id().sub(-2);
+        auto fn_id = id->id().sub(-1);
+
+        if ( module_id.empty() ) {
+            // Functions declared in this module do not include a module name in their ID.
+            if ( auto module = p.findParent<Module>() )
+                module_id = module->get().id();
+        }
```
rsmmr, line 353: This works, though I've been doing elsewhere is a have a visitor method for `Module` and just save the `module_id` for later use.

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -37,35 +42,86 @@ std::optional<std::pair<ModuleID, StructID>> typeID(T&& x) {
     return {{id->sub(-2), id->sub(-1)}};
 }
 
+// Helper function to determine whether a hook should be stored.
```
rsmmr, line 45: Looking at where this is used, I can't quite tell what situation this is catching: in what situation's is it not "good"?

hilti/toolchain/include/compiler/global-optimizer.h
```diff
@@ -21,12 +21,13 @@ struct GlobalOptimizer {
...
 
     struct Uses {
+        bool hook = false;
         bool declared = false;
         bool defined = false;
         bool referenced = false;
     };
 
-    using Hooks = std::map<std::tuple<ModuleID, StructID, FieldID>, Uses>;
+    using Functions = std::map<std::tuple<ModuleID, StructID, FieldID>, Uses>;
```
rsmmr, line 30: This is ok, but as food for thought: I'm back to thinking that my earlier suggestion of using fully qualified/normalized IDs instead of tuples might work well to capture both hooks and functions, without then having to special-case whether the struct ID is available or not. (But fine leaving as is.)

hilti/toolchain/include/ast/module.h
```diff
@@ -31,7 +31,7 @@ class Module : public NodeBase {
     const auto& statements() const { return child<statement::Block>(1); }
     const auto& declarations() const {
         if ( _cache.declarations.empty() )
-            _cache.declarations = childs<Declaration>(2, -1);
+            _cache.declarations = childsOfType<Declaration>();
```
rsmmr, line 34: What's this change needed for?

bbannier: When prune `declaration::Function`s we replace them with `None` nodes. With `childs<Declaration>(2, -1)` we assume that all nodes in `[2, -1]` are `Declaration`s and abort otherwise; the modified form just filters these nodes out.

hilti/toolchain/src/compiler/global-optimizer.cc
bbannier: You are right, this never returns `false` anymore. Removing it.

hilti/toolchain/include/compiler/global-optimizer.h
bbannier: I am leaving this as is for now.

hilti/toolchain/src/compiler/global-optimizer.cc
bbannier: Let's leave this as is so the visitor methods are mostly concerned with extracting information or performing edits. If fear that performing additional work which will introduce ordering dependencies will make both the extraction and the uses harder to understand.

hilti/toolchain/src/compiler/unit.cc
bbannier: I tried that, but simply calling `_resetNodes` seems to do too much. I'd prefer to leave this as is for now.

bbannier: I looked into this again and was able to just reset the scopes. This needed a small fix in codegen where we were retrieving a decl which could get optimized away, but never used it.

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -142,31 +207,55 @@ struct Visitor : hilti::visitor::PreOrder<bool, Visitor> {
...
                 // Nothing.
                 break;
+
+            case Stage::PRUNE_DECLS:
+                const auto& function = _functions->at(function_id);
+
+                if ( function.hook && ! function.defined ) {
+                    if ( ! struct_id.empty() ) {
+                        HILTI_DEBUG(logging::debug::GlobalOptimizer,
+                                    util::fmt("removing declaration for unused function %s::%s::%s", module_id,
```
rsmmr, line 246:
```suggestion
                                    util::fmt("removing declaration for unused hook function %s::%s::%s", module_id,
```

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -142,31 +207,55 @@ struct Visitor : hilti::visitor::PreOrder<bool, Visitor> {
...
+                                              struct_id, field_id));
+                    }
+                    else {
+                        HILTI_DEBUG(logging::debug::GlobalOptimizer,
+                                    util::fmt("removing declaration for unused function %s::%s", module_id, field_id));
```
rsmmr, line 251:
```suggestion
                                    util::fmt("removing declaration for unused hook function %s::%s", module_id, field_id));
```

rsmmr (approved):
This looks good now. There's a bit of nastiness in `function_identifier`, but that's unavoidable right now and we can clean that up later when the AST got more information. The rest looks pretty natural to me now.

PR #950, opened 2021-07-06
Doc fix and submodule bump

rsmmr:


bbannier (dismissed):

bbannier:
Regarding the failure of `protocols.wireguard.wireguard` on freebsd-11, I tried figuring out what could cause it. I was not able to reproduce it within a Vagrant freebsd-11.4 image so there seems to be something special about our CI environment. The test also doesn't look that different to other tests to me.

At this point I'd suggest we mark the test `@TEST-KNOWN-FAILURE` and create a ticket in zeek/spicy-analyzers to track this.

PR #951, opened 2021-07-07
Perform optimizations in a loop.

bbannier:
This patch changes the optimizer to now run a collect/prune loop until
no further AST changes are made. This allows us to remove more code as
we can now remove code called by unused code if it is also unused. The
changes made here also allow other future optimizations (e.g., emptying
out structs to ultimately remove them as well).

For some reason `scope::lookupID` fails to find declarations for
imported modules after a first edit pass which leads to incorrect
function identifiers; we work around this by explicitly inspecting the
imports of the current module.


hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -71,7 +71,12 @@ struct Visitor : hilti::visitor::PreOrder<bool, Visitor> {
         // If the namespace is a single component (i.e., has no namespace itself) we are either dealing
         // with a global function in another module, or a function for a struct in the current module.
         if ( ns_ns.empty() ) {
-            if ( scope::lookupID<declaration::ImportedModule>(ns_local, p, "module").hasValue() )
+            const auto imports = current_module->get().childsOfType<declaration::ImportedModule>();
+            if ( std::any_of(imports.begin(), imports.end(), [&](const auto& imported_module) {
+                     if ( auto&& m = imported_module.module() )
```
rsmmr, line 76:
```suggestion
                     if ( const auto& m = imported_module.module() )
```

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -413,20 +424,30 @@ void GlobalOptimizer::run() {
...
+        for ( auto& unit : units )
+            modified = modified || Visitor(&_hooks).prune_uses(*unit);
+
+        for ( auto& unit : units )
+            modified = modified || Visitor(&_hooks).prune_decls(*unit);
+
+        if ( ! modified )
+            break;
+
+        // After pruning the scopes might contain e.g., references
```
rsmmr, line 442: Seems there's something inconsistent here about the scopes. Clearing them at the end of the loop body means that subsequent iterations will have the scopes emptied, vs. the first iteration which had them still filled. I'd say that either the collection/pruning cycle needs the scopes, then subsequent iterations now have a problem; or they don't, then we could just clear the scopes before we enter the loop.

rsmmr (approved):

PR #952, opened 2021-07-08
Work around for Cirrus Homebrew issue.

bbannier:
The osx machines we currently use in Cirrus CI seem to be affected by
cirruslabs/cirrus-ci-docs#878. In order to work around this this patch
adds a manual Homebrew upgrade before installing any packages. As this
takes time we should probably remove this step once upstream has merged
a fix.


rsmmr:
This is superseded by #950.

PR #964, opened 2021-07-27
Fix issues around generating version information.

### LINKED ISSUE #956: Review version generation
Let's review our we generate version numbers. I keep seeing inconsistencies. 

Most recent example: `brew install spicy —HEAD` installs a `spicyc` that reports itself as `spicyc v1.1.0-branch (701e5137)`, even though that hash corresponds to `1.2.0-dev.87`. Also, sometimes I end up having `VERSION` files even when working from git (which, iirc, then take precendence over git).

Also, a fresh Spicy `main` clone shows in the configure output:

```
====================|  Spicy Build Summary  |====================

Version:               1.2.0
[...]
```

We should then also update minimum Spicy version numbers for the plugin and analyzer repos.

bbannier:
Closes #956.


bbannier:
This is a cleanup related to #956.

PR #968, opened 2021-08-03
Implement removal of unused functions in global optimizer.

bbannier:


hilti/toolchain/include/ast/types/struct.h
```diff
@@ -159,14 +159,6 @@ class Struct : public TypeBase, trait::isAllocable, trait::isParameterized, trai
         return types;
     }
 
-    auto ids() const {
```
rsmmr, line 162: Is this simply unused? (ok to remove then)

hilti/toolchain/include/compiler/global-optimizer.h
```diff
@@ -27,7 +28,9 @@ struct GlobalOptimizer {
         bool referenced = false;
     };
 
-    using Functions = std::map<std::tuple<ModuleID, StructID, FieldID>, Uses>;
+    enum class Type { Method, Function };
+    using ID = std::pair<Type, std::string>;
```
rsmmr: Could the `std:string` be an `ID`?

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -23,10 +24,15 @@
...
 
+std::ostream& operator<<(std::ostream& stream, const GlobalOptimizer::ID& x) { return stream << x.second; }
+
 using ModuleID = GlobalOptimizer::ModuleID;
 using StructID = GlobalOptimizer::StructID;
 using FieldID = GlobalOptimizer::FieldID;
 
+using position_t =
+    visitor::Position<hilti::detail::visitor::Iterator<hilti::Node, hilti::detail::visitor::Order::Pre, false>::E>;
+
```
rsmmr, line 35: Can we avoid this typedef and use the `position_t` as it's available within `Visitor`?

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -87,10 +93,68 @@ struct Visitor : hilti::visitor::PreOrder<bool, Visitor> {
...
     }
 
+    static std::optional<GlobalOptimizer::ID> getID(const type::struct_::Field& x, position_t p) {
+        auto field_id = x.id();
+
+        auto struct_type = typeID(p.parent().as<type::Struct>());
+        if ( ! struct_type )
+            return {};
+
+        auto&& [module_id, struct_id] = *struct_type;
```
rsmmr, line 114: Please use `const auto&` instead of `auto&&` wherever possible (more places below).

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -48,6 +48,17 @@ std::optional<std::pair<ModuleID, StructID>> typeID(T&& x) {
...
 }
 
+// Helper function to detect whether a given type is a filter.
+bool isFilter(const declaration::Type& type) {
+    const auto unit = type.type().as<type::Struct>();
+
+    // By the point we reach the optimizer Spicy unit properties are
+    // not present anymore, so we detect this with a heuristic.
+    //
+    // TODO(bbannier): consider explicitly expressing this information in the AST.
```
rsmmr, line 58: We are using a couple of HILTI-internal attributes already, could add one like `&no-optimize` that Spicy would add to filter types?

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -289,25 +299,35 @@ struct Visitor : hilti::visitor::PreOrder<bool, Visitor> {
 
                 switch ( fn.callingConvention() ) {
                     case function::CallingConvention::ExternNoSuspend:
-                    case function::CallingConvention::Extern:
-                        // If the declaration is `extern` it is part of an externally
-                        // visible API and potentially used elsewhere.
-                        function.referenced = true;
+                    case function::CallingConvention::Extern: {
+                        // If the declaration is `extern` and the unit is `public` or a filter, the function
+                        // is part of an externally visible API and potentially used elsewhere.
```
rsmmr, line 304: I haven't quite understood yet why filter types are special: HILTI doesn't know anything about them, so I would think that normal visibility rules apply. Where's a filter function used that goes around that?

hilti/toolchain/include/ast/types/struct.h
bbannier: Yes, this is not used. I'll move this to its own commit.

hilti/toolchain/include/compiler/global-optimizer.h
bbannier: Yes. The original form in this patch was for some larger set of changes to also track types; I'll add that back if and once needed.

hilti/toolchain/src/compiler/global-optimizer.cc
bbannier: Filters are not in general `public`, so with how this patch is structured their parse functions are candidates for removal if we see no usages. The uses of these functions are typically in code generated for `connect_filter`, but not visible on the AST level.

The right approach is probably to hardcode a whitelist of functions which make use of certain unit member functions (e.g., into the visitor for `MemberCall`), and then explicitly mark the unit member functions as used from there. I'll give that a try.

hilti/toolchain/src/compiler/global-optimizer.cc
bbannier: The approach in this patch is to progressively empty out units so that we can remove empty units in a final pass (that part is not yet implemented), so we'd have to prevent removal of unit member functions with `&no-optimize`; adding that for all filter functions seems very manual. I'll instead try to make the optimizer smarter about what unit member functions are used, see my comment below.

hilti/toolchain/src/compiler/global-optimizer.cc
bbannier: That was a leftover from a version where I needed a `position_t` outside of the visitor, removing it now.

hilti/toolchain/src/compiler/global-optimizer.cc
bbannier: I instead went with the approach discussed offline, namely explicitly marking parse functions of filters as `&always-emit`.

rsmmr (approved):
I like this, looks good now.

PR #971, opened 2021-08-05
Add optimizer pass to remove unused types.

bbannier:


rsmmr (dismissed):
Looks good and pretty straightforward, nice!

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -441,6 +441,121 @@ struct FunctionVisitor : visitor::PreOrder<bool, FunctionVisitor> {
...
+    result_t operator()(const declaration::Type& x, position_t p) {
+        const auto type_id = x.typeID();
+
+        if ( ! type_id )
+            return false;
+
+        switch ( _stage ) {
+            case Stage::COLLECT:
+                // Record the type if not already known. If the type is part of an external API record it as used.
+                _used.insert({*type_id, x.linkage() == declaration::Linkage::Public});
```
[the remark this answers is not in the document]
bbannier: Yeah, this is different since here we only want to add an entry (record a use derived from the type visibility) if non previously exists. The other visitors always record a use. With that actual _uses of a type_ override whatever we derived from just _declarations of the type_.

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -439,6 +441,121 @@ struct Visitor : hilti::visitor::PreOrder<bool, Visitor> {
...
+                break;
+
+            case Stage::PRUNE_USES: break;
+            case Stage::PRUNE_DECLS:
+                if ( ! _used.at(*type_id) ) {
+                    HILTI_DEBUG(logging::debug::GlobalOptimizer, util::fmt("removing unused type '%s'", *type_id));
```
bbannier, line 483: Something is flaky here with the test `spicy.optimizations.unused-types`. I sometimes see `Priv2` and `Priv3` not being optimized out.

bbannier: We should `return true` from this conditional, otherwise the edits will not converge to the same final state.

PR #974, opened 2021-08-11
Add optimizer pass removing unneeded %random-access functionality.

bbannier:


tests/Baseline/spicy.optimization.default-parser-functions/log
```diff
@@ -1,4 +1,21 @@
...
+[debug/global-optimizer] disabling feature 'uses_random_access' of type 'foo::P2' since it is not used
+[debug/global-optimizer] inlining constant 'foo::__feat%foo__P1%supports_filters' at ''
+[debug/global-optimizer] inlining constant 'foo::__feat%foo__P1%supports_filters' at ''
+[debug/global-optimizer] inlining constant 'foo::__feat%foo__P1%supports_filters' at ''
+[debug/global-optimizer] inlining constant 'foo::__feat%foo__P2%supports_filters' at ''
+[debug/global-optimizer] inlining constant 'foo::__feat%foo__P2%supports_filters' at ''
+[debug/global-optimizer] inlining constant 'foo::__feat%foo__P2%supports_filters' at ''
+[debug/global-optimizer] removing conditional code '' which can never be reached
+[debug/global-optimizer] removing conditional code '' which can never be reached
+[debug/global-optimizer] removing conditional code around '' since the condition is always true
```
bbannier, line 13: We should add some dummy location to the conditional, feature-dependent code we generate in the parser builder so we can log something meaningful here.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -1828,3 +1840,27 @@ void ParserBuilder::finishLoopBody(const Expression& cookie, const Location& l)
...
 }
+
+void ParserBuilder::feature(const type::Unit& unit, std::string_view feature, std::function<void()> f) {
+    // FIXME(bbannier): Only handling for the `is_filter` and
+    // `supports_filters` feature flags is currently implemented.
+    if ( feature != "is_filter" && feature != "supports_filters" ) {
+        f();
+        return;
+    }
+
```
bbannier, line 1851: Let's also annotate users of other feature-dependent code so we can emit conditional code for all of them.

tests/Baseline/spicy.optimization.unused-functions/log
```diff
@@ -1,4 +1,22 @@
...
+[debug/global-optimizer] inlining constant 'foo::__feat%foo__F%is_filter' at ''
+[debug/global-optimizer] inlining constant 'foo::__feat%foo__F%is_filter' at ''
+[debug/global-optimizer] removing conditional code around '' since the condition is always true
+[debug/global-optimizer] removing conditional code around '' since the condition is always true
+[debug/global-optimizer] removing conditional code around '' since the condition is always true
+[debug/global-optimizer] removing conditional code around '' since the condition is always true
+[debug/global-optimizer] removing conditional code around '' since the condition is always true
+[debug/global-optimizer] removing conditional code around '' since the condition is always true
+[debug/global-optimizer] removing conditional code around '' since the condition is always true
+[debug/global-optimizer] removing conditional code around '' since the condition is always true
```
bbannier, line 19: There's a lot of churn in these logs since we always run all optimizer passes. This makes it hard to see which pass actually lead to which behavior.

Since this PR introduces a baseclass for optimizer passes adding a way to explicitly enable certain passes from the outside shouldn't be too hard. For that we could e.g., look at an environment variable `HILTI_OPTIMIZER_PASSES` containing a comma-separated list of passes; if the env var is set we'd only enable to given passes.

spicy/lib/spicy_rt.hlt
```diff
@@ -33,11 +33,11 @@ public type HiltiResumable = __library_type("hilti::rt::Resumable");
...
-declare public void filter_disconnect(inout any unit) &cxxname="spicy::rt::filter::disconnect" &have_prototype;
-declare public void filter_forward(inout any filter, bytes b) &cxxname="spicy::rt::filter::forward" &have_prototype;
-declare public void filter_forward_eod(inout any filter) &cxxname="spicy::rt::filter::forward_eod" &have_prototype;
+declare public strong_ref<stream> filter_init(inout any unit &requires_type="supports_filters", inout value_ref<stream> data, view<stream> cur) &cxxname="spicy::rt::filter::init" &have_prototype;
+declare public void filter_connect(inout any unit &requires_type="supports_filters", strong_ref<*> filter &requires_type="is_filter") &cxxname="spicy::rt::filter::connect" &have_prototype;
+declare public void filter_disconnect(inout any unit &requires_type="supports_filters") &cxxname="spicy::rt::filter::disconnect" &have_prototype;
+declare public void filter_forward(inout any filter &requires_type="is_filter", bytes b) &cxxname="spicy::rt::filter::forward" &have_prototype;
+declare public void filter_forward_eod(inout any filter &requires_type="is_filter") &cxxname="spicy::rt::filter::forward_eod" &have_prototype;
```
bbannier, line 40: We should annotate additional function declarations so we can remove the workaround we added in 6160ef603bfa285dccc4ca52bad2d9e422cd1078 which leads us to always emit parser functions for filters, even if unused. Above functions are not sufficient for that.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
bbannier: I simplified this PR to only work on `%random-access` features, and will handle the other features in a follow-up PR.

spicy/lib/spicy_rt.hlt
bbannier: I'll move filter-related work to a follow-up PR.

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -487,8 +487,17 @@ struct TypeVisitor : OptimizerVisitor, visitor::PreOrder<bool, TypeVisitor> {
...
+
                     removeNode(p);
 
+                    if ( auto module_ = p.findParent<Module>() )
+                        // If this type was declared under a top-level module also clear the module declaration
+                        // cache. The cache will get repopulated the next time the module's declarations are
+                        // requested.
+                        //
+                        // TODO(bbannier): there has to be a better way to mutate the module.
+                        const_cast<Module&>(module_->get()).clearCache();
```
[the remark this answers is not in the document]
rsmmr: Another thought: you don't neccessarily need to restrict this to modules and could create a helper functions similar to `Unit::_resetNodes` that just generally clears scopes & caches for an AST; and then call that from different places.

But fine to leave as is as well, as the need for this will go away anyways with the new AST code.

hilti/toolchain/src/compiler/driver.cc
```diff
@@ -682,14 +666,29 @@ Result<Nothing> Driver::run() {
...
+        HILTI_DEBUG(logging::debug::Driver, "performing global transformations");
 
-    HILTI_DEBUG(logging::debug::Driver, "performing global transformations");
+        GlobalOptimizer opt(&_hlts, _ctx);
 
-    GlobalOptimizer opt(&_hlts, _ctx);
+        opt.run();
+    }
+
+    if ( _driver_options.output_hilti ) {
```
rsmmr, line 677: Not sure I like having the printing inside `transformUnits()`. How about moving it up one level to between the calls to `transformUnits()` and `codegenUnits()`?

tests/Baseline/hilti.optimization.unimplemented_hook/output.diff
```diff
@@ -1,168 +0,0 @@
-### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
```
rsmmr, line 1: Does this file need to be deleted? The test doesn't generate it anymore if I'm seeing that right (same presumably for the other tests).

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -594,25 +598,28 @@ void GlobalOptimizer::run() {
...
+    const auto passes__ = rt::getenv("HILTI_OPTIMIZER_PASSES");
+    const auto passes_ =
+        passes__ ? std::optional(util::split(*passes__, ":")) : std::optional<std::vector<std::string>>();
+    const auto passes = passes_ ? std::optional(std::unordered_set<std::string>(passes_->begin(), passes_->end())) :
+                                  std::optional<std::unordered_set<std::string>>();
+
+    const std::unordered_map<std::string, std::function<std::unique_ptr<OptimizerVisitor>()>> creators =
+        {{"functions", []() { return std::make_unique<FunctionVisitor>(); }},
+         {"types", []() { return std::make_unique<TypeVisitor>(); }}};
+
```
rsmmr: This block I find quite hard to read, including the `if ( passes )` section further below. Maybe start with a default `std::vector<std::unique_ptr<OptimizerVisitor>> vs;` that then gets updated if `HILTI_OPTIMIZER_PASSES` is set?

tests/Baseline/hilti.optimization.const/opt.hlt
```diff
@@ -0,0 +1,23 @@
...
+
+hilti::print(True, True);
+hilti::print(False, True);
+
+if ( True ) {
+    0;
+}
+else {
+    1;
+}
```
rsmmr, line 17: This don't look optimized (just the constant is inlined)?

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -623,6 +630,118 @@ struct TypeVisitor : OptimizerVisitor, visitor::PreOrder<bool, TypeVisitor> {
     }
 };
 
+struct ConstantFoldingVisitor : OptimizerVisitor, visitor::PreOrder<bool, ConstantFoldingVisitor> {
+    std::map<uint64_t, bool> _constants;
+
+    void collect(Node& node) override {
```
rsmmr, line 645: Could we move the collect/prune methods into the base class, or are just diverting too much between different optimizer visitors?

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -623,6 +630,118 @@ struct TypeVisitor : OptimizerVisitor, visitor::PreOrder<bool, TypeVisitor> {
...
+            return false;
+
+        switch ( _stage ) {
+            case Stage::COLLECT: {
+                const auto& init = x.init();
+                assert(init);
+
+                if ( auto ctor = init.value().tryAs<expression::Ctor>() )
+                    if ( auto bool_ = ctor->ctor().tryAs<ctor::Bool>() )
+                        _constants[p.node.rid()] = bool_->value();
```
rsmmr, line 685: TODO for the future: Using `rid()` as the index works for now I believe, but it's not very robust as it depends on the the AST not changing. With the new AST code, this can become the canonical ID of the global.

spicy/toolchain/include/compiler/detail/codegen/parser-builder.h
```diff
@@ -454,6 +454,8 @@ class ParserBuilder {
      */
     void finishLoopBody(const Expression& cookie, const Location& l);
 
+    void feature(const type::Unit& unit, std::string_view feature, std::function<void()> f);
```
rsmmr, line 457: I like this API for putting code behind a feature flag. Would suggest to rename it though so that it conveys that `f` is building code behind that feature flag. `guardFeatureCode`? `addCodeForFeature`? `featureBlock`?

Also add Doxygen comment.

tests/Baseline/spicy.optimization.default-parser-functions/noopt.hlt
```diff
@@ -61,6 +61,12 @@ public type P2 = struct {
     method tuple<view<stream>, int<64>, iterator<stream>> __parse_foo_P2_stage2(inout value_ref<stream> __data, copy view<stream> __cur, copy bool __trim, copy int<64> __lah, copy iterator<stream> __lahe);
 } &on-heap;
 
+global auto __feat%foo__P1%supports_filters = True;
```
rsmmr, line 64: The `%` inside the IDs feel a bit funky. :) Just use underscores instead?

spicy/toolchain/src/compiler/codegen/unit-builder.cc
```diff
@@ -201,13 +201,15 @@ Type CodeGen::compileUnit(const type::Unit& unit, bool declare_only) {
     }
 
     if ( unit.usesRandomAccess() ) {
+        auto attr_random_access = Attribute("&requires_type", builder::string("uses_random_access"));
```
rsmmr, line 204: Is `&requires_type` the right name here? The previous patch calls these "features", so would `&requires_feature` be more consistent?

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -753,6 +753,155 @@ struct ConstantFoldingVisitor : OptimizerVisitor, visitor::PreOrder<bool, Consta
...
+        // We only work on feature flags named `__feat*`.
+        if ( ! util::startsWith(id, "__feat") )
+            return;
+
+        const auto& tokens = util::split(id, "%");
+        assert(tokens.size() == 3);
+
+        // The type name is encoded in the variable name with `::` replaced by `__`.
+        const auto& typeID = ID(util::replace(tokens[1], "__", "::"));
+        const auto& feature = tokens[2];
```
rsmmr, line 791: I'm wondering if we, instead of using globals for the feature flags, we should introduce a new HILTI-side declaration `%feature ...` that makes it all more explicit. That would then also simplify the code here  to not have to parse things out the ID name (the feature declaration could store the information as part of its AST node).

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -753,6 +753,155 @@ struct ConstantFoldingVisitor : OptimizerVisitor, visitor::PreOrder<bool, Consta
...
+                if ( ! field )
+                    return;
+
+                // Compute a list of features this use does not activate.
+                // Generated feature-dependent code is always under
+                // conditionals `if (__feat%XYZ%FEATURE) ...` so get all
+                // `FEATURE` for all parents which match this pattern.
+                const std::vector<std::string> ignored_features =
+                    util::transform(util::filter(util::transform(p.path,
+                                                                 [](const visitor::Location<Node&>& n)
```
rsmmr, line 852: Would factor this out into a standalone function, gets pretty deeply nested here.

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -753,6 +753,155 @@ struct ConstantFoldingVisitor : OptimizerVisitor, visitor::PreOrder<bool, Consta
...
+                    auto new_x = declaration::GlobalVariable::setInit(x, builder::bool_(false));
+                    replaceNode(p, new_x);
+                }
+
+                break;
+            }
+        }
+    }
+
+    template<typename MemberAccess>
```
rsmmr, line 817: Does this need to be templated ? (vs either passing in a `ResolvedOperator` or just the specific pieces directly that are needed).

rsmmr (commented):
- "Change optimizer tests to compare HILTI code instead of ASTs": This is for sure better than the previous way, I like it. It would stlll be even nicer, though, to directly see a diff between optimized & non-optimized code. I'm wondering if we can find a format for the diff that doesn't confuse GitHub to mess up the colors, maybe just use `sed` to change diff markers into something else. Then we could go back to committing the diff instead of the two outputs.

- The code coming out of the feature-tracking logic looks nice, both before and after optimization. It makes it pretty clear what's going on there.

hilti/toolchain/src/compiler/global-optimizer.cc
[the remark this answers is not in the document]
bbannier: Unfortunately this only works if we have the position of the module which is not the case here (but works for the other `const_cast` below). Here we call `findParent` to get the `Module` which always returns an optional ref of `const T`. :/

tests/Baseline/hilti.optimization.unimplemented_hook/output.diff
bbannier: I am confused 😅. AFAICT the file _is removed_.

tests/Baseline/spicy.optimization.default-parser-functions/noopt.hlt
bbannier: I went with `%` instead of `_` since we already encode the `::` in type names as `__`; using the same token to separate different components makes extraction (which we still do in too many places) slightly tricky. I'd prefer to leave this as is until we possibly replace these global constants with dedicated `Feature` AST types later like you suggested.

spicy/toolchain/src/compiler/codegen/unit-builder.cc
bbannier: Went with `&requires_type_feature` as discussed offline.

hilti/toolchain/src/compiler/global-optimizer.cc
bbannier: `ResolvedOperator` seems to work fine.

hilti/toolchain/src/compiler/global-optimizer.cc
bbannier: Yeah, I went over board here trying to do `filter_map`/`flatMap` 😄 . Since we would allocate for tons of intermediate `vector`s here, and I'd really prefer a `set` instead, moving this to a function shouldn't just improve readability, but also performance.

tests/Baseline/hilti.optimization.const/opt.hlt
bbannier: Like discussed offline, this is related to the patch you reviewed being unable to remove dead `else` branches (there was a `TODO` for that in the code). I now added that ability and this block gets optimized to a single block `{ 0; }`.

hilti/toolchain/src/compiler/global-optimizer.cc
bbannier: I thought so too initially, but this is tricky. These methods invoke `walk` which comes directly from `Visitor` and can only invoked from something which knows all the matchers the final visitor has. Using straight-forward CRTP for `OptimizerVisitor` makes it impossible to have a container of different visitors below.

I'd prefer to leave this as is for now and possible clean this up later.

hilti/toolchain/src/compiler/global-optimizer.cc
bbannier: Let's leave this for later (again).

hilti/toolchain/src/compiler/global-optimizer.cc
bbannier: Yes, I agree this is pretty verbose. I needed to introduce the additional variables `passes__` and `passes_` since `std::optional` doesn't provide monadic bind (e.g., a `and_then` which allows applying a function over `T` on an `optional<T>`). In addition to unpacking the env var these lines also perform deduplication which could be important.

I simplified the code slightly by ensuring that `passes` is always set, either by the env var, or by overriding with a default set. This is still not trivial as we want to execute the `feature_requirements` visitor differently than other visitors while still allowing to toggle it. Let me know what you think.

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -554,10 +613,297 @@ struct TypeVisitor : visitor::PreOrder<bool, TypeVisitor> {
...
+                if ( ! struct_ )
+                    return;
+
+                auto field = struct_->field(member->id());
+                if ( ! field )
+                    return;
+
+                const auto ignored_features = conditionalFeatures(p);
+
+                for ( const auto& requirement : AttributeSet::findAll(field->attributes(), "&requires_type_feature") ) {
```
bbannier, line 887: Should be `&requires-type-feature` here and elsewhere.

rsmmr (approved):
It's tricky to track what's changed, but I'll say "LGTM!". :-)

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -579,42 +925,67 @@ void GlobalOptimizer::run() {
...
+        for ( const auto& [pass, _] : creators )
+            passes->insert(pass);
+    }
+
     while ( true ) {
         bool modified = false;
 
-        FunctionVisitor function_visitor;
-        TypeVisitor type_visitor;
+        // NOTE: We do not use `util::transform` here to guarantee a consisted order of the visitors.
```
bbannier, line 961: Typo `s/consisted/consistent/`

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -579,42 +925,67 @@ void GlobalOptimizer::run() {
...
+        std::vector<std::unique_ptr<OptimizerVisitor>> vs;
+        vs.reserve(passes->size());
+        for ( const auto& pass : *passes )
+            vs.push_back(creators.at(pass)());
```
bbannier, line 965: As `passes` might come from the user we need to check whether a certain `pass` is known to `creators` here.

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -554,10 +613,297 @@ struct TypeVisitor : visitor::PreOrder<bool, TypeVisitor> {
...
+                    replaceNode(p, new_x);
+                }
+
+                break;
+            }
+        }
+    }
+
+    // Helper function to compute the set of feature flags wrapping the given position.
+    static std::set<std::string> conditionalFeatures(position_t p) {
```
bbannier, line 822: This must return a per-type map.

hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -554,10 +613,300 @@ struct TypeVisitor : visitor::PreOrder<bool, TypeVisitor> {
...
+            case Stage::COLLECT:
+            case Stage::PRUNE_DECLS: return false;
+            case Stage::PRUNE_USES: {
+                auto rid = x.rid();
+
+                if ( const auto& constant = _constants.find(rid); constant != _constants.end() ) {
+                    if ( x.type() == type::Bool() ) {
+                        HILTI_DEBUG(logging::debug::GlobalOptimizer, util::fmt("inlining constant '%s'", x.id()));
+
+                        replaceNode(p, builder::bool_(constant->second));
```
bbannier, line 710: If the `GlobalVariable` is under a `Module` after this replacement its old value will be cached in the `Module`'s `declarations`; we should clear that cache after this edit to make sure we see consistent results.

PR #977, opened 2021-08-25
Optimize out unused filter and sink functionality

bbannier:


hilti/toolchain/src/compiler/global-optimizer.cc
```diff
@@ -826,6 +826,143 @@ struct FeatureRequirementsVisitor : visitor::PreOrder<void, FeatureRequirementsV
         }
     }
 
+    void operator()(const operator_::function::Call& x, position_t p) {
```
bbannier, line 829: As per @rsmmr's comment [here](https://github.com/zeek/spicy/pull/977#pullrequestreview-740204794), it would be nice to emit comments in the generated code indicating which features are enabled for a particular unit.

For that we need to actually emit the comments associated with a node in its `Meta::Comments` (fine to only add that for `declaration::Type`s for now), and edit the comments here in the `TRANSFORM` phase of this visitor. We currently cannot replace these nodes since they are potentially referenced via `NodeRef`s in e.g., the `Module`'s `Scope`, but this should be possible to do once @rsmmr AST rework has landed.

bbannier: Let's move this to a follow-up, #981.

rsmmr (approved):

Issue #982, opened 2021-08-30
Model feature flags with actual constants


bbannier:
For Spicy units we emit feature constants `__feat%%XYZ%%abc` as `GlobalVariable`s. We use `GlobalVariable`s so we can update their `init` expression; we currently cannot `Constant`s here since we cannot really edit their values as their parent `Module` holds a list of `NodeRef`s to its declarations, and performing such edits (which currently means replacing the node) isn't really possible without leaving outdated or dangling `NodeRef`s around.

Once #984 has landed we should be able to edit nodes in place so the issue of orphaned `NodeRef`s will go away. Once that work has landed we should replace these feature globals with proper `Constant`s and allow more constants to be inlined.

PR #983, opened 2021-08-30
Disable constant folding for anything but feature variables.

bbannier:
We previously would inline init expressions for global variables, even
though the values of globals could have later been updated. Ideally
these feature flags should be constants so we know they cannot be
updated on the script side, but this is currently not straightforward,
see #982.

This patch disables constant folding for anything but our feature flags
which we control and from which we know that they cannot be updated at
runtime (their IDs cannot be used in userland Spicy scripts). As a
workaround for testing we also treat names starting with `feat________`
as feature flags; uses of globals with such names can still show the
problematic runtime behavior. We should rewrite our feature flags as
`Constants` in a follow-up patch.


rsmmr (approved):
This is ok given that we want to clean this up anyways, but also: we do have "real" constants as well, could use `declaration::Constant` instead of `declaration::GlobalVariable`.

PR #984, opened 2021-08-31
Major overhaul of the AST infrastructure

rsmmr:
High-level summary of the most important changes:

- Restructure the main AST pass loop. Instead of running passes from all
  plugins interleaved at the same time, we now run plugins sequentially;
  specifically first Spicy, then HILTI. After each plugin's run, we have
  a fully resolved AST that's then transformed into the starting AST for
  the subsequent plugin.

  Note that the Spicy plugin still calls into the HILTI plugin for
  resolving its AST: that's because the two representations share a lot
  of semantics so that we can outsource a bunch of stuff that the Spicy
  AST needs to do to the HILTI plugin --- it's the same logic.

  As part of this, we regroup the visitors: the
  `{resolver,operator-resolver,id-resolver}.cc` are mostly merged into a
  single `resolver.cc`, plus a new `normalizer.cc` where some stuff
  moved that enforces some constraints on the AST before anything
  else. `visitors/apply-coercions.cc` moved to `visitors/coercer.cc`,
  and the former `coercer.cc` moved out of the visitors directory one
  level up into `coercion.cc` because it's not an AST pass.

- New AST structure that doesn't compute anything dynamically anymore;
  all information is stored in AST nodes. We now almost exclusively pass
  around references to nodes, no longer copies. We got some new helper
  classes for that: `hilti::optional_ref<T>`, `hilti::node::Set<T>`,
  `hilti::node::Range<T>`.

  This especially includes replacing all dynamically computed types with
  AST nodes, getting rid of `type::effectiveType()`.

  No more static setters, we modify nodes directly now. However, when
  using typed nodes, we pass them around as `const`. When one needs to
  modify a node, that therefore usually involves a cast:
  `p.node.as<T>.setXYZ()`. In a few cases, it needs a `const_cast` of a
  typed reference.

  None more (public) setters returning `Node` references. New methods
  with names ending in `*Ref` return `NodeRef` instead.

  For more on the new structure and conventions, also see the Wiki:
  https://github.com/zeek/spicy/wiki/Notes-on-the-new-internal-HILTI-Spicy-AST-structure

- Declarations now get globally unique "canonical IDs'.

- Struct fields are now declarations.

- Functions have a parent type that for methods will point to their
  struct/unit type.

- `typeID()` and `cxxID()` moved up a level to become properties of
  `Type` (vs. being properties of the encapsulated specific type
  instantiations).

- Using singletons for `type::Auto` and `type::Void`.

- Remove "original nodes" and also most uses of "preserved nodes".

- Remove AST-internal node caching.

- AST dumps include node addresses to track identity.

- We pass compiler `Unit`s around through `shared_ptr`, because they have a
  clear identity while being stored at various places.


hilti/toolchain/include/ast/node-collection.h
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    Set& operator=(const Set& other) = default;
+    Set& operator=(Set&& other) = default;
+
+private:
+    std::vector<std::reference_wrapper<const T>> _set;
+};
+
+/**
```
bbannier, line 273: Comment for all these helper functions: I'd much prefer if we'd make the existing functions in `util` generic enough that they can deal with arbitrary containers. This should probably be pretty straight-forward with return type computed with `declval` (and for `transform` additionally `result_of`). Insertion can be performed generically with `back_inserter`.

hilti/toolchain/src/ast/node.cc
```diff
@@ -68,21 +67,47 @@ std::string Node::render(bool include_location) const {
...
 }
 
 void Node::print(std::ostream& out, bool compact) const { detail::printAST(*this, out, compact); }
 
+std::string Node::print() const {
+    std::stringstream out;
+    detail::printAST(*this, out, true);
+    return out.str();
+}
+
```
bbannier, line 110: Not touched here, but below `operator+` seems to be unused.

hilti/toolchain/include/ast/node.h
```diff
@@ -296,16 +315,16 @@ class Node final : public node::detail::Node {
...
+    // `NodeRef` uses to maintain links to it.
+    IntrusivePtr<node_ref::detail::Control> _control() const {
         if ( ! _control_ptr )
             _control_ptr = make_intrusive<node_ref::detail::Control>(this);
 
         return _control_ptr;
     }
 
-    IntrusivePtr<node_ref::detail::Control> _control_ptr = nullptr;
+    mutable IntrusivePtr<node_ref::detail::Control> _control_ptr = nullptr;
```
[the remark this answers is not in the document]
rsmmr: Most nodes won't need a control block, this way avoids creating them unnecessarily.

hilti/toolchain/include/ast/node-collection.h
```diff
@@ -0,0 +1,334 @@
...
+    BaseIterator _iter;
+};
+
+/**
+ * A range of AST nodes, defined by start and end into an existing vector of
+ * nodes. The range creaates a view that can be iterated over, yielding a
+ * reference to each node in turn.
+ */
+template<typename T>
+class Range {
```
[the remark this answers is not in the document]
rsmmr: As we chatted about, yes: this is about storing a collection of references transparently just through their begin/end position in the underlying vector. We could skip this and use just `node::Set` but not sure about performance impact.

hilti/toolchain/include/ast/node-collection.h
```diff
@@ -0,0 +1,334 @@
...
+};
+
+/**
+ * A set of AST nodes. The set creates a view of nodes that can be iterated
+ * over, yielding a reference to each node in turn. In contrast to `Range`, a
+ * set can include nodes that are not all part of a continous slice inside a
+ * vector.
+ */
+template<typename T>
+class Set {
```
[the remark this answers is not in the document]
rsmmr: Hmm, not sure; it's a container for a very specific purpose where elements will normally be distinct automatically (returning a subset of a AST node's children). I think it's fine to call it a set.

hilti/toolchain/include/ast/node-collection.h
rsmmr: I'll create a ticket for this.

hilti/toolchain/include/ast/node-collection.h
```diff
@@ -0,0 +1,334 @@
...
+            return false;
+
+        auto x = begin();
+        auto y = other.begin();
+        while ( x != end() ) {
+            if ( ! (*x++ == *y++) )
+                return false;
+        }
+
+        return true;
```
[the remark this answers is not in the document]
rsmmr: I don't think so, the set stores references, but we're comparing by value.

hilti/toolchain/include/ast/declaration.h
```diff
@@ -65,16 +71,30 @@ inline bool operator==(const Declaration& x, const Declaration& y) {
...
+ * by all declarations.
+ */
+class DeclarationBase : public NodeBase, public hilti::trait::isDeclaration {
+public:
+    using NodeBase::NodeBase;
+
+    /** Implements the `Declaration` interface. */
+    const ID& canonicalID() const { return _id; }
+    /** Implements the `Declaration` interface. */
+    void setCanonicalID(ID id) { _id = std::move(id); }
```
bbannier, line 94: This is a dangerous interface as it makes it very easy to accidentally create canonical ID which are empty via a the default ctr; in fact I am seeing a number of cases where `canonicalID` returns an empty string which is not at all unique.

Should we maybe instead change `canonicalID` and `_id` to be `optional<ID>`s so users have at least a chance to detect completely unset canonical IDs? That still wouldn't solve the issue that the way these IDs are constructed makes it hard to ensure uniqueness though.

hilti/toolchain/src/compiler/visitors/normalizer.cc
```diff
@@ -0,0 +1,209 @@
...
+    void operator()(const type::Struct& t, position_t p) { type::Struct::setSelf(&p.node); }
+};
+
+} // anonymous namespace
+
+struct VisitorClearCanonicalIDs : public visitor::PreOrder<void, VisitorClearCanonicalIDs> {
+    result_t operator()(const Declaration& d, position_t p) { p.node.as<Declaration>().setCanonicalID(ID()); };
+};
+
+struct VisitorComputeCanonicalIDs : public visitor::PreOrder<ID, VisitorComputeCanonicalIDs> {
```
bbannier, line 151: Like discussed offline, this seems to compute different IDs for method definitions and their declarations, e.g., given a module

```
module bar;
public type X = unit {
    on %done {}
};
```

I see a method declaration with `[canon-id: bar::bar::X::__on_0x25_done]`

```
[debug/ast-final]       | bar::X::__on_0x25_done -> declaration::Function %241 <linkage="struct" parent_type="%1"> [canon-id: bar::bar::X::__on_0x25_done] [@d:7fb0f3c35520] ([@d:7fb0f3c35520])
```

for the field declaration `[canon-id: bar::X::__on_0x25_done]`

```

```
declaration::Field <cc="<standard>"> (bla.spicy:2:17-4:2) [canon-id: bar::X::__on_0x25_done] [@d:7fb0f3c35e70]
```

Even though these are different declarations, semantically they refer to the same entity.

Not necessarily an issue, but related it also seems hard to connect these to the related `UnitHook` which has neither a canon-id nor any other fields which seems to tie it to the field or function declaration (at least in what is printed to the ast debug stream).

File: hilti/toolchain/include/ast/node.h [hunk 9]
```diff
@@ -483,6 +504,260 @@ class None : public NodeBase, public util::type_erasure::trait::Singleton {
...
+     **/
+    std::vector<T> copy() const {
+        std::vector<T> x;
+        for ( auto i = _begin; i != _end; i++ )
+            x.push_back(*i);
+
+        return x;
+    }
+
+    const T& operator[](size_t i) const { return *(_begin + i); }
```
bbannier, line 608: Let's add some form of check here that `i< std::distance(_begin, _end` (e.g., always check, or just debug mode assertion). I just ran into a situation where after (probably incorrect) optimizations `i` pointed to a non-existing element.

hilti/toolchain/src/compiler/visitors/normalizer.cc
```diff
@@ -0,0 +1,218 @@
...
+            type.setCxxID(ID(t.cxxName()));
+    }
+
+    void operator()(const type::Struct& t, position_t p) { type::Struct::setSelf(&p.node); }
+};
+
+} // anonymous namespace
+
+struct VisitorClearCanonicalIDs : public visitor::PreOrder<void, VisitorClearCanonicalIDs> {
+    result_t operator()(const Declaration& d, position_t p) { p.node.as<Declaration>().setCanonicalID(ID()); };
```
bbannier, line 148: I see inconsistent IDs for hooks implemented in another module.

Example:

```
$ cat a.spicy
module a;

public type X = unit {
    x: uint8;
};
```

```
$ cat b.spicy
module b;

import a;

on a::X::x { print "from b", $$; }
```

```
$ spicyc a.spicy b.spicy  -D ast-final  -p -o /dev/null 2>&1 -g | grep b::a::X::
[debug/ast-final]     - spicy::declaration::UnitHook (b.spicy:5:1) [canon-id: b::a::X::x] [@s:7f82b3706080]
[debug/ast-final]         | __dd -> declaration::Expression %163 <linkage="private"> [canon-id: b::a::X::x::__dd] [@d:7f82a3535bd0] (type: uint<8> @t:7f82b37042f0)
[debug/ast-final]           | __dd -> declaration::Expression %163 <linkage="private"> [canon-id: b::a::X::x::__dd] [@d:7f82a3535bd0] (type: uint<8> @t:7f82b37042f0)
[debug/ast-final]         - declaration::Expression %163 <linkage="private"> [canon-id: b::a::X::x::__dd] [@d:7f82a3535bd0]
[debug/ast-final]       | a::X::__on_x -> declaration::Function %305 <linkage="struct" parent_type="%1"> [canon-id: b::a::X::__on_x] [@d:7f82a34250c0] ([@d:7f82a34250c0])
[debug/ast-final]     - declaration::Function %305 <linkage="struct" parent_type="%1"> (b.spicy:5:1) [canon-id: b::a::X::__on_x] [@d:7f82a34250c0]
[debug/ast-final]         | __dd -> declaration::Parameter %306 <is_type_param="false" kind="in"> [canon-id: b::a::X::__on_x::__dd] [@d:7f82a3421670] ([@d:7f82a3421670])
[debug/ast-final]           - declaration::Parameter %306 <is_type_param="false" kind="in"> (b.spicy:5:1) [canon-id: b::a::X::__on_x::__dd] [@d:7f82a3421670]
```

Here the symbols in `a` as seen from `b` sometimes are prefixed with `b::`.

For `$$` this seems to be clearly incorrect as `$$` in either `a` or `b` should always have the same identity.

I am not so sure about the name of the hook: for the optimizer I would like to have a simple way to detect whether that function impl corresponds to some decl for which the canon id seem nice, but currently does not work. OTOH one could also argue that the decl and the impl are actually different entities and one would need to do extra work to connect them.

hilti/toolchain/src/compiler/visitors/normalizer.cc
rsmmr: This has been fixed.

hilti/toolchain/include/ast/declaration.h
rsmmr: I'm adding asserts, as discussed.

hilti/toolchain/src/compiler/unit.cc
```diff
@@ -399,322 +282,92 @@ Result<CxxCode> Unit::cxxCode() const {
...
         closest_location = n.location();
 
-    for ( const auto& c : n.childs() ) {
-        auto [normal, low] = _recursiveValidateAST(c, closest_location, errors);
-        have_normal = (have_normal || normal);
-        have_low = (have_low || low);
+    if ( ! n.pruneWalk() && n.childs().size() ) {
+        auto oprio = prio;
+        for ( const auto& c : n.childs() )
+            prio = std::max(prio, _recursiveValidateAST(c, closest_location, oprio, level + 1, errors));
```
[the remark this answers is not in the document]
rsmmr: I gave this a quick try, but it's not trivial to change the logic so that we don't pass the priority down. I agree this could be cleaned up, but I'll leave it as it is for now; it's just an internal implementation detail.

hilti/toolchain/include/ast/node.h
rsmmr: Adding an assert.

hilti/toolchain/src/compiler/visitors/normalizer.cc
rsmmr: Ok, fix coming.

bbannier: Still seeing some incorrect canon IDs, e.g., example reduced from `tests/hilti/types/struct/params.hlt`:

```
$ cat x.hlt
module Foo {

type S = struct {
    method void test(string A, real B);
};

method void S::test(string A, real B) {}

global auto s1 = new S();
s1.test("a", 3);

}
```

```
$ hiltic -g -p -o /dev/null -D ast-final x.hlt 2>&1 | grep '::test]'
[debug/ast-final]       | S::test -> declaration::Function %2 <linkage="struct" parent_type="%1"> [canon-id: Foo::S::test] [@d:7f8f00c12360] ([@d:7f8f00c12360])
[debug/ast-final]               - declaration::Field <cc="<standard>"> (x.hlt:4:5) [canon-id: Foo::test] [@d:7f8f00c117e0]
[debug/ast-final]         - declaration::Field <cc="<standard>"> (x.hlt:4:5) [canon-id: Foo::test] [@d:7f8f00c117e0]
[debug/ast-final]     - declaration::Function %2 <linkage="struct" parent_type="%1"> (x.hlt:7:1) [canon-id: Foo::S::test] [@d:7f8f00c12360]
```

I see here both of `Foo::S::test` and `Foo::test` while there probably should only be the former. Curiously, when removing parameters from the function or just removing the implicit conversion this would have worked.

hilti/toolchain/src/compiler/visitors/normalizer.cc
```diff
@@ -0,0 +1,236 @@
...
+            p.node.as<Declaration>().setCanonicalID(id);
+
+        assert(d.canonicalID());
+        return d.canonicalID();
+    }
+};
+
+// Visitor double-checking that all declarations have their canonical IDs set.
+struct VisitorCheckCanonicalIDs : public visitor::PreOrder<void, VisitorCheckCanonicalIDs> {
+    result_t operator()(const Declaration& d, position_t p) { assert(d.canonicalID()); };
```
bbannier, line 199: If I also add the following check here, e.g. the test `spicy.types.unit.hooks-priorities` fails with an assertion error:

```.cpp
    result_t operator()(const expression::ResolvedID& x, position_t p) { assert(x.declaration().canonicalID()); }
```

hilti/toolchain/src/compiler/visitors/normalizer.cc
rsmmr: Ah, that's a regression, I took out a check to prevent this while I was trying to make sure all canonical IDs are set. However, turns out that's not a feasible objective: we have a legitimate case without a canonical IDs, so I just reverted that: https://github.com/zeek/spicy/pull/984/commits/30395bc979b90002da4c0af0724ea5d5c03149dc

hilti/toolchain/src/compiler/visitors/normalizer.cc
rsmmr: Ok, pushed a fix. Not seeing any other missing canon IDs right now other than then struct constructor (which, as said, is expected).

bbannier: As of 1091d61 this became much more unstable and I see a lot more tests fail now with optimizations. I checked that even without my patches already your branch leaves a number of canon IDs unset which probably should be set, e.g., just for the HILTI tests:

```
$ for f in $(find tests/hilti -name '*.hlt'); do 
    build/bin/hiltic -p $f -o /dev/null -D ast-final 2>&1 \
        | grep -q 'canon-id: not set' && echo "UNSET IN $f"
done
```

I set hits from tests not involving structs here, e.g., `tests/hilti/expressions/list-comprehension.hlt`.

hilti/toolchain/src/compiler/visitors/normalizer.cc
rsmmr: Seems finally fixed now.

hilti/toolchain/src/compiler/visitors/normalizer.cc
rsmmr: Likewise fixed now.

rsmmr:
This merged, GitHub didn't close it for some reason.

PR #986, opened 2021-09-02
Support building on arm64

bbannier:


bbannier:
For some reason the CI setup I added here doesn't trigger a Cirrus CI arm64 build, https://cirrus-ci.com/build/5283602276548608.
Screen Shot 2021-09-07 at 4 35 30 PM

I opened #993 to cover only the source level changes, leaving out the CI parts.

PR #987, opened 2021-09-02
Allow use of private units as sinks

bbannier:


bbannier:
As currently written this PR emits a lot of additional code for `parseN` functions which now need to be marked `&always-emit`. Once #984 lands it should be possible to remove that attribute again so no new code should be emitted then (we might end up emitted potentially unused members).

PR #995, opened 2021-09-08
Disable leak tests on macos; correctly handle NULL symbols in `Library::symbol`.

bbannier:


hilti/runtime/src/library.cc
```diff
@@ -98,10 +98,12 @@ hilti::rt::Result<void*> hilti::rt::Library::symbol(std::string_view name) const
...
 
+    // Clear any library errors.
+    ::dlerror();
+
     auto* symbol = ::dlsym(_handle, name.data());
-    // auto* symbol = ::dlsym(RTLD_SELF, name.data());
 
-    if ( symbol == nullptr )
+    if ( auto error = ::dlerror() )
         return result::Error(fmt("symbol '%s' not found", name));
```
rsmmr, line 108: Let's add an `assert(symbol)` here, just to be sure.

rsmmr (approved):

hilti/runtime/src/library.cc
bbannier: The point of this patch is that in general we cannot exclude `symbol` actually being `NULL`. We now follow the approach documented here, https://man7.org/linux/man-pages/man3/dlsym.3.html#NOTES.

I think we should not add that assertion.

rsmmr: Then I'm not sure what problem we're solving here? Can the callers handle it if there's a null coming back? (I guess  in principle it could be legitimate in the case of "can be placed at zero address by the linker", but that seems far fetched).

bbannier: I did some digging to get a feel for "proper" use of symbols returned from `dlsym` and it looks like most examples seem to follow the error checking pattern from the `dlsym` documentation, and use the returned symbol without null check. That being said, we currently only use `symbol` for testing to check whether something or an error was returned. This patch is a pure cleanup and solves no problem we currently have.

rsmmr: Alrighty. :-)

PR #997, opened 2021-09-08
Document initializers for unit variables.

### LINKED ISSUE #939: Add validation that unit var initializers are constants
Unit var initializers are set when a units begins parsing, not where they are declared in a unit. This means that it e.g., is not possible to use other parsed results (not sure if we need to restrict to just constants, though).

```.r
module Foo;

public type X = unit {
    x: uint8;
    var y: uint8 = self.x;
};
```

```
$ echo 123 | spicy-driver bla.spicy
[error] terminating with uncaught exception of type hilti::rt::UnsetOptional: unset optional value (bla.spicy:5:20)
```

We should add validation that unit var initializers do not use unset values.

bbannier:
Close #939.

### DIFF
--- a/doc/programming/parsing.rst
+++ b/doc/programming/parsing.rst
@@ -598,7 +598,7 @@ variables to a unit type to store arbitrary state:
 Here, we assign a string value to ``a`` once we have parsed ``x``. The
 final ``print`` shows the expected value. As you can also see, before
 we assign anything, the variable's value is just empty: Spicy
-initializes instances variables with well-defined defaults. If you
+initializes unit variables with well-defined defaults. If you
 would rather leave a variable unset by default, you can add
 `&optional`:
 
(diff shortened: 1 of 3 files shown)


doc/programming/parsing.rst
```diff
@@ -621,6 +621,30 @@ would rather leave a variable unset by default, you can add
 You can use the ``?.`` unit operator to test if an optional unit
 variable remains unset.
 
+Unit variables can also be initialized with arbitrary expression when being
+defined. The initialization is performed *before* the containing unit starts
+parsing so this form of initialization cannot be used to access parse results
+of the unit itself.
```
rsmmr, line 627:
```suggestion
Unit variables can also be initialized with custom expressions when being
defined. The initialization is performed just before the containing unit starts
parsing (implying that the expressions cannot access parse results
of the unit itself yet).
```

doc/programming/parsing.rst
```diff
@@ -621,6 +621,30 @@ would rather leave a variable unset by default, you can add
...
+    module Test;
+
+    public type Broken = unit {
+        x: int8;
+
+        # This initialization succeeds as the initializer can be converted to the variable type.
+        var b: int8 = 123;
+
+        # This initialization never succeeds as when 'a' is initialized 'self.x' is always unset.
+        var a: int8 = self.x;
```
rsmmr, line 640: Let's skip this, we shouldn't be showing code that's wrong.

rsmmr (changes requested):
I can see where this is coming from but let's focus the documentation on the general use case (initializing variables with expressions), and not on a specific case that's not supported (accessing parse results from the init expression).

rsmmr (approved):

PR #998, opened 2021-09-08
Allow use of units with all defaulted parameters as entry points.

### LINKED ISSUE #941: Units with default'ed parameters cannot be used as external entry points
Units which take parameters which all have default arguments can currently not be external entry points, e.g., given

```.ruby
module foo;

public type X = unit(b: bool = False) {
    x: uint8;
};
```

we cannot use this as a parser via `spicy-driver`, 

```
$ echo '\x42' | spicy-driver -d foo.spicy
[error] unit type 'foo::X' cannot be used as external entry point because it requires arguments
```

The code in questions looks for different parser methods which are not set.

In principle it should be possible to allow such parsers as external entry points. This could e.g., make it easier to reuse units.

bbannier:
This patch adds support to allow units with all defaulted parameters as
entry points. We previously did explicitly not allow that, but had
inconsistent code in place for handling any parameters. We fix these
instances.

Closes #941.


spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -1294,9 +1295,6 @@ hilti::type::Struct ParserBuilder::addParserMethods(hilti::type::Struct s, const
          builder::parameter("cur", type::Optional(type::stream::View()), builder::optional(type::stream::View())),
          builder::parameter("context", type::Optional(builder::typeByID("spicy_rt::UnitContext")))};
 
-    for ( auto p : t.parameters() )
-        params.emplace_back(p);
-
```
rsmmr, line 1299: Hmm, I don't think I see a replacement for this, don't we need this?

bbannier: This adds all unit parameters as parameters of the `parse*` functions. AFAICT this is not need as the parse methods either default-construct units (`parse1` or `parse3`), or work on already initialized units passed as parameters (`parse2`).

Since we previously disallowed units with parameters as entry points this code was up until now only called with an empty `parameters`.

rsmmr: ok, see it now.

rsmmr (approved):

PR #1001, opened 2021-09-14
Fix parsers for bison-3.8.1.

bbannier:
With bison-3.8.1 redefining `YYSTYPE` instead of `%define
api.value.type` has been deprecated and now triggers a warning.

### DIFF
--- a/hilti/toolchain/include/compiler/detail/parser/driver.h
+++ b/hilti/toolchain/include/compiler/detail/parser/driver.h
@@ -25,8 +25,6 @@
                                         hilti::detail::parser::location* yylloc,                                       \
                                         hilti::detail::parser::Driver* driver)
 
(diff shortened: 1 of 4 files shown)


rsmmr (commented):
Does this work with older Bisons (our docs say >= 3.0).

bbannier:
> Does this work with older Bisons (our docs say >= 3.0).

The [Bison changelog](http://git.savannah.gnu.org/cgit/bison.git/tree/NEWS#n40) make it sound as if using `%define api.value.type` has pretty much always been preferred and and they only now softly start to enforce this.

I have explicitly triggered the `docker_debian9` job for this PR which today uses `bison-3.0.4`; that job passed.

rsmmr (approved):

PR #1006, opened 2021-09-15
Clean up optimizer after AST rework.

bbannier:
This patch simplifies code by using AST canonical IDs, and by accessing
a unit's method via `parent`. Since `Field`s are now declarations we can
also simplify code related to that.

### DIFF
--- a/hilti/toolchain/src/compiler/optimizer.cc
+++ b/hilti/toolchain/src/compiler/optimizer.cc
@@ -310,8 +310,6 @@ struct FunctionVisitor : OptimizerVisitor, visitor::PreOrder<bool, FunctionVisit
         assert(x.hasOp0());
 
         auto type = x.op0().type();
-        while ( type::isReferenceType(type) )
-            type = type.dereferencedType();
 
         auto struct_ = type.tryAs<type::Struct>();
         if ( ! struct_ )
--- a/hilti/toolchain/src/compiler/optimizer.cc
+++ b/hilti/toolchain/src/compiler/optimizer.cc
@@ -141,7 +141,7 @@ struct FunctionVisitor : OptimizerVisitor, visitor::PreOrder<bool, FunctionVisit
         if ( ! p.parent().isA<type::Struct>() )
             return {};
 
-        auto function_id = x.canonicalID();
+        const auto& function_id = x.canonicalID();
         assert(function_id);
 
         switch ( _stage ) {
@@ -196,7 +196,7 @@ struct FunctionVisitor : OptimizerVisitor, visitor::PreOrder<bool, FunctionVisit
     }
 
     result_t operator()(const declaration::Function& x, position_t p) {
-        const auto function_id = x.canonicalID();
+        const auto& function_id = x.canonicalID();
         assert(function_id);
 
         switch ( _stage ) {
@@ -631,7 +631,7 @@ struct ConstantFoldingVisitor : OptimizerVisitor, visitor::PreOrder<bool, Consta
         if ( x.type() != type::Bool() )
             return false;
 
-        auto id = x.canonicalID();
+        const auto& id = x.canonicalID();
         assert(id);
 
         switch ( _stage ) {
@@ -871,7 +871,7 @@ struct FeatureRequirementsVisitor : visitor::PreOrder<void, FeatureRequirementsV
 
                 // Check if access to the field has type requirements.
                 if ( auto type_id = type.typeID() )
-                    for ( auto requirement : AttributeSet::findAll(field->attributes(), "&needed-by-feature") ) {
+                    for ( const auto& requirement : AttributeSet::findAll(field->attributes(), "&needed-by-feature") ) {
                         const auto feature = *requirement.valueAsString();
                         if ( ! ignored_features.count(*type_id) || ! ignored_features.at(*type_id).count(feature) )
                             // Enable the required feature.
@@ -900,7 +900,7 @@ struct FeatureRequirementsVisitor : visitor::PreOrder<void, FeatureRequirementsV
                         const auto& param = parameters[i];
 
                         if ( auto type_id = type.typeID() )
-                            for ( auto requirement :
+                            for ( const auto& requirement :
                                   AttributeSet::findAll(param.attributes(), "&requires-type-feature") ) {
                                 const auto feature = *requirement.valueAsString();
                                 if ( ! ignored_features.count(*type_id) ||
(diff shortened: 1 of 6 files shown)


bbannier:
This patch is based on #984.

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -47,6 +47,19 @@ Type innermostType(Type type) {
     return type;
 }
 
+// Wrapper around `canonicalID` to handle malformed IDs.
+//
+// FIXME(bbannier): remove this code once `canonicalID` handles this well.
```
bbannier, line 52: Check if this will be needed in the end.

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -254,13 +138,15 @@ struct FunctionVisitor : OptimizerVisitor, visitor::PreOrder<bool, FunctionVisit
         if ( ! x.type().isA<type::Function>() )
             return false;
 
-        auto function_id = getID(x, p);
-        if ( ! function_id )
-            return false;
+        if ( auto parent = p.parent().tryAs<Type>(); ! parent || ! parent->isA<type::Struct>() )
```
rsmmr, line 141:
```suggestion
        if ( ! p.parent().isA<type::Struct>() )
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -421,29 +304,51 @@ struct FunctionVisitor : OptimizerVisitor, visitor::PreOrder<bool, FunctionVisit
...
 
     result_t operator()(const operator_::struct_::MemberCall& x, position_t p) {
-        auto function_id = getID(x, p);
+        if ( ! x.hasOp1() )
+            return false;
+
+        assert(x.hasOp0());
+
+        auto type = x.op0().type();
+        while ( type::isReferenceType(type) )
```
rsmmr, line 313: I'd be curious to know if we still need this.

tests/Baseline/spicy.optimization.default-parser-functions/log
```diff
@@ -78,6 +78,12 @@
 [debug/optimizer] removing field for unused method foo::P2::__on_0x25_undelivered
 [debug/optimizer] removing field for unused method foo::P2::__on_x
 [debug/optimizer] removing field for unused method foo::P2::__str__
+[debug/optimizer] removing field for unused method foo::__register_foo_P1::<anon-struct-5>::parse1
+[debug/optimizer] removing field for unused method foo::__register_foo_P1::<anon-struct-5>::parse2
+[debug/optimizer] removing field for unused method foo::__register_foo_P1::<anon-struct-5>::parse3
+[debug/optimizer] removing field for unused method foo::__register_foo_P2::<anon-struct-8>::parse1
+[debug/optimizer] removing field for unused method foo::__register_foo_P2::<anon-struct-8>::parse2
+[debug/optimizer] removing field for unused method foo::__register_foo_P2::<anon-struct-8>::parse3
```
rsmmr, line 86: I'm wondering what's happening here: these anon structs should appear only for struct constructors, but we'd be using their fields when we coerce it to the actual type. Would suggest to take a quick look at the before-/after-optimization code here to double-check what's changing.

rsmmr (approved):
Nice to see that the code is indeed getting quite a bit simpler with the new infrastructure in place.

hilti/toolchain/src/compiler/optimizer.cc
bbannier: Looks like this can go now. We still need the `innermostType` helper though.

tests/Baseline/spicy.optimization.default-parser-functions/log
bbannier: Yes, this one is weird, and it seemed surprising that this even works. The reason for this is in how struct ctrs are represented in the AST, i.e., code of this form:

```
foo::P1::__parser = [
  $name="foo::P1",
  $parse1=foo::P1::parse1,
  $parse2=foo::P1::parse2,
  $parse3=foo::P1::parse3,
  $context_new=Null, $type_info=typeinfo(P1),
  $description="",
  $mime_types=vector(),
  $ports=vector()
];
```

Simplified this looks like this:
```
- statement::Expression [@s:7fdad2171960]
  - expression::Assign (non-const) (resolved) [@e:7fdad2171650]
    - expression::ResolvedID <rid="250"> (type: spicy_rt::Parser [@t:7fdad043dd80]) (non-const) (resolved) [@e:7fdac05621b0]
      - ID <name="foo::P1::__parser"> [@i:7fdac0562140]
    - expression::Ctor (default-parser-functions.spicy:14:18) (const) (resolved) [@e:7fdad0616a40]
      - ctor::Coerced (default-parser-functions.spicy:14:18) [@c:7fdad0617200]
        - ctor::Struct (default-parser-functions.spicy:14:18) [@c:7fdad21a8570]
          - type::Struct %413 (default-parser-functions.spicy:14:18) (non-const) (resolved) [@t:7fdac0566530]
            - declaration::Expression %422 <linkage="private"> (default-parser-functions.spicy:14:18) [canon-id: foo::__register_foo_P1::<anon-struct-5>::self] [@d:7fdad0613e40]
              - ID <name="self"> (default-parser-functions.spicy:14:18) [@i:7fdad0613b20]
            - declaration::Field <cc="<standard>"> (default-parser-functions.spicy:14:18) [canon-id: foo::__register_foo_P1::<anon-struct-5>::name] [@d:7fdac0564850]
              - ID <name="name"> [@i:7fdac0564740]
            - declaration::Field <cc="<standard>"> (default-parser-functions.spicy:14:18) [canon-id: foo::__register_foo_P1::<anon-struct-5>::parse1] [@d:7fdac0564ce0]
              - ID <name="parse1"> [@i:7fdac0564900]
            [MORE FIELDS]

          - ctor::struct_::Field (default-parser-functions.spicy:14:18) [@c:7fdad21a9050]
            - ID <name="name"> [@i:7fdad2171ae0]
            - expression::Ctor (const) (resolved) [@e:7fdad216f3b0]
              - ctor::String <value="foo::P1"> [@c:7fdad216e250]
                - type::String (const) (resolved) [@t:7fdad216f570]
          - ctor::struct_::Field (default-parser-functions.spicy:14:18) [@c:7fdad21a9110]
            - ID <name="parse1"> [@i:7fdad2179ca0]
            - expression::ResolvedID <rid="314"> (type: function method view<stream> (inout value_ref<stream> data, optional<view<stream>> cur = Null, optional<spicy_rt::UnitContext> context) [@t:7fdad21725a0]) (const) (resolved) [@e:7fdac0563460]
              - ID <name="foo::P1::parse1"> [@i:7fdac05633f0]
            [MORE FIELDS]
```

We are removing some `declaration::Field`s which mirror data also present in the `ctor::struct_::Field`s so the data we need is still there in the end.

I am not really sure what to do here since this seems to actually remove dead code, but still surprising. Suggest we leave as is.

PR #1007, opened 2021-09-15
Use safe integers for runtime size functions.

bbannier:
In fd5c53cb8ffd1a18aadb7d21b927bda801e264c8 we changed the size
functions of runtime containers to use safe integers. This patch adjusts
the remaining size functions to use safe integers as well.

We also add a test for `RegExp::find` with empty needle. While the
behavior of that function should not have changed with this patch, we
still had to adjust its implementation.


rsmmr (approved):
Looks good.

rsmmr (approved):

PR #1010, opened 2021-09-22
Fix some performance regressions after AST overhaul

bbannier:


hilti/toolchain/src/compiler/context.cc
```diff
@@ -176,7 +176,8 @@ void Context::dumpUnitCache(const hilti::logging::DebugStream& stream) {
                                           d->module().renderedRid(), d.get()));
         }
 
-        hilti::render(stream, unit->module(), true);
+        if ( logger().isEnabled(stream) )
```
rsmmr, line 179: I think you can move this to the very top:

```
if ( ! logger().isEnabled(stream) )
    return
```

rsmmr (dismissed):

PR #1012, opened 2021-09-24
Speed up source parsing

### LINKED ISSUE #46: Switch the Bison files' struct value type to variants
The current struct is pretty inefficient. We can't use a union, but Bison has a variant type: https://www.gnu.org/software/bison/manual/bison.html#C_002b_002b-Variants

Bison maintainer Akim Demaille has an example of efficient C++ use
in flex & bison here:
https://gitlab.lrde.epita.fr/vcsn/vcsn/tree/master/lib/vcsn/dot

bbannier:


bbannier:
The Bison `emplace` syntax I am using here only appeared with bison-3.2.1, we'll need to use the more verbose `build` method which also existed before that, http://git.savannah.gnu.org/cgit/bison.git/commit/?id=3ead69b1c1241100b7a2d2b6c6fd89d545c23ca4.

bbannier:
Even after switching to `build` syntax we still fail with Bison versions missing [this patch](http://git.savannah.gnu.org/cgit/bison.git/commit/?id=e3fdc370495ffdedadd6ac621e32e34a0e1a9de0). I had a look whether we somewhere rely on default actions, but even after e.g., fixing the single instance where we do in the HILTI parser this still segfaults.

bbannier:
FTR, the issue of the segfaults seems to (at least from looking at it with a debugger) be related to us not always initializing variant. I added 0d92823 to fix issues in our parsers, but the issue seems to persist.

For testing I used a HILTI reproducer so I only needed to build `hiltic`:

```
module x {
y;
}
```

Replacing `y` with e.g., `1` passes.

hilti/toolchain/src/compiler/parser/parser.yy
```diff
@@ -27,7 +27,9 @@ namespace hilti { namespace detail { class Parser; } }
 %define api.namespace {hilti::detail::parser}
 %define api.parser.class {Parser}
 %define parse.error verbose
-%define api.value.type {yystype_hilti}
+%define api.value.type variant
+%define parse.assert
+%define parse.trace
```
bbannier, line 32: @rsmmr remove this if you run into issues.

hilti/toolchain/src/compiler/parser/parser.yy
```diff
@@ -789,7 +793,7 @@ expr_f        : ctor                             { $$ = hilti::expression::Ctor(
                                                  { $$ = hilti::expression::ListComprehension(std::move($6), std::move($2), std::move($4), {},  __loc__); }
               | '[' expr FOR local_id IN expr IF expr ']'
                                                  { $$ = hilti::expression::ListComprehension(std::move($6), std::move($2), std::move($4), std::move($8),  __loc__); }
-              | expr_g
+              | expr_g                           { $$ = std::move($1); }
```
bbannier, line 796: 🤦 , of course.

This patch looks like what you want. Let's merge it after cleanup and pulling back in the unordered collection for the location cache (a benchmark shows that that's worthwhile).

rsmmr: Yep, will do, reverted it only for debugging.

rsmmr (approved):

Issue #1009, opened 2021-09-21
Investigate performance


rsmmr:
With the merge of #984. compiler performance has changed a bit: it improved for an individual `spicyc` run, but worsened for the whole testsuite; see below. We need to investigate what's going on.

Also, let's dig into the new `-R` output:

- what's the `renderer` doing in there at that level?
- why's the `printer` called that often?
- why is about 1/3 of the time not accounted for?
- can get the optimizer further down?

```
+++++ main

# ./bin/spicyc  -R -c ../../spicy-analyzers/spicy-analyzers/protocol/dhcp/dhcp.spicy >/dev/null

=== Execution Time Summary ===

  25.13%     0.39s   hilti/compiler/optimizer (#1)
  19.86%     0.31s   hilti/compiler/operator-resolver (#13)
   9.47%     0.15s   hilti/compiler/scope-builder (#5)
   6.77%     0.11s   hilti/compiler/id-resolver (#13)
   5.41%     0.08s   hilti/compiler/apply-coercions (#13)
   5.24%     0.08s   spicy/compiler/id-resolver (#13)
   4.98%     0.08s   spicy/compiler/codegen (#13)
   4.43%     0.07s   hilti/compiler/parser (#3)
   4.31%     0.07s   hilti/compiler/importer (#13)
   3.98%     0.06s   spicy/compiler/validator (#19)
   3.03%     0.05s   spicy/compiler/apply-coercions (#13)
   3.01%     0.05s   hilti/compiler/codegen (#3)
   2.61%     0.04s   spicy/compiler/scope-builder (#5)
   0.08%     0.00s   hilti/printer (#447)
   0.03%     0.00s   spicy/printer (#925)

Total time: 1.56s


++++++ /topic/robin/gh-784-type-resolution

# ./bin/spicyc -R -c ../../spicy-analyzers/spicy-analyzers/protocol/dhcp/dhcp.spicy >/dev/null

=== Execution Time Summary ===

  26.60%     0.34s   hilti/compiler/optimizer (#1)
  22.64%     0.29s   hilti/compiler/ast/resolver (#28)
  20.46%     0.26s   hilti/renderer (#7)
  10.12%     0.13s   hilti/compiler/ast/normalizer (#28)
   6.53%     0.08s   hilti/compiler/parser (#3)
   3.82%     0.05s   hilti/compiler/ast/scope-builder (#28)
   3.79%     0.05s   hilti/compiler/ast/coerce (#28)
[...]

PR #1014, opened 2021-09-29
Fix overly generous generation of type information.

rsmmr:
We recently started including type information for all types that a
module included. That doesn't harm functionally, but increased code
size and compilation time. This patch reverts that to the old behavior.

Doing so required one additional fix for exceptions, which had
generation of their destructor tied to the generation of type
information; which now doesn't work anymore, but also seems unnecessary
in the first place.

Addresses #1009.

### DIFF
--- a/hilti/toolchain/src/compiler/codegen/codegen.cc
+++ b/hilti/toolchain/src/compiler/codegen/codegen.cc
@@ -149,7 +149,8 @@ struct GlobalsVisitor : hilti::visitor::PreOrder<void, GlobalsVisitor> {
     void operator()(const declaration::Type& n, position_t p) {
         assert(n.typeID());
         cg->compile(n.type(), codegen::TypeUsage::Storage);
-        cg->addTypeInfoDefinition(n.type());
+        if ( include_implementation )
+            cg->addTypeInfoDefinition(n.type());
     }
 };
 
--- a/hilti/toolchain/include/compiler/unit.h
+++ b/hilti/toolchain/include/compiler/unit.h
@@ -205,6 +205,12 @@ public:
     bool requiresCompilation();
 
     /**
+     * Explicily marks the unit as requiring compilation down to C++, overiding
+     * any automatic determination.
+     */
+    void setRequiresCompilation() { _requires_compilation = true; }
+
+    /**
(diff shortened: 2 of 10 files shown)


hilti/toolchain/src/compiler/codegen/codegen.cc
```diff
@@ -149,7 +149,8 @@ struct GlobalsVisitor : hilti::visitor::PreOrder<void, GlobalsVisitor> {
     void operator()(const declaration::Type& n, position_t p) {
         assert(n.typeID());
         cg->compile(n.type(), codegen::TypeUsage::Storage);
-        cg->addTypeInfoDefinition(n.type());
+        if ( include_implementation )
+            cg->addTypeInfoDefinition(n.type());
```
bbannier, line 153: I revisited the example we had also looked at offline,

``` $ spicyc -C tests/spicy/types/real/coercion.spicy -j```

Before #984 we emitted only `dbg.Foo.cc` (w/o any type infos), and `dbg.__linker__.cc` (w/o type infos). After #984 we emitted an additional `dbg.hilti.cc` which included type infos in addition to now also emitted type infos in `dbg.Foo.cc`. The hypothesis was that the additional type infos slowed down the compilation.

Now looking at this patch I see that the type infos in `dbg.Foo.cc` are gone and we are pretty much back to the state before #984 which is great. However, I still see `dbg.hilti.cc` being emitted, with type infos which seem unused. This seems to be due to how we attach type infos to units here and below, so we do not check from the top-level user code which types are used, but instead bubble them up from whatever code we see. I suspect that us emitted that additional `.cc` file for hilti still contributes significantly to the overall time it takes to compile, not only due to the extra type infos, but more so since it requires us to start another compilation job.

Do you think we could collect type infos at a higher level were we can actually check whether it would be possible to reference them in generated C++ code?

rsmmr: Typeinfos aren't represented in the AST at all currently, so I suspect that would be a major undertaking. Let me see if I can make it stop generating that additional compilation unit, like it used to be.

bbannier (approved):

Issue #1013, opened 2021-09-24
Consider bringing back pre-transform validation


bbannier:
Before #984 we ran an addition pre-transform validation hook before e.g., fully expanding imports or resolving. It seems that with that approach we were able to handle sources with errors much faster. Now the corresponding validator is folded into the regular validator which runs late, after the AST is fully formed, see https://github.com/zeek/spicy/issues/1009#issuecomment-925809426.

While writing these validators for incomplete ASTs is tricky, providing feedback quickly can be very beneficial e.g., during parser development, or speed up our test suite which contains a larger sample of faulty code. We should consider bringing back some for of pre-transform validation.

rsmmr:
Done.

PR #1015, opened 2021-09-30
Split validation into two passes.

rsmmr:
One validation pass runs before AST resolving, one afterwards. The first
can be used for any checks that don't need any further worn on that AST
so that they can abort processing quickly.

Right now, the pre-resolver validators are empty no-ops, nothing has
moved over yet.

Addresses #1013.

### DIFF
--- a/hilti/toolchain/src/compiler/visitors/validator.cc
+++ b/hilti/toolchain/src/compiler/visitors/validator.cc
@@ -16,28 +16,34 @@ using util::fmt;
 
 namespace {
 
-struct Visitor : public visitor::PostOrder<void, Visitor> {
+struct VisitorBase {
     // Record error at location of current node.
-    void error(std::string msg, position_t& p, node::ErrorPriority priority = node::ErrorPriority::Normal) {
+    void error(std::string msg, const visitor::Position<Node&>& p,
+               node::ErrorPriority priority = node::ErrorPriority::Normal) {
         p.node.addError(msg, p.node.location(), priority);
         ++errors;
     }
 
     // Record error with current node, but report with another node's location.
-    void error(std::string msg, position_t& p, const Node& n,
+    void error(std::string msg, const visitor::Position<Node&>& p, const Node& n,
                node::ErrorPriority priority = node::ErrorPriority::Normal) {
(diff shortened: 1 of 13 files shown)


rsmmr:
As discussed, I'm suggesting to leave this at this: putting the infrastructure in place for pre-resolver validation, but not porting existing checks over because that would not only be quite a bit of work but also make the code harder to maintain, because the current structure of the checks doesn't prepare for the split. 

For the record: the new split isn't the same as the old one prior to introducing the new AST code. In the old code, we did a "pre-transform" validation, which happened on a *partially* resolved AST. I believe we were just lucky in some cases that that actually worked and short-circuited the process.

hilti/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -16,28 +16,33 @@ using util::fmt;
 
 namespace {
 
-struct Visitor : public visitor::PostOrder<void, Visitor> {
+struct VisitorBase {
     // Record error at location of current node.
-    void error(std::string msg, position_t& p, node::ErrorPriority priority = node::ErrorPriority::Normal) {
+    void error(std::string msg, operator_::position_t& p, node::ErrorPriority priority = node::ErrorPriority::Normal) {
```
bbannier, line 21: I am not sure whether using an `operator_::position_t` here is always safe (the `position_t` in spicy's `VisitorBase` seems to resolve to the same type).

Can you instead make it accept whatever `position_t` the deriving visitor has, e.g.,

```.cpp
template <typename Position>
void error(std::string msg, Position& p, node::ErrorPriority priority = node::ErrorPriority::Normal) {
```

Here, below for the other methods, and also in spicy's `VisitorBase`.

bbannier (dismissed):

hilti/toolchain/src/compiler/visitors/validator.cc
rsmmr: I don't think we need the template, we really just instantiate position with `Node`, I'm changing it to make it explicit through `const visitor::Position<Node&>` (as we do elsewhere).

bbannier (approved):

PR #1041, opened 2021-11-04
Fix build for RTD.

bbannier:
This resolves the build problems due to sphinx-doc/sphinx#9727 we've
been seeing.


bbannier:
This commit adapts the changes from zeek/package-manager#126 for Spicy where we have seen similar problems.

rsmmr (approved):

PR #1048, opened 2021-11-11
Bump zkg in published Docker image.

bbannier:
In order to use packages with dependencies we require
zeek/package-manager#106 which is only available with zkg-2.12.0.


rsmmr (approved):
Ok assuming this passes CI.

PR #1060, opened 2021-12-01
Cleanup Docker publish workflow & publish to ECR as well

bbannier:


.github/workflows/docker.yml
```diff
@@ -0,0 +1,87 @@
+# NOTE: keep this action in sync with `docker-tags.yml`.
+
+name: Publish Docker
+on:
+  pull_request:
```
bbannier, line 5: Let's check whether we can make this not required or maybe even trigger-only for PRs to not add too much time before we get a ✅ from CI.

rsmmr (approved):

PR #1066, opened 2021-12-07
Fix code generator to emit types correctly for transitive dependencies.

### LINKED ISSUE #1057: Transitive includes not handled correctly
Under certain setups imports of imported modules are not implicitly imported. It is important that this works 😅.

```.console
$ tail *.spicy
==> a.spicy <==
module a;

import spicy;

%byte-order=spicy::ByteOrder::Little;

public type A = unit {};

==> b.spicy <==
module b;

import a;
import spicy; # FIXME(bbannier): This should be implicit when importing `a`.

public type B = unit {
    a_: a::A;
};

==> c.spicy <==
module c;
import b;

==> d.spicy <==
module d;

import c;
```

When compiling both `c` and `d` this fails:

```.console
spicyc c.spicy d.spicy -j
/private/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/d_dc75e53109471175-38c9e38328e44a4.cc:13:58: error: no member named 'a' in namespace '__hlt'
        std::optional<::hilti::rt::ValueReference<__hlt::a::A>> a_{};
                                                  ~~~~~~~^
```

As marked in the source for `b.spicy` there also seems to be an issue of `a` correctly exporting its dependency on `spicy`.

rsmmr:
Closes #1057.

### DIFF
--- a/hilti/toolchain/include/compiler/unit.h
+++ b/hilti/toolchain/include/compiler/unit.h
@@ -7,6 +7,7 @@
 #include <functional>
 #include <memory>
 #include <string>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
@@ -172,8 +173,13 @@ public:
      */
     Result<CxxCode> cxxCode() const;
 
-    /** Returns the list of dependencies registered for the unit so far. */
-    const auto& dependencies() { return _dependencies; }
+    /**
+     * Returns the list of dependencies registered for the unit so far.
+     *
+     * @param recursive if true, return the transitive closure of all
+     * dependent units, vs just direct dependencies of the current unit
+     */
+    std::vector<std::weak_ptr<Unit>> dependencies(bool recursive = false) const;
 
     /** Removes any dependencies registered for the unit so far. */
     void clearDependencies() { _dependencies.clear(); };
@@ -384,6 +390,9 @@ private:
 
     // Recursively destroys the module's AST.
     void _destroyModule();
+
+    // Helper for depencencies() to recurse.
+    void _recursiveDependencies(std::vector<std::weak_ptr<Unit>>* dst, std::unordered_set<const Unit*>* seen) const;
 
     // Parses a source file with the appropriate plugin.
     static Result<hilti::Module> _parse(const std::shared_ptr<Context>& context,
--- a/hilti/toolchain/src/compiler/unit.cc
+++ b/hilti/toolchain/src/compiler/unit.cc
@@ -266,7 +266,7 @@ Result<Nothing> Unit::codegen() {
     //
     // TODO(robin): Would be nice if we had a "cheap" compilation mode that
     // only generated declarations.
-    for ( const auto& unit : dependencies() ) {
+    for ( const auto& unit : dependencies(true) ) {
         HILTI_DEBUG(logging::debug::Compiler, fmt("importing declarations from module %s", unit.lock()->id()));
         auto other = detail::CodeGen(context()).compileModule(unit.lock()->module(), unit.lock().get(), false);
         c->importDeclarations(*other);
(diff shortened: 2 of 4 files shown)


hilti/toolchain/include/compiler/unit.h
```diff
@@ -173,7 +173,7 @@ class Unit {
     Result<CxxCode> cxxCode() const;
 
     /** Returns the list of dependencies registered for the unit so far. */
-    const auto& dependencies() { return _dependencies; }
+    std::vector<std::weak_ptr<Unit>> dependencies(bool recursive = false);
```
bbannier, line 176: This should be `const`.

hilti/toolchain/include/compiler/unit.h
```diff
@@ -385,6 +385,9 @@ class Unit {
     // Recursively destroys the module's AST.
     void _destroyModule();
 
+    // Helper for depencencies() to recurse.
+    void _recursiveDependencies(std::vector<std::weak_ptr<Unit>>* dst, std::vector<const Unit*>* seen);
+
```
bbannier, line 390: Can we instead move this to a helper function in the `.cc` file and pass in the `Unit` arg via a parameter `const Unit&`?

hilti/toolchain/src/compiler/unit.cc
```diff
@@ -307,6 +307,30 @@ Result<CxxCode> Unit::cxxCode() const {
...
+
+    for ( const auto& d : _dependencies ) {
+        if ( std::find(seen->begin(), seen->end(), d.lock().get()) != seen->end() )
+            continue;
+
+        dst->push_back(d);
+        seen->push_back(d.lock().get());
+        d.lock()->_recursiveDependencies(dst, seen);
+    }
+}
```
bbannier, line 322: Couple of comments:

- I think this would be better as a standalone, hidden function
- a `vector` `seen` seems like an unneeded pessimization for repeated searches, let's use a `unordered_set` instead (which also makes the code for search much shorter).
- let's avoid repeatedly locking the `weak_ptr` as this could become expensive

```suggestion
void _recursiveDependencies(const Unit& unit, std::vector<std::weak_ptr<Unit>>* dst,
                            std::unordered_set<const Unit*>* seen) {
    // This uses two vectors because the weak_ptr are a bit tough to work with,
    // in particular they can't be compared through std::find().

    for ( const auto& d : unit.dependencies(false) ) {
        auto* dep = d.lock().get();
        if ( seen->count(dep) )
            continue;

        dst->push_back(d);
        seen->insert(dep);
        _recursiveDependencies(*dep, dst, seen);
    }
}
```

tests/spicy/types/unit/transitive-imports.spicy
```diff
@@ -0,0 +1,35 @@
+# @TEST-EXEC: spicyc c.spicy d.spicy -j
```
bbannier, line 1:
```suggestion
# @TEST-EXEC: spicyc -d -j c.spicy d.spicy
```

hilti/toolchain/include/compiler/unit.h
```diff
@@ -173,7 +173,7 @@ class Unit {
     Result<CxxCode> cxxCode() const;
 
     /** Returns the list of dependencies registered for the unit so far. */
```
bbannier, line 175: Let's document the new parameter.

bbannier (approved):

hilti/toolchain/include/compiler/unit.h
rsmmr: The function needs access to the `_dependencies` member, so we'd need to friend it, which doesn't seem like it would help much.

bbannier (approved):

PR #1067, opened 2021-12-07
Add basic formatting and linting hooks for CMake files.

bbannier:


CMakeLists.txt
```diff
@@ -60,9 +60,15 @@ endif ()
...
-    )
+    string(
+        JOIN
+        " "
+        sanitizer_cxx_flags
+        "-fsanitize=${USE_SANITIZERS}"
+        "-fno-omit-frame-pointer"
+        "-fno-optimize-sibling-calls"
+        "-O1"
+        "-shared-libsan")
```
rsmmr, line 71: Curious: why is this better?

bbannier: I did this to make this line shorter, but the code is arguably worse.

Since we already break long lines during formatting I'll just disable the line length lint since it just fires for isolate long strings or variable names at this point which are not too.

rsmmr (approved):

PR #1071, opened 2021-12-09
Show log for configure failures.

bbannier:
In 945dd34a5e49a we attempted to fix the issue of `./configure` failures
not leading to error return codes by explictly capturing the log and showing it
later. This introduced the issue that the configure log was only
shown after `./configure` finished _successfully_; on configure failure
we would correctly return an error code, but never reach the code
displaying the log.

This patch fixes this puzzling behavior so that the log is even shown
if `configure` runs into issues. For that we use the `pipefail` Bash
option.

### DIFF
--- a/configure
+++ b/configure
@@ -175,8 +175,9 @@ else
 fi
 
 cd "${cmake_build_directory}"
-eval "${cmake}" > config.log 2>&1
-cat config.log
+fail=$(mktemp)
+{ eval "${cmake}" 2>&1 || echo > "$fail"; } | tee config.log &&  [ ! -s "$fail" ]
+rm "$fail"
 
 echo "# This is the command used to configure this build" > config.status
 echo "${command}" >> config.status


rsmmr (commented):
Weren't we trying to get away from using bash? 

How about instead removing `set -e` and handling failures as suitable?

bbannier:
> How about instead removing `set -e` and handling failures as suitable?

The issue is not that we error too much due to `set -e`, but that we error too little. We were executing code like 
```
cmake | tee config.log
```
This both shows updating CMake output as well as stores the output in `config.log`. If `cmake` fails, this does not error since the `tee` call in the pipe very likely succeeds. With something like `pipefail` we would propagate a CMake failure up and then have the script about due to `-e`.

Since POSIX sh does not have pipefail we need to explicitly simulate it which is [definitely possible](https://github-wiki-see.page/m/koalaman/shellcheck/wiki/SC3040), but much more lengthy and would in principle need to be used everywhere a pipe is involved.

I'll move this script back to sh and will handle just the pipe failure from `cmake` for now.

> Weren't we trying to get away from using bash?

Yes, probably due to e.g., `bash` being aliased to `dash` on some systems (they have in fact different options). Writing robust shell scripts in POSIX sh is hard, so I wonder whether we shouldn't default to e.g., Python scripts instead when writing new code.

rsmmr (approved):

PR #1078, opened 2021-12-14
Make random-access functionality available by default

### LINKED ISSUE #1074: Explore enabling `%random-access` by default
Now that we can optimize away unused features we try making `%random-access` implicitly on. What is missing here is guarding random random-access dependent branches (e.g., updating offsets, trimming or not). Since in some places in the code we do not generate just extra code for `%random-access`, but different code depending on whether the feature is on or not, we might need to look whether constant folding in conditionals also needs to be updated to allow for that.

bbannier:
Closes #1074.


spicy/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -286,6 +286,8 @@ struct VisitorPost : public hilti::visitor::PreOrder<void, VisitorPost>, public
         if ( i.id().str() == "%random-access" ) {
             if ( i.expression() )
                 error("%random-access does not accept an argument", p);
+
+            hilti::logger().deprecated("%random-access is deprecated", i.meta().location());
```
rsmmr, line 290:
```
"%random-access is no longer needed and hence deprecated"
```

rsmmr (approved):
Nice, this looks straight-forward!

PR #1081, opened 2021-12-15
Document visibility for globals [skip CI].

bbannier:
Globals can have default or `public` visibility. This patch adds
documentation for declaring them `public`.

### DIFF
--- a/doc/programming/language/variables.rst
+++ b/doc/programming/language/variables.rst
@@ -8,16 +8,17 @@ Variables and Constants
 At the global module level, we declare variables with the ``global``
 keyword::
 
-    global NAME: TYPE [= DEFAULT];
+    [public] global NAME: TYPE [= DEFAULT];
 
-This defines a global variable called ``NAME`` with type ``TYPE``. If
-a default is giving, Spicy initialized the global accordingly before
-any code executes. Otherwise, the global received a type-specific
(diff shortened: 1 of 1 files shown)


doc/programming/language/variables.rst
```diff
@@ -8,16 +8,17 @@ Variables and Constants
...
-a default is giving, Spicy initialized the global accordingly before
-any code executes. Otherwise, the global received a type-specific
-default, typically the type's notion of a null value. As a result,
-globals are always initialized to a well-defined value.
+This defines a global variable called ``NAME`` with type ``TYPE``. If the
+variable is declared with ``public`` visibility other modules can reference it.
+If a default is giving, Spicy initialized the global accordingly before any
+code executes. Otherwise, the global received a type-specific default,
+typically the type's notion of a null value. As a result, globals are always
+initialized to a well-defined value.
```
rsmmr, line 18:
```suggestion
This defines a global variable called ``NAME`` with type ``TYPE``. If the
variable is declared with ``public`` visibility other modules can reference it.
If a default is given, Spicy initializes the global accordingly before any
code executes. Otherwise, the global receives a type-specific default,
typically the type's notion of a null value. As a result, globals are always
initialized to a well-defined value.
```

rsmmr (approved):

PR #1083, opened 2021-12-15
Decouple whether Spicy was build with ccache from ccache use during JIT.

### LINKED ISSUE #1045: Current behavior of default HILTI `cxx_launcher` is awkward
We currently bake whatever `CMAKE_CXX_COMPILER_LAUNCHER` was used to build Spicy itself into the HILTI config and then attempt to reuse it during JIT. This e.g., leads to the issue that our packages produced in CI (with as an implementation detail uses ccache) require an installed `ccache` binary at the exact same path to be usable or else JIT fails with an error (e.g., during `spicy-plugin`'s test suite with `fatal error: error during compilation: <Spicy Plugin for Zeek>: process '/usr/bin/ccache ' failed to start: No such file or directory`). This all feels awkward.

We probably should decouple what was used during the Spicy build from what is used during JIT. This could e.g., happen by introducing new CMake flags (e.g., by adding a new user-accessible CMake variable giving the path to the launcher); we could also expose this for `configure`.

bbannier:
Closes #1045.

### DIFF
--- a/configure
+++ b/configure
@@ -50,24 +50,25 @@ usage="\
 Usage: $0 [OPTION]... [VAR=VALUE]...
 
   Build Options:
-    --build-dir=DIR                    Place build files in directory [default: ${cmake_build_directory}]
-    --build-static-libs                Build static libraries instead [default: shared]
-    --build-toolchain={yes,no}         Build the Spicy compiler toolchain [default: ${cmake_build_toolchain}]
-    --build-type=TYPE                  Set build type (Debug,Release,RelWithDebInfo) [default: ${cmake_build_type}]
-    --disable-gold                     On Linux, do not try to use the gold linker
-    --disable-precompiled-headers      Disable use of precompiled headers for developer tests
-    --enable-ccache                    Build using the compiler cache cache if in PATH [default: ${cmake_use_ccache}]
-    --enable-debug                     Compile debug version (same as --build-type=Debug) [default: off]
-    --enable-sanitizer[=<names>]       Enable sanitizer(s), default if not further specified is \"address\"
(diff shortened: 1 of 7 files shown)


doc/toolchain.rst
```diff
@@ -55,10 +55,11 @@ control the compilation process:
     ``HILTI_CXX_COMPILER_LAUNCHER``
         Specifies a command to prefix compiler invocations with during JIT.
         This can e.g., be used to use a compiler cache like
-        `ccache <https://ccache.dev/>`_. If ``ccache`` was used during the
-        build ``ccache`` will automatically be used during JIT as well. Setting
-        this variable to an empty value disables use of ``ccache`` in that
-        case.
+        `ccache <https://ccache.dev/>`_. If Spicy was configured with e.g.,
+        ``--with-hilit-compiler-launcher=ccache`` (the equivalent CMake option
```
rsmmr, line 59:
```suggestion
        ``--with-hilti-compiler-launcher=ccache`` (the equivalent CMake option
```

configure
```diff
@@ -65,6 +65,7 @@ Usage: $0 [OPTION]... [VAR=VALUE]...
     --with-bison=<prefix>              Set prefix of Bison installation
     --with-c-compiler=<path>           Set C compiler to use
     --with-cxx-compiler=<path>         Set C++ compiler to use
+    --with-hilti-compiler-launcher=*)  Set compiler launcher to use during JIT, e.g., ccache.
```
rsmmr, line 68:
```suggestion
    --with-hilti-compiler-launcher=<cmd>  Set compiler launcher to use during JIT, e.g., ccache.
```

rsmmr (approved):

PR #1091, opened 2021-12-20
Ensure that field hooks for lists see correct offset

### LINKED ISSUE #1088: `offset` and `position` in list `%done` hook always corresponds to offset of last element
According to the docs `offset` should return the offset from where the current field was parsed. This does not seem to hold for lists.

```.ruby
# foo.spicy
module foo;

public type X = unit {
    xs: uint8[]
        foreach {
            print "element %s [%d]" % ($$, self.offset());
        }
        {
            print "list %s [%d]" % ($$, self.offset());
        }
};
```

```.console
$ printf '\x01\x02\x03\x04\x05\x06\x07\x08' | spicy-driver -d foo.spicy
element 1 [0]
element 2 [1]
element 3 [2]
element 4 [3]
element 5 [4]
element 6 [5]
element 7 [6]
element 8 [7]
list [1, 2, 3, 4, 5, 6, 7, 8] [7]
```

I would have expected the `%done` hook to see a value of `0` for `self.offset()`.

While having special semantics of `offset` in lists seems interesting, deviating from the usual semantics seems to hold a lot of potential for confusion, especially if the grammar contains aliases for field declarations where fields which look like they hold units might actually hold lists. We should fix this.

bbannier:
Closes #1088.


spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -547,12 +548,21 @@ struct ProductionVisitor
...
         const auto& field = meta.field();
         assert(field); // Must only be called if we have a field.
 
         HILTI_DEBUG(spicy::logging::debug::ParserBuilder, fmt("- pre-parse field: %s", field->id()));
 
+        // If the field holds a container we expect to see the offset of the field, not the individual container
+        // elements inside e.g., this unit's fields hooks. Store the value before parsing of a container starts so we
+        // can restore it later.
+        std::optional<Expression> pre_container_offset;
+        if ( field && type::isIterable(field->parseType()) && ! meta.container() ) {
```
rsmmr, line 561: Not completely sure, but I think it should work to use this:

```suggestion
        if ( field->isContainer() ) {
```

rsmmr (approved):

configure
```diff
@@ -32,13 +32,13 @@ command="$0 $*"
 
 cmake_exe="<no cmake>"
 for i in cmake cmake3; do
-    if command -v $i >/dev/null; then
```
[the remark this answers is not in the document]
bbannier: I would have thought so to, but `checkbashisms.pl` does not like use of `command` with any flag but `-p`,

```
Check for bashisms in /bin/sh scripts....................................Failed
- hook id: checkbashisms
- exit code: 1

possible bashism in configure line 35 ('command' with option other than -p):
    if command -v $i >/dev/null; then
possible bashism in configure line 36 ('command' with option other than -p):
        $i --version | grep -q "cmake.*version 3" && cmake_exe=$(command -v $i)
possible bashism in configure line 41 (type):
type "${cmake_exe}" > /dev/null 2>&1 || {
possible bashism in configure line 125 ('command' with option other than -p):
              command -v "${try_c_compiler}" >/dev/null && cmake_c_compiler="${try_c_compiler}"
```

spicy/toolchain/src/compiler/codegen/parser-builder.cc
bbannier: This works well for the test I added, but behaves differently in that the offsets seen in the test `spicy.types.unit.trimming` does not change anymore. This seems better.

bbannier: I'll check for `field && field->isContainer()` here.

PR #1094, opened 2021-12-22
Do not pass errors inside sink units back up to parent.

### LINKED ISSUE #1076: Parse error from sink propagates to connected unit
Given the following code

```.ruby
module test;

type S = unit {
    x : /A/;
};

public type X = unit {
    : bytes &eod { self.s.write($$); }
    sink s;

    on %init { self.s.connect(new S); }
};
```

we can successfully forward data from `X` into `S`.

```.console
$ echo 'A' | spicy-driver test.spicy  -d
$ echo $?
0
```

If however parsing in the sink fails, we will fail even the outer unit
```.console
$ echo '1' | spicy-driver test.spicy  -d
[error] terminating with uncaught exception of type spicy::rt::ParseError: parse error: failed to match regular expression (test.spicy:4:9)
$ echo $?
1
```

This seems to be counter to the expected decoupling a `sink` should provide.

rsmmr:
Exceptions used to propagate back into the writer unit, which
contradicts the decoupling that sinks are meant to provide.

Closes #1076.


tests/spicy/types/sink/parse-error.spicy
```diff
@@ -0,0 +1,29 @@
+# @TEST-EXEC:  printf '1' | spicy-driver -d %INPUT >>output 2>&1
+# @TEST-EXEC:  printf 'A' | spicy-driver -d %INPUT >>output 2>&1
```
bbannier, line 2: Let's precompile the input into an HLTO file.
```suggestion
# @TEST-EXEC: spicyc -d -j -o test.hlto %INPUT
# @TEST-EXEC:  printf '1' | spicy-driver -d test.hlto >>output 2>&1
# @TEST-EXEC:  printf 'A' | spicy-driver -d test.hlto >>output 2>&1
```

tests/spicy/types/sink/parse-error.spicy
```diff
@@ -0,0 +1,29 @@
...
+module test;
+
+type S1 = unit {
+    x : /A/;
+    on %error { print "error in S1"; }
+    on %done { print self; }
+};
+
+type S2 = unit {
+    x : /./;
```
bbannier, line 16: Let's use a different name for this field so it is easier to distinguish it from `S1.x`.

tests/spicy/types/sink/parse-error.spicy
```diff
@@ -0,0 +1,29 @@
+# @TEST-EXEC:  printf '1' | spicy-driver -d %INPUT >>output 2>&1
```
bbannier, line 1: Could you add a test for `skip_delivery`, e.g., a test where we check that once a unit behind a sink enters an error state feeding it more data does not trigger additional errors (e.g., `more data after sink's unit has already completed parsing`)?

spicy/runtime/src/sink.cc
```diff
@@ -138,11 +138,26 @@ bool Sink::_deliver(std::optional<hilti::rt::Bytes> data, uint64_t rseq, uint64_
...
+        if ( s->skip_delivery )
+            continue;
+
         if ( s->resumable )
             throw ParseError("more data after sink's unit has already completed parsing");
 
         s->data->append(*data);
-        s->resumable.resume();
+        try {
+            // Sinks are operating indepdently from the writer, so we
```
bbannier, line 149:
```suggestion
            // Sinks are operating independently from the writer, so we
```

tests/spicy/types/sink/parse-error.spicy
rsmmr: Ok, extended the test case. 

Also added one more check for `skip_delivery` at closing time. Not sure how to trigger that, but seems it should be there.

bbannier (approved):

PR #1095, opened 2021-12-22
Fix self-recursive units.

### LINKED ISSUE #1076: Parse error from sink propagates to connected unit

### LINKED ISSUE #1077: Recursive units show weird behavior
It is currently possible to write recursive units, but I am wondering whether we properly support them; @keithjjones also reported running into surprising behavior when using them.

As a minimal example:
```.ruby
module foo;

public type X = unit {
    has_field: uint8 &convert=cast<bool>($$);
    x: X if (self.has_field);
};
```

This behaves as expected for no recursion:
```
$ printf '\x00' | HILTI_DEBUG=spicy-verbose spicy-dump e.spicy -d
[spicy-verbose] - state: type=foo::X input="\\x00" stream=0x7ffd3b004b48 offsets=0/0/1 chunks=1 frozen=yes mode=default trim=yes lah=n/a lah_token="n/a"
[spicy-verbose] - parsing production: Unit: foo_X -> has_field foo_X
[spicy-verbose]   - state: type=foo::X input="\\x00" stream=0x7ffd3b004b48 offsets=0/0/1 chunks=1 frozen=yes mode=default trim=yes lah=n/a lah_token="n/a"
[spicy-verbose]   - parsing production: Variable: has_field -> uint<8>
[spicy-verbose]     - trimming input
[spicy-verbose]   - setting field 'has_field' to 'false'
foo::X {
  has_field: False
}
```

But when adding even a single layer of recursion this seems to see data from the incorrect layer:
```
$ printf '\x01\x00' | HILTI_DEBUG=spicy-verbose spicy-dump e.spicy -d
[spicy-verbose] - state: type=foo::X input="\\x01\\x00" stream=0x7fece6e047a8 offsets=0/0/2 chunks=1 frozen=yes mode=default trim=yes lah=n/a lah_token="n/a"
[spicy-verbose] - parsing production: Unit: foo_X -> has_field foo_X
[spicy-verbose]   - state: type=foo::X input="\\x01\\x00" stream=0x7fece6e047a8 offsets=0/0/2 chunks=1 frozen=yes mode=default trim=yes lah=n/a lah_token="n/a"
[spicy-verbose]   - parsing production: Variable: has_field -> uint<8>
[spicy-verbose]     - trimming input
[spicy-verbose]   - setting field 'has_field' to 'true'
[spicy-verbose]   - state: type=foo::X input="\\x00" stream=0x7fece6e047a8 offsets=1/1/2 chunks=1 frozen=yes mode=default trim=yes lah=n/a lah_token="n/a"
[spicy-verbose]   - parsing production: Unit: foo_X -> has_field foo_X
[spicy-verbose]     - state: type=foo::X input="\\x00" stream=0x7fece6e047a8 offsets=1/1/2 chunks=1 frozen=yes mode=default trim=yes lah=n/a lah_token="n/a"
[spicy-verbose]     - parsing production: Variable: has_field -> uint<8>
[spicy-verbose]     - setting field 'has_field' to 'false'
[fatal error] terminating with uncaught exception of type hilti::rt::UnsetOptional: unset optional value (e.spicy:4:5)
```

We should look into this.

rsmmr:
There's was a plain bug in the parser generator that would create broken
code if a field had the same type as the current unit.

Closes #1077.


tests/spicy/types/unit/self-recursive.spicy
```diff
@@ -0,0 +1,14 @@
+# @TEST-EXEC: spicyc -j -d -o test.hlto %INPUT
+# @TEST-EXEC: ${SCRIPTS}/printf '\x00' | spicy-dump test.hlto >>output
+# @TEST-EXEC: ${SCRIPTS}/printf '\x01\x00' | spicy-dump test.hlto >>output
```
bbannier, line 3: This seems to work well, but I think it would be great to add a test with one more level of nesting given that in the impl we just use a boolean `top_level`, e.g.,

```
# @TEST-EXEC: ${SCRIPTS}/printf '\x01\x01\x00' | spicy-dump test.hlto >>output
```

bbannier (approved):

bbannier:
Let's also fix the commit message and the description to reference the correct issue #1077.

bbannier (approved):

PR #1097, opened 2021-12-22
Fix `inout` parameters for units

### LINKED ISSUE #1069: Cannot pass references down to subparsers over more than one layer
The following code

```.ruby
module test;

public type Ctx = uint32;

public type X = unit {
    %context = Ctx;
    y: Y(self.context());
};

type Y = unit(inout ctx: Ctx&) {
    z: Z(ctx);
};

type Z = unit(inout ctx: Ctx&) {};
```

fails with

```
$ spicyc -d -j -g bla.spicy
/private/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/test_3c791935d043232-3d5fab3cd92df539.cc:514:56: error: no matching conversion for functional-style cast from 'ValueReference<SafeInt<unsigned int, hilti::rt::integer::detail::SafeIntException> >' to '__hlt::test::Z'
    ::hilti::rt::optional::valueOrInit((*__self).z) = (__hlt::test::Z(__self->__p_ctx.derefAsValue()));
                                                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/private/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/test_3c791935d043232-3d5fab3cd92df539.cc:113:9: note: candidate constructor not viable: no known conversion from 'ValueReference<SafeInt<unsigned int, hilti::rt::integer::detail::SafeIntException> >' to 'const __hlt::test::Z' for 1st argument
        Z(const Z&) = default;
        ^
/private/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/test_3c791935d043232-3d5fab3cd92df539.cc:114:9: note: candidate constructor not viable: no known conversion from 'ValueReference<SafeInt<unsigned int, hilti::rt::integer::detail::SafeIntException> >' to '__hlt::test::Z' for 1st argument
        Z(Z&&) = default;
        ^
/private/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/test_3c791935d043232-3d5fab3cd92df539.cc:219:15: note: candidate constructor not viable: no known conversion from 'ValueReference<SafeInt<unsigned int, hilti::rt::integer::detail::SafeIntException> >' to 'const ::hilti::rt::StrongReference< ::hilti::rt::integer::safe<uint32_t> >' (aka 'const StrongReference<SafeInt<unsigned int, detail::SafeIntException> >') for 1st argument
    inline Z::Z(const ::hilti::rt::StrongReference<::hilti::rt::integer::safe<uint32_t>> __p_ctx) : __p_ctx(std::move(__p_ctx)) {
              ^
/private/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/test_3c791935d043232-3d5fab3cd92df539.cc:216:15: note: candidate constructor not viable: requires 0 arguments, but 1 was provided
    inline Z::Z() {
              ^
1 error generated.
```

```
$ spicyc --version
spicyc v1.4.0-dev.95 (a20c2573)
```

This works as expected if I e.g., declare `Y` empty instead,

```.ruby
type Y = unit(inout ctx: Ctx&) {};
```

I am wondering whether this could be related to #674.

In any case, this should either work, or be rejected by us (not the C++ compiler).

rsmmr:
Closes #1069.


hilti/toolchain/src/compiler/codegen/codegen.cc
```diff
@@ -554,6 +554,16 @@ hilti::Unit* CodeGen::hiltiUnit() const {
     return _hilti_unit;
 }
 
+codegen::TypeUsage CodeGen::parameterKindToTypeUsage(declaration::parameter::Kind k) {
+    switch ( k ) {
+        case declaration::parameter::Kind::Copy: return codegen::TypeUsage::CopyParameter;
+        case declaration::parameter::Kind::In: return codegen::TypeUsage::InParameter;
+        case declaration::parameter::Kind::InOut: return codegen::TypeUsage::InOutParameter;
+        case declaration::parameter::Kind::Unknown: logger().internalError("parameter kind not set");
+        default: util::cannot_be_reached();
```
bbannier, line 563: Do we need this here? I would much prefer to reduce the number of `default` cases as much as possible as during updates to the enum they can lead to silently compiling but misbehaving code.

bbannier (approved):

hilti/toolchain/src/compiler/codegen/codegen.cc
rsmmr: I'll move it to behind the switch. The problem is that GCC has trouble recognizing that all enum values have been covered and will warn about not returning something.

rsmmr:
I can't reproduce that CI error locally, running that in Docker works just fine for me now. Let's see if another push helps ...

rsmmr:
> I can't reproduce that CI error locally, running that in Docker works just fine for me now. Let's see if another push helps ...

Oh, I take that back. HTTP works, but I see an issue with IPsec now. Will look into it.

bbannier (approved):

PR #1098, opened 2022-01-06
Fix typos in the source and docs.

bbannier:
This patch was created by running
[typos-cli](https://github.com/crate-ci/typos) from a pre-commit hook,
adjusting the run by creating a whitelist to prevent undesirable
corrections (currently only contains `inout` which by default would
become `input`), and manually dealing with issues where the edit was
incomplete. That spell checker fixes strings in a larger number of
source types and also will check filenames.


doc/programming/language/types.rst
```diff
@@ -290,7 +290,7 @@ Spicy provides POSIX-style regular expressions.
 
 .. rubric:: Constants
 
-- ``/Foo*ba?r/``, ``/X(..)(..)(..)Y/``
+- ``/Foo*bar?/``, ``/X(..)(..)(..)Y/``
```
rsmmr, line 293: (Doesn't really matter, but not a typo.)

hilti/runtime/src/tests/util.cc
```diff
@@ -490,9 +490,9 @@ TEST_CASE("replace") {
 TEST_CASE("rtrim") {
     CHECK_EQ(rtrim("", ""), "");
     CHECK_EQ(rtrim("", "abc"), "");
-    CHECK_EQ(rtrim("4d3c2b1a", "abc"), "4d3c2b1");
-    CHECK_EQ(rtrim("4d3c2b1ba", "abc"), "4d3c2b1");
-    CHECK_EQ(rtrim("4d3c2b1cba", "abc"), "4d3c2b1");
+    CHECK_EQ(rtrim("4d3c2b1c", "abc"), "4d3c2b1");
+    CHECK_EQ(rtrim("4d3c2b1bc", "abc"), "4d3c2b1");
+    CHECK_EQ(rtrim("4d3c2b1abc", "abc"), "4d3c2b1");
```
rsmmr, line 495: What's the reason for this change?

rsmmr (dismissed):
Nice job!

doc/programming/language/types.rst
bbannier: This is working around `typos` considering `ba` to be a misspelling,

```.console
error: `ba` should be `by`, `be`
  --> doc/programming/language/types.rst:293:9
    |
293 | - ``/Foo*ba?r/``, ``/X(..)(..)(..)Y/``
    |          ^^
    |
```

The other place you flagged below has a similar reason. I also fixed a few places where we used `nope` and `noe` to use more common placeholder strings (`typos` seems to understand e.g., `foo` and `bar`).

I decided to work around such diagnostics instead of whitelisting whole files. At least from this initial pass that seemed acceptable; I'd suggest we try going with this and possibly using different strings where possible and only look into whitelisting if things become too annoying.

hilti/runtime/src/tests/util.cc
bbannier: This is a workaround for this diagnostic:

```.console
error: `ba` should be `by`, `be`
  --> hilti/runtime/src/tests/util.cc:494:27
    |
494 |     CHECK_EQ(rtrim("4d3c2b1ba", "abc"), "4d3c2b1");
    |                            ^^
    |
```

Issue #1084, opened 2021-12-16
Incorrect key type in `map` index operator causes compiler error


bbannier:
Currently the following code passes validation, but fails during compilation:

```.ruby
module foo;

global m: map<tuple<uint64, uint64>, uint64> = map((0,0): 0);
m[0];
```

The error users see looks something like this:

```.console
$ spicyc -d j foo.spicy
/private/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/hilti-rt-test-pn4imh/foo.cc:31:24: error: no viable overloaded operator[] for type '::hilti::rt::Map<std::tuple< ::hilti::rt::integer::safe<uint64_t>, ::hilti::rt::integer::safe<uint64_t> >,
      ::hilti::rt::integer::safe<uint64_t> >' (aka 'Map<tuple<SafeInt<unsigned long long, detail::SafeIntException>, SafeInt<unsigned long long, detail::SafeIntException> >, SafeInt<unsigned long long, detail::SafeIntException> >')
    foo::__globals()->m[::hilti::rt::integer::safe<std::uint64_t>{1u}];
    ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/local/Cellar/spicy/1.3.0/include/hilti/rt/types/map.h:231:11: note: candidate function not viable: no known conversion from '::hilti::rt::integer::safe<std::uint64_t>' (aka 'SafeInt<unsigned long long, detail::SafeIntException>') to 'const
      std::__1::tuple<SafeInt<unsigned long long, hilti::rt::integer::detail::SafeIntException>, SafeInt<unsigned long long, hilti::rt::integer::detail::SafeIntException> >' for 1st argument
    auto& operator[](const K& k) & { return this->get(k); }
          ^
/usr/local/Cellar/spicy/1.3.0/include/hilti/rt/types/map.h:239:17: note: candidate function not viable: no known conversion from '::hilti::rt::integer::safe<std::uint64_t>' (aka 'SafeInt<unsigned long long, detail::SafeIntException>') to 'const
      std::__1::tuple<SafeInt<unsigned long long, hilti::rt::integer::detail::SafeIntException>, SafeInt<unsigned long long, hilti::rt::integer::detail::SafeIntException> >' for 1st argument
    const auto& operator[](const K& k) const& { return this->get(k); }
                ^
[...]

PR #1103, opened 2022-01-18
Specify type of key with map vector operations.

rsmmr:
The fix itself for #1084 was straight-forward: in the operator signature, we need to derive the real key type instead of just say `any`. But there was a reason why we didn't do that already: it led to an ambiguity that now the operator resolver couldn't sometimes tell anymore which index operator to use (const vs non-const). That's been long-time shortcoming that this PR fixes now: it allows to prioritize one operator over the other to resolve such ambiguities.


hilti/toolchain/src/compiler/visitors/resolver.cc
```diff
@@ -928,36 +928,57 @@ std::vector<Node> Visitor::matchOverloads(const std::vector<Operator>& candidate
...
+                if ( priority != c.priority() )
+                    // Not looking at operators of this priority right now.
+                    continue;
 
-        for ( const auto& c : candidates ) {
-            HILTI_DEBUG(logging::debug::Operator, util::fmt("candidate: %s", c.typename_()));
-            logging::DebugPushIndent _(logging::debug::Operator);
+                if ( priority == operator_::Priority::Low && kinds_resolved->find(c.kind()) != kinds_resolved->end() )
+                    // Already have a higher priority match for this operator kind.
+                    continue;
```
bbannier, line 947: We already enforce some ordering outside, suggest to drop the check for `Low` here,
```suggestion
                if ( kinds_resolved->find(c.kind()) != kinds_resolved->end() )
                    // Already have a previous match for this operator kind.
                    continue;
```

bbannier (approved):

hilti/toolchain/src/compiler/visitors/resolver.cc
bbannier: You are right, we mutate `kinds` below.

PR #1104, opened 2022-01-18
Fix import of global constants.

### LINKED ISSUE #1079: Referencing `public` `const` from other modules triggers C++ compiler error
We can reference a `global` from another module if it is declared `public`. The docs suggest that a `const` should behave similarly, but this leads to a C++ compiler error.

```.ruby
# a.spicy
module a;

public global x: uint32 = 64;
public const y: uint32 = 64;
```

```.ruby
# b.spicy
module b;

import a;
print(a::x); # Works on its own.
print(a::y); # Fails during C++ compile.
```

```.console
$ spicyc -j b.spicy
/private/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/hilti-rt-test-IUGmp0/b.cc:29:25: error: no member named 'y' in namespace '__hlt::a'; did you mean simply 'y'?
    ::hilti::rt::print((__hlt::a::y), ::hilti::rt::Bool(true));
                        ^~~~~~~~~~~
                        y
/private/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/hilti-rt-test-IUGmp0/b.cc:20:48: note: 'y' declared here
    const ::hilti::rt::integer::safe<uint32_t> y = ::hilti::rt::integer::safe<std::uint32_t>{64u};
                                               ^
1 error generated.
```

I would have expected a `const` to behave identical to a `global` here.

rsmmr:
We were treating them as globals (instead of constants) and weren't
emitting them into the generated C++ code.

Closes #1079.

### DIFF
--- a/tests/hilti/codegen/imported-constants.hlt
+++ b/tests/hilti/codegen/imported-constants.hlt
@@ -1,0 +1,16 @@
+# @TEST-EXEC: hiltic -j %INPUT
+# @TEST-DOC: Check usage imported constants. Regression test for #1079.
+
+module B {
+
+import A;
+
+assert A::x == 64;
+
+}
+
+@TEST-START-FILE a.hlt
+module A {
+public const uint<32> x = 64;
+}
+@TEST-END-FILE
--- a/hilti/toolchain/src/compiler/codegen/codegen.cc
+++ b/hilti/toolchain/src/compiler/codegen/codegen.cc
@@ -38,7 +38,7 @@ struct GlobalsVisitor : hilti::visitor::PreOrder<void, GlobalsVisitor> {
     bool include_implementation;
 
     std::vector<cxx::declaration::Global> globals;
-    std::vector<cxx::declaration::Global> constants;
+    std::vector<cxx::declaration::Constant> constants;
 
     static void addDeclarations(CodeGen* cg, const Node& module, const ID& module_id, cxx::Unit* unit,
                                 bool include_implementation) {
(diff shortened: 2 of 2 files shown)


tests/hilti/codegen/imported-constants.hlt
```diff
@@ -0,0 +1,16 @@
+# @TEST-EXEC: hiltic -j %INPUT
```
bbannier, line 1: Small suggestion, let's extend this test to check we (at least for some example basic type like an integer) we export enough information on the C++ level to generate constants which contain expressions containing e.g., non-public constants. This works right now since visibility seems to only play into validation, but not directly into the generated code.

```.ruby
module A {

public const uint<32> x = 1;
const uint<32> y = 2;
public const uint<32> z = x + y;

}

module B {

import A;

assert A::x == 1;
assert A::z == 3;

}
```

bbannier (approved):

tests/hilti/codegen/imported-constants.hlt
rsmmr: Just tried that with an additional non-public const, and it doesn't work--which isn't really surprising I guess.

 I think what we should do is enforce that expressions initializing constants are sufficiently self-contained so that they work anywhere. The alternative would requite pre-computing them and storing them somewhere, which would essentially make them globals that don't have the performance benefit of constants anymore.

I'll file a separate ticket for that.

rsmmr: #1106

PR #1117, opened 2022-02-02
Add documentation for error recovery

bbannier:
This adds documentation for the feature implemented in #1054.


bbannier:
I have folded this PR into #1054.

PR #1128, opened 2022-02-21
Disallow enum declarations with non-unique values.

### LINKED ISSUE #1072: Enum with duplicate values not rejected
We currently allow the following code

```.ruby
# foo.spicy
module foo;

public type E = enum {
    A = 1,
    B = 1,
};
```

It is unclear what this code should produce when constructing from an integer `1`.

Additionally this also seems to cause issue in zeek/spicy-plugin when translating this `public` enum to a Zeek enum,

```.console
$ spicyz foo.spicy -o foo.hlt
$ ZEEK_SPICY_MODULE_PATH=$PWD zeek -Cr ../tests/traces/trace.pcap
error in /usr/local/Cellar/zeek/4.1.1_1/share/zeek/base/init-bare.zeek, line 1: enumerator value in enumerated type definition already exists
...
```

Instead of trying to work around this on the plugin side we probably should just reject such enum declarations.

bbannier:
It is unclear what code should be generated when requested to convert an
integer value to the following enum:

    type E = enum {
        A = 1,
        B = 2,
        C = 1,
    };

For `1` we could produce either `E::A` or `E::C` here.

Instead of allowing that ambiguity we disallow enums with non-unique
values with this patch.

Closes #1072.


rsmmr (approved):
Let's make sure to mention this in NEWS.

Issue #1050, opened 2021-11-16
Locations for attributes are off by one


bbannier:
For the following unit 

```
module test;

public type X = unit {
    x: uint8 { self.x = 0; }
    y: bytes &size=self.x - 1;
};
```
we show the following output:

```
$ printf '\x00\x00' | spicy-driver test.spicy
[error] terminating with uncaught exception of type hilti::rt::Overflow: integer overflow (test.spicy:4:15)
```

The underflowing code is on line 5, but the error messages references line 4. This seems to be due to us computing the attribute value before pushing a new location.

* * *
We see a similar issue in [this patch](https://github.com/zeek/spicy/pull/1130#discussion_r812100890). Ideally we could clean this up with this issue as well.

PR #1130, opened 2022-02-21
Add validation rejecting untyped variable decl from struct initializer.

### LINKED ISSUE #1043: Add validation rejecting untyped use of struct initializers
With #1039 we added struct initializers to Spicy (previously they were only available in HILTI code).

We currently do not reject the following code which has no type:

```.r
global foo = [$x=1];
```

This fails at C++ codegen time with the following error:

```.console
<...>/test_3ed9cdf152bbd684-f65cd59cac3632fc.cc:10:5: error: declaration of anonymous struct must be a definition
    struct ;
    ^
```

bbannier:
In order to create a concrete value we require a type. Since struct
initializers are untyped one either needs to wrap the initializer in a
type, or type the LHS.

Closes #1043.


tests/Baseline/spicy.types.struct.init-fail/output
```diff
@@ -0,0 +1,4 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+[error] <...>/init-fail.spicy:12:25-13:22: declaration needs a concrete struct type
```
rsmmr, line 2: Unrelated to this change, but that error location is a bit off.

rsmmr (approved):

tests/Baseline/spicy.types.struct.init-fail/output
bbannier: Not sure it is exactly the same issue, but I linked this in #1050.

PR #1144, opened 2022-03-01
Add support for recoverying from gaps

### LINKED ISSUE #1136: Relax handling of partial connections in spicy-plugin
Now that a first iteration of error recovery has [landed in in Spicy](https://github.com/zeek/spicy/pull/1054) we should revisit how spicy-plugin handles partial connections/undelivered data. It might even be possible to implement the first pieces of this before zeek/spicy#1135 has been addressed.

The old prototype implemented this by injecting a special stream chunk for missing data. Once the parser hits that chunk it could automatically enter search mode (i.e., w/o waiting to first hit the parser error).

Implementation plan for zeroth iteration:
- teach gaps to `Stream`
- if parser hits gap trigger parse error
- spicy-plugin passes on data with undelivered data/partial connections
- trigger usual sync handling if unit supports it, or propagate up

bbannier:
Closes #1136.


hilti/runtime/include/types/stream.h
```diff
@@ -85,8 +90,7 @@ class Chunk {
     using Array = std::pair<Size, std::array<Byte, SmallBufferSize>>;
     using Vector = std::vector<Byte>;
 
-    Chunk(Offset o, std::array<Byte, SmallBufferSize>&& d, Size n)
-        : _offset(o), _data(std::make_pair(n, std::move(d))) {}
+    Chunk(Offset o, std::array<Byte, SmallBufferSize> d, Size n) : _offset(o), _data(std::make_pair(n, d)) {}
```
rsmmr, line 93: Why this change?

hilti/runtime/include/types/stream.h
```diff
@@ -105,24 +112,28 @@ class Chunk {
         _offset = other._offset;
         _data = std::move(other._data);
         _next = std::move(other._next);
-        _chain = std::move(other._chain);
+        _chain = other._chain;
```
rsmmr, line 115: Similar question: what's the reason for this change?

hilti/runtime/include/types/stream.h
```diff
@@ -105,24 +112,28 @@ class Chunk {
...
         return *this;
     }
 
     ~Chunk() = default;
 
     Offset offset() const { return _offset; }
     Offset endOffset() const { return _offset + size(); }
-    bool isCompact() const { return std::holds_alternative<Array>(_data); }
+    bool isCompact() const { return ! std::holds_alternative<Vector>(_data); }
```
rsmmr, line 123: According to a quick grep, it doesn't look like we're using this method anywhere. Given that the semantics are getting a bit odd now, I'd just remove it.

hilti/runtime/include/types/stream.h
```diff
@@ -220,16 +237,16 @@ class Chunk {
         if ( n <= Chunk::SmallBufferSize ) {
             std::array<Byte, Chunk::SmallBufferSize> x{};
             std::copy(ud, ud + n.Ref(), x.data());
-            return Chunk(o, std::move(x), n);
+            return Chunk(o, x, n);
```
rsmmr, line 240: Once more: wondering about this change?

hilti/runtime/include/types/stream.h
```diff
@@ -1410,9 +1434,9 @@ class Stream {
      */
     explicit Stream(const char* d) : Stream(Chunk(0, d, strlen(d))) {}
 
-    /** Creates an instance from an existing memory block. The data will be copied. */
-    Stream(const char* d, Size n) : Stream(Chunk(0, d, n)) {}
-
+    /** Creates an instance from an existing memory block. The data will be copied if set, otherwise a gap will be
+     * recorded. */
```
rsmmr, line 1443: (Fix comment formatting.)

hilti/runtime/src/types/stream.cc
```diff
@@ -173,6 +173,27 @@ int Chain::numberOfChunks() const {
...
+        if ( ! c )
+            break;
+
+        // Try again at the offset of the next chunk.
+        o = c->offset();
+    }
+
+    // If we haven't found a non-gap chunk simply return a view at the next
+    // byte. Since this is a gap chunk this can cause recovery in the caller.
+    return advance(1U);
```
rsmmr, line 194: I find this logic hard to follow because it's mixing operation on chunks with offset-based computation. I believe it's also not the most efficient. Would it be easier to create an iterator `i = unsafeBegin(); ++i` at the beginning, then work with the chunk `c` that that's pointing to & update `i` accordingly each round; and eventually return a new view `(i, end())`?

rsmmr (commented):
For the 1st commit, I ended up wondering about a couple of move-semantics changes; no need to change anything there if those are all as intended, just checking.

hilti/runtime/src/types/stream.cc
bbannier: Simplified liked discussed out of band.

hilti/runtime/include/types/stream.h
bbannier: This fixes a clang-tidy warning about an unnecessary `move` since the array here is trivially copyable. I'll move this to its own patch.

hilti/runtime/include/types/stream.h
bbannier: This is related to the signature change in the constructor above which cleans up a clang-tidy warning. I'll move this to its own patch.

hilti/runtime/include/types/stream.h
bbannier: This fixes a clang-tidy warning about an unnecessary `move` since the array here is trivially copyable. I'll move this to its own patch.

hilti/runtime/include/context.h
```diff
@@ -85,20 +85,13 @@ extern Context*& current();
 extern Context* master();
 
 /** Returns the context set for the current hardware thread. */
-inline auto get() {
-    assert(current());
```
rsmmr, line 89: Suggestion: Rather than completely disabling the null check, how about adding a boolean parameter `allow_missing_context` with default `false`. We'd then set that to true from just those two locations where it can happen for benign reasons.

hilti/runtime/include/types/stream.h
```diff
@@ -1389,7 +1389,14 @@ class View {
...
 inline std::ostream& operator<<(std::ostream& out, const View& x) {
-    out << escapeBytes(x.data().str());
+    Bytes data;
+    try {
+        data = x.data();
+    } catch ( hilti::rt::MissingData& ) {
+        data.append("<missing data>");
+    }
+
+    out << escapeBytes(data.str());
```
rsmmr, line 1399: As discussed, we should change this to insert `<gap>` (or so) into the rendered data, instead of fully replacing it.

rsmmr (dismissed):

rsmmr:
"Approve" is for "changes so far", let's still to the review of if we're handling exceptions correctly internally,  potentially with extending/changing the internal API for better control.

rsmmr (dismissed):
Looks good on the final changes. Only thing is that it'd be good to add a test for `dataForPrint()`, didn't see that.

PR #1148, opened 2022-03-08
Fix matching of multiple regexps to prefer shortest match.

### LINKED ISSUE #1114: Lookahead parsing prefers longest instead of shortest match
Given the following unit

```.ruby
module x;

public type X = unit {
	switch {
		-> a: /A/;
		-> b: /ABC/;
	};

	rest: bytes &eod;

	on %done { print self; }
};
```

the generated lookahead parsing seems to prefer to match the longest, not the shortest match,

```.console
$ printf 'ABC' | spicy-driver x.spicy
[$a=(not set), $b=b"ABC", $rest=b""]
```

I see this behavior also if using incremental parsing with chunks with `size=1`.

bbannier:
If multiple regexps match at the same left-most offset, justrx would
previously prefer the longest instead of the shortest match. These are
not the expected and documented semantics for Spicy.

This patch bumps justrx to a version which has the expected semantics.

Closes #1114.

### DIFF
--- a/hilti/runtime/src/tests/regexp.cc
+++ b/hilti/runtime/src/tests/regexp.cc
@@ -70,6 +70,11 @@ TEST_CASE("match") {
         CHECK_EQ(RegExp("ab+c", regexp::Flags{.use_std = 1}).match("abbbcdef"_b), 1);
         CHECK_EQ(RegExp("ab+c", regexp::Flags{.use_std = 1}).match("012abbbc345"_b), 0);
     }
+
+    SUBCASE("prefers shorted match") {
+        CHECK_EQ(RegExp(std::vector<std::string>{"A{#1}", "ABC{#2}"}).match("ABC"_b), 1);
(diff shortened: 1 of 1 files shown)


hilti/runtime/src/tests/regexp.cc
```diff
@@ -70,6 +70,11 @@ TEST_CASE("match") {
         CHECK_EQ(RegExp("ab+c", regexp::Flags{.use_std = 1}).match("abbbcdef"_b), 1);
         CHECK_EQ(RegExp("ab+c", regexp::Flags{.use_std = 1}).match("012abbbc345"_b), 0);
     }
+
```
bbannier, line 73: We still need to update the revision we bump the justrx submodule to after https://github.com/rsmmr/justrx/pull/2 is merged.

rsmmr:
Do we want to close this for now? Seems there's more going on for #1114.

bbannier:
> Do we want to close this for now? Seems there's more going on for #1114.

Will close this and rsmmr/justrx#2 for now.

PR #1149, opened 2022-03-09
Switch to ECR images in documentation [skip CI].

### LINKED ISSUE #1073: Document move of containers to ECR
We currently publish container images to both the Docker and ECR registry. With spicy-1.4 we switch over to ECR only, and also document this.

bbannier:
Closes #1073.


bbannier:
The image for release tags `zeek/spicy` is only published for tags by CI. Since at least I cannot push that image manually we should hold off merging this until around a time where published images are actually available to users, e.g., to after we have cut the release.

rsmmr (approved):
I'll clean up the ECR views a bit.

PR #1152, opened 2022-03-21
Preserve additional permissions from umask when generating HLTO files.

### LINKED ISSUE #1150: generated hlto file permissions are 711 (-rwx--x--x)
This issue cropped up with the latest release of Spicy (v1.4.0) on Debian Linux 64-bit. I'm using [my own dockerfile](https://github.com/mmguero/zeek-docker/blob/main/base.Dockerfile) to reproduce it but I imagine it would be similar on other environments installing the official Zeek and Spicy release debs.

To illustrate the issue:

Run a docker container with Zeek 4.2.0 and Spicy 1.4.0 and verify my umask isn't weird:

```bash
$ docker run -t -i --rm --entrypoint=/bin/bash ghcr.io/mmguero/zeek:latest
.root 914eea642877 ~> umask 
0022
```

Install any spicy package.

```bash
.root 914eea642877 ~> zkg install --force --skiptests spicy-http
Installing "zeek/zeek/spicy-http"..........................
Installed "zeek/zeek/spicy-http" (v0.0.1)
Loaded "zeek/zeek/spicy-http"
```

Look at the permissions of the hlto files:

```bash
.root 914eea642877 ~> find / -name "*.hlto" -exec ls -l "{}" \;
-rwx--x--x 1 root root 29463088 Mar 15 21:11 /opt/zeek/lib/zeek/plugins/packages/spicy-http/http.hlto
-rwx--x--x 1 root root 29463088 Mar 15 21:11 /opt/zeek/var/lib/zkg/clones/package/spicy-http/build/spicy-modules/http.hlto
```

You'll notice they're `-rwx--x--x` (711) instead of (I would imagine) `-rw-r--r--` (644). This causes non-root users to be unable to run zeek as the htlo files fail with permission errors.

bbannier:
Before e1be41c069e4 when generating HLTO files we just used whatever
permissions the used compiler generated. With that patch we started
using deterministic output paths which for injected `mkstemp` into the
JIT path which on POSIX-compliant platforms uses permissions `0500` for
security reasons. At least clang added an executable bit to this. This
lead to a regression where HLTO files where not user-readable anymore.

With this patch we now add additional permissions allowed by the calling
process' umask. We add permissions instead of replacing them to keep
working even in setups where the umask would e.g., disallow reading or
writing of files, even for the owning user.

Closes #1150.


hilti/runtime/src/library.cc
```diff
@@ -130,5 +131,18 @@ hilti::rt::Result<hilti::rt::Nothing> hilti::rt::Library::save(const hilti::rt::
...
 
+    // Query the current umask. This is safe since we always query and set the umask from a single thread.
+    auto default_perms = ::umask(::mode_t());
+    ::umask(default_perms);
+
+    // Create the file taking into account the active umask. We only add
+    // permissions and never remove masked ones since the user might have set
+    // one which masks required permissions.
+    hilti::rt::filesystem::permissions(path, hilti::rt::filesystem::perms(default_perms),
+                                       hilti::rt::filesystem::perm_options::add, ec);
```
rsmmr, line 142: Let's say our umask is `022`. Without this patch, the ticket says that `path` is created with `-rwx--x--x`. Wouldn't this then add write permissions for group/other to that, so we'd end up with `-rwx-wx-wx`. But what we'd want with that umask is `rwxr-xr-x`. What am I missing?

bbannier: Fixed as discussed offline.

rsmmr (approved):

PR #1153, opened 2022-03-28
Allow building with gcc-8.3.

bbannier:


rsmmr (dismissed):
Approving with a heavy heart. :-)

rsmmr (dismissed):

PR #1159, opened 2022-04-05
Add support for synchronize-at and synchronize-after properties.

### LINKED ISSUE #1134: Add synchronization for unit attributes `&synchronize-at` and `&synchronize-after`
The first implementation of error recovery with #1054 didn't include support for the unit attributes `&synchronize-at` and `&synchronize-after`. We should implement support for them.

bbannier:
Closes #1134.


spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -982,15 +982,49 @@ struct ProductionVisitor
...
+                const auto ctor_ = e->tryAs<hilti::expression::Ctor>();
+                assert(ctor_);
+                auto ctor = production::Ctor(cg()->uniquer()->get(id), ctor_->ctor(), ctor_->meta().location());
+                getLookAhead({ctor}, id, ctor.location(), LiteralMode::Search);
+
+                if ( synchronize_after )
+                    pb->consumeLookAhead();
+
+                return;
+            }
```
rsmmr, line 1021: Oh, this is nice! Looks like our machinery is indeed pretty flexible.

One question: in the standard path below, we check for EOD after `getLookAhead()`. No need for that here?

doc/programming/parsing.rst
```diff
@@ -861,6 +861,16 @@ properties:
     the module level; use ``Null`` to reset the property, i.e., not skip
     anything.
 
+``%synchronize-at = EXPR;``
+    Specifies a literal to synchronize on if the unit is used as a
+    synchronization point during :ref:`error recovery <error_recovery>`.
+    The literal is left in the input stream.
+``%synchronize-after = EXPR;``
```
rsmmr, line 868: Formatting nit: let's add empty lines between the properties (between the `%skip` props above too).

doc/programming/parsing.rst
```diff
@@ -861,6 +861,16 @@ properties:
...
+    Specifies a literal to synchronize on if the unit is used as a
+    synchronization point during :ref:`error recovery <error_recovery>`.
```
rsmmr, line 870: We should probably explain these  in `error recovery`, too?

spicy/toolchain/src/compiler/codegen/parser-builder.cc
bbannier: > One question: in the standard path below, we check for EOD after getLookAhead(). No need for that here?

Yes, this is definitely needed; I'll add the same check here as well.

doc/programming/parsing.rst
bbannier: I'll add a detailed item to the list of possible synchronization points there.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -982,15 +982,49 @@ struct ProductionVisitor
...
+            const auto synchronize_at = type.propertyItem("%synchronize-at");
+            const auto synchronize_after = type.propertyItem("%synchronize-after");
+
+            std::optional<Expression> e;
+
+            if ( synchronize_at )
+                e = synchronize_at->expression();
+
+            if ( synchronize_after )
+                e = synchronize_after->expression();
```
bbannier, line 1008: This will behave in unexpected ways if both `%synchronize-at` and `%synchronize-after` are given in the same unit. We should reject such setups.

PR #1162, opened 2022-04-08
Fix clang-tidy `HeaderFilterRegex`.

### LINKED ISSUE #1142: clang-tidy warnings from HILTI and Spicy headers
We noticed that if one views some headers in an editor with clangd integration, clang-tidy warnings are triggered, e.g.,

```
/spicy/hilti/runtime/include/types/vector.h|111 col 5-10 error| [clang-tidy readability-const-return-type] Return type 'const Iterator<T, Allocator>' is 'const'-qualified at the top level, which may reduce code readability without improving const correctness (fix available) [E]
/spicy/hilti/runtime/include/types/vector.h|187 col 5-10 error| [clang-tidy readability-const-return-type] Return type 'const ConstIterator<T, Allocator>' is 'const'-qualified at the top level, which may reduce code readability without improving const correctness (fix available) [E]
/spicy/hilti/runtime/include/types/vector.h|368 col 13-21 error| [clang-tidy cert-oop54-cpp] Operator=() does not handle self-assignment properly [E]
/spicy/hilti/runtime/include/types/vector.h|476 col 36-38 error| [clang-tidy readability-uppercase-literal-suffix] Integer literal has suffix 'u', which is not uppercase (fix available) [E]
/spicy/hilti/runtime/include/types/vector.h|479 col 48-50 error| [clang-tidy readability-uppercase-literal-suffix] Integer literal has suffix 'u', which is not uppercase (fix available) [E]
/spicy/hilti/runtime/include/types/vector.h|482 col 49-51 error| [clang-tidy readability-uppercase-literal-suffix] Integer literal has suffix 'u', which is not uppercase (fix available) [E]
```

Since headers are not compiled directly clangd seems to infer the flags to use from nearby source files.

We do not see these warnings in CI which is surprising since these headers should be used in source files we do check. It could be that the logic in `scripts/run-clang-tidy` somehow filters out warnings from files not in the compilation database.

While we clean this up we should also look into whether it is possible to get rid of the script and instead use `run-clang-tidy` from the LLVM distribution. This might require overriding the clang-tidy config on a per-directory basis to filter out 3rdparty and generated source files. The way to do this seems to be to use a per-directory `.clang-tidy` file which disables checks for the directory and its subdirectories.

bbannier:
In general LLVM's regexp engine does not seem to support negative
matches at all. We previously tried to set this up, but this also
filtered many headers we would have liked to lint.

This patch sets a regexp under the assumptions that our paths do not
contain numbers. With that many more diagnostics are triggered which we
also fix.

Closes #1142.


.clang-tidy
```diff
@@ -48,7 +51,7 @@ Checks: '
   -readability-function-size
   '
 
-HeaderFilterRegex: '.*(/include/(hilti|spicy)/(^3).*/[^_][^_][^/]*$)|(^3rdparty.*)'
```
rsmmr, line 51: Oh,  that old expression was clearly broken, that `(^3)` doesn't work, that's supposed to be `[^3]`.  However, that was a recent change introduced in 598da755835f878387257f0668e563f1b2255515. Did that change break the headers, or weren't they not working before either?

I'm fine with the new expression too, just wondering if we wouldn't need to be quite as strict ("no digits").

rsmmr (approved):
I just quickly skimmed over the changes, thanks for going through that all.

PR #1165, opened 2022-04-19
Make compiler plugin initialization explicit.

### LINKED ISSUE #1164: Cannot compile Spicy files with `spicyz` if linking against static Spicy libs
If I build a static Spicy (e.g., Spicy configured with ` --build-static-libs`) I can successfully link `spicyz`, but it lacks support for Spicy files:

```console
$ ./bin/spicyz /tmp/bla.spicy -o /tmp/bla.hlto
[error] <Spicy Plugin for Zeek>: unsupported file type (/private/tmp/bla.spicy)
```

The issue seems to that the dynamic registration of the plugin adding support for Spicy files is not linked in, https://github.com/zeek/spicy/blob/5df264f5f02ecb0aa0afcc407553b22369391376/spicy/toolchain/src/compiler/plugin.cc#L69.

bbannier:
Support for working with Spicy and HILTI source files is handled by
plugins. We previously registered these plugins via hidden static
variables. This had issues if Spicy is linked statically and the hidden
registration variable was not linked into the final output (in that case
not registration would happen).

In order to better control the language registration this patch
introduces dedicated functions `hilti::init` and `spicy::init` which
currently only register support for HILTI and Spicy source files,
respectively. We also adjust our executables to invoke these functions
now.

Closes #1164.

### DIFF
--- a/hilti/toolchain/include/compiler/plugin.h
+++ b/hilti/toolchain/include/compiler/plugin.h
@@ -287,21 +287,10 @@ private:
     std::vector<Plugin> _plugins;
 };
 
-namespace plugin {
-/**
- * Helper class to register a plugin at startup. To add a plugin, create a
- * global `Register` instance to register it.
- */
-class Register {
-public:
-    /**
-     * Registers a plugin with the global `registry()`.
-     *
-     * @param p plugin to register
-     */
-    Register(const Plugin& p) { registry().register_(p); }
-};
+namespace detail {
 
-} // namespace plugin
+Plugin create_hilti_plugin();
+
+} // namespace detail
 
 } // namespace hilti
--- a/spicy/toolchain/src/compiler/init.cc
+++ b/spicy/toolchain/src/compiler/init.cc
@@ -1,0 +1,18 @@
+// Copyright (c) 2020-2021 by the Zeek Project. See LICENSE for details.
+
+#include "compiler/init.h"
+
+#include <hilti/compiler/plugin.h>
+
+#include <spicy/compiler/plugin.h>
+
+void spicy::init() {
+    static bool initialized = false;
+
+    if ( initialized )
+        return;
+
+    hilti::plugin::registry().register_(detail::create_spicy_plugin());
+
+    initialized = true;
+}
--- a/hilti/toolchain/CMakeLists.txt
+++ b/hilti/toolchain/CMakeLists.txt
@@ -74,6 +74,7 @@ set(SOURCES
     src/compiler/cxx/linker.cc
     src/compiler/cxx/unit.cc
     src/compiler/driver.cc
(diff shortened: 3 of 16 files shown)


hilti/toolchain/include/compiler/plugin.h
```diff
@@ -304,4 +304,6 @@ class Register {
 
 } // namespace plugin
 
+Plugin hilti_plugin();
```
rsmmr, line 307: Let's move this into `detail` and rename to something like `create_hilti_plugin` to avoid somebody calling this to get access to an already existing plugin instance. Same below for Spicy.

rsmmr (approved):
I'm not quite sure why this actually makes a difference, isn't `plugin.cc` always pulled in as well whenever `driver.cc` is? But anyways, if it fixes the problem, fine with me.

bbannier:
> I'm not quite sure why this actually makes a difference, isn't `plugin.cc` always pulled in as well whenever `driver.cc` is? But anyways, if it fixes the problem, fine with me.

This was enough for making this work for linking spicyz against a statically linked Spicy. When building spicyz and Spicy as part of Zeek in zeek/zeek#2006 this is still insufficient though, so I am still looking into this.

rsmmr (commented):
I know this has been tricky to get working, but I have to say I don't really like the final solution because it's become pretty intransparent what's going on: creating a plugin with each driver instance, registering that, but then skipping any duplicates because we only need one. Plus, now having to create driver instances to make the plugins available, even if one doesn't really need a driver (like in `spicy-doc`) .

I have a different suggestion, which I think you had also already considered as well: it looks like we need to move away from using `hilti::plugin::Register` because it's so hard to control/ensure its construction. Why don't we just do traditional-style `init()` functions that each `main()` using HILTI/Spicy needs to call, similar to how we do it for the runtime system already. So we'd have `hilti:init()` and `spicy::init()`. These then create the plugin instances and register them with the plugin registry (and guard against multiple execution directly, so that `init()` can be called only once).  And `hilti::plugin::Register` goes away.

bbannier:
The accompanying fix for spicy-plugin is zeek/spicy-plugin#111.

spicy/toolchain/src/compiler/init.cc
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2020-2021 by the Zeek Project. See LICENSE for details.
+
+#include "compiler/init.h"
+
+#include <hilti/compiler/plugin.h>
+
+#include <spicy/compiler/plugin.h>
+
+void spicy::init() { hilti::plugin::registry().register_(detail::create_spicy_plugin()); }
```
rsmmr, line 9: Same as above.

rsmmr (dismissed):

PR #1167, opened 2022-04-28
Add binary packaging mode.

bbannier:
This patch adds a binary packaging mode which mirrors the behavior of
that flag. In this mode Spicy gets linked statically and no RPATHs are
embedded in binaries. This should make it easier to package Spicy (e.g.,
alongside Zeek) for typical distributions which tend to disallow
embedded RPATHs.

Note that we still emit linker flags with an RPATH via e.g.,
`spicy-config --ldflags-hlto`. This RPATH is however not embedded into
distributed binary artifacts.


rsmmr (dismissed):

0xxon:
Cool - thank you very much.

PR #1178, opened 2022-05-16
﻿﻿﻿Only check whether Zeek docs are up-to-date from cron builds.

bbannier:


bbannier:
The original target branch for this `release/1.4` where only the last two patches would need to go in, but I'll add them to `main` as well.

rsmmr (approved):

PR #1180, opened 2022-05-16
Cap parallelism use for JIT background jobs.

bbannier:


hilti/toolchain/src/compiler/jit.cc
```diff
@@ -470,19 +473,41 @@ Result<JIT::JobRunner::JobID> JIT::JobRunner::_spawnJob(const hilti::rt::filesys
...
+    // Cap parallelism for background jobs.
+    //
+    // - if `HILTI_JIT_SEQUENTIAL` is used all parallelism is disabled and
+    //   exactly one job is used.
+    // - if `HILTI_JIT_PARALLELISM` is set it is interpreted as the maximum
+    //   number of parallel jobs to use
+    // - by default we use one job per available CPU (on some platforms
+    //   `std::thread::hardware_concurrency` can return 0, so use one job
+    //   there)
+    uint64_t parallelism = std::max(std::thread::hardware_concurrency(), 1U);
```
rsmmr, line 495: Want to add warning here if we get `0`? Running with 1 is hardly what one wants, and we could want to look for other ways to get `nproc` on those platforms if it actually happened.

hilti/toolchain/src/compiler/jit.cc
```diff
@@ -538,6 +563,9 @@ Result<Nothing> JIT::JobRunner::_waitForJobs() {
                 _jobs.erase(id);
             }
         }
+
+        if ( _jobs_pending.empty() && _jobs.empty() )
```
rsmmr, line 567: Can this become the condition of the while loop (instead of `true`). Could then also use that same check at the beginning of the method.

rsmmr (dismissed):

hilti/toolchain/src/compiler/jit.cc
bbannier: I'll add a warning and keep defaulting to `1` in this case `std::thread::hardware_concurrency()` returns `0`. I'll add code to emit a warning in that case which also points that the default can be overridden with `HILTI_JIT_PARALLELISM`.

I tested whether we see it returning `0` in any of our CI platforms (see #1182) and that seems to be not the case anywhere. I am unsure why the standard makes such weak guarantees.

PR #1181, opened 2022-05-16
﻿Remove optional braces around control statements.

bbannier:


rsmmr (commented):
Not fully convinced about this change yet. I'm fine (and prefer) no braces around single-line statements, but for my taste it's more readable to retain braces around multi-line statements. E.g., I prefer

```
while ( x != _end ) {
    if ( ! (*x++ == *y++) )
        return false;
}
```

over 

```
while ( x != _end )
    if ( ! (*x++ == *y++) )
        return false;
```

bbannier:
> Not fully convinced about this change yet. I'm fine (and prefer) no braces around single-line statements, but for my taste it's more readable to retain braces around multi-line statements.

Fine with me to drop this PR. I noticed that this formatting rule became available with clang-format-14 and wanted to check with you whether it would be worthwhile to add, especially since I personally struggle with the no-braces style (e.g., I sometimes left braces in by accident when doing `printf` debugging in single-line loops during development).

PR #1186, opened 2022-05-23
Do not artificially suppress uses of `(u)int8` in some cases.

### LINKED ISSUE #1174: Integer cast returns incorrect type
I would expect that this sample prints `int8`:

```ruby
module foo;

function f(i: int8) { print("int8"); }
function f(i: int16) { print("int16"); }

f(cast<int8>(1));
```

Instead it prints `int16`.

Looking at the generated C++ code this is due to `cast<int8>` creating an `int16`, not the requested `int8`,

```cpp
extern void __hlt::foo::__init_module() {
    f(::hilti::rt::integer::safe<int16_t>(::hilti::rt::integer::safe<std::uint64_t>{1U}));
}
```

bbannier:
We previously would uses at least `(u)int16`, even though the the user
explicitly requested `(u)int8`. This could e.g., lead to surprising
behavior when interacting with function overloading. This patch removes
the special handling and we now use the requested type

Closes #1174.

### DIFF
--- a/hilti/toolchain/src/compiler/codegen/coercions.cc
+++ b/hilti/toolchain/src/compiler/codegen/coercions.cc
@@ -125,10 +125,10 @@ struct Visitor : public hilti::visitor::PreOrder<std::string, Visitor> {
             return fmt("static_cast<bool>(%s)", expr);
 
         if ( auto t = dst.tryAs<type::SignedInteger>() )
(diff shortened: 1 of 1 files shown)


bbannier:
> I'm just realizing I used the wrong type in my code snippet, it should be `global x: int8 = 42;`. Mind giving that a quick try as well?

Same result, `42`.

rsmmr (approved):

PR #1193, opened 2022-05-27
Optimize AST build throughput

### LINKED ISSUE #1177: Investigate JIT throughput for huge grammars
It seems that for huge grammars (hundreds of units, thousands of LOC) JIT throughput can be very slow. We should benchmark such grammars to make sure we are not behaving worse than absolutely necessary.

Things which would be interesting to look at:

- does the resolver time scale at most linear with the number of names/types?
- does the number of template instantiations scale linear with the number of types?

bbannier:
This work is part of #1177.

### DIFF
--- a/hilti/toolchain/src/ast/scope.cc
+++ b/hilti/toolchain/src/ast/scope.cc
@@ -2,6 +2,8 @@
 
 #include <utility>
 
+#include <hilti/rt/util.h>
+
 #include <hilti/ast/declaration.h>
 #include <hilti/ast/declarations/expression.h>
 #include <hilti/ast/declarations/imported-module.h>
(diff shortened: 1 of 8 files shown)


hilti/toolchain/src/ast/scope.cc
```diff
@@ -11,20 +11,12 @@
...
-
-    // Filter out duplicates
-    for ( const auto& i : nodes ) {
-        if ( i && i->as<Declaration>() == d )
-            return;
-    }
-
-    nodes.push_back(std::move(n));
+bool Scope::ItemCmp::operator()(const NodeRef& a, const NodeRef& b) const {
+    return a->as<Declaration>().meta().location() < b->as<Declaration>().meta().location();
```
rsmmr, line 15: I'm not sure this is safe: we don't guarantee that nodes must have a location, and different nodes could also end up with the same location (locations are not only coming in from source code, but also programmatically).

bbannier:
Still need to look into the failures.

hilti/toolchain/src/ast/scope.cc
```diff
@@ -11,20 +11,21 @@
...
-    auto& nodes = _items[std::string(id)];
-
-    // Filter out duplicates
-    for ( const auto& i : nodes ) {
-        if ( i && i->as<Declaration>() == d )
-            return;
-    }
+std::size_t Scope::NodeRefHash::operator()(const hilti::NodeRef& n) const {
+    if ( n )
+        return reinterpret_cast<size_t>(&*n);
```
rsmmr, line 16: So the hash is the address of the node, right? That's changing semantics slightly, in the case that two declarations exist that are semantically equivalent but stored in separates nodes. But I think that's fine, I can't really see how that would become a problem.

rsmmr (approved):

hilti/toolchain/src/ast/scope.cc
bbannier: It turns out that this approach does not work reliably and by using the address we might end up hashing node refs which should be considered equal to different buckets.

I adjusted this to instead use a) the presence of a node behind the ref, and b) a hash of the location information. This seems to work.

PR #1194, opened 2022-06-01
Switch to vendored nlohmann/json.

bbannier:


rsmmr (dismissed):
Looks good, let's just add a couple things:

- The original LICENSE file (I know its in LICENSE.3rdparty, but still)
- A "README.spicy" that points to the repository where this code is from and, just for the record, explains why we're not using a submodule like we do elsewhere.

PR #1197, opened 2022-06-06
Make sanitizer features more granular

bbannier:
Closes zeek/zeek#2153.

### DIFF
--- a/tests/Scripts/have-sanitizer
+++ b/tests/Scripts/have-sanitizer
@@ -6,4 +6,4 @@ test "${USING_BUILD_DIRECTORY}" = 1 || exit 1
 
 base=$(cd $(dirname $0) && pwd)
 cache=$(${base}/build-directory)/CMakeCache.txt
-test -e ${cache} && grep -q "HILTI_HAVE_SANITIZER.*=yes" ${cache}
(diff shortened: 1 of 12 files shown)


tests/Scripts/have-sanitizer
```diff
@@ -6,4 +6,4 @@ test "${USING_BUILD_DIRECTORY}" = 1 || exit 1
 
 base=$(cd $(dirname $0) && pwd)
 cache=$(${base}/build-directory)/CMakeCache.txt
-test -e ${cache} && grep -q "HILTI_HAVE_SANITIZER.*=yes" ${cache}
+test -e ${cache} && grep -q "USE_SANITIZER.*=.*" ${cache}
```
timwoj, line 9: Should this just be `USE_SANITIZERS` instead of the wildcard? I don't see it defined any other way.

bbannier: Thanks for the careful review. The variable is indeed called `USE_SANITIZERS`, but I needed to match `USE_SANITIZER:STRING=...` here. I now use a more narrow pattern.

timwoj (approved):

PR #1203, opened 2022-06-09
Allow changing `DESTDIR` between configure and install time.

bbannier:
We previously would back to value of `DESTDIR` in at configure time.
This made it impossible to change its value at install time so that
e.g., typical uses like

    make DESTDIR=<dest> install

did not perform like expected, i.e., symlinks which were supposed to be
removed at install time did not get removed due to incorrect paths. With
this patch we now evaluate the value of `DESTDIR` at install time
instead.


0xxon (approved):
Thank you :).

This should make it into the 5.0 Zeek release.

PR #1213, opened 2022-06-17
Provide error message to `%error` handler.

### LINKED ISSUE #1209: Access the parse error in %error unit hook
The **spicy-verbose** debug output generates helpful error messages. However, these messages are not accessible in the **%error** unit hook. For example, the debug message `parse error in connected unit ... during close (&size amount not consumed (spicy filename and location))` indicates that there were fewer bytes to consume than specified in the grammar. It would be helpful to expose such error messages to Zeek scripts in events generated upon parsing of units.

rsmmr:
We now allow to optionally provide a string parameter with `%error`
that will receive the associated error message:

    on %error(msg: string) { print msg; }

Closes #1209.


spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -189,7 +190,9 @@ struct ProductionVisitor
                     // is that we want a reliable point of error handling
                     // no matter what kind of trouble a Spicy script runs
                     // into.
-                    auto catch_ = try_->addCatch();
+                    auto catch_ =
+                        try_->addCatch({builder::parameter("__excpt", builder::typeByID("hilti::SystemException"))});
```
bbannier, line 194: Could we go with `__except` here or does that clash with something?

bbannier (dismissed):
Only minor nits. This will be a useful feature e.g., with backtracking (filed #1214 while playing with this PR).

PR #1226, opened 2022-06-27
Mention Spicy feature in zkg package template in Zeek docs

bbannier:


rsmmr (commented):
I don't think this is the right place for this, the tutorial would be better (at least in addition). Let me give this a try, I'll push an update to this branch.

rsmmr:
Extended this, see what you think. 

One question about the template: I see files called like `analyzer/MyOrg.*`, shouldn't those be using the protocol name instead of the organization name?

And a request: let's redo the spicy-tftp repository to start from the template, so that it can act as a complete example (as the text already says :-) ).

bbannier:
> Extended this, see what you think.

This looks great and much better than the minimally invasive change I put up. By adding that extra section we are able to go into more detail, too.
 
> One question about the template: I see files called like `analyzer/MyOrg.*`, shouldn't those be using the protocol name instead of the organization name?

We translate zkg's "namespace" to a Spicy module name, and I believe we'd want the files named like the module; "name" OTOH we translate into the name of the top-level analyzer which we use in a number of places. We could also introduce new names for the Spicy-side constructs instead of repurposing the identifiers already provided by the default template, but I think that we wont be able to get rid of the existing variables. 

> And a request: let's redo the spicy-tftp repository to start from the template, so that it can act as a complete example (as the text already says :-) ).

rsmmr:
Updated as discussed offline.

PR #1228, opened 2022-06-29
Fix code generation when a module's file could be imported through

### LINKED ISSUE #1227: Chained import with `import ... from ...` generates invalid C++ names
The following reproducer with chained imports produces names which cannot be found. Looking at the generated name it looks like this is due to us generating namespace names for hooks by concatenating the module names in the import chain, but use types in simple namespaces.

```ruby
# @TEST-EXEC: spicyc -d -j z.spicy -o z.hlto

module z;
import y;

@TEST-START-FILE y.spicy
module y;
import x from bla;
@TEST-END-FILE

@TEST-START-FILE bla/x.spicy
module x;

type X = unit {
	on %done {}
};
@TEST-END-FILE
```

Bisecting with this reproducer points to #1195 as the change introducing this regression.

rsmmr:
different means.

This can happen when mixing normal import with  `import ... from`. A
previous change to support multiple units of the same name, missed
that we could run into namespacing trouble at code generation time in
this case. The problem was that we included the `from ...` scope into
the namespace, which doesn't work when we import the same module both
w/ and wo/ scope. Now we generate unique module IDs differently
through global state independent of file system paths.

Close #1227.

### DIFF
--- a/hilti/toolchain/include/compiler/unit.h
+++ b/hilti/toolchain/include/compiler/unit.h
@@ -7,6 +7,7 @@
 #include <functional>
 #include <memory>
 #include <string>
+#include <unordered_map>
 #include <unordered_set>
 #include <utility>
 #include <vector>
@@ -75,16 +76,13 @@ public:
      * Returns an ID for the unit's module that's globally unique across all
      * units processed within the current context. This ID will often be the
      * same as returned by `id()`, but it may include an additional string for
-     * unification, in particular if the module was imported through `import
-     * ... from ...`.
+     * unification, in particular if there are more than one module with the
+     * same ID.
      *
      * @returns globally unique ID for the module; the method guaranteees that
      * the ID represents a valid C++ identifier
      */
-    ID uniqueID() const {
-        return _index.scope.empty() ? _index.id :
-                                      ID(util::fmt("%s_%s", util::replace(_index.scope, ".", "_"), _index.id));
-    }
+    ID uniqueID() const { return _unique_id; }
 
     /** Returns the path associated with the unit's module. */
     const auto& path() const { return _index.path; }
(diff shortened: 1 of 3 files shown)


bbannier:
@rsmmr, now to `v1.5.0` has been released, can we merge this into `main`? That would allow us to soak this patch for possible inclusion in `v1.5.1` as well.

hilti/toolchain/src/compiler/unit.cc
```diff
@@ -47,6 +49,16 @@ bool runHook(bool* modified, const Plugin& plugin, const Node* module, const std
     return logger().errors() == 0;
 }
 
+ID Unit::_makeUniqueID(const ID& id) {
+    ID uid;
```
bbannier, line 53: This seems to be unused.
```suggestion
```

hilti/toolchain/include/compiler/unit.h
```diff
@@ -407,13 +410,16 @@ class Unit {
...
+    ID _unique_id;                                  // globally unique ID for this module
     hilti::rt::filesystem::path _extension;         // AST extension, which may differ from source file
     std::optional<Node> _module;                    // root node for AST (always a `Module`), if available
     std::vector<std::weak_ptr<Unit>> _dependencies; // recorded dependencies
     std::weak_ptr<Context> _context;                // global context
     std::optional<detail::cxx::Unit> _cxx_unit;     // compiled C++ code for this unit, once available
     bool _resolved = false;                         // state of resolving the AST
     bool _requires_compilation = false;             // mark explicitly as requiring compilation to C++
+
+    static std::map<ID, unsigned int> _uid_cache; // cache storing state for generating globally unique IDs
```
bbannier, line 422: Since `ID` is hashable we could use a `std::unordered_map` here. This wouldn't cost us anything, but could give minor speedups for compilations using a lot of units.

bbannier (approved):

bbannier:
Landed in both `main` and `release/1.5`.

PR #1244, opened 2022-07-07
Add support for adding link dependencies via `--cxx-link`.

### LINKED ISSUE #1236: Make it possible to specify link dependencies when building HLTO files
We currently allow specifying include directories when building HLTO files via `HILTI_CXX_INCLUDE_DIRS`. We should consider adding a similar mechanism to specify link dependencies. This would allow users to directly expose library code in Spicy without having to [build HLTO files themselves](https://zeekorg.slack.com/archives/C012CQV1K29/p1657006784013199).

bbannier:
We already supported specifying additional include directories, but
injecting additional link dependencies into the link always required
manually compiling and linking files. So while users could already
declare e.g., existing functions with `&cxxname`, we provided no tooling
to use such functions if we were defined in external libraries or static
archives.

This patch adds support for that. One can now reference static archives
or shared libraries to include in the link by passing them with
`--cxx-link <lib>`. This will copy symbols for static archives or
introduce a shared library dependency into the created HLTO file and
needs only to be specified for the step creating the HLTO file.

Closes #1236.


bbannier:
@rsmmr, I still need to figure out how to make the test case work on `freebsd13`, but I think the idea for the user API is where I want it to be. Could you take a look?

rsmmr:
I think my main question is if an environment variable is the right mechanism here. On the plus side, it matches what we do for include paths. On the other hand, it seems unusual to pass specific libraries/targets through an environment variables (whereas it's common for a search path like the include dirs). The alternative would be a command line option (and we could add one for include directories, too, then on top of the env var).

bbannier:
> I think my main question is if an environment variable is the right mechanism here. On the plus side, it matches what we do for include paths. On the other hand, it seems unusual to pass specific libraries/targets through an environment variables (whereas it's common for a search path like the include dirs). The alternative would be a command line option (and we could add one for include directories, too, then on top of the env var).

I think we want to support cases where Spicy analyzers use e.g., the CMake macro `spicy_add_analyzer` which does not invoke `spicyz` directly (this is e.g., used in the package template). If we add a command line argument for libs we sjould add a new `LIBS` arg to that macro as well. Let me check on that.

bbannier:
Updated this as discussed offline. Libs can now be passed with the flag `--cxx-link` (potentially used multiple times).

I also tried adding the global way to pass arbitrary command line args to our drivers that you mentioned (e.g., `HILTI_OPTIONS=`-c --cxx-link foo.so` would set the two command line options). This will be rather tricky to implement an use since we somehow would need to reconstruct `int argc` and `char** argv` from that which becomes hard when arguments need escaping e.g., due to embedded whitespace. I'd suggest we do not attempt this, but instead add a new named argument to `spicy_add_analyzer` over in [zeek/spicy-plugin](https://github.com/zeek/spicy-plugin) as a follow-up.

hilti/toolchain/include/compiler/context.h
```diff
@@ -45,8 +45,9 @@ struct Options {
         "hlt"; /**< CXX namespace for generated C++ code accessible to the host application */
     std::string cxx_namespace_intern = "__hlt"; /**< CXX namespace for generated internal C++ code */
     std::vector<hilti::rt::filesystem::path>
-        cxx_include_paths;  /**< additional C++ directories to search for #include files */
-    bool keep_tmps = false; /**< if true, do not remove generated files on exit */
+        cxx_include_paths;         /**< additional C++ directories to search for #include files */
+    bool keep_tmps = false;        /**< if true, do not remove generated files on exit */
+    std::vector<std::string> link; /**< additional static archives or shared libraries to link during JIT */
```
rsmmr, line 50:
```suggestion
    std::vector<std::string> cxx_link; /**< additional static archives or shared libraries to link during JIT */
```

For clarity, and to match `cxx_include`.

hilti/toolchain/src/compiler/jit.cc
```diff
@@ -382,6 +382,12 @@ hilti::Result<std::shared_ptr<const Library>> JIT::_link() {
         }
     }
 
+    // Add additional shared libraries or static archives to the link. This needs to happen
+    // after we added the objects to make sure we pull in symbols used in the objects.
+    for ( auto&& lib : options().link )
```
rsmmr, line 387:
```suggestion
    for ( const auto& lib : options().link )
```

hilti/toolchain/src/compiler/driver.cc
```diff
@@ -35,9 +35,12 @@ inline const DebugStream Compiler("compiler");
 inline const DebugStream Driver("driver");
 } // namespace hilti::logging::debug
 
+constexpr int OPT_LINK = 1000;
```
rsmmr, line 38:
```suggestion
constexpr int OPT_CXX_LINK = 1000;
```

rsmmr (dismissed):

PR #1246, opened 2022-07-20
Fix validator for field `&default` expression types for constness.

bbannier:
When checking types of field `&default` expressions we previously would
also consider their constness. This breaks e.g., cases where the used
expression is not a LHS like the field the `&default` is defined for,

    type X = unit {
       var x: bytes = b"" + a;
    };

With this patch we do not consider constness in the type check anymore.
Since field are never const this allows us to set a `&default` with
const expressions as well.


rsmmr (approved):

bbannier:
Also backported to `release/1.5`/

PR #1253, opened 2022-08-08
Fix `stop` for unbounded loop.

rsmmr:
More precisely, for anything using a `while` production.


bbannier (approved):

bbannier:

PR #1255, opened 2022-08-09
When decoding bytes into a string using a given charset, allow caller…

### LINKED ISSUE #1251: `bytes::[lower|upper]` can fail in undocumented ways
We currently provide `lower` and `upper` methods for `bytes`. These methods take some encoding and will fail with an exception if a given `bytes` does not conform to it.  We should at the very least document this, and possibly improve this API.

It seems weird that for methods from `bytes` to `bytes` we should fail here; the failure would seem more natural to me if they returned a `string`. We could also consider moving the methods into `string` altogether so they cannot fail (but `decode` to get the the `string` would fail). If `bytes` provided a way to sanitize an input for decoding users could even work around this, e.g., with a method which for a given encoding replaces unrepresentable subsequences with a user-provided input. The downside of this is that it would still allow users to generate strings which again cannot be decoded.

rsmmr:
… to control error handling.

All methods taking a charset parameters now take an additional enum
selecting 1 of 3 possible error handling strategies in case a
character can't be decoded/represented: `STRICT` throws an error,
`IGNORE` skips the problematic character and proceeds with the next,
and `REPLACE` replaces the problematic character with a safe
substitute. `REPLACE` is the default everywhere now, so that by
default no errors are triggered.

This comes with an additional functional change for the ASCII
encoding: we now consistently sanitize characters that ASCII can't
represent when in `REPLACE`/`IGNORE` modes (and, hence, by default),
and trigger errors in `STRICT` mode. Previously, we'd sometimes let
them through, and never triggered any errors. This also fixes a bug
with the ASCII encoding sometimes turning a non-printable character
into multiple repeated substitutes (which our unit tests actually then
checked for; fixed the tests, too).

Closes #1251.


hilti/runtime/include/types/bytes.h
```diff
@@ -315,9 +319,11 @@ class Bytes : protected std::string {
      * specified character set.
      *
      * @param cs character set to assume the binary data to be encoded in
+     * @param errors how to handle errors when decoding the data
      * @return UTF8 string
      */
-    std::string decode(bytes::Charset cs) const;
+    std::string decode(bytes::Charset cs,
```
bbannier, line 325: With replacement here users could decode e.g., to `ASCII` and end up with a non-ASCII result (containing UTF8 substitute characters). I think that if we want to keep this API we really need to specify the replacement character very explicitly as part of the API so users can sanitize results after the fact.

hilti/runtime/include/types/string.h
```diff
@@ -12,27 +12,42 @@ namespace hilti::rt {
...
-/** Computes the length of a UTF8 string.
+/* When processing UTF8, how to handle invalid data not representing UTF8 codepoints. */
+enum class DecodeErrorStrategy : int64_t {
+    IGNORE,  // skip data
+    REPLACE, // replace with a place-holder
+    STRICT   // throw a runtime error
+};
+
+/**
+ * Computes the length of a UTF8 string.
```
bbannier, line 23: Let's be very explicit how replacement characters factor into the length. If I understand the implementation correctly we do not return an UTF8 string length, but the number of code points (e.g., we do not handle combining characters). Each inserted replacement characters should contribute one additional code point.

hilti/runtime/include/types/bytes.h
bbannier: Was incorrect on this one.

bbannier (approved):

PR #1259, opened 2022-08-10
Add coercion for optionals that only differ in constness of their inner types.

### LINKED ISSUE #1143: optional<bytes> differs between struct and tuple only as a parameter
This came up when trying to convert a struct to a tuple in Spicy to pass to Zeek. This same construct worked with `uint64` but failed with a cryptic error when I tried the same thing with `bytes`. It's possible other types would meet the same fate. Code and error follow:

```ruby
module Foo;

type Bar = struct {
  wutang: bytes &optional;
};

type BarTuple = tuple<
  wutang: optional<bytes>
>;

public function make_bar_tuple(bar: Bar): BarTuple {
  local nope: optional<bytes>;
  return (
    bar?.wutang ? optional(bar.wutang) : nope
  );
}
```

```
¡ spicy-build -o foo foo.spicy
[error] foo.spicy:14:5: types of alternatives do not match in ternary expression (optional<bytes> vs. optional<bytes>)
```

However, if `bar` is instead declared locally, it compiles without issue:

```ruby
module Foo;

type Bar = struct {
  wutang: bytes &optional;
};

type BarTuple = tuple<
  wutang: optional<bytes>
>;

public function make_bar_tuple(): BarTuple {
  local bar: Bar;
  local nope: optional<bytes>;
  return tuple(bar?.wutang ? optional(bar.wutang) : nope);
}
```

(Fixed code to actually return tuple instead instead of `optional` - Robin).

### LINKED ISSUE #1220: Missing coercion in ternary operator
When using a `Null` as a value in a ternary operator it does not seem to be coerced to the correct type.

```ruby
module nil;

function g(): optional<uint64> {
	return True ? optional(10) : Null;
}
```

I would expect `Null` to be coerced to an `optional<uint8>` here but instead I see

```console
$ spicyc -j nil.spicy
[error] nil.spicy:4:9: types of alternatives do not match in ternary expression (optional<uint<64>> vs. <null type>)
```

At least in this example we should be able derive some target type from the first branch of the ternary. The same approach seems to break down for ternaries where `Null` appears as first branch; if we do not want to change how types for ternaries are resolved (e.g., from the use of the whole expression), we might need to introduce support for users to specify a type of a `Null` (currently we only support creating set optionals -- still with implicitly derived type).

The workaround is to use a typed `Null` value,

```ruby
function g(): optional<uint64> {
	local nil: optional<uint64> = Null;
	return True ? optional(10) : nil;
}
```

rsmmr:
There are some situations where that's fine to coerce, so adding that.
Also adding a coercion for ternary expressions that coerces the 2dn
expression into the type of the 1st (which we already assume provides
the result type).

Closes #1143.
Closes #1220.

### DIFF
--- a/hilti/toolchain/src/compiler/coercion.cc
+++ b/hilti/toolchain/src/compiler/coercion.cc
@@ -424,6 +424,16 @@ struct VisitorType : public visitor::PreOrder<std::optional<Type>, VisitorType> 
     }
 
     result_t operator()(const type::Optional& r) {
+        if ( auto t = dst.tryAs<type::Optional>() ) {
+            const auto& s = r.dereferencedType();
+            const auto& d = t->dereferencedType();
+
+            if ( type::sameExceptForConstness(s, d) && (style & CoercionStyle::Assignment) )
+                // Assignments copy, so it's safe to turn  into the
+                // destination without considering constness.
+                return dst;
+        }
+
         if ( auto t = dst.tryAs<type::Bool>(); (style & CoercionStyle::ContextualConversion) && t )
             return dst;
 
--- a/hilti/toolchain/src/compiler/visitors/validator.cc
+++ b/hilti/toolchain/src/compiler/visitors/validator.cc
@@ -276,10 +276,11 @@ struct VisitorPost : public hilti::visitor::PreOrder<void, VisitorPost>, public 
     }
 
     void operator()(const expression::Ternary& n, position_t p) {
-        if ( ! hilti::type::sameExceptForConstness(n.true_().type(), n.false_().type()) )
+        if ( ! hilti::type::sameExceptForConstness(n.true_().type(), n.false_().type()) ) {
             error(fmt("types of alternatives do not match in ternary expression (%s vs. %s)", n.true_().type(),
                       n.false_().type()),
                   p);
+        }
     }
 
     void operator()(const expression::UnresolvedID& n, position_t p) {
--- a/tests/spicy/types/optional/coercion.spicy
+++ b/tests/spicy/types/optional/coercion.spicy
@@ -1,0 +1,18 @@
+# @TEST-EXEC: ${SPICYC} -j %INPUT >output
+#
+# @TEST-DOC: Regression test for #1143.
+
+module Foo;
+
+type Bar = struct {
+  wutang: bytes &optional;
+};
+
+type BarTuple = tuple<
+  wutang: optional<bytes>
+>;
+
+public function make_bar_tuple(bar: Bar): BarTuple {
+  local nope: optional<bytes>;
+  return tuple(bar?.wutang ? optional(bar.wutang) : nope);
+}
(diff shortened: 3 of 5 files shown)


hilti/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -276,10 +276,17 @@ struct VisitorPost : public hilti::visitor::PreOrder<void, VisitorPost>, public
...
 
     void operator()(const expression::Ternary& n, position_t p) {
-        if ( ! hilti::type::sameExceptForConstness(n.true_().type(), n.false_().type()) )
+        if ( ! hilti::type::sameExceptForConstness(n.true_().type(), n.false_().type()) ) {
+            hilti::render(std::cerr, n.true_().type());
+            std::cerr << "----\n";
+            hilti::render(std::cerr, n.false_().type());
+            std::cerr << "---- " << hilti::type::sameExceptForConstness(n.true_().type(), n.false_().type())
+                      << std::endl;
+
```
bbannier, line 285: Is this leftover debug code?

tests/spicy/types/optional/coercion.spicy
```diff
@@ -0,0 +1,19 @@
+# @TEST-EXEC: ${SPICYC} -j %INPUT >output
+# @TEST-EXEC: btest-diff output
```
bbannier, line 2: No need to diff anything as the test doesn't produce any output.
```suggestion
```

tests/Baseline/spicy.types.optional.coercion/output
```diff
@@ -0,0 +1 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
```
bbannier, line 1: We probably don't need this file, see below.

hilti/toolchain/src/compiler/coercion.cc
```diff
@@ -424,6 +424,22 @@ struct VisitorType : public visitor::PreOrder<std::optional<Type>, VisitorType>
...
+            const auto& d = t->dereferencedType();
+
+            if ( type::sameExceptForConstness(s, d) ) {
+                if ( style & CoercionStyle::Assignment )
+                    // Assignments copy, so it's safe to turn even into the
+                    // destination without considering constness.
+                    return dst;
+
+                if ( type::isConstant(s) && type::isConstant(d) )
+                    // It's safe to turn a non-const value into a const value.
```
bbannier, line 438: I am confused by the comment as the condition seems to be _safe to turn const value into const value_. If I change the code to match the comment `if ( ! type::isConstant(s) && type::isConstant(d) )` the tests still pass, so it would be great if we had something in the test suite which checks non-assignment behavior if possible.

bbannier (changes requested):

hilti/toolchain/src/compiler/coercion.cc
rsmmr: Yeah, that's a typo. That said, turns out I can't come up with an example where this is actually needed. So I'll just take this out until we find a case where's it's needed.

bbannier (approved):

bbannier:

PR #1260, opened 2022-08-11
Fixing issue with `inout` parameters.

### LINKED ISSUE #1170: Cannot pass context as `inout` parameter
In spicy-1.3.0 it was possible to pass a context as an `inout` parameter. This allowed users to work around #1059 in a supposedly forward-compatible way. That workaround is broken with spicy-1.4.1.

```ruby
# analyzer.spicy
module foo;

type Ctx = unit { };

type X = unit(inout ctx: Ctx) { };

public type Y = unit {
	%context = Ctx;
	: X(self.context());
};
```

```console
$ spicyc -dj analyzer.spicy
/private/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/foo_203c58e6432c1c5-ea163d6951a7f53a.cc:187:25: error: no matching
      conversion for functional-style cast from 'ValueReference<__hlt::foo::Ctx>' to '__hlt::foo::X'
    __transient_anon = (__hlt::foo::X((*__self).__context.derefAsValue()));
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/private/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/foo_203c58e6432c1c5-ea163d6951a7f53a.cc:43:9: note: candidate
      constructor not viable: no known conversion from 'ValueReference<__hlt::foo::Ctx>' to 'const __hlt::foo::X' for 1st
      argument
        X(const X&) = default;
        ^
/private/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/foo_203c58e6432c1c5-ea163d6951a7f53a.cc:44:9: note: candidate
      constructor not viable: no known conversion from 'ValueReference<__hlt::foo::Ctx>' to '__hlt::foo::X' for 1st argument
        X(X&&) = default;
        ^
/private/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/foo_203c58e6432c1c5-ea163d6951a7f53a.cc:119:15: note: candidate
      constructor not viable: expects an l-value for 1st argument
    inline X::X(::hilti::rt::ValueReference<__hlt::foo::Ctx>& __p_ctx) : __p_ctx(std::move(__p_ctx)) {
              ^
/private/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/foo_203c58e6432c1c5-ea163d6951a7f53a.cc:116:15: note: candidate
      constructor not viable: requires 0 arguments, but 1 was provided
    inline X::X() {
              ^
1 error generated.
```

Bisecting Spicy points to 29054778950981cda2a5adbae3d66cae953cc795 as the commit where this broke.

rsmmr:
Fix contexts not allowing being passed `inout`.

This turned into a larger fix needing some internal changes, but it's a good change because it unifies a couple of hacks we had so far, and probably also prevents some other trouble we hadn't even found yet.

The problem was that `inout` parameters require a LHS value that can be passed by reference, but our C++ code generator had no way of telling if a C++-side expression ended up representing a RHS or LHS value; and hence couldn't turn a RHS into a LHS when it was needed for passing by reference.

We had a couple of places in the code generator already were we accounted for this manually, but generally the problem wasn't solvable without consistency tracking the LHS/RHS information. This change now adds  that state to `cxx::Expression` so that it can be set when the expression is created and then follows it around to where it's needed.


hilti/toolchain/src/compiler/codegen/types.cc
```diff
@@ -1111,6 +1111,26 @@ const CxxTypeInfo& CodeGen::_getOrCreateTypeInfo(const hilti::Type& t) {
         });
 }
 
+cxx::Expression CodeGen::_applyLhs(cxx::Expression expr, const Type& type, bool lhs) {
```
bbannier, line 1114: WDYT about moving `lhs` into the caller so that this function becomes a simple conversion to a LHS? We'd need the same boilerplate in the callers, but this would still feel cleaner to me, e.g.,

```cpp
    if ( auto x = Visitor(this).dispatch(c) ) {
        if ( ! lhs )
            return *x;
        else
            return _applyLhs(*x, c.type());
    }
```

hilti/toolchain/src/compiler/codegen/types.cc
```diff
@@ -1111,6 +1111,26 @@ const CxxTypeInfo& CodeGen::_getOrCreateTypeInfo(const hilti::Type& t) {
...
+
+    if ( type.isA<type::ValueReference>() )
+        result = cxx::Expression{fmt("(%s=(%s).asSharedPtr())", tmp, expr), cxx::Side::LHS}; // avoid copy
+    else
+        result = cxx::Expression{fmt("(%s=(%s))", tmp, expr), cxx::Side::LHS};
+
+#if 0
+    // This can help show where LHS conversions happen unexpectedly (they should be very rare).
+    std::cerr << "LHS| " << type.typename_() << ": " << expr << " -> " << result << std::endl;
+#endif
```
bbannier, line 1129: Does it make sense to add this to a debug stream?

bbannier (dismissed):

hilti/toolchain/src/compiler/codegen/types.cc
rsmmr: We don't have a `codegen` debug stream yet, but I'll create one, makes sense.

bbannier (approved):

PR #1262, opened 2022-08-22
Allow combining `&eod` with `&until` or `&until-including`.

### LINKED ISSUE #1249: Support parsing of `bytes` with `&until` of optional sequence
I am trying to extract a `bytes` until a potentially multi-byte pattern is seen or up to EOD, i.e., the delimiter is optional. This seems harder than needed

- it seems I cannot use a regexp since we do not support negative lookahead, e.g., `x: /(?!MY_DELIM)*/` is not supported
- I cannot use `x: bytes &until=DELIM` and recover from the EOD error since the field will only get updated once `DELIM` was seen; it might be possible to on recovery invoke a parser for `bytes` with know `&size` (needs computation of `&size` and probably also an arbitrary sync point following the `bytes` field, e.g., `: b"" &synchronize`). 
- with multi-byte patterns for `DELIM` one could extract a list of single `bytes` and accumulate the full data on the side. One would then need to strip the `DELIM` from the accumulated data again and reset the input position

I believe we might have all the pieces in place to allow combining `[&until|&until-including]` and `&eod` which is currently not allowed. We should check whether we can enable this in some form or another for all container types.

bbannier:
In case one of `&until` or `&until-including` and `&eod` is given,
parsing will continue until either the given expression or EOD is
reached.

Closes #1249.


tests/spicy/types/bytes/parse-until-eod.spicy
```diff
@@ -0,0 +1,16 @@
+# @TEST-DOC: Validates that `&until` or `&until-including` can be combined with `&eod`.
+#
+# @TEST-EXEC: spicyc -dj %INPUT -o test.hlto
+# @TEST-EXEC: ${SCRIPTS}/printf '0' | spicy-dump -d test.hlto -p Test::X >output
+# @TEST-EXEC: ${SCRIPTS}/printf '0' | spicy-dump -d test.hlto -p Test::Y >output
```
rsmmr, line 5:
```suggestion
# @TEST-EXEC: ${SCRIPTS}/printf '0' | spicy-dump -d test.hlto -p Test::Y >>output
```

rsmmr (approved):

tests/spicy/types/bytes/parse-until-eod.spicy
bbannier: Good catch, I'll also update the baseline.

PR #1263, opened 2022-08-22
Clarify that `%context` can only be used in top-level units. [skip CI]

bbannier:


doc/programming/parsing.rst
```diff
@@ -1908,7 +1908,7 @@ error prone. As an alternative, a unit can make use of a dedicated
 lifetime determined by the host application running the parser. For
 example, Zeek will tie the context to the underlying connection.
 
-A public unit can declare its context through a unit-level property
+A top-level unit can declare its context through a unit-level property
```
rsmmr, line 1911: Does it need to be both? I realize that a top-level unit cannot be used if it's not public, but just from Spicy's perspective, does it compile if it has a `%context` but is not declared `public`?

bbannier: I a not sure how to answer this question on the technical level: I could remove the validator so I can add a context to a non-`public` unit, but we already know that it does not work for non-toplevel units and I cannot use a non-`public` unit as a top-level unit.

I think the important bit here is that the unit is a top-level unit so we correctly manage the context. Since we cannot enforce that at compile time we check for `public`, but that's not the real requirement.

rsmmr: So the thing here is that I think the new wording is slightly misleading at this point in the docs. The fact that it only works for top-level units is something that's relevant really on the side of the whatever code is using the Spicy parser; in our case, that's the Zeek plugin. Indeed, a little below this there's this text: `the Zeek plugin always creates a single context value shared by all top-level units belonging to the same connection, enabling parsers to maintain bi-directional, per-connection state.` I'll take another look later.

rsmmr: I pushed a couple of changes elaborating on this further.

bbannier (commented):
This looks good to me, let's move this in.

PR #1270, opened 2022-09-28
Fix usage of `self` in field convert expressions.

rsmmr:
This is on top of #1269 (commits 1-3). There's one new helper script included (commit 4); plus a small independent fix (commit 5); plus finally the main fix in commit 6.


scripts/ast-diff
```diff
@@ -0,0 +1,23 @@
...
+
+tmp1=/tmp/$(basename $0).$$.1
+tmp2=/tmp/$(basename $0).$$.2
+
+trap "rm -f ${tmp1} ${tmp2}" EXIT
+
+cat $1 | filter >${tmp1}
+cat $2 | filter >${tmp2}
+
+diff -u ${tmp1} ${tmp2}
```
bbannier, line 23: Just FTR, when I tried such a script before I always ran into issues with the `diff` binary behaving differently on different platforms; I then instead implemented the script in Python and used [`difflib`](https://docs.python.org/3/library/difflib.html) (e.g., [`difflib.unifed_diff`](https://docs.python.org/3/library/difflib.html#difflib.unified_diff) here).

This is not an issue now since we do not use this script to compute baselines, but we might need to move this whole script to Python at that point.

hilti/toolchain/src/ast/expression.cc
```diff
@@ -1,3 +1,15 @@
...
 
 #include <hilti/ast/expression.h>
+#include <hilti/ast/expressions/id.h>
+
+using namespace hilti;
+
+bool expression::isResolved(const detail::Expression& e, type::ResolvedState* rstate) {
+    // We always consider `self` expressions as fully resolved to break the
+    // cycle with the type that they are pointing to.
+    if ( auto id = e.tryAs<expression::ResolvedID>(); id && id->id() == ID("self") )
```
bbannier, line 11: With this any identifier `self` will always be considered resolved, right? We currently allow this code and it does what is expected:

```ruby
module foo;
global self = 42;
print self;
```

Even with this patch this example still resolves successfully and behaves the same, but I wonder whether there are other cases which would now be broken.

Maybe we could make this less of an issue if we prevented users from creating their own bindings named `self` (could think of variable and parameter decls, not sure there is more).

bbannier: I guess this works since in this example `self` would always have an id `foo::self`. I also tried the following:

```ruby

module foo;

function x() : uint64 {
	local self = 42;
	return self;
}
print x();
```
This still works even though `self` now has an id `self` 🤯.

rsmmr: I'm adding a general check that declarations can't redefine `self`.

scripts/ast-diff
rsmmr: Ack. This script is meant for manual debugging, not baselining, so something we could change later if we wanted to do that.

hilti/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -107,12 +107,22 @@ struct VisitorPost : public hilti::visitor::PreOrder<void, VisitorPost>, public
 
     ////// Declarations
 
+    // Perform validation of ID names suitable for all types of declarations.
+    void checkDeclarationID(const ID& id, const position_t& p) {
+        if ( id.str() == "self" )
+            error("cannot use 'self' as identifier", p);
+    }
```
bbannier, line 114: Let's instead implement this as

```cpp
    void operator()(const Declaration& d, position_t p) {
        // Perform validation of ID names suitable for all types of declarations.
        if ( d.id().str() == "self" )
            error("cannot use 'self' as identifier", p);
    }
```

and remove the explicit calls. I believe this would catch additional declarations.

We could add more tests for that (e.g., funcion or module decls named `self`), but below test would then also report redefinition errors.

bbannier (changes requested):

hilti/toolchain/src/compiler/visitors/validator.cc
rsmmr: Doesn't work unfortunately because `declaration::Expression` is internally used for mapping `self` to its unit instance. The validator code would trigger for that, too. We could white-list that, but not sure that's really better then.

bbannier: Hmm, I didn't see that.

The current implementation misses some declarations, e.g, `Module`, `Type`, or `Function`, so I think using a generic visitor over `Declaration` with whitelisting for `declaration::Expression` is still better (if only since it allows us to put a comment why we ignore `Expression` instead of no coding being there).

This still has issues visiting a `declaration::Module` and I was also unable to visit it explicitly. I am unsure why that is (such nodes would e.g., not be root nodes).

With a generic visitor all of these test cases but the last one should be rejected:

```ruby
module foo;
function self(xyz: uint8) {}
##########################################
module foo;
type self = struct {
	self: uint8;
};
##########################################
module foo;
type self = unit {
	self: uint8;
};
##########################################
module self;
```

rsmmr: Ok, convinced (but not sure why `Module` doesn't trigger)

PR #1279, opened 2022-10-19
Remove `--optimize` flag.

rsmmr:
This was already a no-op, and just confusing/misleading.


hilti/runtime/src/library.cc
```diff
@@ -42,12 +41,6 @@ void hilti::rt::library::Version::checkCompatibility() const {
     if ( hilti_version != PROJECT_VERSION_NUMBER )
         warning(fmt("module %s was compiled with HILTI version %d, but using HILTI version %d", path.filename(),
                     hilti_version, PROJECT_VERSION_NUMBER));
-
-    if ( hilti::rt::isDebugVersion() && optimize )
```
bbannier, line 46: This removes the only user of `isDebugVersion`, maybe we should remove it as well since it doesn't seem useful anymore. For codegen we have been using `Codegen::context()->options().debug` all along.

bbannier (approved):

bbannier (approved):

PR #1281, opened 2022-10-21
Document how to pull custom C++ code into Spicy.

### LINKED ISSUE #1265: Add documentation on how to expose C or C++ functions or types to Spicy
We should document how to expose user-provided C or C++ functions and types to Spicy. Typical use cases would be

- implement a function in C++ and directly link it into the HLTO file
- expose a library function in an external library (could be static archive or shared library) to Spicy code, and how to link it

We might also want to document type mappings, or suggest a workflow which inspects generated C++ for Spicy-side declarations to guide users on what types to use.

rsmmr:
Closes #1265.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -53,17 +53,17 @@ todo_include_todos = True
 
 # Extlinks extension
 extlinks = {
-    "repo":  ("https://github.com/zeek/spicy/blob/main/%s", "#"),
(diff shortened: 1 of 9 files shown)


bbannier (dismissed):
This reads great, given how challenging this is.

Do we have tickets for the TODOs?

rsmmr:
> Do we have tickets for the TODOs?

I'll create them.

bbannier (approved):

PR #1287, opened 2022-10-26
Relax JIT hash inputs to improve ccache hitrate in CI

bbannier:


hilti/toolchain/src/compiler/jit.cc
```diff
@@ -169,8 +169,7 @@ bool CxxCode::save(std::ostream& out) const {
 JIT::JIT(const std::shared_ptr<Context>& context, bool dump_code)
     : _context(context),
       _dump_code(dump_code),
-      _hash(rt::hashCombine(hilti::configuration().hash(),
-                            std::hash<std::string>{}(hilti::rt::filesystem::current_path().string()))) {}
+      _hash(std::hash<std::string>{}(hilti::rt::filesystem::current_path().string())) {}
```
rsmmr, line 172: I'm not quite sure anymore what our hash is now protecting against? Just the current path isn't really anything unique. Do we still need the hash then?

bbannier: With this change it only protects against concurrent JIT jobs working on identical files from racing on creating the file (all temporary files go into the same directory). This can happen a lot with e.g., the linker file which has the same contents for the majority of our btests. The thinking at the time was that a typical user wouldn't JIT the same code from the same directory simultaneously.

Even if we implemented proper file locking we probably would still want different output files for same identical files for throughput.

rsmmr (approved):

PR #1288, opened 2022-10-26
Replace `Lambda` class.

### LINKED ISSUE #1273: `hilti/runtime/include/lambda.h` is licensed under CPOL license which is not a recognized license
The file `hilti/runtime/include/lambda.h` is licensed under the CPOL license which is not a recognized open source license according to e.g., [OSI](https://opensource.org/licenses/alphabetical), mostly [due to the restrictions it places on how the software can be used](https://en.wikipedia.org/wiki/Code_Project_Open_License#Status_as_an_open-source_license). It seems this license is incompatible with how Spicy or other Zeek software are released and [could prevent users from deploying  Spicy or zeek-5 in its default configuration](https://zeekorg.slack.com/archives/C012CQV1K29/p1665161354250959). We should look into replacing that code.

bbannier:
We originally introduced `Lambda` to reduce compilation overhead from use of `std::function`, but the code we pulled in was under a license which is incompatible with the licenses typically used for other Zeek projects.

Since we only ever used `Lambda` to store exactly one type of callback we now replace it with a much smaller custom implementation. We still require type-erase to store lambdas, but know both argument as well as return types.

Closes #1273.


rsmmr (approved):

bbannier:
I also backported this PR to `release/1.5`.

PR #1289, opened 2022-10-26
Downgrade required Flex version.

0xxon:
At least for the release in Zeek 5.0.2 this compiled fine with the older Flex for me :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ cmake_minimum_required(VERSION 3.15.0)
 
 project(spicy LANGUAGES ASM C CXX)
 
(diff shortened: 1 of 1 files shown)


bbannier (approved):
Thanks for digging into this @0xxon.

I confirmed that with this change I can build with the flex which comes with centos7.

Note to @bbannier: this should be backported to `release/1.5` as well.

bbannier:
I also backported this PR to `release/1.5`.

0xxon:
One more note - once the submodule pointer in Zeek is updated, we can change ci there to use the onboard-Centos7-flex.

PR #1291, opened 2022-10-30
cmake replace quote string

nabbi:
CMake Error
string sub-command REPLACE requires at least four arguments.

cmake version 3.24.2


bbannier:
This looks good to me, but as a user of cmake-3.24.2 I wonder how to trigger the issue this fixes. Could you explain what you did to run into the error you posted? I'd assume that the same could also trigger the issue fixed by zeek/spicy-plugin#148.

nabbi:
Hi @bbannier -
I encountered this during Gentoo's sandbox build process using Portage 3.0.38.1
Been working on an [updated ebuild](https://github.com/nabbi/oubliette-overlay/tree/master/net-analyzer/zeek) for this distro and found I needed to patch this out. Yes, both spicy and spicy-plugin were encountered with the same build process. Without digging too much into this, I suspect CMAKE_CXX_FLAGS_RELWITHDEBINFO is being manipulated by portage (as manually configure does not trigger the behavior)

Thank you :)

bbannier:
@nabbi, thanks for the context, especially given how easy it is to add patches to ebuilds. I'll take a look.

As a heads up, If you are preparing an ebuild for the latest release, consider backporting [this Spicy patch](https://github.com/zeek/spicy/commit/5fdb77db6af28b4c88b807ada9759c1335c7121e) which has already landed in Spicy `main`; with that the package should be fully licensed under a OSI-approved Open Source licenses.

bbannier:
Landed in both `main` and `release/1.5`, available in spicy-1.5.2.

PR #1292, opened 2022-11-07
Log successful synchronization.

bbannier:


rsmmr (commented):
This is good, one more idea though: I think all the the sync logs would git better into `spicy-verbose` because (1) that gives the context of where parsing it at, and (2) "spicy" is mean primarily for understanding parsing results (vs. parsing mechanics). How about changing that here as well?

bbannier:
> This is good, one more idea though: I think all the the sync logs would git better into `spicy-verbose` because (1) that gives the context of where parsing it at, and (2) "spicy" is mean primarily for understanding parsing results (vs. parsing mechanics). How about changing that here as well?

Done.

Just to make sure, do we consider such a change breaking so that this patch should not be backported?

.update-changes.cfg
```diff
@@ -1,4 +1,4 @@
-new_commit_msg="Update CHANGES. [skip ci]"
+new_commit_msg="Update CHANGES."
```
rsmmr, line 1: Unintended change?

rsmmr (dismissed):

rsmmr:
> Just to make sure, do we consider such a change breaking so that this patch should not be backported?

I'd generally exclude debugging output from any stability commitments, but still no need to backport I think; not a bugfix nor crucial to have in other ways.

.update-changes.cfg
bbannier: Yes, this slipped in when I cut the release.

PR #1295, opened 2022-11-12
Add `pack` and `unpack` operators to Spicy.

rsmmr:


doc/programming/language/packing.rst
```diff
@@ -0,0 +1,57 @@
...
+  expect for ``DATA``. The ``unpack`` operator returns a 2-tuple
+  ``(VALUE, REMAINDER)`` where ``VALUE`` is the parsed value, and
+  ``REMAINDER`` is any bytes left over from ``DATA`` that weren't used
+  for parsing the value. If parsing fails, ``unpack`` throws an
+  ``InvalidValue`` exception.
+
+  As an example, ``unpack<uint16>(b"\x02\x01XYZ",
+  spicy::ByteOrder::Network)`` returns the tuple ``(513, b"XYZ")``.
+
+The following table summarizes the types that currently support
```
bbannier, line 31: I wonder whether we should also add `pack` (and for symmetry `unpack`) for `bytes`. The intended use case would be to change the byte order of existing `bytes` values.

doc/programming/language/packing.rst
```diff
@@ -0,0 +1,57 @@
+
+.. _packing:
+
+==========================
+Packing / Unpacking Values
+==========================
+
+A few of Spicy's atomic types support low-level conversion from, or
+into, a binary representation through two operators:
```
bbannier, line 9: Regarding the byte order argument to `pack` and `unpack`, for parsing there is an implicit byte order. Do you think it would make sense to also have a default byte order for `pack` and `unpack`?

hilti/toolchain/include/ast/operators/generic.h
```diff
@@ -17,29 +23,150 @@
...
+    bool isLhs() const { return false; }
+    auto priority() const { return hilti::operator_::Priority::Normal; }
+
+    const std::vector<Operand>& operands() const {
+        static std::vector<Operand> _operands = {Operand{{}, type::Tuple(type::Wildcard())}};
+        return _operands;
+    }
+
+    void validate(const expression::ResolvedOperator& i, operator_::position_t p) const {
+        const auto args = i.op0().type().template as<type::Tuple>().elements();
```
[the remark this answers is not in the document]
rsmmr: Honestly I'm not even sure what exactly prevents us from using per-type overloads here instead of a single generic operator; I indeed just adapted this from what `unpack` already does. It could be that the resolver is still not able to match against specific enum types (we tend to use wildcard enums elsewhere). I think it's ok for now.

doc/programming/language/packing.rst
rsmmr: Not sure a notion of byte-order makes sense for just a raw sequence of bytes? "Byte order" needs some kind of more-than-1-byte semantical unit. `pack` is about transforming something into bytes, so I think a `pack` *for* bytes could really just be a no-op.

doc/programming/language/packing.rst
rsmmr: I prefer leaving this explicit. It's pretty low-level functionality, so I think we can require people to specify what they want here for clarity.

hilti/runtime/src/tests/address.cc
```diff
@@ -123,6 +123,18 @@ TEST_CASE("asInAddr") {
...
+    CHECK_EQ(address::pack(Address("1.2.3.4"), ByteOrder::Big), "\x01\x02\x03\x04"_b);
+    CHECK_EQ(address::pack(Address("4.3.2.1"), ByteOrder::Little), "\x01\x02\x03\x04"_b);
+    CHECK_EQ(address::pack(Address("1.2.3.4"), ByteOrder::Big), "\x01\x02\x03\x04"_b);
+    CHECK_EQ(address::pack(Address("102:304:102:304:506:708:901:203"), ByteOrder::Big),
+             "\x01\x02\x03\x04\x01\x02\x03\x04\x05\x06\x07\x08\x09\x01\x02\x03"_b);
+    CHECK_EQ(address::pack(Address("302:109:807:605:403:201:403:201"), ByteOrder::Little),
+             "\x01\x02\x03\x04\x01\x02\x03\x04\x05\x06\x07\x08\x09\x01\x02\x03"_b);
+    CHECK_EQ(address::pack(Address("102:304:102:304:506:708:901:203"), ByteOrder::Big),
+             "\x01\x02\x03\x04\x01\x02\x03\x04\x05\x06\x07\x08\x09\x01\x02\x03"_b);
+}
```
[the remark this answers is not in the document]
rsmmr: Line 141 is missing one `\x20` at the end, Hard to see in the output because that happens to be a space.

bbannier (approved):
This looks good. The `lint` job seems to fail since the `clang-tidy-12` we use in the job has trouble with the macro, but this works fine with newer versions. Let me try to bump it in a different PR.

bbannier:
> This looks good. The `lint` job seems to fail since the `clang-tidy-12` we use in the job has trouble with the macro, but this works fine with newer versions. Let me try to bump it in a different PR.

I have pushed an updated clang-tidy setup for CI to #1318. This has tested successfully locally, but I would like to see it pass in CI. You'd want to cherry-pick the two clang-tidy-related commits.

PR #1297, opened 2022-11-12
Allow activating JIT debug streams via `HILTI_DEBUG`.

### LINKED ISSUE #1296: Cannot activate debug streams like e.g., `compiler` via `HILTI_DEBUG`
While looking at zeek/spicy-dhcp#8 I would have liked to activate the compiler debug stream. While when building manually I could pass e.g., `-D compiler` to a `spicyc` invocation, it seems like I cannot use `HILTI_DEBUG` since it is currently only evaluated when setting up an environment  for executing JIT'ed code.

bbannier:
Previously to activate debugs streams during JIT one needed to pass streams to activate via a command line argument `-D`. This can make it hard to debug compilation issues if one cannot edit the compiler invocation (e.g., the compiler invocation for out-of-tree Spicy analyzers in Zeek lives in the installation of spicy-plugin).

With this patch we now support activating streams listed under a compiler's `-D help` via `HILTI_DEBUG`. Since `HILTI_DEBUG` is already used to activate HILTI runtime flags not necessarily supported by a compiler, we ignore activation errors from unknown streams.

Closes #1296.

### DIFF
--- a/hilti/toolchain/src/compiler/driver.cc
+++ b/hilti/toolchain/src/compiler/driver.cc
@@ -10,6 +10,7 @@
 
 #include <hilti/rt/json.h>
 #include <hilti/rt/libhilti.h>
+#include <hilti/rt/util.h>
 
 #include <hilti/ast/declaration.h>
 #include <hilti/ast/detail/visitor.h>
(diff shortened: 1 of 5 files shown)


rsmmr (commented):
Actually we already have two ways to do that: `ZEEK_SPICY_PLUGIN_OPTIONS` and (less useful) the script-level option `Spicy:: codegen_debug`, see https://docs.zeek.org/projects/spicy/en/latest/zeek.html#options. I'd prefer to keep `HILTI_DEBUG` limited to runtime execution output.

bbannier:
> 

Amazing, thanks.

PR #1301, opened 2022-11-15
Add library function to parse an address from string or bytes.

### LINKED ISSUE #1294: Convert Spicy string to IP address
It would be nice to have a built-in function to parse a Spicy string (or bytes) as an IP address.

rsmmr:
## Parses an address from a string. The address can be in standard IPv4 or IPv6
    ## ASCII representation. The function raises ``InvalidArgument`` if the string
    ## could not be parsed.
    public function parse_address(s: string) : addr;

    ## Parses an address from a bytes instance. The address can be in standard IPv4
    ## or IPv6 ASCII representation. The function raises ``InvalidArgument`` if the
    ## string could not be parsed.
    public function parse_address(b: bytes) : addr;

Closes #1294.

### DIFF
--- a/tests/spicy/lib/parse-address.spicy
+++ b/tests/spicy/lib/parse-address.spicy
@@ -1,0 +1,14 @@
+# @TEST-EXEC: ${SPICYC} -j %INPUT
+#
+# @TEST-DOC: Exercise ``spicy::parse_address()``
+
+module Test;
+
+import spicy;
+
+# We only test basic functionality because the parsing itself is unit-tested
+# inside the runtime librart.
+assert spicy::parse_address("2001:0db8:85a3:0000:0000:8a2e:0370:7334") == [2001:0db8:85a3:0000:0000:8a2e:0370:7334];
+assert spicy::parse_address(b"192.168.1.1") == 192.168.1.1;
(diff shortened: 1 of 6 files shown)


tests/spicy/lib/parse-address.spicy
```diff
@@ -0,0 +1,13 @@
...
+# @TEST-DOC: Exercise ``spicy::parse_address()``
+
+module Test;
+
+import spicy;
+
+# We only test basic functionality because the parsing itself is unit-tested
+# inside the runtime librart.
+assert spicy::parse_address("2001:0db8:85a3:0000:0000:8a2e:0370:7334") == [2001:0db8:85a3:0000:0000:8a2e:0370:7334];
+assert spicy::parse_address(b"192.168.1.1") == 192.168.1.1;
```
bbannier, line 12: I don't see use parsing dual IPv6 address anywhere, could you sneak in a test for that here? This already works.
```suggestion
assert spicy::parse_address(b"192.168.1.1") == 192.168.1.1;
assert spicy::parse_address(b"2001:db8::FFFF:192.168.0.5") == [2001:db8::ffff:c0a8:5];
```

bbannier (approved):

PR #1304, opened 2022-11-23
Fix incorrect offset computation in `advanceToNextData`.

### LINKED ISSUE #1303: Synchronization too late if gap in input
@Mohan-Dhawan reported that he sees synchronization on gap skipping too far in the input. This causes him to loose more data than needed (in fact a previous binpac implementation sees more data).

This seems to reduce down to the behavior already encoded in the test `spicy.types.unit.synchronize-on-gap`: while the grammar there [can parse any of `(A|B|C)`](https://github.com/zeek/spicy/blob/05bda615d16a262df5ec6d06555107b4a2c8baa6/tests/spicy/types/unit/synchronize-on-gap.test#L14), inserting [an unrelated gap](https://github.com/zeek/spicy/blob/05bda615d16a262df5ec6d06555107b4a2c8baa6/tests/spicy/types/unit/synchronize-on-gap.test#L30-L32) causes the parser [synchronize on `C` instead of `B` which is also in the input](https://github.com/zeek/spicy/blob/05bda615d16a262df5ec6d06555107b4a2c8baa6/tests/Baseline/spicy.types.unit.synchronize-on-gap/output#L21).

We should look into whether we can synchronize earlier in the presence of gaps so that in above example we recover `ABC`.

bbannier:
To compute the next `View` in `advanceToNextData` after a gap chunk we would previously compute offsets incorrectly in that we would implicitly assume that the begin of the original `View` started with zero offset. This is unlikely true for any but the most trivial cases and often caused e.g., resynchronization to jump to incorrect offsets much after the correct position.

With this patch we now use a correct offset computation.

Closes #1303.

### DIFF
--- a/hilti/runtime/src/tests/regexp.cc
+++ b/hilti/runtime/src/tests/regexp.cc
@@ -356,6 +356,43 @@ TEST_CASE("advance") {
 
         CHECK_EQ(RegExp("[ \\n]*", {}).tokenMatcher().advance(s.view()), std::make_tuple(1, stream::View()));
     }
+
+    SUBCASE("advance into gap") {
+        // This is a regression test for GH-1303.
+        auto s = Stream();
+        s.append("A");
+        s.append(nullptr, 1024);
+        s.append("BC");
+        s.freeze();
+
+        const auto re = RegExp("(A|B|C)");
+
+        auto cur = s.view();
+
+        // Match on `A`.
+        {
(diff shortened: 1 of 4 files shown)


bbannier:
This should probably be backported to all affected branches.

rsmmr (approved):

bbannier:
Also backported to  `release/1.5`.

PR #1308, opened 2022-12-01
cmake/FindGoldLinker: Put -fuse-ld=gold before existing flags

awelzel:
When building Zeek and specifying LDFLAGS=-fuse-ld=lld to prefer lld over gold, the linker invocations for spicy continue to use gold due to fuse-ld=gold being appended whenever it's detected.

Maybe USE_GOLD should be forced to off when compiled within Zeek, anyway.

Still, when configuring just spicy standalone with "LDFLAGS=-fuse-ld=lld ./configure", currently gold detection will overwrite that, unless one explicitly passes --disable-gold, as well. That's minimally surprising, too.


cmake/FindGoldLinker.cmake
```diff
@@ -1,6 +1,6 @@
 ## Enable Gold linker if available.
 
-option(USE_GOLD "Use Gold linker" ON)
+option(USE_GOLD "Use Gold linker" OFF)
```
bbannier, line 3: Note to self: this preserves the original behavior since `./configure` defaults to `USE_GOLD=ON`.

bbannier (approved):

bbannier:

PR #1312, opened 2022-12-04
Validate that unit using `self.context()` declares `%context`.

### LINKED ISSUE #1311: Segmentation fault running spicy-driver using %context / self.context()
Observing a segfault with the following parser when running it through spicy-driver:

```
$ cat context-crash.spicy 
module Test;

type State = tuple<ssl_requested: uint8>;

type X = unit {
  val: bytes &size=3 {
    if ( self.val == b"ssl" )
        self.context().ssl_requested = 1;
  }
};

public type Y = unit {
  %context = State;
  x: X;
};
```

```                                                                                                                                                        
$ docker pull zeekurity/spicy:latest                                                               
latest: Pulling from zeekurity/spicy                                                                                                                                    
Digest: sha256:0bab492a0e2a68d05cf69c2aa56f2e9c56a7f66162c3acf280acc665593d4254                                                                                         
Status: Image is up to date for zeekurity/spicy:latest                            
docker.io/zeekurity/spicy:latest                                                    
$ docker run -v $(pwd):/src -w /src  -it --rm  zeekurity/spicy-dev:latest
# spicyc --version
spicyc v1.6.0-dev.118 (258565a4)
# spicy-driver ./context-crash.spicy
Segmentation fault (core dumped)
```

Maybe self.context() isn't allowed in non-public types, but the segfault does not help to figure out what I'm doing wrong :-)

bbannier:
When declaring `context` methods we previously would `assert` that any unit using `self.context()` actually would declare a `%context`. This is not true and lead to segfaults in release builds which do not execute code in `assert`.

This patch adds new validation which rejects uses of `self.context()` in units which did not declare a context. Since validation happens after declaration of the `context` methods has executed (the part which previously had the `assert`), we return a dummy type (a void) for such units, and then reject the code later.

Closes #1311.


spicy/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -863,6 +863,16 @@ struct VisitorPost : public hilti::visitor::PreOrder<void, VisitorPost>, public
             error("unit type cannot be a filter, %filter missing", p);
     }
 
+    void operator()(const operator_::unit::ContextConst& n, position_t p) {
```
bbannier, line 866: I added this operator for symmetry, but have no idea how to trigger it.

spicy/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -863,6 +863,16 @@ struct VisitorPost : public hilti::visitor::PreOrder<void, VisitorPost>, public
...
+        if ( auto x = n.op0().type().tryAs<type::Unit>(); x && ! x->contextType() )
+            error("context() used in a unit which did not declare %context", p);
```
rsmmr, line 868: Suggest slight clarification:

```suggestion
            error("context() used with a unit which did not declare %context", p);
```

spicy/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -863,6 +863,16 @@ struct VisitorPost : public hilti::visitor::PreOrder<void, VisitorPost>, public
...
+    }
+
+    void operator()(const operator_::unit::ContextNonConst& n, position_t p) {
+        if ( auto x = n.op0().type().tryAs<type::Unit>(); x && ! x->contextType() )
+            error("context() used in a unit which did not declare %context", p);
```
rsmmr, line 873:
```suggestion
            error("context() used with a unit which did not declare %context", p);
```

tests/Baseline/spicy.types.unit.context-undeclared/output
```diff
@@ -0,0 +1,3 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+[error] <...>/context-undeclared.spicy:9:15: context() used in a unit which did not declare %context
```
rsmmr, line 2:
```suggestion
[error] <...>/context-undeclared.spicy:9:15: context() used with a unit which did not declare %context
```

rsmmr (dismissed):

PR #1313, opened 2022-12-05
Allow expressions with type constructors

rsmmr:


hilti/toolchain/src/compiler/visitors/constant-folder.cc
```diff
@@ -0,0 +1,70 @@
...
+namespace {
+
+// For now, this is only a very basic constant folder that only covers cases we
+// need to turn type constructor expressions coming with a single argument into
+// ctor expressions.
+struct VisitorConstantFolder : public visitor::PreOrder<Ctor, VisitorConstantFolder> {
+    // Helper to cast an uint64 to int64, with range check.
+    int64_t to_int64(uint64_t x, position_t& p) {
+        if ( x > std::numeric_limits<int64_t>::max() )
+            logger().error("signed integer value out of range 2", p.node.location());
```
bbannier, line 25: Probably 
```suggestion
            logger().error("unsigned integer value out of range", p.node.location());
```

This is untested.

Similary for the two error messages below (these should indeed by `signed integer` though.

hilti/toolchain/src/compiler/visitors/constant-folder.cc
```diff
@@ -0,0 +1,70 @@
...
+    result_t operator()(const operator_::signed_integer::SignNeg& n, position_t p) {
+        if ( auto op = detail::foldConstant<ctor::SignedInteger>(n.op1()) ) {
+            if ( op->value() >= 0 ) {
+                if ( op->value() < -std::numeric_limits<int64_t>::min() ) {
+                    logger().error("signed integer value out of range X", p.node.location());
+                    return {};
+                }
+            }
+            else {
+                if ( -op->value() > std::numeric_limits<int64_t>::max() ) {
```
bbannier, line 41: I am wondering whether these comparisons against `min`/`max` values are actually needed or correct (e.g., could `-std::numeric_limits<int64_t>::min()` overflow?). It would really be great if we had tests for both branches.

hilti/toolchain/src/compiler/visitors/printer.cc
```diff
@@ -799,7 +799,7 @@ struct Visitor : visitor::PreOrder<void, Visitor> {
 
     void operator()(const type::Interval& n) { out << const_(n) << "interval"; }
 
-    void operator()(const type::Member& n) { out << const_(n) << "<member>"; }
+    void operator()(const type::Member& n) { out << const_(n) << n.id(); }
```
bbannier, line 802: Not sure this is exactly due to this line, but the printing of the interval operators does not seem to round-trip, e.g.,

```console
$ cat x.hlt
module x {
global i = interval(10);
global j = interval_ns(10);
}
$ hiltic -p x.hlt
module x {

global interval i = 10.000000s;
global interval j = 0.000000s;

}
```

In addition to this not compiling, we also loose the value for `interval_ns` since the default real formatting doesn't have enough precision (probably preexisting).

hilti/toolchain/include/ast/operators/port.h
```diff
@@ -2,15 +2,25 @@
...
 #include <hilti/ast/types/port.h>
 
 namespace hilti::operator_ {
 
 STANDARD_OPERATOR_2(port, Equal, type::Bool(), type::Port(), type::Port(), "Compares two port values.")
 STANDARD_OPERATOR_2(port, Unequal, type::Bool(), type::Port(), type::Port(), "Compares two port values.")
 
+BEGIN_KEYWORD_CTOR(port, Ctor, "port", type::Port(), "Creates a port instance.")
+    std::vector<Operand> parameters() const {
+        return {{"port", hilti::type::UnsignedInteger(16)}, {"protocol", type::Enum(type::Wildcard())}};
```
bbannier, line 20: I was wondering why we weren't more restrictive with the port type arg like e.g.,
```suggestion
         return {{"port", hilti::type::UnsignedInteger(16)}, {"protocol", builder::typeByID("hilti::Protocol")}};```
```

When I tried this the `spicy.types.port.ops` passed, but `hilti.types.port.ops` failed,

```console
$ btest -dv hilti.types.port.ops
hilti.types.port.ops ...
  > ${HILTIC} -j %INPUT >output
... hilti.types.port.ops failed
  % '${HILTIC} -j /Users/bbannier/src/spicy/tests/.tmp/hilti.types.port.ops/ops.hlt >output' failed unexpectedly (exit code 1)
  % cat .stderr
  [error] /Users/bbannier/src/spicy/tests/.tmp/hilti.types.port.ops/ops.hlt:14:14: unknown ID 'p3'
  [error] /Users/bbannier/src/spicy/tests/.tmp/hilti.types.port.ops/ops.hlt:17:14: unknown ID 'p3'
  [error] /Users/bbannier/src/spicy/tests/.tmp/hilti.types.port.ops/ops.hlt:21:8: unknown ID 'p3'
  [error] /Users/bbannier/src/spicy/tests/.tmp/hilti.types.port.ops/ops.hlt:25:8: unknown ID 'p3'
  [error] /Users/bbannier/src/spicy/tests/.tmp/hilti.types.port.ops/ops.hlt:10:13: unsupported operator: port(<uint<64>>, <hilti::Protocol>)
  [error] hiltic: aborting after errors

1 of 1 test failed

**Review by bbannier — APPROVED** [review]

File: hilti/toolchain/src/compiler/visitors/constant-folder.cc (hunk 2 above)
Comment (rsmmr) [reply hilti/toolchain/src/compiler/visitors/constant-folder.cc:41]: Hmm, they might indeed not be correct yet, I'll take another look.

File: hilti/toolchain/src/compiler/visitors/constant-folder.cc (hunk 1 above)
Comment (rsmmr) [reply hilti/toolchain/src/compiler/visitors/constant-folder.cc:25]: Turns out this method isn't used anymore, removing.

File: hilti/toolchain/include/ast/operators/port.h (hunk 4 above)
Comment (rsmmr) [reply hilti/toolchain/include/ast/operators/port.h:20]: Yeah, this is a shortcoming of our resolver: `typeByID` creates an `UnresolvedID`, which we'd never resolve because it's not part of any AST; and then the resolver doesn't know what to do with it for matching against it.

File: hilti/toolchain/src/compiler/visitors/printer.cc (hunk 3 above)
Comment (rsmmr) [reply hilti/toolchain/src/compiler/visitors/printer.cc:802]: That's elsewhere in the printer, but good point,  I'll tweak it to print out the constructors so that the types match the original ones.

**Comment by rsmmr** [comment]
I'm going to iterate a bit more on this. Turns out we do need to do constant folding at more places to really cover the cases we need.

File: hilti/toolchain/src/compiler/visitors/constant-folder.cc (hunk 2 above)
Comment (rsmmr) [reply hilti/toolchain/src/compiler/visitors/constant-folder.cc]: Fixed now (I hope :)).

File: hilti/toolchain/src/compiler/visitors/printer.cc (hunk 3 above)
Comment (rsmmr) [reply hilti/toolchain/src/compiler/visitors/printer.cc:825]: It's now printing the types when it's an integer type with width less then 64, that avoids any ambiguity.

**Comment by rsmmr** [comment]
@bbannier This is now ready for review again. The changes are in commits on top of the previous version, hope that works.

File: hilti/toolchain/src/compiler/codegen/operators.cc [hunk 5]
```diff
@@ -371,7 +404,7 @@ struct Visitor : hilti::visitor::PreOrder<cxx::Expression, Visitor> {
     /// Real
 
     result_t operator()(const operator_::real::CastToInterval& n) {
-        return fmt("::hilti::rt::Interval(%f, hilti::rt::Interval::NanosecondTag())", op0(n));
+        return fmt("::hilti::rt::Interval(%f, hilti::rt::Interval::SecondTag())", op0(n));
```
bbannier, line 407: Good catch! Is this an API breaking change or something we should backport to `release/1.5` (and possibly also `release/1.6` so we can put out a quick bugfix release)? I tend towards breaking change.

hilti/toolchain/src/compiler/visitors/constant-folder.cc
```diff
@@ -0,0 +1,357 @@
...
+
+        return ctor::Bool(! op->value(), p.node.meta());
+    }
+
+    result_t operator()(const expression::ResolvedID& n, position_t p) {
+        // We cannot fold the optimizer's feature constants currently because
+        // that would mess up its state tracking. We continue to let the
+        // optimizer handle expressions involving these.
+        //
+        // TODO: Can we unify this?
```
bbannier, line 132: The issue here is that feature constants are not really constants, but instead flags we toggle in the optimizer if we can detect that they are unused. If we would move this detection before this folder we could treat them like any other constant, but that seems tricky.

tests/Baseline/hilti.optimization.const/noopt.hlt
```diff
@@ -6,43 +6,43 @@ import hilti;
...
-if ( t ) {
+if ( True ) {
     4;
 }
 
 
-if ( f ) {
+if ( False ) {
     5;
 }
```
bbannier, line 35: Just guessing, but you probably didn't move folding for `statement::If` into the new pass since we do it in the optimizer as well to remove dead code (which allows us to simplify parsers). This bothers me slightly more than not being able to handle feature constants there since even if you added it here, we also need that pass to happen in the optimizer. 😕 Not sure there's anything beyond explicitly calling the new folder again from the optimizer to get around that.

hilti/runtime/include/types/interval.h
```diff
@@ -31,7 +31,7 @@ class Interval {
      *
      * @param nsecs interval in nanoseconds.
      */
-    explicit Interval(hilti::rt::integer::safe<int64_t> nsecs, NanosecondTag /*unused*/) : _nsecs(nsecs) {}
+    explicit Interval(const hilti::rt::integer::safe<int64_t>& nsecs, NanosecondTag /*unused*/) : _nsecs(nsecs) {}
```
bbannier, line 34: All the changes in this patch going from passing a safe int by value to passing by ref seem completely unnecessary to me since a `safe<int64_t>` should have exactly the same size as a `int64_t`. My clang-tidy version also seems to be happy with the existing code. What warning did this trigger and what version of clang-tidy were you running?

bbannier (approved):

.gitmodules
```diff
@@ -3,7 +3,7 @@
 	url = https://github.com/onqtam/doctest.git
 [submodule "3rdparty/SafeInt"]
 	path = 3rdparty/SafeInt
-	url = https://github.com/dcleblanc/SafeInt
```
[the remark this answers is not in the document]
rsmmr: This was showing up in CI on Alpine and CentOS Stream 8. I could reproduce it locally on CentOS 8, and that was indeed using gcc 8.5 (after installing packages per our Dockerfile).

hilti/toolchain/src/compiler/codegen/operators.cc
rsmmr: I would say backport because it's fixing a bug.

tests/Baseline/hilti.optimization.const/noopt.hlt
rsmmr: I didn't. I think we'll need to revisit the relationship between the new folding and optimizer at some point, but I left that for later.

hilti/runtime/include/types/interval.h
rsmmr: I saw this on CI originally, see https://cirrus-ci.com/task/4783218440798208. Locally I'm using clang-tidy 15.0.7 currently.

rsmmr:
I'm going to hold off merging this for a little bit more to not block CI, assuming, @bbannier, you're cutting a release.

.gitmodules
```diff
@@ -3,7 +3,7 @@
...
+	url = https://github.com/rsmmr/SafeInt.git
```
bbannier, line 6: @rsmmr, do you plan to upstream your patch? Also, using a private fork is not ideal as it makes pulling in upstream fixes impossible for anyone but you.

PR #1314, opened 2022-12-05
Have &cxxname automatically prefix IDs with `::`

### LINKED ISSUE #1285: Have `&cxxname` automatically prefix IDs with `::`.
This is to avoid scope lookup errors inside namespace that happen to have a different definition for the same ID. Then also remove the `::` from `cxxname` in our library files.

Update TODO in `doc/extending.rst` once done.

bbannier:
Closes #1285.


hilti/runtime/include/logging.h
```diff
@@ -99,5 +89,17 @@ inline void setLocation(const char* l = nullptr) {
         context->source_location = l;
 }
 
+/**
+ * Prints a string, or a runtime value, to a specific debug stream. This is a
+ * macro wrapper around `debug::detail::print(*)` that avoids evaluation of
```
rsmmr, line 94:
```suggestion
 * wrapper around `debug::detail::print(*)` that avoids evaluation of
```

hilti/toolchain/src/compiler/visitors/normalizer.cc
```diff
@@ -142,6 +143,19 @@ struct VisitorNormalizer : public visitor::PreOrder<void, VisitorNormalizer> {
...
             type::Struct::setSelf(&p.node);
     }
+
+    void operator()(const Attribute& n, position_t p) {
+        // Normalize values passed as `&cxxname` so they always are interpreted as FQNs by enforcing leading `::`.
+        if ( const auto& tag = n.tag(); tag == "&cxxname" && n.hasValue() ) {
+            if ( const auto& value = n.valueAsString(); value && ! util::startsWith(*value, "::") ) {
+                auto new_n = Attribute(tag, builder::string(util::fmt("::%s", *value)));
+                HILTI_DEBUG(logging::debug::Normalizer,
+                            util::fmt("[%s] %s -> %s (%s)", *value, n, new_n, n.value().location()));
```
rsmmr, line 153: Suggest to add new `log_change` method here for attributes.

hilti/toolchain/src/ast/types/library.cc
```diff
@@ -0,0 +1,19 @@
...
+
+#include <utility>
+
+#include <hilti/base/util.h>
+
+using namespace hilti;
+using namespace hilti::type;
+
+static std::string normalize(std::string cxx_name) {
+    if ( cxx_name.find(' ') == std::string::npos && ! util::startsWith(cxx_name, "::") )
```
rsmmr, line 13: What's the `find(' ')` for?

rsmmr (dismissed):
Nice, this has been bothering me for a while!

hilti/toolchain/src/ast/types/library.cc
bbannier: This is a dedicated workaround for `__library_type("const hilti::rt::TypeInfo*")`. We can get rid of treatment of spaces if we use `__library_type("hilti::rt::TypeInfo* const")` instead; this seems fine to me as `__library_type` is an internal API anyway.

bbannier: Did just that.

PR #1315, opened 2022-12-06
Improve efficiency of `startsWith` for long inputs.

bbannier:
We previously would always search for _any occurrence_ of `prefix` in the given string and only later check whether it was at the beginning; this could have meant potentially lots of work for long input strings.

With this patch we examine at most the first `prefix.size()` characters of the string.

### DIFF
--- a/hilti/runtime/src/util.cc
+++ b/hilti/runtime/src/util.cc
@@ -377,6 +377,18 @@ std::string hilti::rt::replace(std::string s, std::string_view o, std::string_vi
     return s;
 }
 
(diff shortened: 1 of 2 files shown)


hilti/runtime/src/util.cc
```diff
@@ -1,5 +1,7 @@
 // Copyright (c) 2020-2021 by the Zeek Project. See LICENSE for details.
 
+#include "util.h"
```
rsmmr, line 3:
```suggestion
#include <hilti/rt/util.h>
```

For consistency with other files.

rsmmr (dismissed):

PR #1324, opened 2022-12-13
Bump Cirrus macos CI

bbannier:


doc/installation.rst
```diff
@@ -166,6 +166,14 @@ macOS versions:
       - :package-release-tgz:`TAR <macos_big_sur>`
       - :package-dev-tgz:`TAR <macos_big_sur>`
 
+    * - Monterey (12)
```
rsmmr, line 169: Should we remove the dev entries for Catalina and Big Sur then?

rsmmr (dismissed):

doc/installation.rst
bbannier: Good catch, I'll remove them.

PR #1325, opened 2022-12-13
Fix docs for enum using semicolon instead of comma.

awelzel:
This seems too small to open a PR, but, oh well.

### DIFF
--- a/doc/programming/language/types.rst
+++ b/doc/programming/language/types.rst
@@ -88,7 +88,7 @@ Enum types associate labels with numerical values.
 
 .. rubric:: Type
 
-- ``enum { LABEL_1; ... LABEL_N }``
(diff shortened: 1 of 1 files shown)


bbannier (dismissed):
Thanks! I left a minor suggestion, but can fix that while committing.

doc/programming/language/types.rst
```diff
@@ -88,7 +88,7 @@ Enum types associate labels with numerical values.
 
 .. rubric:: Type
 
-- ``enum { LABEL_1; ... LABEL_N }``
+- ``enum { LABEL_1, ... LABEL_N }``
```
bbannier, line 91: While we are at it, let's add the second missing separator,

```suggestion
- ``enum { LABEL_1, ..., LABEL_N }``
```

bbannier (approved):

PR #1329, opened 2022-12-13
Audit previously disabled clang-tidy-15 checks

bbannier:


rsmmr (approved):
Thanks for going through!

PR #1331, opened 2022-12-14
Push container images to Docker Hub zeek org, too.

awelzel:


awelzel:
I've created the two repos and given zeekbot permissions to push there.

PR #1336, opened 2022-12-19
Reject uses of imported module IDs as expression.

### LINKED ISSUE #1330: Dereferencing a module leads to an internal error
We should catch this invalid code:

```
module Test;

import spicy;

print spicy.version(); # should be spicy::version()
```

```
# spicyc -c test.spicy
[internal-error] test.spicy:6:7: unsupported declaration type hilti::declaration::ImportedModule
...
```

bbannier:
Closes #1330.


hilti/toolchain/src/ast/scope-lookup.cc
```diff
@@ -31,7 +32,7 @@ std::pair<bool, Result<std::pair<NodeRef, ID>>> hilti::scope::detail::lookupID(c
         }
 
         if ( auto d = r.node->template tryAs<Declaration>() ) {
-            if ( auto c = d->tryAs<declaration::Module>() ) {
+            if ( d->isA<declaration::Module>() || d->isA<declaration::ImportedModule>() ) {
```
rsmmr, line 35: Don't think this is the right place for the check: why shouldn't `lookupScope()` be able to look up a module ID? Seems it's the caller's context to decide if a module is permissible or not. How about a validator check for resolved expression IDs instead?

bbannier: This was my gut feeling as well.

I initially tried to catch this up the chain, but ran into issues since callers of this function typically want to go from `UnresolvedID` to `ResolvedID` (both of which implement `isExpression`) and `ImportedModule` (and also `Module` already rejected here) doesn't really fit the what is expected for it, e.g., `ResolvedID` currently needs to have a definite type. As a minimal fix we could work around this by using `type::Null` for `ImportedModule` and `Module`, but the real fix would probably be to have `UnresolvedID` and `ResolvedID` implement a more open API (`isStatement`?) and only go to a type implementing `isExpression` after validation.

This all seemed bigger than I wanted it to be, so I instead extended the existing check which rejects similar bad uses of the current module name as an ID.

rsmmr: Given that the existing check is already at this place, ok to leave of course if not easy to fix.

bbannier: Okay, I am merging this as is. We can revisit this if we ever need to have more fine-grained control.

PR #1340, opened 2023-01-05
Fix references to zeekurity for container images.

awelzel:
The project is called zeek and the pull command ends up with
zeek/spicy, too, so let go of zeekurity.


bbannier:
Thanks for the patch. Since Spicy is now available in the Zeek image, let's point users to that instead. I'll open a different PR.

bbannier:
Closing this one in favor of #1341 which removes the offending section.

PR #1341, opened 2023-01-05
Update repo to not publish Docker images anymore.

bbannier:
Since Spicy is included in the Docker image published for zeek/zeek we do not need to publish separate Spicy Docker images anymore (the analyzers we included should be easy to install for users).

This patch removes the CI pipeline for publishing the images, and updates the documentation to now largely point to the Zeek documentation when it comes to Docker images.


bbannier:
After merging this we should keep the existing images in place, but remove the `latest` tags. That way existing versioned builds keep working, but we notify anyone who still might expect automatic updates.

rsmmr (approved):

PR #1346, opened 2023-01-12
Fix pathological performance for `_haveEod` on highly chunked streams

### LINKED ISSUE #1345: Pathological performance in parsing `bytes` of known size from highly chunked input
A user privately reported extremely bad performance when parsing data from highly chunked data.

#### Reproducer

```ruby
module x;

public type X = unit {
	xs: bytes &size=99999999999999;
};
```

In their actual parser the `bytes` are used as a fallback `default` case in a unit `switch` statement over protocol message types; `size` is previously parsed data.

#### Sample input

```console
$ dd if=/dev/zero of=data bs=1 count=1000000
1000000+0 records in
1000000+0 records out
1000000 bytes transferred in 3.046939 secs (328198 bytes/sec)
```

On my machine parsing the input (until failure) takes roughly 35ms:

```console
$ hyperfine -i 'cat data | spicy-driver x.hlto '
Benchmark 1: cat data| spicy-driver x.hlto
  Time (mean ± σ):      34.4 ms ±   2.4 ms    [User: 29.6 ms, System: 6.1 ms]
  Range (min … max):    30.3 ms …  45.0 ms    75 runs

  Warning: Ignoring non-zero exit code
```

The same driver invocation working on 1 byte chunks takes more than 30s on my machine:
```console
$ cat data| timeout -s KILL 30s spicy-driver x.hlto -i 1
[2]    63331 broken pipe  cat data |
       63332 killed       timeout -s KILL 30s spicy-driver x.hlto -i 1
```

Profiling shows that most of the time is being spent in `Chain::findChunk` which is called when creating unsafe end iterators.

Screenshot 2023-01-12 at 3 02 07 PM

That this is not simply due to e.g., overhead from allocation many chunks can also be seen by changing above grammar to use `&eod` instead of `&size`.

```ruby
module x;

public type X = unit {
	xs: bytes &eod;
};
```

```console
$ hyperfine -i 'cat data| spicy-driver x.hlto -i 1'
Benchmark 1: cat data| spicy-driver x.hlto -i 1
  Time (mean ± σ):      2.130 s ±  0.063 s    [User: 2.063 s, System: 0.045 s]
  Range (min … max):    2.073 s …  2.293 s    10 runs
```

Unfortunately adding `&max-size` alongside `&eod` introduces the pathological performance again.

bbannier:
Closes #1345.

### DIFF
--- a/hilti/runtime/include/logging.h
+++ b/hilti/runtime/include/logging.h
@@ -26,7 +26,17 @@ void fatalError(const std::string& msg) __attribute__((noreturn));
 /** Reports a warning. */
 void warning(const std::string& msg);
 
-#define HILTI_RT_DEBUG(stream, msg) hilti::rt::debug::print(stream, msg);
+/**
+ * Prints a string, or a runtime value, to a specific debug stream. This is a
+ * macro wrapper around `debug::detail::print(*)` that avoids evaluation of
+ * the arguments if nothing is going to get logged.
+ */
+#define HILTI_RT_DEBUG(stream, msg)                                                                                    \
+    {                                                                                                                  \
+        if ( ::hilti::rt::detail::globalState()->debug_logger &&                                                       \
+             ::hilti::rt::detail::globalState()->debug_logger->isEnabled(stream) )                                     \
+            ::hilti::rt::debug::detail::print(stream, msg);                                                            \
+    }
 
 /** Shortcut to `hilti::rt::debug::setLocation`. */
 #define __location__(x) ::hilti::rt::debug::setLocation(x);
--- a/hilti/runtime/include/logging.h
+++ b/hilti/runtime/include/logging.h
@@ -12,9 +12,6 @@
 
 namespace hilti::rt {
 
-/** Reports an internal error and aborts execution. */
-// void internalError(const std::string& msg) __attribute__((noreturn)); // Declared in util.h
-
 /**
  * Reports a fatal error and immediately aborts execution. This skips all
  * cleanup and should be used only for catastrophic library issues; not for
--- a/spicy/runtime/src/parser.cc
+++ b/spicy/runtime/src/parser.cc
@@ -29,7 +29,7 @@ static bool _haveEod(const hilti::rt::ValueReference<hilti::rt::Stream>& data, c
     if ( cur.isOpenEnded() )
         return false;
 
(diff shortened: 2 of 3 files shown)


spicy/runtime/src/parser.cc
```diff
@@ -29,7 +29,7 @@ static bool _haveEod(const hilti::rt::ValueReference<hilti::rt::Stream>& data, c
     if ( cur.isOpenEnded() )
         return false;
 
-    return cur.unsafeEnd().offset() <= data->unsafeEnd().offset();
+    return cur.end().offset() <= data->end().offset();
```
rsmmr, line 32: Yeah, this looks right, given the implementation of  `end()`. I'm wondering if we have more cases like this elsewhere. The unsafe itereators are supposed to be more efficient than the safe versions, so the code usually prefers using/creating them.

hilti/runtime/include/logging.h
```diff
@@ -26,7 +26,17 @@ void fatalError(const std::string& msg) __attribute__((noreturn));
...
 
-#define HILTI_RT_DEBUG(stream, msg) hilti::rt::debug::print(stream, msg);
+/**
+ * Prints a string, or a runtime value, to a specific debug stream. This is a
+ * macro wrapper around `debug::detail::print(*)` that avoids evaluation of
+ * the arguments if nothing is going to get logged.
+ */
+#define HILTI_RT_DEBUG(stream, msg)                                                                                    \
+    {                                                                                                                  \
+        if ( ::hilti::rt::detail::globalState()->debug_logger &&                                                       \
```
rsmmr, line 33: Good catch.

rsmmr (approved):

spicy/runtime/src/parser.cc
bbannier: When making this patch I looked at other users of `unsafeEnd` and this one seemed the most problematic since it is repeatedly called from generated code. There are other questionable uses of `unsafeEnd`, but they didn't seem to be so easy to trigger and also didn't show up in profiling output for test runs. I still filed #1347 to potentially clean up this API since it seems easy to introduce performance problems.

hilti/runtime/include/logging.h
bbannier: This seems like a pretty bad problem so I'll backport it as well.

PR #1351, opened 2023-01-19
Fix incremental regexp matching for potentially empty results.

### LINKED ISSUE #1349: Assertion failure when matching regex with overlapping alternatives
For the following grammar a debug build of Spicy triggers an assertion failure in `regexp.cc`

```ruby
module foo;

public type X = unit {
    x: /([^\xff]|(\xff\xff))*/;
};
```

```console
printf '\xff\xfe"' | spicy-driver -d foo.spicy  -i 1
Assertion failed: (_pimpl->_ms.match_eo >= start_ms_offset), function _advance, file regexp.cc, line 179.
[2]    65515 done       printf '\xff\xfe"' |
       65516 abort      spicy-driver telnet.hlto -i 1
```

The incremental feeding via `-i 1` is critical to reproduce the issue.

For a release build of Spicy we likely produce completely bogus offsets which leads to a very confusing error messages being shown to users:

```console
[error] terminating with uncaught exception of type hilti::rt::Overflow: integer overflow (foo.spicy:4:8)
```

* * *

This issue was reported privately by a user working on a grammar where `\xff\xff` is used as end-of-message marker. For such cases a workaround would be to explicitly match the end marker instead of using regexp alternatives, e.g.,

```ruby
module foo;

public type X = unit {
    ended: /(\xff\xff)?/ &convert=|$$|>0;
    x: /[^\xff]*/ if (!self.ended);
};
```

rsmmr:
When feeding data incrementally into the regexp matcher, we could run into an assert trying to ensure that the end of a match wasn't in the previous chunk already. Turns out, however, that *is* possible with empty matches that are determined only later (and maybe it's possible in other cases, too). To fix this, we need to account for potentially backtracking into data from a previous round. This could be a bit surprising to callers of the `advance()` method, but should work as long as they aren't trimming the stream in between (which in our use cases shouldn't be the case).

Closes #1349.


bbannier:
Let's make sure to backport this to 1.5 and 1.6.

hilti/runtime/src/types/regexp.cc
```diff
@@ -102,15 +102,21 @@ std::tuple<int32_t, stream::View> regexp::MatchState::advance(const stream::View
 
     auto [rc, offset] = _advance(data, data.isFrozen());
 
+    stream::View ndata;
+    if ( offset >= 0 )
+        ndata = data.trim(data.begin() + offset);
+    else
+        ndata = data.trim(data.begin() - static_cast<uint64_t>((-offset)));
```
bbannier, line 109: Nit: this is awkward and needed since `SafeConstIterator` implements both `operator-` and `operator+` for `uint64_t` while all we would have need here is `operator+` for `int64_t` (but probably do not want to add since it makes the interface even more complicated). Probably a good idea to at least add a comment so we don't simplify this once we have swapped out the context, e.g.,

```suggestion
    // `SafeConstIterator` implements both `operator+` and `operator-` for `uint64_t`
    // while `offset` is a `int64_t`. Make sure we trim in the correct direction.
    if ( offset >= 0 )
        ndata = data.trim(data.begin() + offset);
    else
        ndata = data.trim(data.begin() - static_cast<uint64_t>((-offset)));
```

bbannier (approved):
Looks good to me. I'll take over merging this to `main`, `release/1.5` and `release/1.6`.

hilti/runtime/src/types/regexp.cc
rsmmr: Yeah, I was thinking about changing the operators, but didn't want to do that as part of this fix to avoid breaking something else.

bbannier: Fixed during merge.

bbannier:
Merged into `main` and also backported to `release/1.5` and `release/1.6`.

PR #1356, opened 2023-01-24
Add stringification to UnitContext.

bbannier:
Not adding this in the first place broke e.g., `-X flow` or `-X trace`.


rsmmr (approved):
We should add some tests for `-X flow` and `-X trace` as well, but can do that independently too.

PR #1359, opened 2023-01-25
Add DPD-style `spicy::accept_input` and `spicy::decline_input`. 

rsmmr:
This generalizes the confirmation/rejection from being a Zeek thing to
being part of Spicy. Any Spicy grammar can call these functions as to
tell its host application if it recognizes the input format. The host
application can (but doesn't have to) install callbacks to receive
that information and react appropriately. For example, the Zeek
plugin will just forward it to Zeek's standard DPD hooks.

For testing, `spicy-driver` gets a new option `--require-accept`
that lets it exit with success only if `spicy::accept_input()` has
been called, and  `spicy::decline_input()` hasn't.


tests/Baseline/spicy.rt.confirmation/output
```diff
@@ -0,0 +1,13 @@
...
+---
+ok
+---
+fail
+---
+ok
+fail
+---
+fail
+ok
```
bbannier, line 11: I am confused by the semantics of (the absence of) `--require-input` here. Shouldn't the driver stop feeding input after `spicy::reject_input` is called?

spicy/runtime/include/configuration.h
```diff
@@ -0,0 +1,46 @@
...
+     * `spicy::confirm_input()`.
+     **/
+    std::optional<std::function<void()>> hook_confirm_input;
+
+    /**
+     * Optional callback to execute when a Spicy parser calls
+     * `spicy::reject_input()`. This string argument is the reason provided by
+     * the caller.
+     */
+    std::optional<std::function<void(const std::string&)>> hook_reject_input;
```
bbannier, line 26: WDYT about passing e.g., the global stream offset where this was called to the hooks? I am not exactly sure how this would be used by users, but a simple use case would be e.g., showing where the Spicy parser gave up?

The tricky bit would be in coming up with the offset which would not be passed directly by the user.

spicy/lib/spicy.spicy
```diff
@@ -159,3 +159,18 @@ public function parse_address(s: string) : addr &cxxname="hilti::rt::address::pa
...
+## appears to be processing the expected input format. It's up to the host
+## application how to use this information.
+public function confirm_input() : void &cxxname="spicy::rt::confirm_input";
+
+## Reports a violation to the host application indicating that the parser
+## appears not to be processing the expected input format.  It's up to the
+## host application how to use this information.
+##
+## Note that this does not automatically abort processing. If that's desired,
+## you need to trigger a parse error as well.
```
bbannier, line 173:
```suggestion
## Note that this does not automatically abort processing. If that's desired,
## you need to trigger a parse error as well, e.g., by throwing an error,
## `throw "<error message>"`.
```

bbannier (approved):
This looks mostly good for me but for minor issues. The bigger question is whether we'd also want to let the host application know where in the stream input was rejected (possibly also confirmed?). I am not a huge fan of the names, but I don't have a better idea either 🤷 .

spicy/runtime/src/parser.cc
```diff
@@ -17,6 +18,16 @@ HILTI_EXCEPTION_IMPL(Backtrack)
 HILTI_EXCEPTION_IMPL(MissingData);
 HILTI_EXCEPTION_IMPL(ParseError)
 
+void spicy::rt::confirm_input() {
+    if ( const auto& hook = globalState()->configuration->hook_confirm_input )
```
awelzel, line 22: I'm wondering if it would be possible/easy to pass more context like the active `ParsedUnit` and maybe even `Parser` instance?

I once tried to write a host application and am wondering how I could, for example, print info about the unit or parser without needing to reach to static / globals? And seems spicy might have that info and could inject it?

awelzel: (And no, I don't have a direct use case, was just curious)

rsmmr:
> I am not a huge fan of the names, but I don't have a better idea either 🤷 .

Yeah, I know. How about `accept_format` & `decline_format` to avoid overloading confirm/reject?

spicy/runtime/include/configuration.h
rsmmr: I like that idea, but yeah, I don't see a way to access that information from where we call the hook, because it's just a function call and all we have are the user arguments. We could do it if we turned this into full Spicy statements, but not sure I like that. That said, the host application will have some idea where this is coming from, because it knows what data it has passed in most recently.

spicy/runtime/src/parser.cc
rsmmr: Similar answer unfortunately than above to Benjamin's suggestion: we don't have that information available because it's the user calling our new functions; we only have access to whatever is passed as parameters there. 

However, in this case I think there is an even better salutation: the host application can track whatever state it needs, and it will be accessible by the callback implementations through the current `cookie`. That's for example how the Zeek plugin knows what the current connection/file/packet is. This is "even better" I think because we don't need to hardcode some specific choice of what to make available on the Spicy side.

bbannier:
> How about `accept_format` & `decline_format` to avoid overloading confirm/reject?

I do like that since if avoids confusing use of `confirm`/`reject` used elsewhere. We could even use `accept_parse` and `decline_parse`.

rsmmr:
> > How about `accept_format` & `decline_format` to avoid overloading confirm/reject?
> 
> I do like that since if avoids confusing use of `confirm`/`reject` used elsewhere. We could even use `accept_parse` and `decline_parse`.

`_parse` sounds a bit to technical/low-level to me. I'm torn between `accept_format` vs `accept_input` now (and still open to other suggestions as well).

bbannier:
> > > How about `accept_format` & `decline_format` to avoid overloading confirm/reject?
> > 
> > 
> > I do like that since if avoids confusing use of `confirm`/`reject` used elsewhere. We could even use `accept_parse` and `decline_parse`.
> 
> `_parse` sounds a bit to technical/low-level to me. I'm torn between `accept_format` vs `accept_input` now (and still open to other suggestions as well).

Not sure you wanted to start a poll, but 👍 for `accept_input` from me sine we are not formatting anything.

spicy/runtime/src/parser.cc
awelzel: I guess I'll need to look at the spicy-plugin :-)

Probably `std::function` can just hold a callable object, too, so that getting back to the cookie or some other state doesn't necessarily mean using global variables.

spicy/runtime/src/configuration.cc
```diff
@@ -0,0 +1,23 @@
...
+        globalState()->configuration = std::make_unique<spicy::rt::Configuration>();
+
+    return *globalState()->configuration;
+}
+
+void configuration::set(Configuration cfg) {
+    if ( isInitialized() )
+        hilti::rt::fatalError("attempt to change configuration after library has already been initialized");
+
+    *globalState()->configuration = std::move(cfg);
```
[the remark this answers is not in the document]
rsmmr: Changing this into 

```
 globalState()->configuration = std::make_unique<spicy::rt::Configuration>(std::move(cfg));
```

spicy/runtime/src/configuration.cc
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2020-2021 by the Zeek Project. See LICENSE for details.
+
+#include <hilti/rt/logging.h>
+#include <spicy/rt/configuration.h>
```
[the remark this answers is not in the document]
rsmmr: I think last time we talked about it I didn't like it. :-) My point is that it's not very useful to do this for just a handful of files (like, new files). I'd be fine doing a general change across all files, but having just a few inconsistent ones doesn't seem to improve the situation.

tests/Baseline/spicy.rt.confirmation/output
rsmmr: You mean `--require-confirmation` (or now: `—require-accept`)? That's there, but the driver doesn't stop right now, it's only used for deterring success in the end. We could change that, but not immediately sure if that's better.

PR #1361, opened 2023-01-25
Support Zeek-style documentation strings in Spicy source code.

rsmmr:
This adds support for extracting and retaining Zeek-style
documentation strings (`##`, `##!`) in Spicy source code to all
declarations: if they are prefixed with blocks of such comments, that
information is retained along with the declaration for later access.

This commit also prepares for supporting `##<` to document unit and
struct fields, but that isn't implemented yet.

We don't actually use the extracted information further yet, but may
later leverage it for auto-generating Spicy-side documentation (which
we currently do through a Python script ingesting source code
directly.)

Originally, this was implemented to support the Zeek plugin providing
documentation to Zeekygen, but we ended up using a different approach
there (prototyping events Zeek-side). Still, seems worth keeping this
capability.


spicy/toolchain/src/compiler/parser/driver.cc
```diff
@@ -133,3 +133,18 @@ void Driver::processPreprocessorLine(const std::string_view& directive, const st
...
+void Driver::docField(const std::string& s) {
+    // TODO
+}
+
+void Driver::docDeclaration(Declaration* d) {
+    if ( _doc )
+        d->setDocumentation(std::move(_doc));
+
+    _doc.clear();
+}
```
bbannier, line 148: Just out of curiosity, why are we passing a pointer in the parser instead of just a value? That seems to be the only reason we need `clear` and `docClear` in the first place (I initially read this is _clear must be a nop since we moved from the value_ -- but it is actually a pointer 😬).

I'd probably prefer this taking a `DocString`, and the caller modifying its state.

hilti/toolchain/include/ast/declaration.h
```diff
@@ -45,6 +47,54 @@ namespace linkage {
...
+    /** Empties out all content. */
+    void clear();
+
+    /**
+     * Renders the comment back into a multi-line string. This is primarily for debugging.
+     */
+    void render(std::ostream& out) const;
+
+    /** Returns true if any summary or documentation text has been added. */
+    operator bool() const { return ! (_summary.empty() || _text.empty()); }
```
bbannier, line 88: Let's disable conversion of this type to e.g., `int`.

```suggestion
    explicit operator bool() const { return ! (_summary.empty() || _text.empty()); }
```

bbannier (approved):
I like this a lot. Could you create a ticket to rewrite `spicy-doc` to use this infrastructure in the future (might need adjusting our docs)?

spicy/toolchain/src/compiler/parser/driver.cc
rsmmr: Fair. Let me revisit this.

rsmmr:
There are couple things here that still need tweaking, will work on that, but taking back to draft for now.

rsmmr:
Turns out the needed a larger refactor of the Spicy Bison grammar to track the doc strings reliably. I also cleaned it up a bit more overall. 

Downside is that this needs to raise the Bison requirement from 3.0 to 3.1 now, due to https://www.appsloveworld.com/cplus/100/267/bison-c-mid-rule-value-lost-with-variants. I don't know if that's an issue on our platforms, waiting for CI to tell me.

rsmmr:
> Downside is that this needs to raise the Bison requirement from 3.0 to 3.1 now, due to https://www.appsloveworld.com/cplus/100/267/bison-c-mid-rule-value-lost-with-variants. I don't know if that's an issue on our platforms, waiting for CI to tell me.

Of course, my special friend CentOS 8 Stream is objecting.

rsmmr:
This is ready for another look (except that CentOS keeps OOMing, but I don't think that's related to this change; just restarted once more).

rsmmr:
Also, no longer need for Bison 3.1, back to 3.0 now.

spicy/toolchain/src/compiler/parser/parser.yy
```diff
@@ -89,6 +89,20 @@ static uint64_t check_int64_range(uint64_t x, bool positive, const hilti::Meta&
 
 static int _field_width = 0;
 
+// We keep a stack of doc strings here that's maintained during parsing. There
+// woiuld actually be a much nicer way of doing this through Bison's mid-action
```
bbannier, line 93:
```suggestion
// would actually be a much nicer way of doing this through Bison's mid-action
```

hilti/toolchain/include/ast/doc-string.h
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * Appends a line of documentation text to the documentation.
+     *
+     * @param line line to add, with optional comment prefix (which will be removed)
+     */
+    void addText(const std::string& line) { _text.push_back(normalize(line)); }
+
+    /** Empties out all content. */
+    void clear();
+
```
bbannier, line 45: Not needed, unused.
```suggestion
```

hilti/toolchain/include/ast/node.h
```diff
@@ -484,6 +485,26 @@ class NodeBase : public trait::isNode {
...
+    /** Returns the documentation associated with the declaration, if any. */
+    const std::optional<DocString>& documentation() const { return _doc; }
+
+    /** Clears out any documentation associated with the declaration. */
+    void clearDocumentation() { _doc.reset(); }
+
+    /** Sets the documentation associated with the declaration. */
+    void setDocumentation(DocString doc) {
+        if ( doc )
+            _doc = std::move(doc);
```
bbannier, line 501: This is slightly confusing: if one passes a falsy `doc`, nothing will get changed. Maybe instead clear the `_doc` in that case instead.

```suggestion
        if ( doc )
            _doc = std::move(doc);
        else
            _doc.reset();
```

bbannier (dismissed):
I took a more detailed look. This seems like it can do most things we want `spicy-doc` to do. For the future where this could be used by developers to document their parsers it could also be interested to document type fields, e.g., the following are currently not extracted:

```public type X = enum {
	##! A.
	a,
	b, ##< B.
	c ##< C.
};

public type Y = unit {
	##! A.
	a: uint8;
	b: uint8; ##< B.
};
```

Not a blocker, but something we should keep on our radar if this ever gets traction with users.

hilti/toolchain/include/ast/node.h
rsmmr: Ah, yes, I had that at some point, got lost it looks like.

hilti/toolchain/include/ast/node.h
```diff
@@ -484,6 +485,26 @@ class NodeBase : public trait::isNode {
     NodeRef _orig;
 };
 
+
+/** * Common mix-in class for nodes storing doc strings. */
+class NodeWithDocString {
```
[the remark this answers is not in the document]
rsmmr: Sounds good, but I'll move it into the `node` namespace then, otherwise it's not clear what it refers to.

PR #1362, opened 2023-01-25
Streamline runtime exception hierarchy.

rsmmr:
We were a bit inconsistent in the structure of exceptions that could
happen during runtime, making it hard for host applications for
consistently handle trouble. We tweak this, as follows:

- We now only have two general subcategories of exceptions under the
  top-level `hilti::rt::Exception`:

    - `RuntimeError`: non-fatal errors that mean a host
      applications cannot complete the current line of processing,
      but can otherwise continue. This is the main exception that
      host applications should catch, and then proceed
      appropriately.

    - `UsageError`: fatal errors that indicate something about
      using the runtime environment was fundamentally wrong, and a
      host application should abort.

- Almost all other exceptions are now derived from `RuntimeError`,
  including Spicy's `ParseError` and `RecoverableFailure`, so that
  they can treated consistently.

### DIFF
--- a/hilti/runtime/include/exception.h
+++ b/hilti/runtime/include/exception.h
@@ -81,14 +81,14 @@ inline std::ostream& operator<<(std::ostream& stream, const Exception& e) { retu
 
 #define HILTI_EXCEPTION_IMPL(name) name::name::~name() = default;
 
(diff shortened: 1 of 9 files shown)


bbannier (dismissed):
This makes a lot of sense a was a surprisingly small change.

PR #1363, opened 2023-01-25
 Provide ability to initiate runtime's module-pre-init phase manually.

rsmmr:
This can be used if triggering it through a static constructor isn't reliable.

(No test, we'll use this from inside Zeek and will notice there if
something breaks.)

### DIFF
--- a/hilti/runtime/src/init.cc
+++ b/hilti/runtime/src/init.cc
@@ -7,6 +7,8 @@
 #include <cinttypes>
 #include <clocale>
 #include <cstring>
+#include <memory>
+#include <vector>
 
 #include <hilti/rt/configuration.h>
 #include <hilti/rt/context.h>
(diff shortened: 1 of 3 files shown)


hilti/runtime/src/init.cc
```diff
@@ -94,3 +94,22 @@ void hilti::rt::detail::registerModule(HiltiModule module) {
 
     globalState()->hilti_modules.emplace_back(module);
 }
+
+static std::unique_ptr<std::vector<void (*)()>> _registered_preinit_functions;
```
bbannier, line 98: Needs `include <memory>` and `#include <vector>`.

hilti/runtime/src/init.cc
```diff
@@ -94,3 +94,22 @@ void hilti::rt::detail::registerModule(HiltiModule module) {
...
+}
+
+void hilti::rt::executeManualPreInits() {
+    if ( ! _registered_preinit_functions )
+        return;
+
+    for ( const auto& f : *_registered_preinit_functions )
+        (*f)();
+
+    _registered_preinit_functions->clear();
```
bbannier, line 114: Nit
```suggestion
    _registered_preinit_functions.reset();
```

bbannier (approved):

PR #1369, opened 2023-01-31
Documentation updates for recent plugin changes

rsmmr:
- Update auto-generated docs.
- Update Zeek docs for JIT removal and `accept_input`/`decline_input`.
- Documenting the new type export feature.


doc/zeek.rst
```diff
@@ -554,27 +558,91 @@ the pieces going into such an event definition:
...
+will make a ``record`` type of the same name, and with the same
+fields, available in Zeek. If you instead use ``export TFTP::Request
+as TheOtherTFTP::Request``, it will be placed into a different
+namespace instead.
+
+Exporting types generally works for most Spicy types as long as
+there's an ID associated with them in your Spicy code. However,
+exporting is most helpful with user-defined types, such as ``enum``
+and ``unit``, because it can save you quite a bit of typing there. We
+ discuss the more common type conversions in more detail below.
```
bbannier, line 603: Sphinx seems to warn for this leading space.
```suggestion
discuss the more common type conversions in more detail below.
```

bbannier (approved):
LGTM modulo CI.

doc/zeek.rst
```diff
@@ -409,6 +409,8 @@ As a full example, here's what a new GIF analyzer could look like:
         parse with GIF::Image,
         mime-type image/gif;
 
+.. _zeek_events:
```
bbannier, line 412: Just as a heads-up: While adding anchors to the FAQ rubrics I noticed that Sphinx automatically adds anchors derived from the text to headings (but not rubrics), this one would be `#event-definitions`. This is insufficient if we need to refer to sections with identical title, but often good enough.

rsmmr: Yeah, I know about that but I don't like to use it because it's so brittle: changing the header not only breaks internal links that then need fixing, but also invalidates external links to that section existing elsewhere.

Issue #1347, opened 2023-01-12
Consider more efficient reimplementation of `Chain` and `Chunk` 


bbannier:
Currently `Chain` is a thin wrapper around `Chunk` which provides singly linked list functionality. The current implementation makes it harder than needed to e.g., implement an efficient `findChunk` on top of a binary search. The currently implementation likely also causes fragmented use of memory.

We should consider reimplementing the needed functionality on top of e.g., a `std::deque` and revist the implementation of the interface of the classes as we probably could offload lots of manual manipulations to standard library algorithms.

PR #1371, opened 2023-02-07
Fix a few remaining uses of `unsafeEnd`.

bbannier:


bbannier:
Similar to #1345, `unsafeEnd` again showed up as a bottleneck (I had initially missed this function since it is in a template function). There are still other uses of `unsafeEnd` in the code base which I probably also need to look at more carefully.

rsmmr (commented):
I need to go through in more detail but quick note: I'm not sure that it's good to change all of these: the reason that we have `UnsafeIterator` in the 1st place is to avoid unnecessary bounds checking that `SafeIterator` would do. So while switching from safe to unsafe helps with the issue we found the other day, it adds overhead when dereferencing the iterator. It would be ideal if we could get the best of both worlds: a non-checking iterator that doesn't have that original problem for `.unsafeEnd()`

bbannier:
>  It would be ideal if we could get the best of both worlds: a non-checking iterator that doesn't have that original problem for `.unsafeEnd()`

I agree, the current situation is not good since it is very hard to tell which API would be better for a certain situation. From a quick check, it looks like currently it is not straight-forward to make `unsafeEnd` fast since the underlying implementation uses a linked list. I filed #1347 to make use something slower there instead.

bbannier:
Closing this in favor of #1374.

PR #1372, opened 2023-02-08
Make `printParserState` cheaper to call if debug logging is disabled.

bbannier:
Currently `printParserState` is called at least during synchronization. Previously it would always comput its log message, even if that string was never logged. Since computing this log message involved string formatting it came with a noticable performance penalty.

With this patch we only compute the log message if it would actually be logged using our debug log macros.


bbannier:

PR #1373, opened 2023-02-10
[WIP] Introduce `barrier` type

### LINKED ISSUE #1328: Provide a mechanism to synchronize two parse streams
Use case is two sides of a connection where the parsing results from one side determine how the other side should proceed (e.g., `STARTTLS` by the client needs to be acked by the server before the TLS actually kicks in). Currently one needs to manually buffer any further data until the information about how to proceed becomes available, which is both very cumbersome and inefficient. The current thought is to introduce a semaphore-like mechanism that allows one side to wait for the other, yielding in the meantime. Details to be figured out.

rsmmr:
The new `barrier` types provides a synchronization point for multiple parsing streams, such as the two sides of a connection. It comes with an expected number `N` of parties to reach the synchronization point. Each party must call `arrive()` to signal it has done so. Each party may call `wait()` to block until all parties have arrived.

This remains experimental for now, feedback welcome. The Zeek plugin still needs to be adapted for this.

This will close #1328.


rsmmr:
Closing this for now. I still like the idea and think it would be quite a valuable addition to the language, but the implemented in WIP and never fully worked in the current form.

PR #1374, opened 2023-02-10
Optimizing stream performance

rsmmr:
- Revert "Fix pathological performance for `_haveEod` on highly chunked streams."
- Apply alternative fix for #1345.
- Shortcut some offset computations.
- Simplify extract methods.


rsmmr:
@bbannier what do you think of this, in particular the alternative fix in the 2nd commit?

hilti/runtime/src/types/stream.cc
```diff
@@ -59,6 +59,9 @@ const Chunk* Chain::findChunk(const Offset& offset, const Chunk* hint_prev) cons
 
     const Chunk* c = _head.get();
 
+    if ( ! hint_prev )
+        hint_prev = _tail; // see if null means last chunk
```
bbannier, line 63: This is pretty subtle, so let's expand the comment a bit, maybe
```suggestion
        // A very common way this function gets called without `hint_prev` is
    // `Stream::unsafeEnd` via `Chain::unsafeEnd` in construction of an
    // `UnsafeConstIterator` from a `SafeConstIterator`; in this case the chunk
    // for `end()` will be `nullptr`. Optimize for that case by assuming we
    // always want a chunk near the end if no hint is given.
    if ( ! hint_prev )
        hint_prev = _tail;
```

hilti/runtime/src/types/stream.cc
```diff
@@ -76,6 +79,9 @@ Chunk* Chain::findChunk(const Offset& offset, Chunk* hint_prev) {
 
     Chunk* c = _head.get();
 
+    if ( ! hint_prev )
+        hint_prev = _tail; // see if null means last chunk
```
bbannier, line 83: Same.

hilti/runtime/include/types/stream.h
```diff
@@ -1038,6 +1020,12 @@ class View {
      */
     Offset offset() const { return _begin.offset(); }
 
+    /**
+     * Returns the offset of the view's end location within the associated
+     * stream instance. For an open-ended view, returns the `uint64_t` maximum.
+     */
+    Offset endOffset() const { return _end ? _end->offset() : Offset(std::numeric_limits<uint64_t>::max()); }
```
bbannier, line 1027: This is great since it allows removing all instances of `unsafeEnd().offset()`.

We already had `Chain::endOffset` in a `detail` namespace with similar signature, but this adds similar code to the public API. With this signature and its magic value when using this function to compute lengths one always needs to check in parallel whether `isOpenEnded` which seems error-prone.

WDYT about instead returning a `std::optional<Offset>` here instead to handle the opened-ended case? If we'd go that we probably should use the same return type for `Stream::offsetEnd` (which would also take care of the documentation not mentioning the magic value there).

hilti/runtime/include/types/integer.h
```diff
@@ -158,7 +158,7 @@ inline Result<std::tuple<integer::safe<T>, D>> unpack(D b, ByteOrder fmt) {
         return result::Error("insufficient data to unpack integer");
 
     uint8_t raw[sizeof(T)];
-    b = b.extract(raw);
+    b = b.extract(raw, sizeof(raw));
```
bbannier, line 161: Let's be a little bit more verbose to make 110% clear we want the number of elements:

```suggestion
    b = b.extract(raw, sizeof(raw)/sizeof(raw[0]));
```

hilti/runtime/include/types/stream.h
```diff
@@ -1280,16 +1268,23 @@ class View {
...
+
+        if ( n > size() )
+            throw WouldBlock("end of stream view");
+
+        auto p = unsafeBegin();
+        for ( int i = 0; i < n; ++i )
+            dst[i] = *p++;
+
+        return View(SafeConstIterator(p), _end);
     }
```
bbannier, line 1288: While we are at it, let's move this into the `.*cc` file.

hilti/runtime/src/tests/bytes.cc
```diff
@@ -84,13 +84,13 @@ TEST_CASE("decode") {
 TEST_CASE("extract") {
     SUBCASE("sufficient data") {
         unsigned char dst1[3] = {0};
-        CHECK_EQ("123456"_b.extract(dst1), "456"_b);
+        CHECK_EQ("123456"_b.extract(dst1, 3), "456"_b);
```
bbannier, line 87: Let's get rid of the magic values here and below,

```suggestion
        CHECK_EQ("123456"_b.extract(dst1, sizeof(dst1) / sizeof(dst1[0])), "456"_b);
```

hilti/runtime/src/tests/stream.cc
```diff
@@ -960,25 +960,25 @@ TEST_CASE("View") {
 
         SUBCASE("1") {
             Byte dst[1] = {'0'};
-            CHECK_EQ(v.extract(dst), "234567890"_b);
+            CHECK_EQ(v.extract(dst, 1), "234567890"_b);
```
bbannier, line 963: Let's get rid of the magic value here and below,

```suggestion
            CHECK_EQ(v.extract(dst, sizeof(dst) / sizeof(dst[0])), "234567890"_b);
```

hilti/runtime/src/init.cc
```diff
@@ -29,12 +29,14 @@ void hilti::rt::init() {
     if ( ! globalState()->configuration )
         globalState()->configuration = std::make_unique<hilti::rt::Configuration>();
 
-    if ( auto debug_out = globalState()->configuration->debug_out )
-        globalState()->debug_logger = std::make_unique<hilti::rt::detail::DebugLogger>(*debug_out);
-    else
-        globalState()->debug_logger = std::make_unique<hilti::rt::detail::DebugLogger>("/dev/stderr");
+    if ( ! globalState()->configuration->debug_streams.empty() ) {
```
bbannier, line 32: LGTM.

Re: not installing loggers at all in `Release` builds, I agree, having them available there is still useful.

bbannier (approved):
This definitely does also fix the runtime behavior for the use case which triggered the investigation in #1345.

hilti/runtime/include/types/stream.h
rsmmr: I left this in the header to the compiler a chance to inline.

hilti/runtime/include/types/integer.h
rsmmr: I'll stay with just `sizeof` and will use that in the tests, too. I think that's pretty clear.

hilti/runtime/include/types/stream.h
rsmmr: I'm fine going with optional for the new method. However, a stream does always have an end offset, so don't see a reason to change that.

bbannier:

Issue #1310, opened 2022-12-03
Zeek ASAN build with builtin-spicy/stack-use-after-scope: __fiber_switch_trampoline


awelzel:
Hey - running an ASAN built of latest Zeek with builtin spicy, I'm running into stack-user-after-scope splashes..

I've double checked the environment and it should be clean, so I'm a bit at loss what's going on here.

```
$ which zeek; which spicyz; zeek --version; spicyz --version
/opt/zeek-dev/bin/zeek                                                                                                                                                  
/opt/zeek-dev/bin/spicyz                                                                                                                                                
zeek version 5.2.0-dev.398-debug                                                    
1.4.0  
# Within a spicy-http checkout
$ rm -rf build && mkdir build && (cd build && cmake ../ && make -j4)
$ ZEEK_SPICY_MODULE_PATH=$(pwd)/build/spicy-modules zeek -r tests/traces/http-post.pcap 
=================================================================
==3905188==ERROR: AddressSanitizer: stack-use-after-scope on address 0x7f4eb315d3e0 at pc 0x5567a793450b bp 0x7f4eb315d020 sp 0x7f4eb315d018
WRITE of size 8 at 0x7f4eb315d3e0 thread T0
    #0 0x5567a793450a in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_Alloc_hider::_Alloc_hider(char*, std::allocator<char> const&) /usr/include/c++/10/bits/basic_string.h:157
    #1 0x5567a793450a in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string<std::allocator<char> >(char const*, std::allocator<char> const&) /usr/include/c++/10/bits/basic_string.h:526
    #2 0x5567a793450a in __fiber_switch_trampoline ../auxil/spicy/spicy/hilti/runtime/src/fiber.cc:122
    #3 0x5567aa9292f1 in fiber_asm_invoke (/opt/zeek-dev/bin/zeek+0x81482f1)

0x7f4eb315d3e0 is located 1047520 bytes inside of 1048576-byte region [0x7f4eb305d800,0x7f4eb315d800)
allocated by thread T0 here:
    #0 0x7f4eb9739e8f in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x5567aa928f49 in fiber_alloc ../auxil/spicy/spicy/3rdparty/fiber/src/fiber.c:207
[...]

PR #1375, opened 2023-02-11
ASAN fixes

rsmmr:
- Skip clang-specific ASAN flags with other compilers.
- Fix ASAN false positive with GCC.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,9 +74,14 @@ endif ()
 if (USE_SANITIZERS)
     # Recommended flags per https://github.com/google/sanitizers/wiki/AddressSanitizer
     set(sanitizer_cxx_flags
-        "-fsanitize=${USE_SANITIZERS} -fno-omit-frame-pointer -fno-optimize-sibling-calls -O1 -shared-libsan"
-    )
-    set(sanitizer_ld_flags "-fsanitize=${USE_SANITIZERS} -frtlib-add-rpath -shared-libsan")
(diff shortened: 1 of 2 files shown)


rsmmr:
@awelzel see if this fixes #1310 for you

bbannier (approved):
This seems very subtle and a very specific workaround (we e.g., do not remove all allocations on these debug functions, only a few), so I'd really be curious if @awelzel can confirm the fix. Let's hope we do not break this again.

On merge this should be backported to `release/1.5` and `release/1.6`.

awelzel:
> @awelzel see if this fixes https://github.com/zeek/spicy/issues/1310 for you

It does, thanks. The syslog and finger tests work without ASAN complaining.

Digression: I could now also use that Zeek build for analyzer development. However, I just observed that when using an ASAN/Debug Zeek install, compile times for a small analyzer go from ~50seconds (Release, no ASAN build) to 1min 40 seconds. So, probably need to keep a release build for that around. Neither 1min 40sec nor 50sec is pleasant.

awelzel (approved):

rsmmr:
> This seems very subtle and a very specific workaround (we e.g., do not remove all allocations on these debug functions, only a few)

My best guess is that we're hitting undefined behavior here, where the compiler just orders stuff differently. If that's the case, I'm not sure there's really a way to use those ASAN API functions for reporting stack switching robustly.

rsmmr:
And to be clear, I do think this is a false positive, I don't see anything we were doing wrong.

bbannier:

PR #1378, opened 2023-02-20
Build debian-11 instead of centos stream 8 in CI.

bbannier:
Since the Zeek container image is based on debian-11 this allows testing against Spicy versions from branches in zeek/spicy-plugin.

Closes zeek/spicy-plugin#169.


rsmmr:
Actually I think leaving CentOS 8 in CI would be helpful: that's the platform where there's regularly something breaking.

PR #1379, opened 2023-02-20
Switch SafeInt back to upstream.

rsmmr:
This pulls in https://github.com/dcleblanc/SafeInt/pull/50.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,7 +3,7 @@
 	url = https://github.com/onqtam/doctest.git
 [submodule "3rdparty/SafeInt"]
 	path = 3rdparty/SafeInt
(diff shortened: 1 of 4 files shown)


rsmmr:
@bbannier any idea why the lint check is failing?

bbannier:
> any idea why the lint check is failing?

The SafeInt changes in this bump probably make `stream::Offset` (which is just a `integer::safe<uint64_t>`) trivially copyable, so now `UnsafeConstIterator` becomes trivially copyable as well (it only non-pointer member is an `Offset`). I think the linter is right here, it does not make sense to `move` these values anymore.

bbannier (approved):
LGTM, but let's wait for CI.

PR #1382, opened 2023-02-22
Set reupload_on_changes for ccache caches.

awelzel:
Without explicitly setting reupload_on_changes, the caches are never updated due to a constant fingerprint_script.

Found while looking at the CI output and finding 46Mb being a rather small ccache size.

     echo $CIRRUS_TASK_NAME-$CIRRUS_OS
     docker_debian11-linux
     Downloaded 46Mb in 0.484147s.
     Cache hit for ccache-05b8329b...

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -24,6 +24,7 @@ lint_task:
     ccache_cache:
       folder: /tmp/ccache
       fingerprint_script: echo $CIRRUS_TASK_NAME-$CIRRUS_OS
+      reupload_on_changes: true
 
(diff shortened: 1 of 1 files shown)


bbannier (approved):
Thank you for fixing this! It looks like even with this change the upload is still not huge, so the size was a red herring which still lead to a better solution.

I retriggered the freebsd12 task and saw no speedup, but this was probably due to the PR CI not using the cache it previously uploaded (also confirmed by comparing the previous and updated cache sizes).

PR #1383, opened 2023-02-24
Fix potential use-after-move bug.

bbannier:


rsmmr (approved):

bbannier:
Also backported to all supported release branches.

PR #1384, opened 2023-03-01
Implement bytes::DecodeErrorStrategy::ESCAPE

simeonmiteff:
Implement a new `bytes::DecodeErrorStrategy` that behaves the same way as `print` which is also what `bytes.decode()` used to return before #1255.

Very much WIP. Lots missing:
- Handling of UTF8 charset
- Handling places `bytes::DecodeErrorStrategy` applies other than `bytes.decode()` (like the `Bytes` constructor? Maybe other types?)
- Documentation updates
- ~~Check of `std::stringstream` is the best way to do the hex conversion~~
- Check that existing tests pass
- Add new tests
- ~~Fix lint errors~~

### DIFF
--- a/hilti/runtime/src/types/bytes.cc
+++ b/hilti/runtime/src/types/bytes.cc
@@ -104,6 +104,9 @@ std::string Bytes::decode(bytes::Charset cs, bytes::DecodeErrorStrategy errors) 
             return Bytes(str(), cs, errors).str();
 
         case bytes::Charset::ASCII: {
+            if ( errors.value() == DecodeErrorStrategy::ESCAPE )
+                return escapeBytes(str());
+
             std::string s;
             for ( auto c : *this ) {
                 if ( c >= 32 && c < 0x7f )
@@ -113,6 +116,7 @@ std::string Bytes::decode(bytes::Charset cs, bytes::DecodeErrorStrategy errors) 
                         case DecodeErrorStrategy::IGNORE: break;
                         case DecodeErrorStrategy::REPLACE: s += "?"; break;
                         case DecodeErrorStrategy::STRICT: throw RuntimeError("illegal ASCII character in string");
+                        case DecodeErrorStrategy::ESCAPE: cannot_be_reached();
                     }
                 }
             }
@@ -232,6 +236,7 @@ std::string to_string(const bytes::DecodeErrorStrategy& x, tag /*unused*/) {
         case bytes::DecodeErrorStrategy::IGNORE: return "Charset::IGNORE";
         case bytes::DecodeErrorStrategy::REPLACE: return "Charset::REPLACE";
         case bytes::DecodeErrorStrategy::STRICT: return "Charset::STRICT";
+        case bytes::DecodeErrorStrategy::ESCAPE: return "Charset::ESCAPE";
     }
 
     cannot_be_reached();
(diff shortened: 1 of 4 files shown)


hilti/runtime/include/types/bytes.h
```diff
@@ -7,6 +7,8 @@
 #include <string>
 #include <tuple>
 #include <utility>
+#include <iomanip>
+#include <sstream>
```
bbannier, line 11: Not used in header, let's move this to CC file actually requiring these.

I made a suggestion below which might make this unneeded.

hilti/runtime/src/types/bytes.cc
```diff
@@ -113,6 +113,12 @@ std::string Bytes::decode(bytes::Charset cs, bytes::DecodeErrorStrategy errors)
                         case DecodeErrorStrategy::IGNORE: break;
                         case DecodeErrorStrategy::REPLACE: s += "?"; break;
                         case DecodeErrorStrategy::STRICT: throw RuntimeError("illegal ASCII character in string");
+                        case DecodeErrorStrategy::ESCAPE: {
+                            std::stringstream stream;
+                            stream << std::setfill('0') << std::setw(2) << std::hex << (c|0);
+                            s += "\\x";
+                            s += stream.str();
```
bbannier, line 120: This duplicates what we already implement in `hilti::rt::escapeBytes`. Since it builds a new `stringstream` for each byte iterated over it might also become fairly expensive.

Let's instead use that library function before we create `s` above

```cpp
            if ( errors.value() == DecodeErrorStrategy::ESCAPE )
                return escapeBytes(str());
```
and mark this branch here as unreachable
```suggestion
                            cannot_be_reached();
```

simeonmiteff:
@bbannier quick question:

```c++
std::string to_string(const bytes::DecodeErrorStrategy& x, tag /*unused*/) {
    switch ( x.value() ) {
        case bytes::DecodeErrorStrategy::IGNORE: return "Charset::IGNORE";
        case bytes::DecodeErrorStrategy::REPLACE: return "Charset::REPLACE";
        case bytes::DecodeErrorStrategy::STRICT: return "Charset::STRICT";
        case bytes::DecodeErrorStrategy::ESCAPE: return "Charset::ESCAPE";
    }

    cannot_be_reached();
}
```

☝🏻 should the returned strings be `Charset::...` or `DecodeErrorStrategy:...`? Feels like a  cut&paste bug.

hilti/runtime/src/types/bytes.cc
simeonmiteff: 👍🏻

bbannier:
> @bbannier quick question:
> 
> ```c++
> std::string to_string(const bytes::DecodeErrorStrategy& x, tag /*unused*/) {
>     switch ( x.value() ) {
>         case bytes::DecodeErrorStrategy::IGNORE: return "Charset::IGNORE";
>         case bytes::DecodeErrorStrategy::REPLACE: return "Charset::REPLACE";
>         case bytes::DecodeErrorStrategy::STRICT: return "Charset::STRICT";
>         case bytes::DecodeErrorStrategy::ESCAPE: return "Charset::ESCAPE";
>     }
> 
>     cannot_be_reached();
> }
> ```
> 
> ☝🏻 should the returned strings be `Charset::...` or `DecodeErrorStrategy:...`? Feels like a cut&paste bug.

🤦 , a copy & paste bug indeed. These should be e.g., `DecodeErrorStrategy::IGNORE` indeed. If you make that fix in a separate patch it will be much easier to backport it.

simeonmiteff:
Closing as per discussion in !1385

PR #1389, opened 2023-03-06
Do not force locale on users of libhilti.

bbannier:
Setting a locale from a library can interfere negatively with the rest of the application. We previously would set a locale inside `hilti::rt::init`, but we probably do not need it at all.


awelzel:
There's a clocale include further up, maybe that can go, too?

rsmmr (approved):

PR #1392, opened 2023-03-07
Only run lint CI task from scheduled builds or PRs.

bbannier:
With proper caching now implementing, the linting task takes most of the time in a CI run since it executes uncacheable clang-tidy jobs. This PR only executes clang-tidy for PRs and in scheduled builds (currently: just a nightly build). This should make execute CI much faster will.


rsmmr (approved):
Thanks!

PR #1393, opened 2023-03-08
Update types.rst

tbfhg:
This commit adds a change that corrects a suspected typo.


bbannier (approved):
Thanks for the patch, that was a pretty blatant copy-and-paste error 😅.

PR #1395, opened 2023-03-09
Fix normalization issues around enums

bbannier:


hilti/toolchain/src/compiler/codegen/ctors.cc
```diff
@@ -195,7 +195,7 @@ struct Visitor : hilti::visitor::PreOrder<cxx::Expression, Visitor> {
 
     result_t operator()(const ctor::Enum& n) {
         auto id = cg->compile(n.type(), codegen::TypeUsage::Storage);
-        return fmt("%s(%s::%s)", id, id, n.value().id());
+        return fmt("%s(%s::%s)", id, id, cxx::normalize_id(n.value().id()));
```
rsmmr, line 198: Better to create a `cxx::ID` here, that will automatically normalize (that's how `type::Enum` does it too).

rsmmr (dismissed):

PR #1403, opened 2023-03-22
Fiber updates

rsmmr:


hilti/runtime/src/fiber.cc
```diff
@@ -289,14 +289,31 @@ size_t detail::StackBuffer::liveRemainingSize() const {
...
-    auto [lower, upper] = activeRegion();
+    auto want_buffer_size = std::max(activeSize(), configuration::get().fiber_shared_stack_swap_size_min);
 
-    _buffer = ::realloc(_buffer, (upper - lower));
-    if ( ! _buffer )
-        throw RuntimeError("out of memory when saving fiber stack");
+    // Round to KB boundary to avoid frequent reallocations.
+    want_buffer_size = ((want_buffer_size >> 10) + 1) << 10;
+
+    if ( want_buffer_size != _buffer_size ) {
```
bbannier, line 300: This pretty aggressively always uses a buffer size very close the current requirements, and allocates new memory whenever there is a mismatch. I would have thought that this would introduce overhead.

Couple things I was wondering:

- do we need the ability to shrink fibers stacks at all, or could we just grow capacity until we reach some steady state (e.g., similar to how a `vector` wouldn't automatically `shrink_to_fit`)?
- we don't really have histograms of typical buffer sizes, but I wonder whether instead of the KB-sized buckets we could reduced the number of sizes even more  by only allocating sizes of e.g. 1KB, 1MB, 5MB, 10MB and only after that go with the actual requirement. I would expect this to reduce the number of new allocations we need to make.

bbannier (approved):

hilti/runtime/src/fiber.cc
```diff
@@ -621,6 +639,11 @@ void detail::checkStack() {
     if ( fiber->type() == Fiber::Type::Main )
         return;
 
+    if ( fiber->type() == Fiber::Type::IndividualStack || fiber->type() == Fiber::Type::SharedStack ) {
+        if ( auto size = fiber->stackBuffer().activeSize(); size > detail::Fiber::_max_stack_size )
+            detail::Fiber::_max_stack_size = size;
```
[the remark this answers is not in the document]
rsmmr: But it always does an additional assignment even though most of the time it won't need it.

hilti/runtime/src/fiber.cc
rsmmr: This might not be the end of it. I am planning on profiling the buffer sizes further, but this at least improves things. To your questions:

> * do we need the ability to shrink fibers stacks at all, or could we just grow capacity until we reach some steady state 

Yeah, we do because we are caching and reusing fibers. If we never shrank their buffer, they'd eventually all be at their all-time max without releasing any memory anymore. If that was ok (which I think it isn't), we could just allocate them very large from the get-go. This is the reason why I introduced the minimum size, so that at least we don't go below that all time only to then grow again.

> * we don't really have histograms of typical buffer sizes, but I wonder whether instead of the KB-sized buckets we could reduced the number of sizes even more  by only allocating sizes of e.g. 1KB, 1MB, 5MB, 10MB and only after that go with the actual requirement.

Yeah, but not sure exponential growth is the right strategy for stack size. We're trading of the amount of memory allocated here with the number of allocations. Keep in mind that we can have many many of fibers in parallel, so we shouldn't be too wasteful either (that's the reason for having this swap buffer in the first place; there's a 2nd mode in the fiber code that gives each fiber it's own static stack memory; then we don't need to memcpy at all but in return we now need to pre-allocate stacks at their estimated worst-case size.)

So not saying the current way is the best, but I find it really hard to predict what's the best trade-off. Agree that profiling buffer sizes further is a TODO.

PR #1404, opened 2023-03-22
Fix generated parser so filter support can be optimized out

### LINKED ISSUE #1399: `supports_filter` optimization doesn't kick in
```
module Test;

public type Foo = unit {
    bar: Bar;
};

type Bar = unit {
    x: uint32;
    y: uint32;
};
```

No filters in use, but we still generate this code:

```
method method tuple<view<stream>, int<64>, iterator<stream>, optional<hilti::RecoverableFailure>> Test::Bar::__parse_stage1(inout value_ref<stream> __data, copy view<stream> __cur,>
[...]
            if ( filtered = spicy_rt::filter_init(self, __data, __cur) ) {
                local value_ref<stream> filtered_data = filtered;
                (*self).__parse_Test_Bar_2_stage2(filtered_data, (*filtered_data), __trim, __lah, __lahe, __error);
                __cur = __cur.advance(|__cur|);

                if ( __trim )
                    (*__data).trim(begin(__cur));

                __result = (__cur, __lah, __lahe, __error);
            }



        if ( ! filtered )
            __result = (*self).__parse_Test_Bar_2_stage2(__data, __cur, __trim, __lah, __lahe, __error);

    }
[...]
}
```

In fact, if I remember right, originally (i.e., before we switched to automatic optimization), we wouldn't be generating the `stage1` function at all if the unit didn't need to support filtering.

bbannier:
Closes #1399.


rsmmr (approved):
Nice!

PR #1411, opened 2023-04-03
Fix signed-unsigned comparison.

bbannier:


awelzel (approved):
Fixes the following compile error here:
```
../auxil/spicy/spicy/hilti/runtime/src/fiber.cc:316:25: warning: comparison of integer expressions of different signedness: ‘size_t’ {aka ‘long unsigned int’} and ‘long
 int’ [-Wsign-compare]                                                                                                                                                  
  316 |     assert(_buffer_size >= (upper - lower));                                
      |            ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~       
```

PR #1414, opened 2023-04-06
Performance improvements.

rsmmr:


hilti/runtime/src/context.cc
```diff
@@ -13,8 +13,14 @@ using namespace hilti::rt::detail;
 
 namespace hilti::rt::context::detail {
 
-// Not part of global state, it's per thread.
+// Not part of global state, it's per thread. (See comment in `logging.h` about
+// platform differences for TLS.)
+#if defined(__linux__)
+__thread Context* __current = nullptr;
```
bbannier, line 19: If we had some `#define THREAD_LOCAL ...` in some header (also a nice place to put documentation) we could avoid some of the checks added in this patch.

hilti/toolchain/src/compiler/codegen/codegen.cc
```diff
@@ -40,74 +40,149 @@ struct GlobalsVisitor : hilti::visitor::PreOrder<void, GlobalsVisitor> {
...
                     body.addStatement(fmt("__globals()->%s = {%s}", g.id.local(), util::join(g.args, ", ")));
             }
+        }
+        else {
+            for ( const auto& g : globals ) {
+                auto cxx_g = g;
+                cxx_g.type = fmt("std::unique_ptr<%s>", g.type);
+                cxx_g.init =
+                    fmt("std::make_unique<%s>()",
+                        g.type); // initialize to a default value at first to match semantics of dynamic globals
```
bbannier, line 119: Naively I would have expected the naked values here. Could you add a comment why we need dynamic allocation instead? All runtime types should be default-constructible and to catch uninitialized access I'd image that a `std::optional` on the stack might perform better.

bbannier (approved):
LGTM, only left a minor clarification request below.

Re: the additional locations passed as `std::string` I mentioned: I filed #1415 for that.

hilti/runtime/src/context.cc
rsmmr: Ack, I guess that's what `util.h` is for.

hilti/toolchain/src/compiler/codegen/codegen.cc
rsmmr: I had that at first, but we can't control the time of destruction then, which needs to happen in `rt::done()` to be well- defined and match the existing semantics. Will add a comment.

bbannier: I see, if the reason was deterministic cleanup couldn't we with a `std::optional` instead to avoid the heap allocation and access? An optional can be `reset` which might be enough for our cleanup needs.

rsmmr: I just changed it to `optional`, that's a good thought.

PR #1418, opened 2023-04-11
Use safe helper to get runtime configuration.

bbannier:
We previously would directly access a value from the configuration in the global state in functions for `accept_input` and `decline_input`. This is only safe if the configuration was created previously, but would read uninitialized memory if e.g., a parser using these functions is embedded in code not doing that.

With this patch we use a safe accessor for the configuration which ensures that it always contains a good value.


rsmmr (approved):

bbannier:
Also backported to `release/1.7`.

PR #1419, opened 2023-04-12
Add missing license headers.

bbannier:


rsmmr (approved):
Fine of course. We could update the license headers too while we are at it. There's is script for that.

PR #1423, opened 2023-04-14
Omit safety checks on `self` values at runtime where possible

bbannier:


bbannier:
Closing this since we went with #1428 instead.

PR #1424, opened 2023-04-18
Series of smaller performance improvements.

rsmmr:
I'm not 100% sure if all these are are strictly necessary but for most I could measure improved performance (sometimes hard to tease their contribution apart reliably).


hilti/runtime/include/configuration.h
```diff
@@ -64,11 +65,34 @@ struct Configuration {
...
+/** The runtime's configuration singleton. */
+extern std::unique_ptr<hilti::rt::Configuration> __configuration;
+
+/**
+ * Returns the current global configuration without checking if it's already
+ * initialized. This is only safe to use if the runtime is already fully
+ * initialized, and should be left to internal use only where performance
+ * matters.
+ */
+inline const Configuration& unsafeGet() {
```
bbannier, line 79: I think there are a few users of `configuration::get` which could be `configuration::unsafeGet` now since outside of code changing the configuration we should be able to assume that the configuration has been set up by `rt::init`.

hilti/runtime/include/global-state.h
```diff
@@ -97,6 +97,17 @@ inline auto globalState() {
     return createGlobalState();
 }
 
+/**
+ * Returns the current global configuration without checking if it's already
+ * initialized. This is only safe to use if the runtime is already fully
+ * initialized, and should be left to internal use only where performance
+ * matters.
+ */
+inline const GlobalState* unsafeGlobalState() {
```
bbannier, line 106: Similar argument as for `configuration::unsafeGet`, I think most users outside of `init` of `globalState` should be able to use the unsafe variant (e.g., in logging or profiler code).

bbannier (approved):

hilti/runtime/include/configuration.h
rsmmr: Not I'd sure I'd generally switch but I went through and updates places where it could make a performance difference.

rsmmr:
This now needs a plugin update as well (which is coming anyways) because of an include-change.

PR #1428, opened 2023-04-21
Performace optimizations around `self` and `ValueReference`.

bbannier:


hilti/runtime/include/types/reference.h
```diff
@@ -298,44 +299,51 @@ class ValueReference {
...
+    inline const T* _get() const noexcept {
         if ( auto ptr = std::get_if<T*>(&_ptr) )
             return *ptr;
 
-        if ( auto ptr = std::get_if<std::shared_ptr<T>>(&_ptr) )
-            return (*ptr).get();
-
-        cannot_be_reached();
+        assert(std::holds_alternative<std::shared_ptr<T>>(_ptr));
+        return std::get_if<std::shared_ptr<T>>(&_ptr)->get();
```
rsmmr, line 311: Here and elsewhere: The commit message says "replace std::get_if with std::get", but I don't see that change anywhere? Looks like we could do that in most (all?) places.

rsmmr (dismissed):

Issue #1427, opened 2023-04-21
`void` field with `&size` behaves incorrectly


bbannier:
Given the following grammar

```ruby
# file: foo.spicy
module foo;

public type X = unit {
	: void &size=42;
	xs: bytes &eod;
	on %done { print self; }
};
```

- When giving this less than 42 bytes of input I would expect a parse error since the input contains insufficient data. Instead I get

    ```console
    $ printf '1234567890' | spicy-driver -dj foo.spicy
    [$xs=b""]
    ```

- When parsing with incremental input `&size` seems to also be interpreted incorrectly
    ```console
    $ printf '1234567890' | spicy-driver -dj foo.spicy -i 1
    [$xs=b"234567890"]
    ```
    I would have expected the same parse error due to insufficient input here, but instead the `void` field skips only the chunk size is was fed (if the incremental feeding size would have been `2` `xs` would have been `b"34567890"`).

Issue #1051, opened 2021-11-17
Streamline syntax to use `&requires` for precondition checks on unit parameters


bbannier:
We allow `&requires` as a unit attribute, but will only evaluate it as a post-condition. This makes writing precondition checks for unit parameters slightly awkward, e.g., one needs to write syntax like this as a workaround:
```.r
type X = unit(length: uint8) {
     : bytes &size=4711 &requires=length-1; # Take onto arbitrary field in unit.
};
```

It would be nice if it was possible to make this more explicit, e.g., with the following hypothetical syntax:
```.r
type Y = unit(length: uint8) {
     : void &requires=(length>1); # Just specify precondition independent of other parsing.
     : bytes &size=4711;
};
```

Currently `&requires` is not supported for `void` fields.

rsmmr:
Not sure I like the syntax, because it still leaves the `&requires` at a place where semantically it doesn't seem right. Couple other ideas:

```
type X = unit(length: uint8) &requires=length>1 {
     : bytes &size=0; 
     : bytes &size=4711;
};
```

```
type X = unit(length: uint8) {
     %requires = length>1
     : bytes &size=0; 
     : bytes &size=4711;
};
```

That said, allowing `&requires` with `void` fields still makes sense I think, just for consistency.

bbannier:
Another, even more explicit option would be to add an explicit attribute for pre-conditions. That would allow to place such attributes in the place where we already place other attributes (which might also enable new use cases). If we decide to do that, we should also consider renaming `&requires` to e.g., `&post-condition` (could keep the old name as a deprecated alias) to make this symmetric.

rsmmr:
Good thought, I like that for units. I'm unsure if we'd do that for fields as well, or stay with the `&requires` there?

PR #1431, opened 2023-04-24
Add `skip` fields

rsmmr:
- Add new `skip` unit item that efficiently skips over uninteresting data.
- Remove `void` parsing with any attributes.
- Map `void` parsing to skip (but deprecate).
- Add documentation for `skip`.


bbannier (commented):
This looks like it gets the job done for scoping down what could previously be done with `void` fields. At the same time, I wonder whether treating `skip` like a "field type" instead of an attribute will restrict too much where this could be used in the future. What you have currently implemented here could also be written like e.g.,

```ruby
: bytes &size=4711 &skip;
```

i.e., we support skipping of `bytes`-like fields. With `skip` as a type it seems it would be harder to support more complex use cases in the future, e.g., zeek/spicy-http uses regexps fields pretty heavily which currently would be hard to switch to `skip`. If `skip` was an attribute this should be easier.

tests/spicy/types/unit/skip-item.spicy
```diff
@@ -0,0 +1,28 @@
...
+
+type Foo = unit {
+    a: uint8;
+    : skip &size=2 { print "skip b"; }
+    c: uint8;
+    : skip &size=2 if ( True );
+    : skip &size=2 if ( False );
+    f: uint8; # == 0x07
+    : skip &until=b"ABC";
+    eod: skip &eod;
```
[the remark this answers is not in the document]
rsmmr: I like being able to give them names, at it makes for easy documentation of what's being skipped without much downside (e..g, `padding: skip ...`). But I'm sure we can improve error message, I'll check.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -1632,6 +1632,77 @@ struct ProductionVisitor
...
+            Expression until_expr = builder::coerceTo(*until_attr->valueAsExpression(), hilti::type::Bytes());
+            auto until_bytes_var = builder()->addTmp("until_bytes", until_expr);
+            auto until_bytes_size_var = builder()->addTmp("until_bytes_sz", builder::size(until_bytes_var));
+
+            auto body = builder()->addWhile(builder::bool_(true));
+            pushBuilder(body, [&]() {
+                pb->waitForInput(until_bytes_size_var, "end-of-data reached before &until expression found",
+                                 until_expr.meta());
+
+                auto find = builder::memberCall(state().cur, "find", {until_bytes_var});
```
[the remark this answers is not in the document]
rsmmr: As we chatted about already, this should be fine thanks for `find` telling us what we can trim (this is code used elsewhere as well).

rsmmr:
> ```ruby
> : bytes &size=4711 &skip;
> ```

As we had chatted about, I don't think I like the attribute approach because (1) it's more hidden/implicit, vs `skip` upfront making the purpose of the field very clear, and (2) it's less obvious/explainable what field types we support for skipping (we need to implement dedicated skip codegen for each case we want to support, so it won't be complete; and right now it's just 3 cases). 

But I do like the other option we came up with, putting a `skip` modified up front:

```
: skip bytes &size=4711;
```

That generalizes to arbitrary fields, avoids new field syntax just for skipping (other than the modifier), and is still quite explicit. It does still suffer from the "explainability" issue I think,  but at least I think it's easier to document this way (because `skip` can remain its own section, and there we list what we support).

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -1632,6 +1632,77 @@ struct ProductionVisitor
...
+        auto until_attr = AttributeSet::find(p.skip().attributes(), "&until");
+
+        if ( auto c = p.skip().condition() )
+            pushBuilder(builder()->addIf(*c));
+
+        if ( size_attr ) {
+            auto n = builder()->addTmp("skip", *size_attr->valueAsExpression());
+            auto loop = builder()->addWhile(builder::greater(n, builder::integer(0U)));
+            pushBuilder(loop, [&]() {
+                pb->waitForInput(builder::integer(1U), "not enough bytes for skipping", p.location());
```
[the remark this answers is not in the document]
bbannier: The `waitForInput` advances the input for the loop which took me a second to realize. Maybe just add a comment pointing that out (`// Advance input.`).

bbannier:
I also thought about possible other syntax and we probably also could in the future and when needed add a new attribute for `skip` fields to support literals or regexps which could ensure that we skip over an expected pattern, e.g., with a hypothetical `&pattern` attribute

```ruby
: skip &size=4711;  # This patch.
: skip &pattern=b"\x47\x11"; 
: skip &pattern=/A{4}/;
```

rsmmr:
I changed this as discussed, but its not quite ready yet: couple of test failures and docs need updating.

rsmmr:
Next iteration on this one with one major change: It now supports `skip` for all fields. For those where we don't have a dedicated codeine path, we just fall back to the standard parsing. Limiting `skip` just to some fields where we happen to have an implementation for, felt confusing to me from a user perspective.

spicy/toolchain/src/compiler/visitors/resolver.cc
```diff
@@ -268,6 +269,17 @@ struct Visitor : public hilti::visitor::PreOrder<void, Visitor> {
     }
 
     void operator()(const type::unit::item::UnresolvedField& u, position_t p) {
+        if ( u.type() && u.type()->isA<type::Void>() && u.attributes() ) {
+            // Transparently map void field into skipping bytes fields.
+            hilti::logger().deprecated(
+                "using `void` fields with attributes is deprecated and support will be removed in a future "
+                "release; replace 'void' with 'skip bytes ...'",
+                u.meta().location());
+
```
bbannier, line 278: If I understand correctly this is inconsistent with what you did over in `spicy/toolchain/src/compiler/visitors/validator.cc` where you disallow `void` fields with attributes completely. While I would prefer a deprecation where we still allow `void` fields with attributes we also have #1427 so maybe we should just completely pull support for that so the issue wouldn't apply anymore and adjust the code here.

tests/spicy/types/unit/skip-field-fail.spicy
```diff
@@ -0,0 +1,16 @@
...
+
+type X = unit {
+    a: skip bytes &size=1 { print $$; }
+    b: skip bytes &size=1 &requires=($$ == b"x");
+    c: skip bytes &size=1 &convert=$$;
+};
+
+on X::a {
+    print $$;
+}
```
bbannier, line 16: For the following grammar

```ruby
module foo;

public type X = unit {
    b: skip bytes &eod { print self.b; }
};
```

I get the following very generic error message:

```
[error] /tmp/foo.spicy:5:22: type does not have field 'b'
```

This makes sense since `b` is not really there, but still seems confusing to me. If it doesn't require too much work it would be nice if we could somehow teach the resolver to emit a different error message for `skip` fields, e.g., `cannot reference field 'b' since the field is a skip field`.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -1632,6 +1633,85 @@ struct ProductionVisitor
...
+                    pb->advanceInput(consume);
+                    builder()->addAssign(n, builder::difference(n, consume));
+                    builder()->addDebugMsg("spicy-verbose", "- skipped %u bytes (%u left to skip)", {consume, n});
+                });
+            }
+
+            else if ( eod_attr ) {
+                builder()->addDebugMsg("spicy-verbose", "- skipping to eod");
+                pb->waitForEod();
+                pb->advanceInput(builder::size(state().cur));
```
bbannier, line 1667: This comes straight from the existing production for `void &eod`, but as a comment, we are leaving some performance on the table here by keeping all data until EOD in the stream before attempting to trim, e.g., for a field `nope: skip bytes &eod` we generate

```python
# Begin parsing production: Skip: nope -> skip: nope
spicy_rt::waitForEod(__data, __cur, Null);
__cur = __cur.advance(|__cur|);

if ( __trim ) 
    (*__data).trim(begin(__cur));

(*self).__error = __error;
default<void>();
__error = (*self).__error;
# End parsing production: Skip: nope -> skip: nope
```

We couldn't really do much better before, but with `skip` we have a chance. We could e.g., use `waitForInputOrEod` in a loop here and trim on each iteration.

bbannier (dismissed):
This looks and feels good, I like how extensible this turned out.

I only left smaller comments.

spicy/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -661,15 +656,17 @@ struct VisitorPost : public hilti::visitor::PreOrder<void, VisitorPost>, public
         auto repeat = f.repeatCount();
         auto is_sub_item = p.parent().isA<spicy::type::unit::item::Field>();
 
+        if ( f.isSkip() ) {
+            if ( ! f.sinks().empty() )
+                error("skip field cannot have sinks attached", p);
+        }
```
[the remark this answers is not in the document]
rsmmr: Turns out this was a bug: skipped unit fields only worked wo/ a name. Fixing.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
rsmmr: Good catch, will change.

spicy/toolchain/src/compiler/visitors/resolver.cc
rsmmr: This PR continues to support attributes with void for now but it does indeed deprecate them. It's handled by mapping them into `skip` fields before the validator sees them. The existing tests for that feature also continue to pass.

rsmmr:
Ok, one more iteration. Review feedback addressed, and I sneaked it support for #1051. Also updated the `void` documentation.

rsmmr:
Updates are all in new commits.

bbannier (approved):

Issue #1416, opened 2023-04-07
Build fails on GCC 13


timwoj:
From a recent Zeek build on GCC 13:

```
In file included from ../../../../../../../auxil/spicy/spicy/3rdparty/reproc/reproc++/src/reproc.cpp:1:
../../../../../../../auxil/spicy/spicy/3rdparty/reproc/reproc++/include/reproc++/reproc.hpp:95:5: error: declaration of ‘reproc::options::<unnamed struct> reproc::options::env’ changes meaning of ‘env’ [-Wchanges-meaning]
   95 |   } env = {};
      |     ^~~
../../../../../../../auxil/spicy/spicy/3rdparty/reproc/reproc++/include/reproc++/reproc.hpp:91:5: note: used here to mean ‘class reproc::env’
   91 |     env::type behavior;
      |     ^~~
In file included from ../../../../../../../auxil/spicy/spicy/3rdparty/reproc/reproc++/include/reproc++/reproc.hpp:11:
../../../../../../../auxil/spicy/spicy/3rdparty/reproc/reproc++/include/reproc++/env.hpp:8:7: note: declared here
    8 | class env : public detail::array {
      |       ^~~
../../../../../../../auxil/spicy/spicy/3rdparty/reproc/reproc++/include/reproc++/reproc.hpp:107:5: error: declaration of ‘reproc::options::<unnamed struct> reproc::options::redirect’ changes meaning of ‘redirect’ [-Wchanges-meaning]
  107 |   } redirect = {};
      |     ^~~~~~~~
../../../../../../../auxil/spicy/spicy/3rdparty/reproc/reproc++/include/reproc++/reproc.hpp:102:5: note: used here to mean ‘struct reproc::redirect’
  102 |     redirect err;
      |     ^~~~~~~~
../../../../../../../auxil/spicy/spicy/3rdparty/reproc/reproc++/include/reproc++/reproc.hpp:68:8: note: declared here
   68 | struct redirect {
      |        ^~~~~~~~
make[2]: *** [auxil/spicy/spicy/3rdparty/reproc/reproc++/CMakeFiles/reproc++.dir/build.make:76: auxil/spicy/spicy/3rdparty/reproc/reproc++/CMakeFiles/reproc++.dir/src/reproc.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:5000: auxil/spicy/spicy/3rdparty/reproc/reproc++/CMakeFiles/reproc++.dir/all] Error 2
[...]

PR #1432, opened 2023-04-26
Bump reproc dependency to fix building with gcc-13.

bbannier:
The recently released gcc-13 is supported by Zeek LTS releases and users reported issues from building Spicy with that compiler, see #1416.

This patch bump the reproc dependency to a recent development snapshot which addresses that (currently their latest `main` still cannot build without hard errors). Since this could pull in breaking changes in `3rdparty/reproc` this is technically a breaking change, but after discussion the consensus among the merge masters was that this was acceptable, e.g., since this we hope and expect this to not be directly visible by users.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -9,7 +9,7 @@ environment:
     IMAGE_MACOS_VENTURA: ghcr.io/cirruslabs/macos-ventura-base:latest
 
     # Branches to use for spicy-plugin and spicy-analyzers tests.
-    ZEEK_SPICY_BRANCH:      main
+    ZEEK_SPICY_BRANCH:      release/1.3
     ZEEK_ANALYZERS_BRANCH:  main
 
     # Cache HILTI C++ compilation.


awelzel (commented):
After looking a bit, we have the following logic on the Zeek side to avoid the error, wouldn't it be simpler to backport that to Zeek 5.0 if we're not 100% comfortable bumping reproc?

https://github.com/zeek/zeek/blob/dd269904102e214835e25d008b27eb62aea7ea01/auxil/spicy/CMakeLists.txt#L32-L36

awelzel (approved):
Approving after chatting with @bbannier - it probably was clear to everyone but me that the Zeek side `-Wno-changes-meaning` fix isn't reasonable for LTS.

bbannier:
Closing this in favor #1434. gcc-13 support actually needs more than this patch (include & test fixes). It also raises now interesting warnings about actual uses of dangling references which we should fix.

PR #1437, opened 2023-04-28
Fix stdlib exception leak in `bytes::sub`

### LINKED ISSUE #1436: `bytes::sub` can throw non-HILTI exception
The following directly throws a `std::out_of_range` while it should throw a HILTI exception:

```ruby
# file: foo.spicy
module foo;

b"".sub(10, 11);
```

```console
$ spicyc -dj foo.spicy
[error] uncaught exception of type std::out_of_range: basic_string
```

bbannier:
Closes #1436.

### DIFF
--- a/hilti/runtime/include/types/bytes.h
+++ b/hilti/runtime/include/types/bytes.h
@@ -4,6 +4,7 @@
 
 #include <cstring>
 #include <memory>
+#include <stdexcept>
 #include <string>
 #include <tuple>
 #include <utility>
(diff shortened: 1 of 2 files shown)


hilti/runtime/include/types/bytes.h
```diff
@@ -294,7 +295,13 @@ class Bytes : protected std::string {
      * @param offset of one byeond end of subrage
      * @return a `Bytes` instance for the subrange
      */
-    Bytes sub(Offset from, Offset to) const { return {substr(from, to - from)}; }
+    Bytes sub(Offset from, Offset to) const {
+        try {
+            return {substr(from, to - from)};
+        } catch ( const std::out_of_range& ) {
+            throw InvalidArgument(fmt("start index %s out of range for bytes with length %d", from, size()));
```
rsmmr, line 302: We have our own `OutOfRange`, I'd throw that.

rsmmr (approved):
I don't quite see why we the 1st commit is needed to fix this? (But fine either way.)

bbannier:
> I don't quite see why we the 1st commit is needed to fix this? (But fine either way.)

The idea was to minimize places where we would have called functions of `std::string` and would need a catch block. With that patch we now only call `substr` exactly once and handle the exception there.

bbannier:

PR #1438, opened 2023-04-28
Profile exception activity.

rsmmr:
This is to measure how often they are instantiated. We ignore
execution time.


rsmmr:
This is for catching when large numbers of exceptions get thrown. Overhead should be negligible if profiling isn't enabled.

bbannier (approved):

Issue #1421, opened 2023-04-14
Code generation broken for certain uses of random access types


bbannier:
While debugging a separate issue in spicy-dns I came across a code generation issue for code involving use of another random access type.

```ruby
# file: foo.spicy
module foo;

public type X = unit {
    # Could have the following here, but is not required for reproducer.
    # : A(self);
};

type A = unit(x: X) {
    : bytes &eod &parse-at=x.position();
};
```

Under the covers this produces invalid HILTI code which then fails to type check:

```console
$ spicyc -j foo.spicy
[error] foo.spicy:8:10-10:2: unsupported operator: <view<stream>>.advance(<optional<iterator<stream>>>)
[error] foo.spicy:8:10-10:2: unknown ID '__parse_cur'
[error] spicyc: aborting after errors
```

This also reproduces without optimizations `-g` or in debug mode `-d`.

I checked whether this was broken recently, but code like this doesn't work with versions back to at least spicy-1.1.0.

bbannier:
Fixed by #1439.

PR #1439, opened 2023-05-04
Allow to use `offset()` without enabling full random-access support.

### LINKED ISSUE #1089: Explore pulling functionality for tracking offsets into its own unit feature
With #1074 we recently made `%random-access` an implicit unit feature so that its code is only active if any related functionality is used. We then e.g., emit additional code to track offsets, or change trimming behavior so that users can set the input position to an earlier position.

Changing the trimming behavior can have noticeable effects on unit behavior as we might now need to hold potentially a lot of data; computing offsets on the other hand is pretty cheap. We should explore whether we can pull the computations for offsets and accessing it via `offset` into its own unit feature, or even make it available by default. If fact `spicy-dump` already includes an option to dump offsets with `-Q` so the code might already be there, or could be unified with whatever we do for a potential offset feature.

bbannier:
Closes #1089.


bbannier:
The patch inlining `__offset` will also fix #1421.

rsmmr (approved):

PR #1455, opened 2023-06-07
Replace obsolete uses of `egrep` with `grep -E`.

bbannier:


rsmmr (approved):
LGTM, but why/since when is `egrep` obsolete?

bbannier:
> LGTM, but why/since when is `egrep` obsolete?

It looks like `egrep` was provided at a time when POSIX `grep` did not support extended regular expressions, but that does not seem to be an issue on modern platforms anymore. I noticed this from shellcheck, https://www.shellcheck.net/wiki/SC2196.

Additionally, uses of `egrep` now trigger warnings on some newer platforms like e.g., fedora-38.

PR #1458, opened 2023-06-12
Include condition in `&requires` failure message.

### LINKED ISSUE #1343: Include expression text in "&requires failed" message
While writing up zeek/spicy-plugin#160.

Nice-to-have: In the `&requires` violation messages, having the textual version of the expression may provide more information without resorting to opening the involved spicy analyzer (which may not be at hand depending on where one looks at those messages).

So `&requires failed` might be nicer as `&requires=(self.length >= 9) failed`.

```
violation, Analyzer::ANALYZER_SPICY_MINI, C21rnr2KzAmlQ38u52, &requires failed (/home/awelzel/corelight-oss/projects/spicy-mini/analyzer/mini.spicy:4:24)
```

bbannier:
With this patch we now include the `&requires` condition in the failure message in case the condition is not met. Since we do not have direct access to the Spicy source code anymore we render HILTI code for now, but undo the most common desugaring by replacing `__dd` with `$$`.

Closes #1343.


rsmmr (dismissed):
LGTM. Couple thoughts: (1) we could probably preserve the Spicy AST node and print that, but not sure it's worth the trouble. (2) As a quick hack we could do a string replacement `__dd` -> `$$` on the message and would probably have the most common difference covered. (Both just optional ideas, fine as os).

bbannier:
I went with the simple replacement of `__dd` -> `$$` (captured in a helper).

rsmmr (approved):

PR #1459, opened 2023-06-12
Add bitwise operators `&`, `|`, and `^` for booleans.

### LINKED ISSUE #1435: Support the XOR operator for bool
Xor (`^`) is not available for the `bool` type, while being available for other types, like uint.

https://docs.zeek.org/projects/spicy/en/latest/programming/language/types.html?highlight=xor#operator-integer::BitXor

It would be neat to have it for bools :)

bbannier:
Closes #1435.

### DIFF
--- a/doc/autogen/types/bool.rst
+++ b/doc/autogen/types/bool.rst
@@ -1,5 +1,17 @@
 .. rubric:: Operators
 
+.. spicy:operator:: bool_::BitAnd bool t:bool <sp> op:& <sp> t:bool
+
+    Computes the bit-wise 'and' of the two boolean values.
+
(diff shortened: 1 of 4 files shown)


rsmmr (commented):
I get that we lacked *xor* but do we really need`|` and `&` (given `||`/`&&`)? Having multiple operators doing the same thing seems unnecessary.

0xxon:
I admittedly did not think about the fact that `&` and `&&` are equivalent for bools when I mentioned this in the ticket. It might be nice to have them for consistency - they do exist for other types - like integers.

(That being said, I actually don't have strong feelings, one way or another one)

PR #1469, opened 2023-06-23
Initialize generated struct members in constructor body.

### LINKED ISSUE #1446: Internal compiler error when combining unit vars and `inout` parameters
@0xxon privately reported an internal compiler error triggered from code similar to the following

```ruby
# file: foo.spicy

module foo;

type A = unit {};

function fun(x: X, inout a: A): bool {
    return True;
}

type X = unit (a: A){
    var z: bool = fun(self, a);
};
```

```console
$ spicyc -j foo.spicy
[internal-error] codegen: cannot add tmp without an active block

--- Aborting
# 0   libhilti.dylib                      0x0000000106e8ff86 _ZN5hilti2rt9BacktraceC1Ev + 166
# 1   libhilti.dylib                      0x000000010698023b _ZN5hilti4util20abort_with_backtraceEv + 139
# 2   libhilti.dylib                      0x000000010697a567 _ZN5hilti6Logger13internalErrorERKNSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEERKNS_8LocationE + 55
# 3   libhilti.dylib                      0x0000000106989e4f _ZN5hilti6detail7CodeGen6addTmpERKNSt3__112basic_stringIcNS2_11char_traitsIcEENS2_9allocatorIcEEEERKNS0_3cxx7ElementILNSB_7element4TypeE1EEE + 719
# 4   libhilti.dylib                      0x0000000106a0fd5a _ZN5hilti6detail7CodeGen8_makeLhsENS0_3cxx10ExpressionERKNS_4TypeE + 90
# 5   libhilti.dylib                      0x00000001069bfbd1 _ZN5hilti6detail7CodeGen7compileERKNS_10expression6detail10ExpressionEb + 6385
# 6   libhilti.dylib                      0x00000001069869cd _ZN5hilti6detail7CodeGen20compileCallArgumentsERKNS_4node5RangeINS_10expression6detail10ExpressionEEERKNS3_INS_11declaration9ParameterEEE + 653
# 7   libhilti.dylib                      0x00000001069d2c8b _ZN5hilti6detail7visitor15do_dispatch_oneINS0_3cxx10ExpressionENS_9operator_8function4CallENS_10expression17resolved_operator6detail16ResolvedOperatorEN12_GLOBAL__N_17VisitorENS1_8IteratorINS_4NodeELNS1_5OrderE0ELb0EEEEENSt3__111conditionalIXsr3stdE9is_void_vIT_EEbNSI_8optionalISK_EEE4typeERT1_RKSt9type_infoRT2_RNT3_8PositionERb + 1099
# 8   libhilti.dylib                      0x00000001069ca68c _ZN5hilti6detail7visitor11do_dispatchINS0_3cxx10ExpressionENS_10expression17resolved_operator6detail16ResolvedOperatorEN12_GLOBAL__N_17VisitorENS1_8IteratorINS_4NodeELNS1_5OrderE0ELb0EEEEENSt3__111conditionalIXsr3stdE9is_void_vIT_EEbNSF_8optionalISH_EEE4typeERT0_RT1_RNT2_8PositionERb + 1964
# 9   libhilti.dylib                      0x00000001069c9910 _ZN5hilti6detail7CodeGen7compileERKNS_10expression17resolved_operator6detail16ResolvedOperatorEb + 976
# 10  libhilti.dylib                      0x00000001069c26e2 _ZN5hilti6detail7visitor15do_dispatch_oneINS0_3cxx10ExpressionENS_10expression17resolved_operator6detail16ResolvedOperatorENS5_6detail10ExpressionEN12_GLOBAL__N_17VisitorENS1_8IteratorINS_4NodeELNS1_5OrderE0ELb0EEEEENSt3__111conditionalIXsr3stdE9is_void_vIT_EEbNSH_8optionalISJ_EEE4typeERT1_RKSt9type_infoRT2_RNT3_8PositionERb + 194
# 11  libhilti.dylib                      0x00000001069bed65 _ZN5hilti6detail7CodeGen7compileERKNS_10expression6detail10ExpressionEb + 2693
# 12  libhilti.dylib                      0x0000000106a24ed6 _ZZN12_GLOBAL__N_118VisitorDeclarationclERKN5hilti4type6StructENS1_7visitor8PositionIRNS1_4NodeEEEENKUlRT_E_clINS1_6detail3cxx11declaration4TypeEEEDaSC_ + 5238
# 13  libhilti.dylib                      0x0000000106a23a22 _ZNSt3__110__function6__funcIZN12_GLOBAL__N_118VisitorDeclarationclERKN5hilti4type6StructENS4_7visitor8PositionIRNS4_4NodeEEEEUlRT_E_NS_9allocatorISG_EEFNS4_6detail3cxx11declaration4TypeERSM_EEclESN_ + 18
# 14  libhilti.dylib                      0x0000000106a236a3 _ZN5hilti4util5CacheINS_6detail3cxx2IDENS3_11declaration4TypeEE11getOrCreateERKS4_RKNSt3__18functionIFS6_vEEERKNSB_IFS6_RS6_EEE + 435
# 15  libhilti.dylib                      0x0000000106a0e8d2 _ZN5hilti6detail7visitor7VisitorINS0_3cxx11declaration4TypeEN12_GLOBAL__N_118VisitorDeclarationENS_4NodeELNS1_5OrderE0EE8dispatchERKS8_ + 11778
# 16  libhilti.dylib                      0x0000000106a0f40e _ZN5hilti6detail7CodeGen15typeDeclarationERKNS_4TypeE + 174
# 17  libhilti.dylib                      0x000000010699eb8f _ZN5hilti6detail7visitor7VisitorIvN12_GLOBAL__N_114GlobalsVisitorENS_4NodeELNS1_5OrderE0EE8dispatchERKS5_ + 4287
# 18  libhilti.dylib                      0x000000010698768b _ZN12_GLOBAL__N_114GlobalsVisitor15addDeclarationsEPN5hilti6detail7CodeGenERKNS1_4NodeERKNS1_2IDEPNS2_3cxx4UnitEb + 171
# 19  libhilti.dylib                      0x0000000106986e9c _ZN5hilti6detail7CodeGen13compileModuleERNS_4NodeEPNS_4UnitEb + 1020
# 20  libhilti.dylib                      0x0000000106b215e9 _ZN5hilti4Unit7codegenEv + 953
# 21  libhilti.dylib                      0x0000000106ad2a1f _ZN5hilti6Driver13_codegenUnitsEv + 1295
# 22  libhilti.dylib                      0x0000000106ad3aca _ZN5hilti6Driver12compileUnitsEv + 2186
# 23  libhilti.dylib                      0x0000000106ad434f _ZN5hilti6Driver7compileEv + 31
# 24  libhilti.dylib                      0x0000000106ad3ec8 _ZN5hilti6Driver3runEv + 152
# 25  spicyc                              0x0000000104619167 main + 311
# 26  dyld                                0x00007ff80c46341f start + 1903
```

This appears to be triggered by a path which tries to construct a LHS in a non-block context so codegen fails.

bbannier:
We previously would generate them with default member initializers. This lead to issues since for some constructs we need to create temporaries (e.g., mutable accessors to `self`). With this patch we now always emit a constructor body which ensures we have a block to attach temporaries to.

Closes #1446.


hilti/toolchain/src/compiler/codegen/types.cc
```diff
@@ -208,11 +216,16 @@ struct VisitorDeclaration : hilti::visitor::PreOrder<cxx::declaration::Type, Vis
...
 
-                    auto x = cxx::declaration::Local{cxx::ID(f.id()),
-                                                     t,
-                                                     {},
-                                                     default_,
-                                                     (f.isStatic() ? "inline static" : "")};
+                    cg->popCxxBlock();
+
+                    if ( default_ )
+                        ctor.addStatement(fmt("%s = %s", cxx::ID(f.id()), *default_));
```
rsmmr, line 222: `%s = std::move(%s)`?

rsmmr (commented):
Yeah, this looks like the right solution. I don't see a regression test though?

bbannier:
> Yeah, this looks like the right solution. I don't see a regression test though?

I added a regression test based on the original issue. It is however pretty hard to make sure that codegen hits a path needing a block, but at least it does now.

hilti/toolchain/src/compiler/codegen/types.cc
bbannier: I would prefer to not move the value here since even with a moved value we generate the same code (`fmt` should bind a const ref), but could introduce e.g., use-after-move issues should we ever do more with `default_` in this scope.

rsmmr (approved):

Issue #1231, opened 2022-07-04
Cannot recover from gap while searching for lookahead


bbannier:
At least in some situations when a parser encounters a gap while searching for a lookahead token one cannot recovery with `&synchronize`/`confirm`, e.g.,

```ruby
# sync.spicy
module foo;

type X = unit {
	a: /A/ &synchronize;

	on %synced { print "sync"; confirm; }
};

public type Xs = unit {
	xs: (X &synchronize)[] { confirm; }
};
```

```cpp
// foo.cc
#include "sync.cc"
#include "sync_linker.cc"

int main() {
    hilti::rt::init();
    spicy::rt::init();

    const spicy::rt::Parser* parser = nullptr;
    for ( auto* p : spicy::rt::parsers() ) {
        parser = p;
        break;
    }

    assert(parser);

    auto xs = hilti::rt::reference::make_value<__hlt::foo::Xs>();

    hilti::rt::ValueReference<hilti::rt::Stream> stream;
    stream->append("A");
    stream->append(nullptr, 1);
    stream->append("A");

     __hlt::foo::Xs::parse2(xs, stream, {}, {});
    std::cerr << xs << '\n';
}
```

```console
$ spicyc -dco sync.cc sync.spicy
$ spicyc -dlo sync_linker.cc sync.spicy
$ $(spicy-config --cxx-launcher --cxx --cxxflags --debug --ldflags) -o foo foo.cc
$ ./foo
llibc++abi.dylib: terminating with uncaught exception of type hilti::rt::MissingData: data is missing (sync.spicy:10:2)
```

Looking at the backtrace, this error originates in stage 1 parsing which has no recovery code,

```
frame #5: 0x00000001000b1895 foo`__hlt::foo::Xs::__parse_xs_stage1(this=0x0000000100706d88, __data=0x00007ffeefbfed28, __cur=View @ 0x00007ffeefbfdad8, __trim=(_value = false), __lah=(m_int = 0), __lahe=SafeConstIterator @ 0x00007ffeefbfdab0, __error= Has Value=false , __dst=0x0000000100706d98) at sync.cc:495:49
   492                      for ( ::hilti::rt::regexp::MatchState ms = __hlt::foo::__re_xs_l1.tokenMatcher(); ::hilti::rt::Bool(true);  ) {
   493                          ::hilti::rt::integer::safe<int32_t> rc;
   494                            __location__("sync.spicy:10:2");
[...]

PR #1470, opened 2023-06-23
Add special handling for potential `advance` failure in trial mode.

### LINKED ISSUE #1464: Synchronizing a vector with a constant value for lookahead
I was pulling my hair out when trying to get something like the following to work:
```
type X = unit {
    a: b"begin-of-X";
    b: bytes &size=10;
};

type Foo = unit {}
    xs: (X &synchronize)[];

    on %synced { confirm; }
};
```

I just couldn't get the parser to synchronize. Then, @Mohan-Dhawan suggested to use a regex for `a`. To my surprise that worked! However, given that the above is taken from the docs, requiring a regex feels like a bug.

bbannier:
In 7c18597f1b95747b0305918eb1777ed22c06b166 for #1231 we added code to handle advance failures in trial mode when synchronizing on regexps, but missed adding the same support for literals. This patch adds that missing part.

Closes #1464.


spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -1021,7 +1022,32 @@ struct ProductionVisitor
...
                 pstate.literal_mode = mode;
                 pushState(std::move(pstate));
+
+                // Since `advance` can trigger recoverable errors when
+                // hitting a gap, bracket the call to it in a `try`/`catch`
+                // block if we are in search mode and attempt to recover.
+                if ( mode == LiteralMode::Search ) {
+                    auto [body, try_] = builder()->addTry();
+
+                    pushBuilder(try_.addCatch(builder::parameter(ID("e"), builder::typeByID("hilti::MissingData"))),
```
rsmmr, line 1032: This is being popped again later? (Can't quite tell)

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -1021,7 +1022,32 @@ struct ProductionVisitor
...
+                                    // // FIXME(bbannier):
+                                    // // We operate on `ncur` while `advanceToNextData`
+                                    // // updates `cur`; copy its result over.
+                                    // builder()->addAssign(ID("ncur"), state().cur);
+
+                                    // Continue incremental matching.
+                                    builder()->addContinue();
+                                });
+
+                    pushBuilder(body);
```
rsmmr, line 1046: Since we have this block of code twice now, I'd factor it out into a function, that will make this all more readable.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
bbannier: This doesn't need to be popped since the overload here is the form which directly takes a callback and handles push/pop for the user.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
bbannier: The code for regexps needs a slight twist (accidentally left in here commented out and marked with `FIXME`), but we can capture that difference with an additional parameter. I added a helper in a new commit.

rsmmr (approved):

bbannier:

PR #1473, opened 2023-06-27
Reject uses of `self` in unit `&size` and `&max-size` attribute

### LINKED ISSUE #1466: Use of unset unit fields not rejected in unit `&size` attribute
The following unit:

```
module Test;

public type test = unit {
	tag: uint8;
	length: uint16;
	data: bytes &size=self.length;
} &size=self.length+3;
```

Terminates with the following error:

```
$ printf '\x00\x00\x03abc' | spicy-driver test.spicy
[error] terminating with uncaught exception of type hilti::rt::UnsetOptional: unset optional value (test.spicy:3:20-6:2)
```

Using a size-check with size being sourced from a field inside a unit is a really useful feature, especially for TLV protocols, where the value can contain other types that can try to exceed the boundaries of the outer type. I encountered this during the reimplementation of the TLS analyzer; the old binpac analyzer uses this kind of construct a lot. The new parser currently doesn’t do as much checking - and I stumbled across this while trying to add it.

bbannier:
Closes #1466.


spicy/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -579,6 +580,22 @@ struct VisitorPost : public hilti::visitor::PreOrder<void, VisitorPost>, public
                 if ( a.tag() == "&size" || a.tag() == "&max-size" ) {
                     if ( ! a.hasValue() )
                         error(fmt("%s must provide an expression", a.tag()), p);
+                    else {
+                        auto v = hilti::visitor::PreOrder<>();
+                        for ( auto i : v.walk(a.value()) ) {
+                            auto&& member = i.node.tryAs<operator_::unit::MemberNonConst>();
+                            if ( ! member || ! member->hasOp0() )
+                                continue;
```
rsmmr, line 588: Is this needed? Wouldn't just any usage of `self` enough for rejection? (i.e., just check if `i` is  the `ResolvedID`.)

spicy/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -579,6 +580,22 @@ struct VisitorPost : public hilti::visitor::PreOrder<void, VisitorPost>, public
...
+
+                            auto&& self = member->op0().tryAs<hilti::expression::ResolvedID>();
+                            if ( ! self || self->id().str() != "self" )
+                                continue;
+
+                            error(fmt("%s expression cannot depend on information available only after parsing of unit",
```
rsmmr, line 594: I'd make this more explicit: "... cannot use `self` because that depends on ...".

spicy/toolchain/src/compiler/visitors/validator.cc
bbannier: You are right, this e.g., fails to capture cases where `self` is passed to a function which accesses fields. I'll simplify this to reject all uses of `self`, regardless of context.

rsmmr (approved):

Issue #1475, opened 2023-06-30
Incorrect internal offset computation when using lists and `&parse-from`


bbannier:
The following code triggers an integer overflow in generated code updating the internal `__offset` field:

```ruby
module foo;

public type X = unit {
    : bytes &size=2; # FF
    : Y;
};

type Y = unit {
    payload: bytes &eod;
    : uint8[] &parse-from=self.payload foreach { print "element %s" % $$; }
};
```

```console
$ echo -n -e 'FF\x01\x02' | spicy-driver -d foo.spicy
element 1
[error] terminating with uncaught exception of type hilti::rt::Overflow: integer overflow (foo.spicy:10:50)
```

If instead `Y` is defined as the following equivalent code, parsing succeeds:

```ruby
type Y = unit {
    : uint8[] foreach { print "element %s" % $$; }
};
```

This is a regression and started to fail with 755437a17c1f80f1345dfb28877a308435ce85e1.

bbannier:
The issue here is in how we safe (and later restore) the parse position, https://github.com/zeek/spicy/blob/8e985bd345b0f39e333788bb8fd7999f14b1a03e/spicy/toolchain/src/compiler/codegen/parser-builder.cc#L2445-L2448.

When redirecting input we pick a different `cur`, so above subtraction works with different iterators (same for `&parse-at`), https://github.com/zeek/spicy/blob/755437a17c1f80f1345dfb28877a308435ce85e1/spicy/toolchain/src/compiler/codegen/parser-builder.cc#L1206-L1221

bbannier:
Fixed by #1476.

PR #1476, opened 2023-07-03
Revert #1439

bbannier:
This works around #1475 via revert. With that we reopen #1421 and #1089.


rsmmr (approved):

bbannier:

PR #1479, opened 2023-07-05
Support skipping `%done` with external hooks

rsmmr:
- Support skipping explicit `%done` in external hooks.
- Produce better error message when a `%XXX` hook is used on a unit field.


tests/Baseline/spicy.types.unit.hooks-fail/output
```diff
@@ -13,11 +13,13 @@
...
 [error] <...>/hooks-fail.spicy:18:11: unknown hook '%x2'
 [error] <...>/hooks-fail.spicy:19:13: cannot use paths in hooks; trigger on the top-level field instead
-[error] <...>/hooks-fail.spicy:23:15: hook '%init' does not take any parameters
-[error] <...>/hooks-fail.spicy:24:15: signature for hook must be: %gap(seq: uint64, len: uint64)
-[error] <...>/hooks-fail.spicy:24:15: cannot use hook '%gap', unit type does not support sinks because it is not public
-[error] <...>/hooks-fail.spicy:25:12: no field 'y1' in unit type
-[error] <...>/hooks-fail.spicy:26:13: unknown hook '%y2'
-[error] <...>/hooks-fail.spicy:27:15: cannot use paths in hooks; trigger on the top-level field instead
-[error] <...>/hooks-fail.spicy:28:11: unknown ID 'XXXX'
+[error] <...>/hooks-fail.spicy:22:16: cannot use '%done' with a unit field
```
awelzel, line 16: Nit, the other messages always mention "hook" before before  '%name' part.

> hook '%init' does not take any parameters
> cannot use hook '%gap', unit type does not support sinks because it is not public

awelzel (dismissed):
Nice :-)

spicy/toolchain/src/compiler/visitors/normalizer.cc
```diff
@@ -128,16 +128,48 @@ struct Visitor : public hilti::visitor::PostOrder<void, Visitor> {
...
+                }
+                else {
+                    // Produce a tailored error message if `%XXX` is used on a unit field.
+                    if ( auto id = ns.namespace_(); id && hilti::util::startsWith(h.id().local(), "0x25_") ) {
+                        if ( auto resolved = hilti::scope::lookupID<hilti::declaration::Type>(id, p, "unit type") ) {
+                            if ( auto utype =
+                                     resolved->first->as<hilti::declaration::Type>().type().tryAs<type::Unit>();
+                                 utype && utype->itemByName(ns.local()) ) {
+                                p.node.addError(hilti::util::fmt("cannot use '%s' with a unit field",
+                                                                 hilti::util::replace(h.id().local(), "0x25_", "%")));
```
bbannier, line 164: Maybe more as a note, this will mutilate valid hook but unused hook names like `%done_0x25_foo` and show `%done_foo`. What we really want here is to replace a prefix `0x25_` with `%`, but we have no tooling for that. Same below.

bbannier (approved):

spicy/toolchain/src/compiler/visitors/normalizer.cc
```diff
@@ -128,16 +128,48 @@ struct Visitor : public hilti::visitor::PostOrder<void, Visitor> {
...
+                if ( unit_type_ref->as<type::Unit>().itemByName(h.id().namespace_().local()) ) {
+                    p.node.addError(hilti::util::fmt("cannot use '%s' with a unit field",
+                                                     hilti::util::replace(h.id().local(), "0x25_", "%")));
+                    return;
+                }
+            }
+        }
+        else {
             // External hook, do name lookuo.
             auto ns = h.id().namespace_();
```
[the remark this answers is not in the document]
rsmmr: Actually I prefer `ns` here, just as a descriptor of where that piece is coming from.

spicy/toolchain/src/compiler/visitors/normalizer.cc
rsmmr: Yeah, not worth the trouble here I think, iirc we do this elsewhere too, and we control the names of hooks anyways.

PR #1482, opened 2023-07-06
Fix typo: maccOS -> macOS

ryandesign:


ryandesign:
...although this PR doesn't really fix the two big problems that occur when this check is triggered:

```
CMake Error at auxil/spicy/cmake/Util.cmake:87 (message):
  Need maccOS version >= 19.0.0, found
```

1. "19.0.0" is not a macOS version; it's a Darwin version
2. It doesn't tell us what OS version it found

So a more comprehensive fix is needed.

bbannier:
Thanks for the patch, I merged it as is.

PR #1487, opened 2023-07-18
Mention unit switch statements in conditional parsing docs.

bbannier:
Instead of decorating many fields with `if` conditions for certain grammars a simpler approach can be to use a unit switch statement with branch-by-expressions. Add a link to unit switch from the conditional parsing section.

### DIFF
--- a/doc/programming/parsing.rst
+++ b/doc/programming/parsing.rst
@@ -1319,6 +1319,9 @@ time the field is next in line.
     :exec: printf '\01\02\03\04' | spicy-driver %INPUT; printf '\02\02\03\04' | spicy-driver %INPUT
     :show-with: foo.spicy
 
+For repeated cases of conditional parsing unit :ref:`parse_switch` statements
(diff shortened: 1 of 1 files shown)


doc/programming/parsing.rst
```diff
@@ -1319,6 +1319,9 @@ time the field is next in line.
     :exec: printf '\01\02\03\04' | spicy-driver %INPUT; printf '\02\02\03\04' | spicy-driver %INPUT
     :show-with: foo.spicy
 
+For certain cases of conditional parsing unit :ref:`parse_switch` statements
+can be used and should allow more compact and easier to maintain code.
```
rsmmr, line 1323: Just word smithing:

```suggestion
For repeated cases of conditional parsing, unit :ref:`parse_switch` statements
might allow for more compact and easier to maintain code.
```

rsmmr (dismissed):

Issue #1484, opened 2023-07-06
`&convert` on `bitfield` seems to be broken


bbannier:
The following code does not compile; I would have expected the `&convert` to have the effect of extracting the element value.

```ruby
module foo;

public type X = unit {
    x: bitfield(8) {
        a: 0..4;
    } &convert=$$.a;
};
```

```console
$ printf '\x00' | spicy-dump -d foo.spicy
[error] foo.spicy:3:17-7:2: cannot coerce expression '(__bits)' of type 'tuple<uint<8>>' to type 'bitfield
(8) {
    a: 0..4; }'
[error] foo.spicy:3:17-7:2: cannot coerce expression '__parsed_x' of type 'bitfield(8) {
    a: 0..4; }' to type 'tuple<a: uint<8>>'
[error] spicy-dump: aborting after errors
```

rsmmr:
I believe this will come with #1032.

Issue #1468, opened 2023-06-23
Allow to map bitfield fields into parent scope


J-Gras:
It would be nice if fields of a bitfield can be mapped in the parent scope:
```
type Foo = unit {
  b: bitfield(8) {
    x: 0..3;
    y: 4..7;
  };
};
```
For example, not naming the bitfield could make `x` and `y` directly accessible from `Foo`.

See [Slack](https://zeekorg.slack.com/archives/C012CQV1K29/p1687505826326719) for a workaround at the cost of performance.

bbannier:
Without major rewrites any fix to this will likely only add syntactic sugar and incur the same runtime cost.

J-Gras:
Given that the manual approach requires ~3x the lines of code, the syntactic sugar would still be helpful I think.

J-Gras:
> Mind elaborating why this would be so useful? Why is foo.x preferable over foo.b.x?

This is mostly for estethic reasons. In the use case this came up, the fields of the bitfield are defined as "first-class members", just like other fields of the protocol specification. Introducing the additional nesting (including meaningful naming) just feels a bit unnatural. Along those lines, I was also aiming to use the automatic export of types, which promises to save a lot of boilerplate code (now wondering whether bitfields are supported). So all in all, I am just trying to minimize work while coding closely along a protocol spec.

rsmmr:
> This is mostly for estethic reasons. In the use case this came up, the fields of the bitfield are defined as "first-class members", just like other fields of the protocol specification. Introducing the additional nesting (including meaningful naming) just feels a bit unnatural.

Ok, got it. Yeah, your suggestion of not naming the bitfield would work well then I think. Like:

```
type Foo = unit {
  : bitfield(8) {
    x: 0..3;
    y: 4..7;
  };
};
```

Then `foo.x` and `foo.y` would be allowed. Performance would remain the same: internally we'd just map the IDs accordingly during compilation.

> (now wondering whether bitfields are supported)

Good question. :-)

bbannier:
We should check what changes this needs for Zeek's Spicy integration.

rsmmr:
Re-assigning to me, I'm adding this to the other bitfield work.

Issue #1032, opened 2021-10-29
Bitfield not rejected in type decl


bbannier:
Currently the following code passes validation, but fails with an internal error when compiled:

```.r
# file: /tmp/bitfield.spicy
module foo;

type X = bitfield(32) {
    x: 0..4;
};
```

```
$ spicyc -j /tmp/bitfield.spicy
|| - spicy::type::Bitfield (bitset.spicy:3:10-5:2) (non-const) (type-id: foo::X) (resolved) [@s:7f81cb412c60]
||  - type::UnsignedInteger <width="32"> (bitset.spicy:3:10-5:2) (const) (resolved) [@t:7f81cb411e70]
||  - type::Tuple <wildcard="false"> (bitset.spicy:3:10-5:2) (const) (resolved) [@t:7f81cb48ee30]
||   - type::tuple::Element (bitset.spicy:3:10-5:2) [@t:7f81cb48f120]
||    - ID <name="x"> (bitset.spicy:4:5) [@i:7f81cb48e010]
||    - type::UnsignedInteger <width="32"> (const) (resolved) [@t:7f81cb4120e0]
||  - spicy::type::bitfield::Bits <field_width="32" lower="0" upper="4"> (bitset.spicy:4:5) [@s:7f81cb412c00]
||   - ID <name="x"> (bitset.spicy:4:5) [@i:7f81cb412240]
||   - declaration::Expression %4 <linkage="private"> (bitset.spicy:4:5) [canon-id: foo::X::__dd] [@d:7f81cb412580]
||    - ID <name="__dd"> [@i:7f81cb4121d0]
||    - expression::Keyword <kind="$$"> (non-const) (resolved) [@e:7f81cb412170]
||     - type::UnsignedInteger <width="32"> (const) (resolved) [@t:7f81cb4120e0]
||    - node::None (bitset.spicy:4:5) [@n:7f81cb405fb0]
||   - type::UnsignedInteger <width="32"> (const) (resolved) [@t:7f81cb4120e0]
||   - node::None (bitset.spicy:4:5) [@n:7f81cb405fb0]
|| [internal-error] /private/tmp/bitset.spicy:3:10-5:2: codegen: type foo::X does not have a visitor
|| 
|| --- Aborting
|| # 0   libhilti.dylib                      0x000000010fef0696 _ZN5hilti2rt9BacktraceC1Ev + 118
[...]

PR #1492, opened 2023-08-04
Bitfield improvements

rsmmr:
This addresses #1032, #1468,  and #1484.


spicy/toolchain/bin/spicy-dump/printer-text.cc
```diff
@@ -116,7 +136,11 @@ void TextPrinter::print(const type_info::Value& v) {
                     if ( y ) {
                         out() << '\n';
                         outputIndent();
-                        out() << f.name << ": ";
+
+                        if ( ! f.isAnonymous() )
+                            out() << f.name;
```
bbannier, line 141: I believe this is still inconsistent and not what we want. It seems that e.g., `spicy-dump` omits any anonymous field unless it is a bitfield, e.g., given the following unit
```ruby
public type X = unit {
    : uint8;
    : bitfield(8) { a: 0..7; };
};
```
I see the following output
```console
$ printf '\x00\x00' | spicy-dump bla.spicy
bla::X {
  : (a: 0)
}
```
I would have expected no output here.

bbannier (changes requested):

spicy/toolchain/src/compiler/visitors/normalizer.cc
```diff
@@ -227,6 +227,48 @@ struct Visitor : public hilti::visitor::PostOrder<void, Visitor> {
...
+                auto bits = t->bits(id);
+                if ( ! bits )
+                    continue;
+
+                if ( found_field ) {
+                    hilti::logger().error(
+                        hilti::util::fmt("unit '%s' has multiple anonymous bitfields with member '%s'", *unit->id(),
+                                         id));
+                    return;
+                }
```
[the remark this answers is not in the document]
rsmmr: > Just wondering whether we should instead reject multiple anon bitfields with fields with same name

Ack, I'll add that validation.

spicy/toolchain/bin/spicy-dump/printer-text.cc
rsmmr: Agree that that's not right. However, in the spirit of our use case here, what I'd expect is actually this:

```
bla::X {
  a: 0
}
```

I'll give that a try.

rsmmr: Adding three new commits at the end implementation this. The first two change how (normal) bitfields and also offsets are printed, the 3rd then implements printing of anonymous bitfields.

bbannier (approved):
🚀

PR #1505, opened 2023-08-28
Use user-inaccessible chars for encoding `::` in feature variables

bbannier:


hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -919,31 +936,30 @@ struct FeatureRequirementsVisitor : visitor::PreOrder<void, FeatureRequirementsV
...
 
-        const auto& tokens = util::split(id, "%");
-        assert(tokens.size() == 3);
-
-        // The type name is encoded in the variable name with `::` replaced by `__`.
-        const auto& typeID = ID(util::replace(tokens[1], "__", "::"));
-        const auto& feature = tokens[2];
+        const auto& id_feature = idFeatureFromConstant(x.id());
+        if ( ! id_feature )
+            return;
```
rsmmr, line 945:
```suggestion
            return;
            
```

rsmmr (dismissed):

PR #1510, opened 2023-08-29
Support anonymous bitfields in switches

### LINKED ISSUE #1503: Unwrapping of `bitfield` into unit broken under unit `switch`
We map anon `bitfield` entries into the parent unit. This seems to be broken if the bitfield is under a unit `switch`.

```ruby
# foo.spicy
module foo;

public type X = unit {
    n: uint8;

    switch (self.n) {
        0 -> : bitfield(8) {
            A: 0..7;
        };
        * -> : bitfield(8) {
            A: 0..7;
        };
    };

    on %done { print self.A; }
};
```

```console
$ spicyc -dj foo.spicy
[error] foo.spicy:15:22: type does not have field 'A'
```

* * *

Additionally, stringification for such units seems to be broken and we leak the anon fields,

```ruby
module foo;

public type X2 = unit {
    n: uint8;

    switch (self.n) {
        0 -> : bitfield(8) {
            A: 0..7;
        };
        * -> : bitfield(8) {
            A: 0..7;
        };
    };

    on %done { print self; }
};
```

```console
$ printf '\x00\x01' | spicy-driver foo.spicy
[$n=0, $_anon=(1), $_anon_2=(not set)]
```

Dumping on the other hand seems to work:

```console
$ printf '\x00\x01' | spicy-dump foo.spicy
foo::X {
  n: 0
  A: 1
}
```

rsmmr:
Closes #1503


spicy/toolchain/src/compiler/visitors/normalizer.cc
```diff
@@ -227,41 +227,84 @@ struct Visitor : public hilti::visitor::PostOrder<void, Visitor> {
...
-        auto unit = o.op0().type().tryAs<type::Unit>();
-        auto id = o.op1().tryAs<hilti::expression::Member>()->id();
-        if ( unit && id && ! unit->itemByName(id) ) {
-            // See if we we got an anonymous bitfield with a member of that
-            // name. If so, rewrite the access to transparently to refer to the
-            // member through the field's internal name.
-            bool found_field = false;
-            for ( const auto& f : unit->items<type::unit::item::Field>() ) {
-                if ( ! f.isAnonymous() )
+    // Helper return the field name containing a given item of an anonymous bitfield.
```
bbannier, line 230:
```suggestion
    // Helper returning the field name containing a given item of an anonymous bitfield.
```

spicy/toolchain/src/compiler/visitors/normalizer.cc
```diff
@@ -227,41 +227,84 @@ struct Visitor : public hilti::visitor::PostOrder<void, Visitor> {
...
-                    return;
+        return {};
+    }
 
+    void operator()(const operator_::unit::MemberNonConst& o, position_t p) {
+        auto unit = o.op0().type().tryAs<type::Unit>();
+        auto id = o.op1().tryAs<hilti::expression::Member>()->id();
+        if ( unit && id && ! unit->itemByName(id) ) {
+            // See if we we got an anonymous bitfield with a member of that
+            // name. If so, rewrite the access to transparently to refer to the
```
bbannier, line 261:
```suggestion
            // See if we got an anonymous bitfield with a member of that
            // name. If so, rewrite the access to transparently refer to the
```

spicy/toolchain/src/compiler/visitors/normalizer.cc
```diff
@@ -227,41 +227,84 @@ struct Visitor : public hilti::visitor::PostOrder<void, Visitor> {
...
+        }
+    }
+
+    void operator()(const operator_::unit::HasMember& o, position_t p) {
+        auto unit = o.op0().type().tryAs<type::Unit>();
+        auto id = o.op1().tryAs<hilti::expression::Member>()->id();
+
+        if ( unit && id && ! unit->itemByName(id) ) {
+            // See if we we got an anonymous bitfield with a member of that
+            // name. If so, rewrite the access to transparently to refer to the
```
bbannier, line 284:
```suggestion
           // See if we got an anonymous bitfield with a member of that
            // name. If so, rewrite the access to transparently refer to the
```

hilti/toolchain/src/compiler/cxx/elements.cc
```diff
@@ -342,7 +342,9 @@ std::string cxx::type::Struct::str() const {
 
     auto fmt_member = [&](const auto& f) {
         if ( auto x = std::get_if<declaration::Local>(&f) ) {
-            if ( ! (x->isInternal() || x->linkage == "inline static") ) // Don't visit internal or static fields.
+            if ( x->isAnonymous() )
+                visitor_calls.emplace_back(fmt("_(\"<anon>\", %s); ", x->id));
```
bbannier, line 346: Nit: This is still slightly weird in that it seems to only apply to anon bitfield fields, so if users mix these switch bitfield constructs and other anon fields the output might given them pause.

I don't have a better suggestion and this is already pretty contrived, so totally fine if we just go with this for now.

spicy/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -571,6 +571,35 @@ struct VisitorPost : public hilti::visitor::PreOrder<void, VisitorPost>, public
         }
     }
 
+    void checkBits(const spicy::type::Unit& u, position_t p, const hilti::node::Set<type::unit::Item>& items,
+                     std::set<ID>* seen_bits) {
+        for ( const auto& item : items ) {
+            if ( auto f = item.template tryAs<spicy::type::unit::item::Field>() ) {
```
bbannier, line 577: We don't seem the `template` here as `item` is not a dependent type.

```suggestion
            if ( auto f = item.tryAs<spicy::type::unit::item::Field>() ) {
```

spicy/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -571,6 +571,35 @@ struct VisitorPost : public hilti::visitor::PreOrder<void, VisitorPost>, public
...
+                if ( ! f->isAnonymous() )
+                    continue;
+
+                auto t = f->itemType().template tryAs<type::Bitfield>();
```
bbannier, line 581: Ditto.

spicy/toolchain/src/compiler/visitors/validator.cc
```diff
@@ -571,6 +571,35 @@ struct VisitorPost : public hilti::visitor::PreOrder<void, VisitorPost>, public
...
+                        error(fmt("bitfield item '%s' shadows unit field", b.id()), p);
+
+                    if ( seen_bits->find(b.id()) != seen_bits->end() )
+                        error(fmt("bitfield item name '%s' appears in multiple anonymous bitfields", b.id()), p);
+
+                    seen_bits->insert(b.id());
+                }
+            }
+
+            else if ( auto f = item.template tryAs<spicy::type::unit::item::Switch>() ) {
```
bbannier, line 596: Ditto.

bbannier (dismissed):

Issue #1489, opened 2023-07-28
Inconsistent Bitfield bitrange behavior


aribl:
Running the following code results in the expected individual bit values based on the specified bit order, but unexpected bitrange values based on the specified bit order. Expected result for bitrange 0..7 in lsb bitfield is `55`.

```
module bitorder;

import spicy;

public type Byte = unit { 
    data: bytes &size=1;

    lsb: bitfield(8) {
        bit0: 0;
        bit7: 7;
        range0_7: 0..7;
    } &parse-from=self.data &bit-order=spicy::BitOrder::LSB0;

    msb: bitfield(8) {
        bit0: 0;
        bit7: 7;
        range0_7: 0..7;
    } &parse-from=self.data &bit-order=spicy::BitOrder::MSB0;

    on %done { 
        print "LSB bit0: %x" % self.lsb.bit0;
        print "LSB bit7: %x" % self.lsb.bit7;
        print "LSB bitfield 0..7: %x" % self.lsb.range0_7;
        print "MSB bit0: %x" % self.msb.bit0;
        print "MSB bit7: %x" % self.msb.bit7;
        print "MSB bitfield 0..7: %x" % self.msb.range0_7;
    }
};
```

```
> printf '\xAA' | spicy-driver bitorder.spicy
LSB bit0: 0
LSB bit7: 1
LSB bitfield 0..7: aa
MSB bit0: 1
MSB bit7: 0
MSB bitfield 0..7: aa
```

rsmmr:
I would actually expect `aa` for `0..7` with LSB0. Bit 7 (i.e., the msb) is one in that case, so the value must be > 0x80. Also, LSB0 is the default so getting a different value than put in when selecting all bits would be surprising.

But I'm wondering if there's a problem in the handling of MSB0: the way it works internally is that it first swaps the bit order of the whole integer value, the extracts the bits to yield the desired value. But that value is now in LSB0 order, so would it need to swap the bit order once more before continuing with the value? That would then yield the `55` value.

aribl:
You're right about LSB0, it seems to be working as intended - my confusion. I see your point about how the bitfields operate - even though the location of each bit is different, the natures of MSB0 and LSB0 mean that the range starting at bit0 is built LSB first (right to left) for an LSB0 bitfield and MSB first (left to right) for an MSB0 bitfield. If the types containing the values respected bitorder, this would be fine, but since the type itself is LSB0 bitorder, the internal representation of the ranges becomes the same. 

[...]

PR #1511, opened 2023-08-31
Bitfield constants

rsmmr:


doc/programming/language/types.rst
```diff
@@ -40,6 +40,17 @@ parsed inside a unit.
...
+.. rubric:: Constants
+
+- ``bitfield(N) { RANGE_1 [= VALUE_1]; ...; RANGE_N [= VALUE_N] }``
+
+A bitfield constant represents expected values for all or some of the
+individual bitranges. They can be used only for parsing inside a unit
+field, not as values to otherwise operate with. To define such a
+constant with expected values, add ``= VALUE`` to the bitranges inside
+the type definition as suitable. See :ref:`parse_bitfield` for more
+information.
```
bbannier, line 52: Maybe a bit of foreshadowing, but could you explicitly spell out that `VALUE` corresponds to the extracted value _after_ applying any `&bit-order` if given (at least that's what I would assume)?

We should we'd also test that.

doc/programming/parsing.rst
```diff
@@ -1037,6 +1037,25 @@ range to an enum, using ``$$`` to access the parsed value:
     :exec: printf '\x21' | spicy-driver %INPUT
     :show-with: foo.spicy
 
+When parsing a bitfield, you can enforce expected values for some
```
bbannier, line 1040: Let's use a ref to the bitfield types section around here so users can have a look at more details.

spicy/toolchain/src/compiler/codegen/parsers/types.cc
```diff
@@ -128,6 +128,8 @@ struct Visitor : public hilti::visitor::PreOrder<Expression, Visitor> {
         performUnpack(target, t, t.width() / 8, {state().cur, fieldByteOrder(), *bitorder}, t.meta(), is_try);
 
         if ( pb->options().debug ) {
+            auto have_value = builder()->addIf(builder::hasMember(target, "__value__"));
+            pushBuilder(have_value);
```
bbannier, line 132: Nit: this would nicely fit use of `pushBuilder` with a lambda so we wouldn't have too use `popBuilder()` below.

tests/spicy/types/bitfield/ctor-fail.spicy
```diff
@@ -0,0 +1,16 @@
...
+
+module Test;
+
+public type Foo = unit {
+  x: bitfield(8) {
+    a: 1..3 = 255; # error
+    b: 1..3 = 8; # error
+    c: 1..3 = 7; # ok
+    d: 4..7 = -1;
+    e: 4..7 = "42";
```
bbannier, line 14:
```suggestion
    d: 4..7 = -1; # error
    e: 4..7 = "42"; # error
```

bbannier (approved):

hilti/toolchain/include/ast/operators/bitfield.h
```diff
@@ -75,4 +77,24 @@ right.
...
+            {{{}, type::constant(type::Bitfield(type::Wildcard())), false, {}, "bitfield"},
+             {{}, type::Member(type::Wildcard()), false, {}, "<field>"}};
+        return _operands;
+    }
+
+    void validate(const expression::ResolvedOperator& i, operator_::position_t p) const {
+        detail::checkName(i.op0(), i.op1(), p.node);
+    }
+
+    std::string doc() const { return "Returns true if the bitfield's element has a value to provide."; }
```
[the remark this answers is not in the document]
rsmmr: I fuzzed around with this a bit because "has a value" might seem a bit odd at first because in the standard use case (mapping an integer to a bitfield) all elements are always available anyways. But yeah,  "to provide" doesn't really help either.

doc/programming/language/types.rst
rsmmr: Ack. Also adding a note to #1489 to test that (don't think it'd help much right now where that isn't settled).

PR #1517, opened 2023-09-05
Do not include generated Doxygen documentation in RTD search results

### LINKED ISSUE #1516: Hilti doxygen indexed in latest RTD
Searching the latest Spicy doc mostly returns hits on Hilti stuff (e.g., https://docs.zeek.org/projects/spicy/en/latest/search.html?q=bool).

bbannier:
Doxygen documentation is likely not relevant for users and hides results they are looking for. Also, Doxygen already comes with its own search which is reachable the developer documentation.

Closes #1516.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -14,3 +14,13 @@ sphinx:
 python:
   install:
     - requirements: doc/requirements.txt
+
+search:
+  ignore:
+    - "doxygen-output/*"
+    - "_static/*"
+    # Defaults.
+    - search.html
+    - search/index.html
+    - 404.html
+    - 404/index.html


.readthedocs.yaml
```diff
@@ -14,3 +14,8 @@ sphinx:
 python:
   install:
     - requirements: doc/requirements.txt
+
+search:
+  ignore:
+    - "doxygen-output/*"
+    - "_static/*"
```
awelzel, line 21: Wonder if the default list should be kept here? Second example of this section:

https://docs.readthedocs.io/en/stable/config-file/v2.html#search-ignore

```
     # Defaults
     - search.html
     - search/index.html
     - 404.html
     - 404/index.html'
```

Otherwise, lgtm and :crossed_fingers:

awelzel (approved):

.readthedocs.yaml
bbannier: Thanks for checking, adjusted.

PR #1526, opened 2023-09-14
Add support for passing arbitrary C++ compiler flags.

bbannier:
This adds a magic environment variable `HILTI_CXX_FLAGS` which if set specifies compiler flags which should be passed during C++ compilation after implicit flags. This could be used to e.g., set defines, or set low-level compiler flags.

Even with this flag, for passing include directories one should still use `HILTI_CXX_INCLUDE_DIRS` since they are searched before any implicitly added paths.

### DIFF
--- a/doc/programming/extending.rst
+++ b/doc/programming/extending.rst
@@ -278,3 +278,13 @@ If your C++ code requires additional include files outside of standard
 include paths, you can set the environment variable
 ``HILTI_CXX_INCLUDE_DIRS`` to a colon-separated list of additional
 directories for ``spicyc`` to use when compiling C++ code.
+
+Passing arbitrary compiler flags
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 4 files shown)


doc/programming/extending.rst
```diff
@@ -278,3 +278,13 @@ If your C++ code requires additional include files outside of standard
 include paths, you can set the environment variable
 ``HILTI_CXX_INCLUDE_DIRS`` to a colon-separated list of additional
 directories for ``spicyc`` to use when compiling C++ code.
+
+Passing arbitrary compiler flags
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you need to set arbitrary compiler flags you can use the environment
+variable ``HILTI_CXX_FLAGS``. If this value is set it will be added verbatim as
+last argument in the compiler invocation used when compiling C++ code.
```
rsmmr, line 287:
```suggestion
the last argument to the compiler invocation when compiling C++ code.
```

rsmmr (approved):

PR #1528, opened 2023-09-15
Use find_package(Python) with version.

awelzel:
Zeek's configure sets Python_EXECUTABLE has hint, but Spicy is using find_package(Python3) and would only use Python3_EXECUTABLE as hint. This results in Spicy finding a different (the default) Python executable when configuring Zeek with --with-python=/opt/custom/bin/python3.

Switch Spicy over to use find_package(Python) and add the minimum version so it knows to look for Python3.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,7 +136,7 @@ if (BISON_ROOT)
         NO_DEFAULT_PATH)
 endif ()
 
-find_package(Python3)
+find_package(Python ${python_minimum_version} COMPONENTS Interpreter)
 find_package(FLEX REQUIRED)
 find_package(BISON REQUIRED)
 find_package(ZLIB REQUIRED)
@@ -163,7 +163,7 @@ if (APPLE)
     require_version("macOS" MACOS_FOUND ${CMAKE_SYSTEM_VERSION} "${macos_minimum_version}" true)
 endif ()
 
-require_version("Python" Python3_FOUND Python3_VERSION "${python_minimum_version}" true)
+require_version("Python" Python_FOUND Python_VERSION "${python_minimum_version}" true)
 require_version("Flex" FLEX_FOUND FLEX_VERSION "${flex_minimum_version}" true)
 require_version("Bison" BISON_FOUND BISON_VERSION "${bison_minimum_version}" true)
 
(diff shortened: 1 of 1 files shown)


awelzel:
"Maybe fix for": https://zeekorg.slack.com/archives/CSZBXF6TH/p1694734672910209

We could also do `set(Python3_EXECUTABLE ${Python_EXECUTABLE})` within Spicy or Zeek as an alternative fix.

zeekctl, zeek-aux, broker and zeek all use `find_package(Python)` so seemed reasonable to align.

bbannier:
> We could also do `set(Python3_EXECUTABLE ${Python_EXECUTABLE})` within Spicy or Zeek as an alternative fix.

I'd vote for getting this PR into Spicy (assuming CI passes), and also setting this in Zeek for some transition period (if only to make bisecting changes easier). We can pull that workaround out with e.g., zeek-6.2, I believe it should have no user-visible effect.

bbannier:
Looking at the Spicy build in more detail, there's no reason to find Python at all, all we do with it currently is log the found versions. Either we merge this as is, and add a cleanup, or you could just now pull out all mention of `[pP]ython` in `CMakeLists.txt`.

awelzel:
> Looking at the Spicy build in more detail, there's no reason to find Python at all, all we do with it currently is log the found version. Either we merge this as is and add a cleanup, or you could just now pull out all mention of `[pP]ython` in `CMakeLists.txt`.

There are to a few scripts/tests that rely on the existence of a Python interpreter though - so Python is a required package for building? There might be pieces missing to actually respect that python interpreter?

bbannier:
> > Looking at the Spicy build in more detail, there's no reason to find Python at all, all we do with it currently is log the found version. Either we merge this as is and add a cleanup, or you could just now pull out all mention of `[pP]ython` in `CMakeLists.txt`.
> 
> There are to a few scripts/tests that rely on the existence of a Python interpreter though - so Python is a required package for building? There might be pieces missing to actually respect that python interpreter?

Are right, these just worked accidentially since I had Python available.

bbannier (approved):
Thanks, this looks good.

bbannier:

PR #1529, opened 2023-09-19
Document using anonymous field for extracting TCP messages

bbannier:


doc/faq.rst
```diff
@@ -120,8 +120,13 @@ allow for that. For example, if the grammar parses messages of the
...
 
-A simple way to accomplish this is to introduce a parser which wraps
-messages of the protocol:
+One way to express this is to introduce a parser which wraps
+messages of the protocol in an :ref:`anonymous field <anonymous_fields>`.
+
+.. warning:: Since in general the number of messages exchanged over a TCP
+  connection is unbounded, an anonymous field should be used. If a named field
+  was used instead the parser would need to store all messages over the
+  connection which would lead to unbounded memory growth.
```
rsmmr, line 129:
```suggestion
  connection, which would lead to unbounded memory growth.
```

rsmmr (approved):

doc/faq.rst
bbannier: Not a native speaker, but if I understand e.g., https://www.grammarly.com/blog/comma-before-which/ correctly, this would be a change, which would be incorrect. 😉

PR #1530, opened 2023-09-19
Declare Spicy pygments extension as parallel-safe. [skip CI]

bbannier:
We previously would not declare that the Spicy pygments highlighter is safe to execute in parallel (reading or writing of sources). Sphinx then assumed that the extension was not safe to run in parallel and instead ran jobs sequentially.

This patch declares the extension as able to execute in parallel. Since the extension does not manage any external state this is safe.


rsmmr (approved):

bbannier:

PR #1531, opened 2023-09-20
Add trait class to bitfield types.

rsmmr:
This lets C++ templates test if a class T is a bitfield.

Needed for upcoming Zeek-side PR.

### DIFF
--- a/hilti/runtime/include/types/bitfield.h
+++ b/hilti/runtime/include/types/bitfield.h
@@ -13,18 +13,22 @@
 
 namespace hilti::rt {
 
+namespace trait {
+struct isBitfield {};
+} // namespace trait
+
 /// A bitfield is just a type wrapper around a tuple of the corresponding field
 /// values (including the hidden additional element storing the bitfield's
 /// original integer value). We wrap it so that we can customize the
 /// printing.
 template<typename... Ts>
-struct Bitfield {
+struct Bitfield : public trait::isBitfield {
     std::tuple<std::optional<Ts>...> value;
 };
 
 template<typename... Ts>
 Bitfield<Ts...> make_bitfield(Ts&&... args) {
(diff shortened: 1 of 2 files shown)


hilti/runtime/include/types/bitfield.h
```diff
@@ -13,18 +13,22 @@
...
 template<typename... Ts>
-struct Bitfield {
+struct Bitfield : public trait::isBitfield {
     std::tuple<std::optional<Ts>...> value;
 };
 
 template<typename... Ts>
 Bitfield<Ts...> make_bitfield(Ts&&... args) {
-    return Bitfield<Ts...>{std::make_tuple(std::forward<Ts>(args)...)};
+    return Bitfield<Ts...>{{}, std::make_tuple(std::forward<Ts>(args)...)};
```
bbannier, line 31: Just a note, no action needed.

Another way do deal with the base class would be to use designated initializers,
```suggestion
    return Bitfield<Ts...>{.value = std::make_tuple(std::forward<Ts>(args)...)};
```

hilti/toolchain/src/compiler/codegen/ctors.cc
```diff
@@ -36,7 +36,7 @@ struct Visitor : hilti::visitor::PreOrder<cxx::Expression, Visitor> {
                 values.emplace_back("std::nullopt");
         }
 
-        return fmt("hilti::rt::Bitfield<%s>{std::make_tuple(%s)}", util::join(types, ", "), util::join(values, ", "));
+        return fmt("hilti::rt::Bitfield<%s>{{}, std::make_tuple(%s)}", util::join(types, ", "), util::join(values, ", "));
```
bbannier, line 39: This is missing from the bump over in zeek/zeek#3315.

bbannier (approved):

PR #1532, opened 2023-09-21
Fix consistency issues for iterator stored in units

### LINKED ISSUE #1478: `offset()` works with incompatible iterators when used with lists/`&parse-[at|from]`
While working on #1475 I noticed that `offset()` seems to return incorrect results, even after pulling out the incorrect changes from #1439. The following reproducer (derived from the test added in #1477) fails in all versions down to v1.2.0.

```ruby
# @TEST-EXEC: spicyc -dj %INPUT -o foo.hlto
# @TEST-EXEC: ${SCRIPTS}/printf 'FF\x01\x02' | spicy-driver foo.hlto >output 2>&1

module foo;

public type X = unit {
    : bytes &size=2; # FF
    : Y;
};

type Y = unit {
    payload: bytes &eod;
    : uint8[2] &parse-from=self.payload foreach { print self.offset(); }
};
```

bbannier:
Closes #1478.


rsmmr (approved):

bbannier:
Backported to `release/1.8`.

PR #1534, opened 2023-09-21
Artifically limit the number of open files.

bbannier:
This works around a silent failure in reproc where it would refuse to run on systems which huge rlimits for the number of open files. We have seen this hit on huge production boxes.

### DIFF
--- a/hilti/toolchain/src/compiler/jit.cc
+++ b/hilti/toolchain/src/compiler/jit.cc
@@ -1,5 +1,7 @@
 // Copyright (c) 2020-2023 by the Zeek Project. See LICENSE for details.
 
+#include <sys/errno.h>
+#include <sys/resource.h>
 #include <unistd.h>
 
 #include <array>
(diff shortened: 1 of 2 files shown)


hilti/toolchain/src/compiler/jit.cc
```diff
@@ -592,6 +594,29 @@ Result<Nothing> JIT::JobRunner::_waitForJobs() {
...
+        limit.rlim_cur = REPROC_MAX_FD_LIMIT;
+        if ( setrlimit(RLIMIT_NOFILE, &limit) != 0 ) {
+            logger().internalError(
+                util::fmt("cannot set limit for number of open files ('ulimit -n %d'), please set it in your "
+                          "environment: %s",
+                          REPROC_MAX_FD_LIMIT, strerror(errno)));
+        }
+    }
+}
+
```
awelzel, line 619: Seems pragmatic. Was tempted to suggest a warning when this is being done, and whether the constructor is the right place, but mostly mood.

Fix tested on Linux by:
1) echo 2000000 > /proc/sys/fs/nr_open  (default per process max was 1048576)
2) ulimit -n 2000000 in the shell
3) compiling the quic analyzer with Spicy 1.8.1 (7e074675) (10801)
```
# make
[ 25%] Building CXX object analyzer/CMakeFiles/decrypt_crypto.dir/decrypt_crypto.cc.o
[ 50%] Linking CXX static library libdecrypt_crypto.a
[ 50%] Built target decrypt_crypto
[ 75%] Compiling QUIC analyzer
[error] <Spicy Plugin for Zeek>: C++ compiler not available or not functioning (looking for "/usr/bin/c++")
make[2]: *** [analyzer/CMakeFiles/QUIC.dir/build.make:76: quic.hlto] Error 1
make[1]: *** [CMakeFiles/Makefile2:100: analyzer/CMakeFiles/QUIC.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```

With Spicy from this PR 1.9.0-dev.93 (d7888753) (10900)
```
[ 25%] Building CXX object analyzer/CMakeFiles/decrypt_crypto.dir/decrypt_crypto.cc.o
[ 50%] Linking CXX static library libdecrypt_crypto.a
[ 50%] Built target decrypt_crypto
[ 75%] Compiling QUIC analyzer
[100%] Preparing dependencies of QUIC
[100%] Built target QUIC
```
:tada:

awelzel (approved):

hilti/toolchain/src/compiler/jit.cc
bbannier: Thanks for testing this out. Like mentioned out of band, this could still run into issues if `rlim_max < REPROC_MAX_FD_LIMIT`, but I believe this would require `rlim_max < rlim_cur` as well which seems extremely weird and is hopefully impossible. Even if it would be possible to trigger, we at least now give the user information on ways to work around the issue.

bbannier:
FTR, this works around DaanDeMeyer/reproc#82.

bbannier:
Backported to `release/1.8`.

PR #1535, opened 2023-09-22
feat: report paths as relative to binary

r-rojo:
* Add a utility function, `fromOrigin`, in hilti/toolchain that returns a normalized path relative to the current process.
* Change hilti/toolchain/config to use relative paths when reporting
  * install prefix
  * lib directory
  * hiltic
* Change spicy/toolchain/config to use relative paths when reporting
  * spicyc

### DIFF
--- a/hilti/toolchain/include/base/util.h
+++ b/hilti/toolchain/include/base/util.h
@@ -417,7 +417,7 @@ using hilti::rt::createTemporaryFile; // NOLINT(misc-unused-using-decls)
 hilti::rt::filesystem::path currentExecutable();
 
 /** Returns the path relative to parent of the current executable. */
-hilti::rt::filesystem::path fromOrigin(const hilti::rt::filesystem::path &relativePath);
+hilti::rt::filesystem::path fromOrigin(const hilti::rt::filesystem::path& relativePath);
 
 /** Dumps a backtrace to stderr and then aborts execution. */
 [[noreturn]] extern void abort_with_backtrace();
--- a/hilti/toolchain/include/base/util.h
+++ b/hilti/toolchain/include/base/util.h
@@ -416,6 +416,9 @@ using hilti::rt::createTemporaryFile; // NOLINT(misc-unused-using-decls)
 /** Returns the path of the current executable. */
 hilti::rt::filesystem::path currentExecutable();
 
+/** Returns the path relative to parent of the current executable. */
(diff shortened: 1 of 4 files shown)


bbannier:
Thank you for opening this PR @r-rojo.

I am still not 100% sure I understand which problem this solves since the commits talk only about the _how_. Is it that paths backed into Spicy (visible e.g., via `spicy-config --include-dirs`) break if one moves (as in `mv`) the prefix Spicy was installed in around? That is currently unsupported. If it is support for that you are after this patch now breaks moving e.g., `spicyc` or other Spicy binaries around (also not explicitly supported).

Note that e.g., `zeek-config` also [bakes in absolute paths ATM](https://github.com/zeek/zeek/blob/b5aecece68f4b00808f71d3f1f8e5d7fce755205/cmake_templates/zeek-config.in#L10), and Spicy behaves identically. Unless there are good reasons I'd prefer to the current approach. A possible workaround for you would be to not move the Spicy prefix around, like you should already do for Zeek as well.

r-rojo:
@bbannier, I tried to summarize the basic problem in the title. To provide more details, the goal of this change is to make it so that moving a spicy installation tree won't impact ability for spicy binaries to find spicy collateral (includes, libraries, modules).
One use case is for adding spicy to a dependency manager (such as conan, vcpkg) that stores dependencies as packages that can be later installed to an arbitrary prefix. By removing the original install prefix from the code and making it relative, spicy binaries can work well with other related packages installed in the same prefix.

Another use case is for customers that want to use spicy tools that we include in our deliverables. They can't use tools, like spicy-dump, without having to pass in new prefix paths in environment and/or command line arguments.

As for the related programs like zeek-config, I also have a similar patch to accomplish the same goal of portability.

bbannier:
Thanks for writing down the motivation @r-rojo, definitely useful information.

> As for the related programs like zeek-config, I also have a similar patch to accomplish the same goal of portability.

Let's merge this and the changes to zeek/zeek at a similar time. Could you ping here again once that part is on its way as well, ideally with a link to the issue?

r-rojo:
Here's the PR for zeek
https://github.com/zeek/zeek/pull/3352
It's failing CI for some distros/OSes. I glanced and think it could be related to those not having c++17.

bbannier:
The top-level change to get this into zeek/zeek was closed, https://github.com/zeek/zeek/pull/3352#issuecomment-1800881834. Closing this for the time being as well. Please feel to reopen when things come together (again).

PR #1537, opened 2023-09-22
Fix access to anonymous bitfield element through a constant value.

### LINKED ISSUE #1533: Field from anonymous bitfield in other unit not resolved
If a bitfield field was lifted into its unit we do not find it when operating from a different unit.

```ruby
module foo;

type X = unit {
    : bitfield(8) {
        a: 0..7;
    };
};

type Y = unit(x: X) {
    on %init { print x.a; }
};
```

```console
$ spicyc -dj foo.spicy
[error] foo.spicy:10:22: type does not have field 'a'
```

rsmmr:
Closes #1533.


bbannier (approved):
Nice, great that this was a pretty isolated issue.

Issue #1549, opened 2023-09-28
Input not trimmed for certain parsers


bbannier:
A user reported running into a stack overflow for their Spicy-based Zeek analyzer on connection closed. Digging in with a debugger this appear to be due to so many `Chunk`s needing cleanup that the stack overflowed.

I believe the following reproduces their scenario. This constructs many small input chunks and feeds them to a parser which consumes exactly one chunk at a time.

```ruby
# @TEST-EXEC: spicyc -djo x.hlto x.spicy
# @TEST-EXEC: bash ./gen.sh
# @TEST-EXEC: spicy-driver -F input.dat x.hlto >>output 2>&1

# @TEST-START-FILE x.spicy
module x;

public type C = unit {
    %port= 80/tcp;
    : B[];
};

type B = unit {
    : bytes &size=1;
};
# @TEST-END-FILE

# @TEST-START-FILE gen.sh
echo '!spicy-batch v2' >>input.dat
echo '@begin-flow id1 stream 80/tcp' >>input.dat
echo '@begin-flow id2 block application/foo' >>input.dat

# Generate 1 million chunks.
for _ in $(seq 0 1000); do
	for _ in $(seq 0 1000); do
		echo '@data id1 1' >>input.dat
                echo 'C'           >>input.dat
	done
done
# @TEST-END-FILE
```

My expectation here would be that as soon as a `B` is parsed the input gets trimmed. Looking at the `spicy-verbose` log this seems to be not the case. Instead this reproducer aborts with a segfault due to a stack overflow and an extremely deep backtrace. This should be fixed.

A potentially separate issue is the behavior if the `B[]` field in `C` was not anonymous and trimming was not desired by the user (e.g., `C` being random access). We should still be able to execute such code like we can deal with huge containers elsewhere, however `Chunk` is implemented as a linked list and cleaned up differently. We might need to address this in the context of #1347.

bbannier:
This got broken by 3e6e1610fc4223575e6e2006809d96f757a61949, part of #1091, between v1.3.0 and v1.4.0.

bbannier:
Also fixed in release/1.8.

bbannier:
The fix from #1551 now surfaces #1554 which was a preexisting, but likely not observable issue.

PR #1555, opened 2023-09-29
Fix potential infinite loop when trimming data before stream.

### LINKED ISSUE #1554: Trimming before the head chunk of a stream leads to infinite loop
With #1551 we now emit more trim calls which is good. It however leads to an infinite loop with the `spicy.gap-recovery` test in the test suite of zeek/zeek. The issue there is that we hit a code path were would would like to trim with an offset before the head chunk of the stream (which was already previously trimmed).

bbannier:
Previously we would trigger an infinite loop if one tried to trim before the head chunk of a stream. In praxis this seem to have been no issue due to #1549 and us emitting way less calls to trim than possible.

This patch adds an explicit check whether we need to trim anything, and exits the low-level function early for such cases.

Closes #1554.

### DIFF
--- a/hilti/runtime/src/types/stream.cc
+++ b/hilti/runtime/src/types/stream.cc
@@ -4,6 +4,7 @@
 #include <hilti/rt/extension-points.h>
 #include <hilti/rt/types/bytes.h>
 #include <hilti/rt/types/stream.h>
+#include <hilti/rt/util.h>
 
 using namespace hilti::rt;
 using namespace hilti::rt::stream;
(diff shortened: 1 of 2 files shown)


hilti/runtime/src/types/stream.cc
```diff
@@ -112,6 +119,11 @@ void Chain::trim(const Offset& offset) {
             assert(_head->offset() == offset);
             break;
         }
+
+        else {
+            // Other offsets are already rejected before entering loop.
+            cannot_be_reached();
+        }
```
rsmmr, line 126:
```suggestion
        else
            // Other offsets are already rejected before entering loop.
            cannot_be_reached();
```

rsmmr (approved):

bbannier:

PR #1563, opened 2023-10-04
Make `offset()` independent of random access functionality

bbannier:


spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -2454,7 +2453,13 @@ void ParserBuilder::saveParsePosition() {
                      [&]() { builder()->addAssign(builder::member(state().self, ID("__begin")), state().begin); });
 
     guardFeatureCode(state().unit_id, {"uses_offset"}, [&]() {
-        builder()->addAssign(builder::member(state().self, ID("__position")), builder::begin(state().cur));
+        pushBuilder(builder()->addIf(state().begin), [&]() {
```
rsmmr, line 2456: One question: Should we be doing anything if `begin` isn't set? I'm wondering if otherwise we could end up with a previous, now stale, value in `__offset` that would be hard to track down if accidentally used in a hook?

bbannier: We basically assume here that if `offset()` is used in user code we always set `begin` in the parser state. If that assumption was violated previously we might have triggered an unset optional exception when the user called `offset`, https://github.com/zeek/spicy/blob/9adbfe592049b44e6e837a2beaba00a17db82b68/spicy/toolchain/src/compiler/codegen/codegen.cc#L198

I think setting `__offset` to a dummy value is not what we want since it would obscure that the value is bogus. If we want to have similar visibility as previously for internal inconsistencies even after this change we could change `__offset` to be an `optional<uint64>` and reset in on the else branch here (the operator impl would deref in that case). That would give the same visibility as the previous implementation, but do more work.

bbannier: Like discussed offline, since we hit the else branch pretty often I looked into whether we'd need to worry about it. For that I prefixed this `if` with a runtime `assert(state().begin)`. Looking at where we hit this it seems to very early on for a public unit when we set it up, but haven't set up any `ParserState` yet, https://github.com/zeek/spicy/blob/9adbfe592049b44e6e837a2beaba00a17db82b68/spicy/toolchain/src/compiler/codegen/parser-builder.cc#L2302

I pushed an additional patch which sets meaningful values also for these cases. With that we should always have a value for `begin` and deref it unconditionally.

rsmmr (approved):

PR #1569, opened 2023-10-09
Explicitly set ASM compiler.

bbannier:
When not setting an ASM compiler we might end up configuring Clang as C compiler, but GCC as ASM compiler. The code in `3rdparty/fiber` then causes Clang-only flags like `-Weverything` to be passed to GCC (which does not understand it).

The way we need to do this is subtle. We need to rely on `3rdparty/fiber` to explicitly `enable_language(ASM)` and not set it as a project language to work around `find_package(BISON)` breaking if `CMAKE_ASM_COMPILER` is set. By removing ASM from project languages and setting the ASM config after Bison was found, but before we configure `3rdparty/fiber` things seem to work as intended.

### DIFF
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -7,11 +7,19 @@ add_subdirectory(doctest)
 
 add_subdirectory(justrx)
 
+
+# Use the configured C compiler as ASM compiler as well. This prevents that we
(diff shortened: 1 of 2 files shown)


bbannier:
I validated manually that this fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=63066.

awelzel (approved):

PR #1572, opened 2023-10-12
Remove trimming inside individual chunks

### LINKED ISSUE #1571: Potentially expensive trimming of `Chunk` bytes
A `Chunk` currently directly wraps [containers holding its data](https://github.com/zeek/spicy/blob/4970368eff32e4e8198d7cbcc752aa085b28e132/hilti/runtime/include/types/stream.h#L91-L92). When calling `Chunk::trim` we directly trim these wrapped values, https://github.com/zeek/spicy/blob/4970368eff32e4e8198d7cbcc752aa085b28e132/hilti/runtime/src/types/stream.cc#L56 https://github.com/zeek/spicy/blob/4970368eff32e4e8198d7cbcc752aa085b28e132/hilti/runtime/src/types/stream.cc#L60

For the Zeek case we typically receive larger chunks and extract small bits from them, e.g., imagine a unit parsing a `uint8[]`. The current trim implementation makes sure that we free as much memory as possible (by removing each individual byte as soon as possible), but potentially incurs a performance overhead for that (e..g, `std::vector::erase` requires touching all bytes in the `Chunk`). The current implementation shows pathological performance for e.g., inputs generated by oss-fuzz.

We should consider using a cheaper implementation where `Chunk::trim` just moves the begin of a `Chunk`, but does not free the memory anymore.

bbannier:
Trimming `Chunk`s (always from the left) causes a lot of internal work with only limited benefit since we manage visbility with `stream::View`s on top of `Chunk`s anyway.

This patch removes trimming inside `Chunk`s so now any trimming only removes `Chunk`s from `Chain`s, but does not internally change individual `Chunk`s anymore. This might lead to slighly increased memory use, but callers usually have that data in memory anyway.

Closes #1571.

### DIFF
--- a/hilti/runtime/include/types/stream.h
+++ b/hilti/runtime/include/types/stream.h
@@ -189,8 +189,6 @@ protected:
     // only from chain so that it can track any changes.
     friend class Chain;
 
-    void trim(const Offset& o);
-
     // Update offset for current chunk and all others linked from it.
     void setOffset(Offset o) {
         auto c = this;
--- a/hilti/runtime/src/types/stream.cc
+++ b/hilti/runtime/src/types/stream.cc
@@ -47,23 +47,6 @@ Chunk::Chunk(const Offset& offset, const std::string& s) : _offset(offset) {
     }
 }
 
-void Chunk::trim(const Offset& o) {
-    assert(o >= _offset && o < _offset + size());
-    if ( auto a = std::get_if<Array>(&_data) ) {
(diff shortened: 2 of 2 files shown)


bbannier:
Benchmarking this against one of the pathological edge cases generated by oss-fuzz reduces processing time for particular input from >25s to 2s.

Benchmarking with a big internal grammar shows a small, but statistically significant performance improvement of about 2%.

rsmmr (approved):

bbannier:

PR #1578, opened 2023-10-19
Disable capturing backtraces with HILTI exceptions in non-debug builds.

### LINKED ISSUE #1565: &try / backtrack() usage too expensive due to unconditional stack unwinding
The initial implementation of the QUIC parser uses two `&try / backtrack()` constructs for parsing the initial byte and a second one for capturing all the data. During tests, the output of `perf report --no-children` indicates 8.5% of the samples fall into `_Unwind_GetTextRelBase` which seems surprising: The pcap in question produces a single analyzer.log entry only.

Concretely, processing a a 274MB pure QUIC traffic pcap with latest Zeek master (v6.1.0-dev-487-ge22bf8ebb6, RelWithDebInfo, frame pointers enabled) and latest spicy-quic parser takes ~30seconds.

Replacing the two `&try / backtrack()` constructs with `&parse-at=self.input()` and resetting the input reduces the processing time to ~24seconds. That's 20% faster for that specific pcap (and not using try/backtrack() looks actually better).

Light grepping through Spicy makes me believe that for every `&try` and `backtrack()` execution a new `Backtrace` object is instantiated, which unconditionally calls `backtrace()` and then `_Unwind_GetTextRelBase`. For the `try / backtrack()` mechanism that is unproductive work. Flamegraphs before/after below. They look quite bit different, so wonder if the backtrack() also triggers something with the fiber execution.

Anyway, `try/backtrack()` should probably not come with such a huge overhead or otherwise cause a warning with discouragement and alternatives to use instead.

---

Flamegraphs
### try-backtrack (before)
try-backtrack

### parse-at (after)
parse-at

rsmmr:
They can be expensive to capture, and aren't used anywhere by default
unless explicitly requested.

Closes #1565.


hilti/runtime/include/exception.h
```diff
@@ -66,7 +92,7 @@ class Exception : public std::runtime_error {
 
     std::string _description;
     std::string _location;
-    Backtrace _backtrace;
+    std::unique_ptr<Backtrace> _backtrace; // null if not available
```
bbannier, line 95: Unless there is a clear performance win when going with allocating this on the heap I'd say go with `std::optional` instead here. That would avoid having to explicitly deal with `std::unique_ptr` not being copyable which IMO is not the semantics we want for an `Exception`.

hilti/runtime/include/exception.h
```diff
@@ -24,18 +25,43 @@ class Exception : public std::runtime_error {
...
+        : std::runtime_error(other), _description(other._description), _location(other._location) {
+        if ( other._backtrace )
+            _backtrace = std::make_unique<Backtrace>(*other._backtrace);
+    }
+
     Exception(Exception&&) noexcept = default;
-    Exception& operator=(const Exception&) = default;
+    Exception& operator=(const Exception& other) {
+        // Note: We don't copy the backtrace here, as it's not clear what
+        // that would mean.
```
bbannier, line 55: Shouldn't an exception be copyable, including its full state? I cannot see where outside of implementation details we care about the backtrace being unique.

bbannier (approved):
I wonder if we shouldn't instead use a `std::optional` to capture the backtraces. This increases space requirements slightly, but I would guess shouldn't be perform horribly. I'll push a fixup to this branch.

awelzel (approved):

PR #1590, opened 2023-11-10
Reduce string allocations for literals [skip CI]

### LINKED ISSUE #1589: Significant std::string overhead for generated waitForInput(...,error, location,...) calls
It's a micro-benchmark, but I believe there's something to it.

Compiling the reproducer `t.spicy` with its absolute filename (assuming an appropriate long parent directory) results in upwards of 10% slower parsing performance due to short-string-optimization of an immediate constructed `std::string` instance not being in effect. Not constructing any immediate `std::string` instances should help parsing performance even more. Particularly if they are not being used and not variable.

Reproducer at the bottom, executed within the latest `zeek/zeek-dev` container, turbo boost disabled.

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `cat test-data.txt \| spicy-driver t.hlto` | 7.324 ± 0.063 | 7.254 | 7.376 | 1.00 |
| `cat test-data.txt \| spicy-driver t-compiled-with-realpath.hlto` | 8.310 ± 0.074 | 8.246 | 8.392 | 1.13 ± 0.01 |

Background: Robin gave a demo showing generated C++ code of Spicy units and what tickled me were the two immediate std::string instances constructed for every call to `waitForInput()` for uint16 parsing:

```
:spicy::rt::detail::waitForInput(__data, __cur, ::hilti::rt::integer::safe<std::uint64_t>{2U}, std::string("expecting 2 bytes for unpacking value"), std::string("t.spicy:4:6"), ::hilti::rt::StrongReference<::spicy::rt::filter::detail::Filters>());
```

Compiling with an absolute path makes the second string exceed 15 bytes and therefore thwarts short-string optimization, causing an extra heap allocation for every `waitForInput()` call. The "expecting 2 bytes for unpacking value" also does an extra heap allocation that should probably be avoided.

---

Reproducer:
```
#!/bin/bash
set -eux

cat <<EOF > t.spicy
module Test;

type Bar = unit {
  g: uint16;
  h: uint16;
  i: uint16;

  j: uint16;
  k: uint16;
  l: uint16;

  m: uint16;
  n: uint16;
  o: uint16;

  p: uint16;
  q: uint16;
  r: uint16;
};

public type Foo = unit {
  : Bar[] &eod;
};
EOF

export HILTI_CXX_COMPILER_LAUNCHER=ccache

spicyc -j t.spicy -o t.hlto
# Encodes the full-filename into generated code, likely 15+ bytes.
spicyc -j $(realpath t.spicy) -o t-compiled-with-realpath.hlto

python3 -c 'print("xxxxxx" * 10000000, end="")' > test-data.txt
ls -lha test-data.txt

hyperfine --export-markdown results.md -M3 'cat test-data.txt | spicy-driver t.hlto' 'cat test-data.txt | spicy-driver t-compiled-with-realpath.hlto'
```

### LINKED ISSUE #1591: std::string parameters for debug::indent() and debug::dedent() in generated code
Separate follow up to #1589 . Looking at the generate code for the following grammar:

```
type K = unit {
  x: skip bytes &size=1;
};

type L = unit {
  k: K;
};

type M = unit {
  l: L;
};

public type N = unit {
  : M[] &eod;
};
```

There's a number of `indent()` and `dedent()` calls generated involving `std::string` instances being constructed at runtime. These are pretty short strings so they shouldn't involve an allocation and might be optimized well, but might still be sensible to not even construct them.
```
    try {
        ::hilti::rt::debug::indent(std::string("spicy"));
        std::optional<::hilti::rt::stream::SafeConstIterator> __begin = std::make_optional(__cur.begin());


    ::hilti::rt::debug::dedent(std::string("spicy"));
```

A flamegraph indicates ~3.3% samples falling into basic_string functions.

flame

Screenshot-20231110-120557

Unfortunately I don't have the means to provide before-after here.

bbannier:
With this PR the benchmark from #1589 shows identical performance for both cases for me. Overall runtime seems to improve as well.

Before:

```console
$ hyperfine --export-markdown results.md -w3 'cat test-data.txt | ./build/bin/spicy-driver t.hlto' 'cat test-data.txt | ./build/bin/spicy-driver t-compiled-with-realpath.hlto'
Benchmark 1: cat test-data.txt | ./build/bin/spicy-driver t.hlto
  Time (mean ± σ):      1.713 s ±  0.018 s    [User: 1.699 s, System: 0.031 s]
  Range (min … max):    1.695 s …  1.738 s    10 runs
 
Benchmark 2: cat test-data.txt | ./build/bin/spicy-driver t-compiled-with-realpath.hlto
  Time (mean ± σ):      2.166 s ±  0.014 s    [User: 2.141 s, System: 0.041 s]
  Range (min … max):    2.147 s …  2.191 s    10 runs
 
Summary
  cat test-data.txt | ./build/bin/spicy-driver t.hlto ran
    1.26 ± 0.02 times faster than cat test-data.txt | ./build/bin/spicy-driver t-compiled-with-realpath.hlto
```

After:

```console
$ hyperfine --export-markdown results.md -w3 'cat test-data.txt | ./build/bin/spicy-driver t.hlto' 'cat test-data.txt | ./build/bin/spicy-driver t-compiled-with-realpath.hlto'
Benchmark 1: cat test-data.txt | ./build/bin/spicy-driver t.hlto
  Time (mean ± σ):      1.218 s ±  0.008 s    [User: 1.204 s, System: 0.029 s]
  Range (min … max):    1.209 s …  1.232 s    10 runs
 
Benchmark 2: cat test-data.txt | ./build/bin/spicy-driver t-compiled-with-realpath.hlto
  Time (mean ± σ):      1.219 s ±  0.005 s    [User: 1.202 s, System: 0.031 s]
  Range (min … max):    1.215 s …  1.231 s    10 runs
 
Summary
  cat test-data.txt | ./build/bin/spicy-driver t.hlto ran
    1.00 ± 0.01 times faster than cat test-data.txt | ./build/bin/spicy-driver t-compiled-with-realpath.hlto
```

Closes #1589.
Closes #1591.

### DIFF
--- a/hilti/runtime/include/debug-logger.h
+++ b/hilti/runtime/include/debug-logger.h
@@ -24,14 +24,15 @@ public:
     bool isEnabled(std::string_view stream) { return _streams.find(stream) != _streams.end(); }
 
     void indent(std::string_view stream) {
-        if ( isEnabled(stream) )
-            _streams[stream] += 1;
+        if ( auto s = _streams.find(stream); s != _streams.end() ) {
+            auto& indent = s->second;
+            indent += 1;
+        }
     }
 
     void dedent(std::string_view stream) {
-        if ( isEnabled(stream) ) {
-            auto& indent = _streams[stream];
-
+        if ( auto s = _streams.find(stream); s != _streams.end() ) {
+            auto& indent = s->second;
             if ( indent > 0 )
                 indent -= 1;
         }
--- a/hilti/runtime/include/debug-logger.h
+++ b/hilti/runtime/include/debug-logger.h
@@ -6,7 +6,7 @@
 #include <map>
 #include <memory>
 #include <optional>
-#include <string>
+#include <string_view>
 
 #include <hilti/rt/filesystem.h>
 #include <hilti/rt/util.h>
@@ -18,17 +18,17 @@ class DebugLogger {
 public:
     DebugLogger(hilti::rt::filesystem::path output);
 
-    void print(const std::string& stream, const std::string& msg);
-    void enable(const std::string& streams);
+    void print(std::string_view stream, std::string_view msg);
+    void enable(std::string_view streams);
 
-    bool isEnabled(const std::string& stream) { return _streams.find(stream) != _streams.end(); }
+    bool isEnabled(std::string_view stream) { return _streams.find(stream) != _streams.end(); }
 
-    void indent(const std::string& stream) {
+    void indent(std::string_view stream) {
         if ( isEnabled(stream) )
             _streams[stream] += 1;
     }
 
-    void dedent(const std::string& stream) {
+    void dedent(std::string_view stream) {
         if ( isEnabled(stream) ) {
             auto& indent = _streams[stream];
 
@@ -41,7 +41,7 @@ private:
     hilti::rt::filesystem::path _path;
     std::ostream* _output = nullptr;
     std::unique_ptr<std::ofstream> _output_file;
-    std::map<std::string, integer::safe<uint64_t>> _streams;
+    std::map<std::string_view, integer::safe<uint64_t>> _streams;
 };
 
 } // namespace hilti::rt::detail
--- a/hilti/runtime/include/logging.h
+++ b/hilti/runtime/include/logging.h
@@ -2,7 +2,7 @@
 
 #pragma once
 
-#include <string>
+#include <string_view>
 #include <utility>
 
 #include <hilti/rt/debug-logger.h>
@@ -18,10 +18,10 @@ namespace hilti::rt {
  * anything that can happen during "normal" operation (which is almost
  * everything).
  */
-void fatalError(const std::string& msg) __attribute__((noreturn));
+void fatalError(std::string_view msg) __attribute__((noreturn));
 
 /** Reports a warning. */
-void warning(const std::string& msg);
+void warning(std::string_view msg);
 
 /**
  * Prints a string, or a runtime value, to a specific debug stream. This is a
@@ -42,39 +42,39 @@ namespace debug {
 
 namespace detail {
 /** Prints a debug message to a specific debug stream. */
-inline void print(const std::string& stream, const char* msg) {
+inline void print(std::string_view stream, const char* msg) {
     if ( ::hilti::rt::detail::globalState()->debug_logger )
         ::hilti::rt::detail::globalState()->debug_logger->print(stream, msg);
 }
 
 /** Print a string to a specific debug stream with proper escaping. */
-inline void print(const std::string& stream, const std::string_view& s) {
+inline void print(std::string_view stream, std::string_view s) {
     if ( ::hilti::rt::detail::globalState()->debug_logger )
         ::hilti::rt::detail::globalState()->debug_logger->print(stream, hilti::rt::escapeBytes(s, false));
 }
 
 template<typename T, typename std::enable_if_t<not std::is_convertible<T, std::string_view>::value>* = nullptr>
 /** Prints the string representastion of a HILTI runtime value to a specific debug stream. */
-inline void print(const std::string& stream, const T& t) {
+inline void print(std::string_view stream, const T& t) {
     if ( ::hilti::rt::detail::globalState()->debug_logger )
         ::hilti::rt::detail::globalState()->debug_logger->print(stream, hilti::rt::to_string_for_print(t));
 }
 } // namespace detail
 
 /** Returns true if debug logging is enabled for a given stream. */
-inline bool isEnabled(const std::string& stream) {
+inline bool isEnabled(std::string_view stream) {
     return ::hilti::rt::detail::globalState()->debug_logger &&
            ::hilti::rt::detail::globalState()->debug_logger->isEnabled(stream);
 }
 
 /** Increases the indentation level for a debug stream. */
-inline void indent(const std::string& stream) {
+inline void indent(std::string_view stream) {
     if ( ::hilti::rt::detail::globalState()->debug_logger )
         ::hilti::rt::detail::globalState()->debug_logger->indent(stream);
 }
 
 /** Decreases the indentation level for a debug stream. */
-inline void dedent(const std::string& stream) {
+inline void dedent(const std::string_view stream) {
     if ( ::hilti::rt::detail::globalState()->debug_logger )
         ::hilti::rt::detail::globalState()->debug_logger->dedent(stream);
 }
@@ -103,7 +103,7 @@ inline void setLocation(const char* l = nullptr) { detail::tls_location = l; }
  * arguments if nothing is going to get logged.
  */
 template<typename T>
-inline void print(const std::string& stream, T&& msg) {
+inline void print(std::string_view stream, T&& msg) {
     if ( ::hilti::rt::detail::globalState()->debug_logger &&
          ::hilti::rt::detail::globalState()->debug_logger->isEnabled(stream) )
         ::hilti::rt::debug::detail::print(stream, std::forward<T>(msg));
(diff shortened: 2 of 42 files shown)


bbannier:
The remaining work here is dealing with the fallout from type mismatches from `const char*` vs `std::string`. This is likely exactly why we introduced the explicit typing. I picked `const char*` instead of `std::string_view` since `const char*` can implicitly convert to `std::string`, but maybe introducing a dedicated type would give us more control.

bbannier:
I slightly reworked this since while we can emit C++ literals for generated (effectively static) strings user-specified literals should still be emitted as mutable strings (e.g., so users can call mutating functions on them). The way I now detect the difference is to check whether a literal has a location.

hilti/toolchain/src/compiler/codegen/ctors.cc
```diff
@@ -181,7 +181,21 @@ struct Visitor : hilti::visitor::PreOrder<cxx::Expression, Visitor> {
...
-    result_t operator()(const ctor::String& n) { return fmt("std::string(\"%s\")", util::escapeUTF8(n.value(), true)); }
+    result_t operator()(const ctor::String& n) {
+        // We emit string constructors for both user-provided literals and
+        // generated strings, e.g., static error messages or identifiers. Only
+        // user-provided literals need to behave like actual our runtime string
+        // type `std::string` since they e.g., need to be usable with functions
+        // expecting that type or might need to be mutable; generated string
+        // OTOH are static and only used with our internal APIs which we can
+        // make usable with actual literals, and we can avoid allocating owning
+        // mutable strings for them.
```
bbannier, line 192: @rsmmr This is a lot of magic here, not sure you feel about this. The upside is that we can keep using `builder::string` without having to think too much about what we stringify. The alternative could be to e.g., introduce a separate builder for generated literals.

bbannier:
I added a couple of other perf cleanups to this PR, please let me know if I should pull them into their own PR.

hilti/runtime/include/logging.h
```diff
@@ -42,39 +42,39 @@ namespace debug {
...
 /** Prints a debug message to a specific debug stream. */
-inline void print(const std::string& stream, const char* msg) {
+inline void print(std::string_view stream, const char* msg) {
     if ( ::hilti::rt::detail::globalState()->debug_logger )
         ::hilti::rt::detail::globalState()->debug_logger->print(stream, msg);
 }
 
 /** Print a string to a specific debug stream with proper escaping. */
-inline void print(const std::string& stream, const std::string_view& s) {
+inline void print(std::string_view stream, const std::string_view& s) {
```
rsmmr, line 51: General question: The new `string_view` are not const, does it help in any way to make them const where we can? Plus, the 2nd argument `s` here remains a `const` `string_view`. If generally use non-const, adapt for consistency.

hilti/runtime/include/debug-logger.h
```diff
@@ -24,14 +24,15 @@ class DebugLogger {
     bool isEnabled(std::string_view stream) { return _streams.find(stream) != _streams.end(); }
 
     void indent(std::string_view stream) {
-        if ( isEnabled(stream) )
-            _streams[stream] += 1;
+        if ( auto s = _streams.find(stream); s != _streams.end() ) {
```
rsmmr, line 27: How about adding a quick check upfront if `_streams` is empty? Or is `find()` good enough in that case? (Same in `dedent`.)

hilti/toolchain/src/compiler/codegen/ctors.cc
rsmmr: Yeah, not feeling very good about this honestly. That'd be very hard to remember and track down if anything goes wrong either way. Either a separate builder or make it explicit somehow in the meta data that's not from a user (i.e., not just non-existence of location). I think I'd indeed prefer the former, as you already suggest.

rsmmr (commented):
Main comment is that "magic" part: not a fan. Rest looks good.

hilti/runtime/include/logging.h
bbannier: Re:constness, I'd be surprised since we already pass by value so it should make no difference, but I'll check. In general `std::string_view` is designed to be very cheap to construct and copy and if used with literals often leads to the literal put into the `.TEXT` segment, e.g., for the following code both Clang and GCC seem to put `abc` into `.TEXT`.

```cpp
int foo() {
    std::string_view x{"abc"};
    puts(x.data());
}
```

The second argument was an oversight, will fix (a guideline is that `std::string_view` should always be passed by copy).

bbannier: I checked the performance of passing both args to this function as `const` values, and like I suspected it makes no difference. Checking the assembly both should produce exactly the same code.

hilti/runtime/include/debug-logger.h
bbannier: I looked into the impl of `find` with the libc++ I have on my system. There `find` uses `lower_bound` which already short-circuits for an empty list, so I'd prefer to keep this as is.

hilti/toolchain/src/compiler/codegen/ctors.cc
bbannier: I cleaned this up somewhat to track in a `ctor::String` whether it was a literal or we'd need to allocate. With what I have now we'd need to make a deliberate choice between `builder::string` and `builder::string_literal`. As we move more internal methods over to accept `string_view` incorrectly using `builder::string` with generated literals would not compile anymore (this is good), but we could still accidentally slip in `builder::string` instead of `builder::string_literal` so we'd do more work.

hilti/toolchain/include/ast/ctors/string.h
```diff
@@ -35,6 +37,7 @@ class String : public NodeBase, public hilti::trait::isCtor {
 
 private:
     std::string _value;
+    bool _isLiteral = false;
```
rsmmr, line 40: Add `is_literal` to `properties()` as well.

rsmmr (dismissed):

hilti/toolchain/src/compiler/codegen/ctors.cc
bbannier: I went with `builder::string_mut`. I also added a new `builder::string` which is immediately marked deprecated, but forwards to `builder::string_mut`.

spicy/runtime/src/sink.cc
```diff
@@ -1,5 +1,7 @@
 // Copyright (c) 2020-2023 by the Zeek Project. See LICENSE for details.
 
+#include "spicy/rt/sink.h"
```
rsmmr, line 3: #include <spicy/rt/sink.h>

rsmmr (approved):

spicy/runtime/src/sink.cc
bbannier: Even though we still do not do this consistently, I would really like to include this with `#include "..."` to make clear that this is [the header related to this .cc file](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes). With that we e.g., get it to sort before other headers which helps make sure that the header is self-contained.

hilti/runtime/include/fiber.h
```diff
@@ -115,8 +115,8 @@ struct StackBuffer {
 
 private:
     const ::Fiber* _fiber;
-    void* _buffer = nullptr; // allocated memory holding swapped out stack content
-    size_t _buffer_size = 0; // amount currently allocated for `_buffer`
+    alignas(16) void* _buffer = nullptr; // allocated memory holding swapped out stack content
```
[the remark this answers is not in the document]
bbannier: That might be the reason I see no effect from this (alternatively: might be something does line up with how the stack on the fiber side is aligned inside the 3rdparty dependency). I'll just drop the patch adding this for now.

spicy/runtime/src/sink.cc
rsmmr: I think we had this discussion a few times already. :-) I won't fight it, I just don't like the inconsistency of doing it randomly for a few implementation files, but not most others.

PR #1604, opened 2023-11-30
RFC: Switch to use of individual stacks for fiber execution.

bbannier:
We would previously use shared stacks to minimize memory overhead from pessimisitcally giving each fiber enough stack space if they were running on individual fibers. This causes overhead since when a fiber suspends or resumes it stack needs to be saved and restored (at least: memcpy of the whole used stack).

On many (most?) systems the `malloc` used by our fiber dependency to allocate memory for fiber stacks is lazy though so the overhead is not really a concern there (we might increase the virtual, but not the resident memory). I see a lazy `malloc` on macos-13, and this also seems to be the default on Linux^[1]. This means that we can avoid the overhead from `memcpy` on these systems.

This patch switches fibers to use individual stacks instead.

---
^[1]: https://www.kernel.org/doc/Documentation/vm/overcommit-accounting

### DIFF
--- a/hilti/runtime/src/fiber.cc
+++ b/hilti/runtime/src/fiber.cc
@@ -21,8 +21,8 @@ using namespace hilti::rt;
 
 #ifndef HILTI_HAVE_ASAN
 // Defaults for normal operation.
-static const auto DefaultFiberType = detail::Fiber::Type::SharedStack; // share stack by default
-static const auto AlwaysUseStackSwitchTrampoline = false;              // use switch trampoline only with shared stacks
+static const auto DefaultFiberType = detail::Fiber::Type::IndividualStack; // individual stack by default
+static const auto AlwaysUseStackSwitchTrampoline = false; // use switch trampoline only with shared stacks
 static const auto FiberGuardFlags = FIBER_FLAG_GUARD_LO | FIBER_FLAG_GUARD_HI;
 
 #define ASAN_NO_OPTIMIZE // just leave empty


bbannier:
Superseded by #1604.

hilti/runtime/include/configuration.h
```diff
@@ -18,7 +18,7 @@ struct Configuration {
     Configuration();
 
     /** Stack size for fibers with individual stacks. */
-    size_t fiber_individual_stack_size = static_cast<size_t>(1 * 1024 * 1024);
+    size_t fiber_individual_stack_size = static_cast<size_t>(256 * 1024 * 1024);
```
bbannier, line 21: On platforms with overcommitment completely disabled this can cause serious memory issues now, especially given the increased size. I wonder whether we should (e.g., on HILTI rt init) check whether overcommitment is enabled, and if not either fall back to shared stacks, or emit a warning.

bbannier: I reverted back to a size of 1MB since e.g., even Cirrus seems to disable memory overcommittment (that's at least how I interpreted the test suite hanging on practically all machines). Back at 1MB it now seems to pass again.

I pull fiber stats from a large internal Zeek analyzer we have at Corelight, and it looks like that analyzer had at most a less than 100 fibers active at a time, so individual stacks cost us worst case 100MB there as well (though in reality: much less due to overcommittment).

bbannier:
In https://github.com/zeek/spicy/pull/1603#issuecomment-1834080505:
> > We also increase the memory allocated for individual fiber stacks from 1MB to 256MB which should allow most parsers to run without running out of stack space
> 
> One question (while applauding the intent!): does this mean that if there are say 10K concurrent connections, there will be 10K separate stacks allocated, one per connection parser? I'm wondering about overwhelming the VM data structures, even if very little winds up being actually allocated.

Yes, that is exactly what would happen in the case @vpax. The current behavior with shared stacks somewhat protects against this, but comes at a cost (around 10-20% overhead). There would also be the possibility to make this a runtime as opposed to compile time flag, but the issue still remains that if something fails due to misbehaving individual stacks there would be little one could do to recover from it beyond crashing.

awelzel:
[I like the 10000 concurrent connections/fibers scenario, with lets say 1 packet per second per connection coming in]

>> I'm wondering about overwhelming the VM data structures, even if very little winds up being actually allocated.
> ...
> The current behavior with shared stacks somewhat protects against this, but comes at a cost (around 10-20% overhead)

The overhead likely depends much on the fiber stack size, no? Not sure about the effective and max stack sizes in the experiments, but expect the overhead significantly higher if stack sizes were 2x, 4x or 10x.

> There would also be the possibility to make this a runtime as opposed to compile time flag, but the issue still remains that if something fails due to misbehaving individual stacks there would be little one could do to recover from it beyond crashing.

If not crashing, to me sounds like the shared stack approach would just be really slow with large stacks. In the 10000 fiber scenario, each growing to an immense 128KB stack, doing 10000 fiber switches per second would result in a constant "non-productive" copying of memory at ~2.4GByte/s (2*(128*1024)*10000)). This seems already a lot of work - compare with 10k packets/s at 1.5k packet size being a mere 14MByte/s. Assuming this worst-case happens in all workers on a system, say 64, at the same time: not sure this is actually manageable.

awelzel:
Hmm, hmm, hmm. Maybe shared stacks was all about the VM structures? Robin would probably know.

Using a pcap with [50k concurrent TCP connections](https://github.com/awelzel/pcaps/raw/master/50k-tcp-conns.pcap.gz) and forcing the existing finger analyzer on it (all connections have resp_p 7000/tcp) with individual stacks this immediately aborts execution with the following trace. `strace` indicates `mprotect()` fails with ENOMEM. Setting `vm.overcommit_memory=1` doesn't help and reducing the stack sizes doesn't either.

Setting `FiberGuardFlags = 0` to skip the guard pages - avoiding `mprotect()` calls allows it to work. Then it runs some 5-10% faster compared to shared stacks. Suspect stacks for the finger analyzer are very small and the traffic isn't matching anyhow, so that number should be taken with a grain of salt.

Ugh. Seems the memcpy() overhead is annoying and difficult to predict while the individual stack approach will easily exhaust some resource. Neither choice appears great.


```
$ /usr/bin/time zeek -b -r /media/awelzel/T7/pcaps/50k-tcp-conns.pcap scripts/base/protocols/finger/ -e 'event zeek_init() { Analyzer::register_for_port(Analyzer::ANALYZER_FINGER, 7000/tcp); }'
[libhilti] Internal error: could not allocate individual-stack fiber

--- Aborting in libhilti
# [0x7fb77cdc2471] hilti::rt::Backtrace::Backtrace()
# [0x7fb77ce1868e] hilti::rt::abort_with_backtrace()
# [0x7fb77cde0e4b] hilti::rt::internalError(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
# [0x7fb77cdc947c] hilti::rt::detail::Fiber::Fiber(hilti::rt::detail::Fiber::Type)
# [0x7fb77cdcd997] hilti::rt::detail::Fiber::create()
# [0x564aadaad6c0] hlt_finger::Finger::Reply::parse1(hilti::rt::ValueReference<hilti::rt::Stream>&, std::optional<hilti::rt::stream::View> const&, std::optional<spicy::rt::UnitContext> const&)
...
```

bbannier:
I am closing this as this is not the direction we want to move in in general.

PR #1607, opened 2023-12-03
Replace Chunk's variant with std::string

awelzel:
@bbannier - here's some commits. For some pcap and parser combination I've tested with, this removes an addition ~5% relative to your branch - probably won't translate.

Pick what makes sense to you. Feel free to redo commits as you see fit and then close this PR :-)

### DIFF
--- a/hilti/runtime/include/types/stream.h
+++ b/hilti/runtime/include/types/stream.h
@@ -1478,7 +1478,7 @@ public:
      * Creates an instance from a bytes instance.
      * @param d `Bytes` instance to the create the stream from
      */
-    explicit Stream(const Bytes& d);
+    explicit Stream(Bytes d);
 
     /**
      * Creates an instance from an existing memory block. The data
--- a/hilti/runtime/include/types/stream.h
+++ b/hilti/runtime/include/types/stream.h
@@ -87,22 +87,11 @@ struct Gap {
  */
 class Chunk {
 public:
-    static const int SmallBufferSize = 32;
-    using Array = std::pair<Size, std::array<Byte, SmallBufferSize>>;
-    using Vector = std::vector<Byte>;
-
-    Chunk(const Offset& o, std::array<Byte, SmallBufferSize> d, const Size& n)
-        : _offset(o), _data(std::make_pair(n, d)) {}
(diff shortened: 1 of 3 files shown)


hilti/runtime/include/logging.h
```diff
@@ -36,7 +36,8 @@ void warning(std::string_view msg);
     }
 
 /** Shortcut to `hilti::rt::debug::setLocation`. */
-#define __location__(x) ::hilti::rt::debug::detail::tls_location = x;
+// #define __location__(x) ::hilti::rt::debug::detail::tls_location = x;
```
bbannier, line 39: This is needed for location reporting for exceptions, so if we wanted to get rid of it we would need to introduce some other, cheaper mechanism.

That being said, I look into what overhead using thread-local storage introduces, and it seems moderate and on a fraction of the total cost (practically just a memory access).

awelzel:
I've tested this with a large analyzer after rebasing on top of the latest Spicy version and using the newest Zeek version, too. Testing indicates it shapes off 3% to 4% of runtime, though likely depends on the input, too.

```
  Time (mean _ _):     20.883 s _  0.178 s    [User: 19.962 s, System: 0.765 s]
  Range (min _ max):   20.670 s _ 21.217 s    10 runs
 ```

After:
```
  Time (mean _ _):     20.134 s _  0.133 s    [User: 19.198 s, System: 0.757 s]
  Range (min _ max):   19.951 s _ 20.361 s    10 runs
```

I'd like to hand this off though as I suspect its current form isn't acceptable for the main tree. Either by creating a ticket or leaving this PR open. Thoughts?

bbannier:
Sorry for dropping the ball on this @awelzel, I think having something like this would be great (if only to get rid of our own SSO in `Chunk`). Let me take this over.

rsmmr (approved):
Certainly makes it easier, and if it makes it faster too, that's a win-win. :-)

bbannier:
> Certainly makes it easier, and if it makes it faster too, that's a win-win. :-)

I benchmarked this with a huge internal parser, and it definitely does not make the code slower. With this patch I see runtime relative to the unmodified baseline "reduced" to (0.97 ± 0.03).

PR #1609, opened 2023-12-05
Allow unsafe Vector iteration over underlying std::vector.

awelzel:
Iterating over hilti::Vector is quite expensive due to the weak_ptr::lock() calls. For unsafe uses, allow access to the underlying std::vector for faster raw iteration speed.

---

Change on the Zeek side: https://github.com/zeek/zeek/pull/3491


bbannier (approved):
Looks good to me, and similar to what we do for other containers.

PR #1610, opened 2023-12-06
Display floats in scientific notation in profiler report.

bbannier:
We previously would display floats as decimals with two decimals in the profiler report. While this makes it easy to get a quick overview of the magnitude from directly looking at the logs, in real parsers many more components might be involved which each only contribute less than 1%; such contributions are currently displayed as 0.00.

This patch switches display of floating point numbers in the profiler report to use scientific notation so we do not loose information for small fractions.

### DIFF
--- a/tests/Baseline/hilti.rt.profiler/output
+++ b/tests/Baseline/hilti.rt.profiler/output
@@ -2,4 +2,4 @@
 #name count avg-% total-%
 hilti/func/Foo::fibo 192
 hilti/func/Foo::y 95
-hilti/total 1 100.00 100.00
+hilti/total 1 1.00e+02 1.00e+02
--- a/hilti/runtime/src/profiler.cc
+++ b/hilti/runtime/src/profiler.cc
@@ -89,7 +89,7 @@ std::optional<Measurement> profiler::get(const std::string& name) {
 
 void profiler::report() {
     static const auto fmt_header = "#%-49s %10s %10s %10s %10s\n";
-    static const auto fmt_data = "%-50s %10" PRIu64 " %10" PRIu64 " %10.2f %10.2f \n";
+    static const auto fmt_data = "%-50s %10" PRIu64 " %10" PRIu64 " %10.2e %10.2e\n";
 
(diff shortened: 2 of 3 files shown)


tests/Baseline/hilti.rt.profiler/output
```diff
@@ -2,4 +2,4 @@
 #name count avg-% total-%
 hilti/func/Foo::fibo 192
 hilti/func/Foo::y 95
-hilti/total 1 100.00 100.00
+hilti/total 1 1.00e+02 1.00e+02
```
rsmmr, line 5: This feels much harder to read. What about instead increasing the precision for the fractional part?

bbannier: I agree that this reads less nicely. I think no matter how much we increase the number of digits we show we will still loose data at some point with that approach for small fractions. With the current output format I see ~1/3 of entries with `tot` serialized as `0.00`. If we want to avoid exponents (which requires slightly more work to compare magnitudes), what do you think about e.g., just showing all double digits?

Overall my gut feeling is that since such output has so much data in it, users would always want to write quick ad-hoc tooling to look at it (I used pandas after converting to csv).

rsmmr: Do those entries really matter? If yes, then how about adding a new mode that provides an alternative output format for parsing, like CSV with full precision?

bbannier: Good question. The perf output I am looking at ATM has >1500 entries, so I hadn't really consider this anything but a "full format" which can only be consumed with additional tooling 😅

rsmmr: For me, that tooling has so far been `grep/awk/sort`, and in the end I still look at it. :-)

bbannier: Fair enough, and if that is the consensus I am totally for keeping the current format. I initially wanted this to fix `avg` which mostly showed `0.00`, but that is easy to compute with better precision offline.

PR #1612, opened 2023-12-07
Vector/Bytes: Lazily allocate shared_ptr _control

awelzel:
Vector and Bytes currently allocate and de-allocated a shared_ptr as a "control block" unconditionally.

This results in an extra malloc and free as well as atomic instructions during destruction. This control block is only required when "safely" iterating, but that may never happen.

Applying this change and testing with a fairly large analyzer, execution time is reduced from 20.6 seconds to ~17.3~ 19.5 seconds, some ~16%~ 5%.

A further step might be to implement a custom control block class to avoid the atomic instructions altogether. Also, the StrongReferene and ValueReference use shared_ptr, showing significantly in profiles. Their usage should likely be revisited, too.

---

mutable and const_cast. Happy to hear how to do that nicer :-)

### DIFF
--- a/hilti/runtime/include/types/bytes.h
+++ b/hilti/runtime/include/types/bytes.h
@@ -48,17 +48,18 @@ class Iterator {
     using B = std::string;
     using difference_type = B::const_iterator::difference_type;
 
-    std::weak_ptr<B*> _control;
+    std::weak_ptr<const B*> _control;
     typename integer::safe<std::uint64_t> _index = 0;
 
 public:
     Iterator() = default;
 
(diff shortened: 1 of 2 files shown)


awelzel:
> Applying this change and testing with a fairly large analyzer, execution time is reduced from 20.6 seconds to 17.3 seconds, some 16%.

Nope, something broke. This is likely off.

awelzel:
> > Applying this change and testing with a fairly large analyzer, execution time is reduced from 20.6 seconds to 17.3 seconds, some 16%.
>
> Nope, something broke. This is likely off.

It's only 20.6 vs 19.5 seconds, so IMO still worth, but not amazing.

bbannier (dismissed):
Thanks you, even with the corrected impl this is still great.

PR #1616, opened 2023-12-08
Optimize C++ tuple element coercions

### LINKED ISSUE #1615: Inefficient code generation for named tuple returns
The following "mktuple()" functions produce very different code. Notice the repeated and unproductive `make_tuple(...)` invocations for each of the offsets for the named variants.  Large named tuples cause unnecessary compile and runtime overhead. With bytes elements, this likely involves many unnecessary memory allocations and `memcpy` calls.

---

```
module Test;

function named_mktuple(p: point): tuple<x: uint8, y: uint8, z: uint8> {
  return (p.x, p.y, p.z);
}

function unnamed_mktuple(p: point): tuple<uint8, uint8, uint8> {
  return (p.x, p.y, p.z);
}

public type point = unit {
  x: uint8;
  y: uint8;
  z: uint8;

  on %done {
    print named_mktuple(self);
    print unnamed_mktuple(self);
  }
};
```

```
static auto __hlt::Test::unnamed_mktuple(const ::hilti::rt::ValueReference<__hlt::Test::point>& p)
    -> std::tuple<::hilti::rt::integer::safe<uint8_t>, ::hilti::rt::integer::safe<uint8_t>,
                  ::hilti::rt::integer::safe<uint8_t>> {
    ::hilti::rt::detail::checkStack();
    __location__("tuple-return.spicy:8:3");
    return std::make_tuple(::hilti::rt::optional::value((*p).x), ::hilti::rt::optional::value((*p).y),
                           ::hilti::rt::optional::value((*p).z));
}
```

```
static auto __hlt::Test::named_mktuple(const ::hilti::rt::ValueReference<__hlt::Test::point>& p)
    -> std::tuple<::hilti::rt::integer::safe<uint8_t>, ::hilti::rt::integer::safe<uint8_t>,
                  ::hilti::rt::integer::safe<uint8_t>> {
    ::hilti::rt::detail::checkStack();
    __location__("tuple-return.spicy:4:3");
    return std::make_tuple(std::get<0>(std::make_tuple(::hilti::rt::optional::value((*p).x),
                                                       ::hilti::rt::optional::value((*p).y),
                                                       ::hilti::rt::optional::value((*p).z))),
                           std::get<1>(std::make_tuple(::hilti::rt::optional::value((*p).x),
                                                       ::hilti::rt::optional::value((*p).y),
                                                       ::hilti::rt::optional::value((*p).z))),
                           std::get<2>(std::make_tuple(::hilti::rt::optional::value((*p).x),
                                                       ::hilti::rt::optional::value((*p).y),
                                                       ::hilti::rt::optional::value((*p).z))));
}
```

bbannier:
This patch introduces an optimization to pass C++ tuples through if they
do not need C++-side coercion. This emits better code if a tuple ctor
already has the right types. We also optimize codegen if we are coercing
from a temporary tuple in that we now capture it in a temporary and
coerce elements from that instead of repeatedly emitting the same tuple
ctor (we scale with the number of tuple elements now not its square
anymore).

Closes #1615.


hilti/toolchain/src/compiler/codegen/coercions.cc
```diff
@@ -163,11 +163,27 @@ struct Visitor : public hilti::visitor::PreOrder<cxx::Expression, Visitor> {
             std::vector<cxx::Expression> exprs;
 
             assert(src.elements().size() == t->elements().size());
-            for ( auto i = 0U; i < src.elements().size(); i++ )
+
+            // Check if a coercion is needed at all.
+            bool allSameTypes = true;
```
rsmmr, line 168:
```suggestion
            bool all_same_types = true;
```

rsmmr (approved):

PR #1619, opened 2023-12-12
Fix handling of `%synchronize-*` attributes for units in lists.

### LINKED ISSUE #1617: Use of `%synchronize-at` unexpectedly rejected
One would expect that the following code can be used to specify a literal to synchronize on independently of usual lookahead parsing:

```ruby
module sync;

public type PDUs = unit {
    : (PDU &synchronize)[];
    on %synced { confirm; }
};

type PDU = unit {
    %synchronize-at = b"\x00";
    : uint32;
};
```

We currently reject this code.

```console
$ spicyc -j sync.spicy
[error] sync.spicy:10:7: &synchronize cannot be used on field, look-ahead contains non-literals
```

bbannier:
We previously would not detect `%synchronize-at` or `%synchronize-from` attributes if the unit was not directly in a field, i.e., we mishandled the common case of synchronizing on a unit in a list.

With this patch we now handle these attributes, regardless of how the unit appears.

Closes #1617.


spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -1141,10 +1142,22 @@ struct ProductionVisitor
...
+        std::optional<type::Unit> unitType;
+        if ( while_ ) {
+            if ( const auto& field = while_->meta().field() )
+                if ( auto unit = field->parseType().elementType().tryAs<type::Unit>() )
+                    unitType = *unit;
+        }
+
+        else if ( const auto& unit = p->tryAs<production::Unit>() ) {
+            unitType = unit->unitType();
+        }
```
rsmmr, line 1156:
```suggestion
        else if ( const auto& unit = p->tryAs<production::Unit>() )
            unitType = unit->unitType();
```

rsmmr (approved):

spicy/toolchain/src/compiler/codegen/parser-builder.cc
bbannier: I actually thought of that, but incorrectly went with symmetry with the first case ;). Fixed.

PR #1620, opened 2023-12-14
Misc performance improvements

bbannier:
The main changes in this patch are adding of move constructors for types used heavily in generated code (in C++17 if a type declares any constructor no implicit move constructor is generated anymore). This cuts about 4% from the runtime for a huge internal parser we have. Changes here are motivated by benchmark results and I didn't attempt a sweeping audit.


rsmmr (approved):

bbannier:
> Changes here are motivated by benchmark results and I didn't attempt a sweeping audit.

I merged this as is, but continue working on auditing the full code. I tried implementing all fixes needed to pass [clang-tidy's Rule of Five checker](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/special-member-functions.html), but seem to have made performance worse along the way. Once I have sorted that out I'll open a new PR.

PR #1623, opened 2023-12-20
Refresh Clang tooling in CI

bbannier:


hilti/runtime/include/types/stream.h
```diff
@@ -856,13 +856,7 @@ class UnsafeConstIterator {
...
 
     Byte _dereference() const {
-        // TODO(bbannier): Ideally we would `assert(_chunk)` here so clang-tidy
-        // knows that `_chunk` is always not null. Unfortunately it looks like
-        // that fails to it pruning that CFG edge so we have to return instead
-        // here. This should be inconsequential to users as this function must
-        // not be called if the data is invalid.
-        if ( ! _chunk )
-            internalError("dereference of invalid iterator");
+        assert(_chunk);
```
bbannier, line 859: This was the main motivation for this patch since this function can get pretty hot for realistic workloads. By replacing the branch with an `assert` we can make the code less branchy and potentially also easier to optimize (viewed from the optimizer `assert` reduces the possible states). Overall the changes in this branch shave off about 1%.

hilti/runtime/include/hilti.h
```diff
@@ -28,7 +28,7 @@ void print(const T& t, bool newline = true) {
     cout << hilti::rt::to_string_for_print(t);
 
     if ( newline )
-        cout << std::endl;
+        cout << '\n';
```
rsmmr, line 31: Note there's semantic difference: `endl` flushes, which for buffered `cout` can make a difference. It's unlikely to matter here, so seems fine, but I am a bit surprised that clang-tidy is generally suggesting that change,

hilti/toolchain/src/compiler/visitors/resolver.cc
```diff
@@ -53,23 +53,6 @@ struct Visitor : public visitor::PostOrder<void, Visitor> {
     bool modified = false;
     std::map<ID, Type> auto_params; // miapping of `auto` parameters inferred, indexed by canonical ID
 
-#if 0
```
rsmmr, line 56: These are left in on purpose, to activate for debugging. 

More, generally don't like disallowing `#if 0`, suggest to drop `0aa5ba4908d` and suppress the warning instead.

rsmmr (dismissed):

hilti/toolchain/src/compiler/visitors/resolver.cc
bbannier: I checked the code I removed again and if activated none of the removed code would compile today. This is exactly one reason [this lint](https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-unconditional-preprocessor-if.html) flagged this an an antipattern:

> Unconditional preprocessor directives, such as `#if 0` for disabled code and `#if 1` for enabled code, can lead to dead code and always enabled code, respectively. Dead code can make understanding the codebase more difficult, hinder readability, and may be a sign of unfinished functionality or abandoned features. This can cause maintenance issues, confusion for future developers, and **potential compilation problems**.

A workable approach they suggest would be to use some debugging define which would allow actually enabling this usually disabled code in e.g., CI to make sure it keep working:

> As a solution for both cases, consider using preprocessor macros or defines, like `#ifdef DEBUGGING_ENABLED`, to control code enabling or disabling. This approach provides better coordination and flexibility when working with different parts of the codebase. Alternatively, you can comment out the entire code using `/* */` block comments and add a hint, such as @todo, to indicate future actions.

We could consider following that suggestion, and add a setup in CI which enables this code and fix the issues the code currently has. I personally never enabled any of these snippets so I am unsure how much value that would provide though.

hilti/runtime/include/hilti.h
bbannier: Yepp, that's exactly [the reason this lint gives](https://clang.llvm.org/extra/clang-tidy/checks/performance/avoid-endl.html). For our code there was only a single place where we wanted the flush which I [now made explicit](https://github.com/zeek/spicy/blob/b763e7c56f99512c429724ea15ed7b72d384b908/hilti/runtime/src/debug-logger.cc#L47).

hilti/toolchain/src/compiler/visitors/resolver.cc
rsmmr: I believe this is overthinking it ... The `#if 0` is just there to save the code in case it's helpful again later, rather than having to come up with it once more then. If it doesn't compile anymore, fine; still better to have a starting point than nothing (I have used some of it multiple times). However, in the grand scheme of things the code is really not important, so custom `#define`, CI, etc. seems too much. 

But if it's better to use comments instead of `#define`, sure, I wont' fight that. :-)

PR #1627, opened 2023-12-22
Enable LTO

bbannier:
For a large internal parser the changes in this PR lead to speedups of up to 8-10% when interfaced with Zeek. I am not 100% certain that this is not due to the `-flto` flags on our object libraries leaking into the `zeek` binary and making things the parsers interact with faster though.

In general such LTO builds seem to take (much) longer to link.


sethhall:
Have you tried parallelizing the linking (to make it finish faster)?

PR #1630, opened 2024-01-02
Fix docs namespace for symbols from `filter` module.

bbannier:
We previously would document these symbols to be in `spicy` even though they are in `filter`.


rsmmr (approved):
I'm wondering if they should actually relocate to `spicy::` but that's a breaking change, so I correcting the docs is fine for now.

PR #1637, opened 2024-01-08
Fix ASAN false positives introduces with d332827aee7d70cdb642631d7a28…

bbannier:
…9751e1d8a36a.

Since the logging changes of d332827aee7d70cdb642631d7a289751e1d8a36a ASAN reports false positives on e.g., macos-13.6.2 with its clang-1500.0.40.1. This patch slightly reorganizes the code so these false positiives are not triggered.


bbannier:
> Thanks for tracking this down! Looks much better, however I still see two tests failing, see below. I assume you aren't seeing those?
> 
> ```
> ctest.all ... failed
>   ...
> spicy.rt.stack-check ... failed
>   ...
> 2 of 2 tests failed
> ```

I see these as well, but I see them all the way back to at least spicy-1.2.0
and not caused by recent work (using macos-13.6.2 with their
`clang-1500.0.40.1`). I spent quite some time with the code and they look like
genuine false positives, triggered by exceptions and stack unwinding. I was not
able to make them go away, and we do not see them in our Linux ASAN CI either.

rsmmr:
Ok, what do you think about disabling those tests on macOS with sanitizer builds then?

bbannier:
> Ok, what do you think about disabling those tests on macOS with sanitizer builds then?

Done. I also added patches to streamline our CI setup.

rsmmr (dismissed):

PR #1642, opened 2024-01-09
Fix infinite loop in regular expression parsing.

### LINKED ISSUE #1634: Parsing trailing null byte padding using a regular expression fails to parse
If the data being parsed contains 0 to 3 bytes of trailing null byte padding and a regular expression is used to parse this data, in certain configurations, the null byte padding fails to get parsed or an infinite parsing loop is entered.  The attached files illustrate this behavior.

Version: `spicy-driver v1.9.0 (7b8eff5)`

=====================
`regex_padding.spicy` 
Desc: Spicy code showing trailing null byte padding failing to parse.
Command: `spicy-driver -f spicy_test_two_null_bytes.in regex_padding.spicy`
Result:
```
clang: warning: -lc++abi: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: -lc++abi: 'linker' input unused [-Wunused-command-line-argument]
unit2_field_1 => 7 
unit2_field_2 => \x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00
unit2_field_3 => ~&%\x89\x00\x0b\x00
unit2_field_4 => 
```

===================== 
`regex_padding_in_line.spicy`
Desc: Spicy code with Unit2 from the above `regex_padding.spicy` removed and the cooresponding field(s) declared in line of Unit1.  This code works as expected.
Command: `spicy-driver -f spicy_test_two_null_bytes.in regex_padding_in_line.spicy`
Result:
```
clang: warning: -lc++abi: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: -lc++abi: 'linker' input unused [-Wunused-command-line-argument]
unit1_field_3 => 7 
unit1_field_4 => \x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00
unit1_field_5 => ~&%\x89\x00\x0b\x00
unit1_field_6 => \x00\x00
```

=====================
`regex_padding_with_sink.spicy` 
Desc: Spicy code similar to the `regex_padding.spicy` code, but with a sink configured.  This code is very close to my actual use case which started my Spicy regex adventure.  This code results in an infinite parsing loop. :-(
Command: `spicy-driver -f spicy_test_two_null_bytes.in regex_padding_with_sink.spicy`
Result:
```
infinite parsing loop
```
Misc Files:
 File | Description |
|--------|--------|
|`spicy_test_no_null_bytes.in` | Test input that contains 0 trailing null bytes |
|`spicy_test_one_null_bytes.in` | Test input that contains 1 trailing null bytes |
|`spicy_test_two_null_bytes.in` | Test input that contains 2 trailing null bytes |
|`spicy_test_three_null_bytes.in` | Test input that contains 3 trailing null bytes |
|`regex_padding-hilti-trace.out` | Hilti trace file from regex_padding.spicy |
|`regex_padding-spicy-verbose.out` | Verbose parsing file from regex_padding.spicy |
|`regex_padding_in_line-hilti-trace.out` | Hilti trace file from regex_padding_in_line.spicy |
|`regex_padding_in_line-spicy-verbose.out` | Verbose parsing file from regex_padding_in_line.spicy |
|`regex_padding_with_sink-hilti-trace.out` | Hilti trace file from regex_padding_with_sink.spicy |
|`regex_padding_with_sink-spicy-verbose.out` | Verbose parsing file from regex_padding_with_sink.spicy |

rsmmr:
Closes #1634.


hilti/runtime/src/types/stream.cc
```diff
@@ -113,6 +113,18 @@ int Chain::numberOfChunks() const {
     return n;
 }
 
+bool View::isComplete() const {
+    _ensureValid();
+
+    if ( _begin.isFrozen() )
```
bbannier, line 119: The only other use of `View::isFrozen` was in `MatchState::advance` which is now gone. Could you remove the global method and inline it here? That would help avoid having two pretty similar functions which could be easy to confuse.

bbannier (approved):

PR #1644, opened 2024-01-10
Attempt to make copying into `Stream`s more lazy

bbannier:
Spicy-generated parser operate on `Stream`s, so in order to parse anything we always need to append input data to the `Stream` a parser operates on. `Stream`s consist of `Chunk`s which own their data, so this always involves a deep copy of the data.

This PR attempts to make this more lazy to explore whether we can get rid of associated overhead. For that I introduced a `Chunk` behavior where it does not own the data; to make this safe (e.g., since the input data might at some point disappear) users of this API need to explicitly make such `Chunk`s owning before their data goes out of scope. For now I encapsulated this in a RAII class which performs this operation when it goes out of scope.

Benchmarking this with a large internal parser the changes here actually make performance _worse_ (by about (3.5±1.2)%), so I am opening this more for reference. The added overhead seems to be due to many `Chunk` methods now having to also check whether a `Chunk` is not owning (previously: only check for gap chunk or not); it probably is worthwhile checking whether e.g., introducing dedicated `Chunk` classes for owning, non-owning, and gaps perform any better.

### DIFF
--- a/hilti/runtime/include/types/stream.h
+++ b/hilti/runtime/include/types/stream.h
@@ -118,17 +118,7 @@ public:
     bool isGap() const { return _gap_size > 0; };
     bool inRange(const Offset& offset) const { return offset >= _offset && offset < endOffset(); }
     bool isLazy() const { return ! _non_owning_data.empty(); };
-    void makeOwning(Offset begin, Offset end) {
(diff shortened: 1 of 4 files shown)


bbannier:
Closing as #1723 implemented something similar.

PR #1646, opened 2024-01-10
Implement `skip` handling for fields of fixed size

### LINKED ISSUE #1640: Optimize `skip` for fixed sized types and units
According to @bbannier, `skip` generates optimized code for `bytes`. It would be great to apply the same optimization to fixed sized basic types like integers as well as fixed size units.

bbannier:
Closes #1640.


spicy/toolchain/src/compiler/codegen/grammar-builder.cc
```diff
@@ -127,16 +127,24 @@ struct Visitor : public hilti::visitor::PreOrder<Production, Visitor> {
                 // Skipping not supported
             }
 
+            else if ( const auto& size = AttributeSet::find(n.attributes(), "&size") )
+                // FIXME(bbannier): the referenced node seems to go away the node in `skip` ends up pointing to junk.
```
bbannier, line 131: Any idea why this is or what I could do here to make sure the ref stays valid @rsmmr?

bbannier: The issue here was how conditional fields were handled. This was already broken for conditional fields with literal values. I added a patch & test which fixes that.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -1699,34 +1699,70 @@ struct ProductionVisitor
...
+                    hilti::rt::cannot_be_reached();
+            }
+
+            else if ( auto t = p.field().parseType().tryAs<type::Real>() ) {
+                auto type = *p.field().attributes()->find("&type")->valueAsExpression();
+                return std::optional(
+                    builder::ternary(builder::equal(type, builder::id("spicy::RealType::IEEE754_Single")),
+                                     builder::integer(4U), builder::integer(8U)));
+            }
+
```
rsmmr, line 1744: How about making this a type property instead: we could add a API method to types that returns how many bytes their unpacking requires if that's a fixed amount. Same method could then be used by the grammar builder to decide if a `Skip` production can be created. That way we centralize the knowledge with the types themselves, which is easier to maintain.

rsmmr (commented):
Good stuff. A further future extension could be inferring static sizes for units that don't have an explicit `&size`: if all their fields are fixed size, the unit is so too.

spicy/toolchain/src/compiler/codegen/grammar-builder.cc
```diff
@@ -127,13 +127,20 @@ struct Visitor : public hilti::visitor::PreOrder<Production, Visitor> {
                 // Skipping not supported
             }
 
+            else if ( const auto& size = AttributeSet::find(n.attributes(), "&size") ||
```
bbannier, line 130: This means that e.g., hooks or other attached stateful computations (e.g., `&requires`, `&convert`) would not be invoked on skipped fields. We could document that with `skip` whether they are invoked is undefined.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
bbannier: The only types here where we know their sizes are the integers, but the size of everything else depends on some field attribute. I'll try to instead add a method to `Field`.

spicy/toolchain/src/compiler/codegen/grammar-builder.cc
bbannier: I updated the docs.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -1699,34 +1699,70 @@ struct ProductionVisitor
...
+            auto n = builder()->addTmp("skip", size);
+            auto loop = builder()->addWhile(builder::greater(n, builder::integer(0U)));
+            pushBuilder(loop, [&]() {
+                pb->waitForInput(builder::integer(1U), "not enough bytes for skipping", p.location());
+                auto consume = builder()->addTmp("consume", builder::min(builder::size(state().cur), size));
+                pb->advanceInput(consume);
+                builder()->addAssign(n, builder::difference(n, consume));
+                builder()->addDebugMsg("spicy-verbose", "- skipped %u bytes (%u left to skip)", {consume, n});
+            });
+        };
```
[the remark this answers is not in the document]
bbannier: I added `ParserBuilder::skip` which just captures this body. I think you mentioned offline that this would allow further optimizations, but I seem unable to see them, could you point me what else would be needed? We e.g., cannot trim unconditionally since the `skip` field might be in a non-trimming context.

doc/programming/parsing.rst
```diff
@@ -328,11 +328,15 @@ prefix corresponding fields with:
...
 avoiding the overhead of storing any data.
 
-``skip`` fields may have conditions and hooks attached, like any other
-fields. However, they do not support ``$$`` in expressions and hook.
+``skip`` fields may have conditions and hooks attached, like any other fields.
+However, they do not support ``$$`` in expressions and hook. Since ``skip``
+fields might be completely elided there is no guarantee that any stateful side
+effects from parsing them happens, e.g. ``&requires`` attributes might be
+ignored, side effects of ``&convert`` attributes might not manifest, and hooks
+could not end up being invoked.
```
rsmmr, line 339: I'd phrase this more directly as "undefined", like:

```suggestion
However, they do not support ``$$`` in expressions and hook. 

Since ``skip``
allows the compiler to optimize the field's parsing code---including completely 
eliding most of it---it remains undefined if any side effects associated with the 
field will take effect. For example, ``&requires`` attributes might be ignored, 
``&convert`` expressions might not be evaluated, and hooks could end up not 
being invoked.
```

hilti/toolchain/include/ast/types/bitfield.h
```diff
@@ -133,6 +135,9 @@ class Bitfield : public hilti::TypeBase,
     auto typeParameters() const { return hilti::util::slice(children(), 2); }
     /** Implements the `Type` interface. */
     auto isWildcard() const { return _wildcard; }
+
+    std::optional<uint64_t> size() const { return width() / 8; }
+
```
rsmmr, line 140:
```suggestion
     /** Implements the `Type` interface. */
    std::optional<uint64_t> size() const { return width() / 8; }
```
(just for consistency)

spicy/toolchain/src/ast/types/unit.cc
```diff
@@ -38,6 +40,27 @@ hilti::optional_ref<const type::unit::Item> type::Unit::itemByName(const ID& id)
...
+    for ( const auto& i : this->itemRefs() ) {
+        if ( const auto& f = i->tryAs<unit::item::Field>() ) {
+            // TODO(bbannier): To support fields with conditions we need to be able to return an runtime value (an
+            // `Expression`) instead of a compile-time `uint64_t` here which requires support in the type API.
+            if ( f->condition() )
+                return {};
+
+            // TODO(bbannier): Add support for returning value of fixed-size lists. These will not be represented as
+            // vectors in the unit but as opaque counters, so this requires some more work elsewhere.
+            if ( auto s = f->parseType().size() )
```
rsmmr, line 54: Couldn't this be `f->size()`?

spicy/toolchain/src/ast/types/unit.cc
```diff
@@ -38,6 +40,27 @@ hilti::optional_ref<const type::unit::Item> type::Unit::itemByName(const ID& id)
     return {};
 }
 
+std::optional<uint64_t> type::Unit::size() const {
+    uint64_t result = 0;
+    for ( const auto& i : this->itemRefs() ) {
+        if ( const auto& f = i->tryAs<unit::item::Field>() ) {
+            // TODO(bbannier): To support fields with conditions we need to be able to return an runtime value (an
+            // `Expression`) instead of a compile-time `uint64_t` here which requires support in the type API.
```
rsmmr, line 48: This seems tricky, not sure if it's even viable at all. Maybe just skip the comment and focus this just on static properties.

spicy/toolchain/src/ast/types/unit-items/field.cc
```diff
@@ -30,3 +33,30 @@ std::optional<std::pair<const Expression, std::optional<const Type>>> spicy::typ
 
     return {};
 }
+
+std::optional<const Expression> spicy::type::unit::item::Field::size() const {
```
rsmmr, line 37: Couple things:

Good point that the size can depend not just on types, but also on field attributes, I didn't think about that. Given that, I'm now wondering if the type-side `size()` method is really as helpful as I originally thought because we now have two levels of `size()`, each with just a couple of cases/implementations (`Type::size()` and (`Field::size`). So maybe putting it all here into `Field::size()` is the better approach.

And then I think I'd use a visitor instead of the if/else-chain, both for readability/consistency-in-style and (I believe ...) performance.

hilti/toolchain/include/ast/type.api
```diff
@@ -2,6 +2,9 @@
 
 /** Interface for HILTI types. */
 class Type(hilti::trait::isType) : hilti::trait::isNode {
+    /** Returns the number of bytes the type consumes if known. */
```
rsmmr, line 5: Wondering what the right phrasing for the description here is. We are in HILTI, so "consume" doesn't match. Maybe: "number of bytes of the type's binary serialization if that's statically determinable." (but also see below on whether we want to even keep this).

spicy/toolchain/src/ast/types/unit.cc
bbannier: We are implementing `Type::size()` here which needs to return a static size; `Field::size()` OTOH hand returns an `Expression` (e.g.., to support `&size`).

What this means is that we can only really compute units which exclusively contain fields with attribute-independent sizes, i.e., only units over integers. This is not great, and in light of your comment below we maybe should remove this code here and only truely skip fields of unit type if they come with e.g., a `&size`.

hilti/toolchain/include/ast/type.api
bbannier: This change is gone now I dropped adding `size()` to `Type`.

spicy/toolchain/src/ast/types/unit-items/field.cc
bbannier: I changed this as suggested.

spicy/toolchain/src/ast/types/unit.cc
bbannier: I now removed this code again.

rsmmr (approved):

bbannier:
I backported 2559ed404aff719d15d4b5f9e2a39e8d74248473) to `release/1.8`.

PR #1647, opened 2024-01-11
Fix `&size` check.

### LINKED ISSUE #1645: Incorrect behavior with nested `&size` attributes
@kkvarfordt reported an issue on Slack where code making use of nested `&size` attributes does not behave correctly:

```ruby
module foo;

public type X = unit {
    x: Y[] &size=3;
};

type Y = unit {
    y: bytes &eod &size=2;
};
```

```console
$ printf '\x01\x02\x03' | spicy-driver foo.spicy
[fatal error] terminating with uncaught exception of type spicy::rt::ParseError: &size amount not consumed (foo.spicy:4:13)
```

The expectation here would be that this parses into `X { x: [ Y { y: b"\x01\x02"}, Y { y: b"\x03"} ] }`. The behavior is the same if `Y` is inlined as `x: (bytes &eod &size=2)[] &size=3`.

This has been an issue since at least `spicy-1.2.0`.

rsmmr:
The current parsing offset could legitimately end up just beyond the
`&size` amount.

Closes #1645.


bbannier (approved):
Looking good. I did a pass of the code and checked whether we also needed to do anything for `&max-size`, but that code already seems to do the right thing.

Let's make sure to backport this to `release/1.8` so we can include it in the next patch release.

PR #1649, opened 2024-01-13
Fix type of generated code for `string::size`.

bbannier:
While we defined `string`'s size operator to return an `uint64` and documented that it returns the length in codepoints, not bytes, we still generated C++ code which worked on the underlying bytes (i.e., it directly invoked `std::string::size` instead of using `hilti::rt::string::size`).


rsmmr (approved):

bbannier:

PR #1650, opened 2024-01-15
Provide meaningful unit `__begin` value when parsing starts.

### LINKED ISSUE #1648: Obscure `UnsetOptional` error on `assert` from `%done` hook
I came across the following failure:

```ruby
module foo;

public type X = unit {
    on %done { assert self.offset() != 0; }
};
```
The assertion always fails since the unit consumes no input. With that I see

```console
$ echo | spicy-driver -d foo.spicy
[error] terminating with uncaught exception of type hilti::rt::UnsetOptional: unset optional value (foo.spicy:4:16)
```

Interestingly this seems to tied to the use of `assert`, if I e.g., `print` the boolean expression this shows the expected `False`.

bbannier:
We previously would not provide `__begin` when starting the initial parse. This meant that e.g., `offset()` was not usable if nothing ever got parsed.

With this patch we provide a meaningful value now.

Closes #1648.


rsmmr (commented):
I'm trying to recall if there was a reason for the current behavior. This is actually  documented even, see https://docs.zeek.org/projects/spicy/en/latest/programming/language/types.html#method-unit::offset and other random access functions. So we will need to update the documentation at least.

bbannier:
> I'm trying to recall if there was a reason for the current behavior. This is actually documented even, see https://docs.zeek.org/projects/spicy/en/latest/programming/language/types.html#method-unit::offset and other random access functions. So we will need to update the documentation at least.

Good question. This docstring seems to have existed before we introduced the new explicit tracking of the offset value, but the parse methods still allow for parsing without specifying the a begin so it seems possible that it could still unset. I'll investigate.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -337,7 +337,7 @@ struct ProductionVisitor
                             builder()->addLocal("filtered_data", type::ValueReference(type::Stream()),
                                                 builder::id("filtered"));
                             args2[0] = builder::id("filtered_data");
-                            args2[1] = builder::optional(type::stream::Iterator());
+                            args2[1] = builder::default_(type::stream::Iterator()); // FIXME(bbannier): sensible data.
```
rsmmr, line 340: What does this mean?

rsmmr (approved):

spicy/toolchain/src/compiler/codegen/parser-builder.cc
bbannier: Thanks for catching this! This was leftover WIP code for getting the typing right which suprisingly seems to have had no effect with our test suite. I added the correct value which we already have available.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -337,7 +337,7 @@ struct ProductionVisitor
                             builder()->addLocal("filtered_data", type::ValueReference(type::Stream()),
                                                 builder::id("filtered"));
                             args2[0] = builder::id("filtered_data");
-                            args2[1] = builder::default_(type::stream::Iterator()); // FIXME(bbannier): sensible data.
+                            args2[1] = builder::begin(builder::deref(args[0]));
```
rsmmr, line 340: I'm wondering if this should be `args[2]` to be relative to the filtered stream? Both semantically, and also because otherwise we might end up comparing iterators for different streams?

Either way, a test using `offset()` on filtered data would be good too if we don't have one.

bbannier: > I'm wondering if this should be args[2] to be relative to the filtered stream?

Duh, good catch! Yes, this definitely was supposed to be `args2[0]` (unless I misunderstood your comment). 

> Either way, a test using offset() on filtered data would be good too if we don't have one.

We didn't have such a test, I added one. I do however have difficulty making it fail with my previous broken code. I discovered #1652 along the way though which might be a reason for that.

rsmmr (approved):

PR #1654, opened 2024-01-22
Fix parser locations

rsmmr:
It has bothered me for a while that sometimes our location information
in error/debug message would be off, and I think I finally found what
was going on there.

- Fix Bison locations.
- Extend location printing to include single-line ranges.
- Adjust end of Bison-generated locations.


hilti/toolchain/src/ast/location.cc
```diff
@@ -18,11 +18,15 @@ std::string Location::render(bool no_path) const {
         else
             lines = util::fmt(":%d", _from_line);
 
-        if ( _to_line >= 0 && _to_line != _from_line ) {
-            if ( _to_character >= 0 )
-                lines += util::fmt("-%d:%d", _to_line, _to_character);
-            else
-                lines += util::fmt("-%d", _to_line);
+        if ( _to_line >= 0 ) {
```
bbannier, line 21: Our `Location` has overloaded semantics since it encodes either a _position_ and a _range_. I'd argue these are the entities downstream tools care about (see e.g., [Vim's `setqflist`](https://vimdoc.sourceforge.net/htmldoc/eval.html#setqflist())). With that use case in mind I think it would be better to use a less "compressed" format, i.e., either emit a position `LINE:COLUMN` or a range `START_LINE:START_COLUMN-END_LINE:END_COLUMN`, and drop the special casing for end on same line as start.

bbannier (approved):
LGTM. The results look resonable to me, but large parts of the impl a opaque to me.

The commit message of the second commit is outdated now (still talks about compression range logging). Could you fix that before merging?

rsmmr:
> The commit message of the second commit is outdated now (still talks about compression range logging). Could you fix that before merging?

Oops, saw that too late.

Issue #1652, opened 2024-01-17
Field of type with filter consumes more data than expected


bbannier:
Given the following code

```ruby
module foo;

const N: uint64 = 1;

public type X = unit {
    y: Y &max-size=N;
    z: bytes &eod &requires=|$$|!=0;
};

type Y = unit {
    on %init {
        self.connect_filter(new F);
    }

    data: bytes &size=N;
};

type F = unit {
    %filter;

    filter_input: bytes &size=N {
        print $$;
        self.forward($$);
    }
};
```

For an input `\x01\x02\x03` I would expect `y` to consume exactly one (`N`) byte since the filter connected to `Y` consumes one byte. It looks however as if that is not the case and the `&max-size` check fails; if I remove the check I still see no data in `z` so it looks to me as if the filter consumes all input.

Printing `filter_input` shows a single byte so I suspect this is happening outside of the parsing of `F` in how we update the view it operates on.

This seems to be broken from at least 1.4.0 on (before d15281b1b836515695c012b29f5190528ee45555 we rejected above reproducer).

bbannier:
The issue here is that we seem to trim too much, at least once in the filter and again for the field. One of these seems to see the wrong `cur`. Currently we do not set up new parser state for parsing in filters which might be required for some functionality, but I suspect contributes to the issue here.

https://github.com/zeek/spicy/blob/daa2e3ce48a92dcf6662df724a8bb1c8d5f8df4b/spicy/toolchain/src/compiler/codegen/parser-builder.cc#L343

PR #1658, opened 2024-01-23
Fix lookahead parsing on filtered views

### LINKED ISSUE #1655: Filter generates code with incompatible iterators
Using the filter or not in the following example should not make a difference, however the version with connected filter throws an error from generated code.

```ruby
# @TEST-EXEC: echo "foo" | spicy-dump -d %INPUT 1>&2

module foo;

public type X = unit {
    headers: Y[];
};

type Y = unit {
    on %init {
        self.connect_filter(new F);
    }

    : b"foo";
};

type F = unit {
    %filter;
    : bytes &eod {
        self.forward($$);
    }
};
```

```
[fatal error] terminating with uncaught exception of type hilti::rt::InvalidIterator: incompatible iterator (../filter-fail.spicy:14:7)
```

This seems to have been broken since at least v1.2.0.

bbannier:
Closes #1655.


bbannier:
I have rebased the fixes for #1668 and #1652, but this still requires work.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -316,34 +316,51 @@ struct ProductionVisitor
...
-                            args2[0] = builder::id("filtered_data");
-                            args2[1] = builder::begin(builder::deref(args2[0]));
-                            args2[2] = builder::deref(args2[0]);
-                            builder()->addExpression(builder::memberCall(state().self, id_stage2, args2));
+                            auto pstate = state();
+                            pstate.data = builder()->addTmp("filtered_data", type::ValueReference(type::Stream()),
+                                                            builder::id("filtered"));
+                            pstate.cur = builder::deref(pstate.data);
+                            pstate.begin = builder::begin(pstate.cur);
+                            // FIXME(bbannier): set `lahead` and `lahead_end` to refer to the filtered view.
```
bbannier, line 351: @rsmmr I am a little stuck what to pass here, in particular for `lahead_end`. Do you have an idea?

bbannier:
Abandoning this per https://github.com/zeek/spicy/issues/1655#issuecomment-1935577766.

PR #1662, opened 2024-01-25
Add bitfield examples.

awelzel:
Add examples to the bitfield docs and remove &bit-order for the individual bitfield ranges, it doesn't seem to have an effect at all, but the `&bitfield-order` on the whole bitfield is actually useful, but wasn't documented.

This was deduced with the following test program:

```
module Test;

import spicy;

public type X = unit {

  #  input:     10000010 00000101   \x82\x05
  #  big/lsb0:  54321098 76543210
  be_lsb0: bitfield(16) {
    x0: 0;                # 1
    x12: 1..2;            # 2  ( 10 )
    x35: 3..5;            # 0
    x79: 7..9;            # 4  ( 100 )
    x1015: 10..15;        # 32 ( 100000 )
    x1015_msb0: 10..15 &bit-order=spicy::BitOrder::MSB0;  # still 32
  };

  #  input:        10000010 00000101   \x82\x05
  #  little/lsb0:  76543210 54321098
  #                       0      1
  le_lsb0: bitfield(16) {
    x0: 0;                # 0
    x12: 1..2;            # 1
    x35: 3..5;            # 0
    x79: 7..9;            # 3 011
    x1015: 10..15;        # 1
    x1015_msb0: 10..15 &bit-order=spicy::BitOrder::MSB0;  # still 1
  } &byte-order=spicy::ByteOrder::Little;

  #  input:        10000010 00000101   \x82\x05
  #  little/msb0:  89012345 01234567
  #                  1      0
  le_msb0: bitfield(16) {
    x0: 0;             # 0
    x12: 1..2;         # 0
    x35: 3..5;         # 1 001
    x79: 7..9;         # 6 110
    x1015: 10..15;     # 2 000010
    x1015_msb0: 10..15 &bit-order=spicy::BitOrder::MSB0;  # still 2
    x1015_lsb0: 10..15 &bit-order=spicy::BitOrder::LSB0;  # still 2
  } &byte-order=spicy::ByteOrder::Little &bit-order=spicy::BitOrder::MSB0;

  on %done {
    print "be_lsb0", self.be_lsb0;
    print "le_lsb0", self.le_lsb0;
    print "le_msb0", self.le_msb0;
  }
};
```

```
$ printf '\x82\x05\x82\x05\x82\x05' | spicy-driver a.spicy 
be_lsb0, (1, 2, 0, 4, 32, 32)
le_lsb0, (0, 1, 0, 3, 1, 1)
le_msb0, (0, 0, 1, 6, 2, 2, 2)
```


awelzel:
I was tempted to mark this as closes #1489 - @rsmmr could you see if you agree that improved docs might be all that's needed? With the mental model I have now things make sense, but maybe that has nothing to do with the actual implementation :-)

doc/programming/parsing.rst
```diff
@@ -1009,17 +1009,112 @@ Generally, a field ``bitfield(N)`` field is parsed like an
...
-Furthermore, each bit range can also specify a ``&bit-order``
-attribute to specify the :ref:`ordering <spicy_bitorder>` for its
-bits; the default is ``spicy::BitOrder::LSB0``.
+
+..
+    This doesn't actually have any effect.
+
+    Furthermore, each bit range can also specify a ``&bit-order``
+    attribute to specify the :ref:`ordering <spicy_bitorder>` for its
+    bits; the default is ``spicy::BitOrder::LSB0``.
```
rsmmr, line 1026: Let's just remove  it.

doc/programming/parsing.rst
```diff
@@ -1009,17 +1009,112 @@ Generally, a field ``bitfield(N)`` field is parsed like an
...
+
+    MSB                           LSB
+         <--    1        <--        0
+    6 5 4 3 2 1 0 9 8 7 6 5 4 3 2 1 0
+    +---------------+---------------+
+    |     byte 1    |   byte 0      |
+    +-------------------------------+
+
+
+This default layout may be surprising given that some RFCs use inverse bit
```
rsmmr, line 1041: If it's indeed surprising, I'm wondering if we should change the default (though that would break things of course).

rsmmr (commented):
Thanks for explaining all this in more detail, this sounds good. 

I'm fine closing #1489 with this, I'm lost on that ticket anyways as I can't really figure out what we'd want. At least I don't think the current approach is clearly wrong, and describing how it all works may be all we need.

Couple more requests:

- If we don't want to support the attribute on the individual fields, we should report them as errors (or for now: deprecate and warn).
- Mind turning your test program into a btest?

doc/programming/parsing.rst
awelzel: Hmm.. Maybe as a module level specification?

Let me address the other points. Thanks!

awelzel:
@rsmmr - I've added the test and a deprecation warning in the spicy validators.

Unfortunately, the `checkBits` function seems to run twice, so this looks a bit clunky, maybe there's a better approach?
```
$ ../../build/bin/spicyc -j a.spicy -o a.hlto
[warning] a.spicy:37:5-37:56: &bit-order on bitfield item 'x1015_msb0' has no effect and is deprecated
[warning] a.spicy:38:5-38:56: &bit-order on bitfield item 'x1015_lsb0' has no effect and is deprecated
[warning] a.spicy:37:5-37:56: &bit-order on bitfield item 'x1015_msb0' has no effect and is deprecated
[warning] a.spicy:38:5-38:56: &bit-order on bitfield item 'x1015_lsb0' has no effect and is deprecated

```

bbannier:
> @rsmmr - I've added the test and a deprecation warning in the spicy validators.

Could you add a test for the new deprecation warning?

tests/spicy/types/bitfield/parse-bitorder-2.spicy
```diff
@@ -0,0 +1,143 @@
+# @TEST-EXEC: echo "" | spicy-driver %INPUT >output
```
rsmmr, line 1: Nit here: better not to use `-<N>` naming on test files. `btest` creates those internally when using `@BTEST-START-NEXT`. Using them for both cases can be confusing, also if one might want to add a `@BTEST-START-NEXT` to an existing test.

rsmmr (approved):

rsmmr:
> Unfortunately, the `checkBits` function seems to run twice, so this looks a bit clunky, maybe there's a better approach?

Let's leave it like that for now. You can file a ticket if you want. Once the new AST infrastructure is in, we can take a stab at fixing various of the smaller language issues.

(btw, I don't see these duplicates in the baseline you committed?)

tests/spicy/types/bitfield/parse-bitorder-2.spicy
awelzel: Ah, yes! parse-bitorder-endianness.spicy it has become.

awelzel:
@bbannier fixed up the duplicated output moving the test into `operator()(Field)`

PR #1666, opened 2024-01-29
Fix running of `codebase` tests.

### LINKED ISSUE #1665: Codebase tests aren't executed
Because `btest.cfg` is missing `codebase`:

```
TestDirs=hilti spicy ctest
```

Not fixing it right now because there are actually test failures that need to be addressed before activating this.

bbannier:
This patch removes linters originally written as tests since they are now executed as pre-commit hooks, and enables running of the remaining test with the btests.

Closes #1665.


rsmmr (approved):
Ah, easier than I thought! :-)

Issue #1668, opened 2024-02-05
`&max-size` has effect on how much data is consumed


bbannier:
I would expect `&max-size` to inject a post-condition into parsing, but not have an effect on how much data is consumed (unless there is an error). This is not the case, e.g., in below reproducer regardless of whether `&max-size` was used or not, `rest` should always be `b"BC"`,

```ruby
# @TEST-EXEC: printf '\x01_BC' | spicy-dump -d %INPUT >>output2 2>&1
# @TEST-EXEC: printf '\x00_BC' | spicy-dump -d %INPUT >>output2 2>&1

module foo;

public type X = unit {
    use_max_size: uint8 &convert=cast<bool>($$);

    switch (self.use_max_size) {
        True -> xs: bytes &until=b"_" &max-size=1;
        False -> xs: bytes &until=b"_";
    };

    rest: bytes &eod;
};
```

Instead the following output is produced:

```
foo::X {
  use_max_size: True
  xs:
  rest: C
}
foo::X {
  use_max_size: False
  xs:
  rest: BC
}
```

The error is in how `&max-size` sets the next position, https://github.com/zeek/spicy/blob/041a0723998185b51d774ed1ef9cb7955102e868/spicy/toolchain/src/compiler/codegen/parser-builder.cc#L704

Contrary to what we currently do, `&max-size` is completely different from `&size` wrt. `ncur` which depends on what was actually parsed and cannot be known or set before.

I came across this while working on #1652.

PR #1670, opened 2024-02-07
Fix filters consuming too much data.

### LINKED ISSUE #1652: Field of type with filter consumes more data than expected

bbannier:
We would previously assume that a filter would consume all available data. This only holds if the filter is attached to a top-level unit, but in general not if some sub-unit uses a filter. With this patch we explicitly compute how much data is consumed.

Closes #1652.


tests/Baseline/spicy.types.unit.filter-size/output
```diff
@@ -0,0 +1,22 @@
...
+    data: A
+  }
+  z: BC
+}
+foo::X {
+  restrict_: MAX_SIZE
+  y: foo::Y {
+    data: A
+  }
+  z: BC
```
bbannier, line 21: This would produce `z: C` until the fix for #1668 is merged.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -335,15 +340,23 @@ struct ProductionVisitor
...
 
-                            // Assume the filter consumed the full input.
-                            pb->advanceInput(builder::size(state().cur));
+                            auto offset2 =
+                                builder()->addTmp("offset2",
+                                                  builder::memberCall(builder::begin(builder::deref(state().data)),
+                                                                      "offset", {}));
+
+                            auto advance = builder::difference(offset2, offset1);
+                            pb->advanceInput(advance);
```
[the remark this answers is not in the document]
bbannier: Unfortunately not. `filter::init` does not modify `state().cur`, but we can observe what is consumed via `state().data`, https://github.com/zeek/spicy/blob/e4704de28e933434b3f9466a2a4df3453b64d12c/spicy/runtime/include/filter.h#L201-L207.

Like discussed offline, there is little use in e.g., capturing the return value of the call to the stage 2 function

tests/Baseline/spicy.types.unit.filter-size/output
bbannier: This has the expected behavior now.

rsmmr (approved):

PR #1672, opened 2024-02-08
Drop overzealous validator.

### LINKED ISSUE #1522: Validation error when using list and `&convert` to list
The following code is rejected by the validator.

```ruby
module foo;

public type A = unit {
    # Parenthesis around X to possibly pass unit parameter.
    # Remove parentheses to let code pass validator.
    xs: (X)[]; # E: use [] syntax to parse vectors
};

type X = unit {
} &convert=vector(1);
```

```console
$ spicyc -dj foo.spicy
[error] foo.spicy:5:10: use [] syntax to parse vectors
```

The code passes if `A.x` is instead declared like so:

```ruby
    xs: X[];
```

I would have expected this code to be valid.

bbannier:
This validator was intended to reject incorrect parsing of vectors but instead ending up rejecting all vector parsing if the vector elements itself produced vectors. Since this code has no test and it seems to have no clear purpose this patch drops this validation.

Closes #1522.

### DIFF
--- a/spicy/toolchain/src/compiler/visitors/validator.cc
+++ b/spicy/toolchain/src/compiler/visitors/validator.cc
@@ -1,6 +1,5 @@
 // Copyright (c) 2020-2023 by the Zeek Project. See LICENSE for details.
 
-#include <algorithm>
 #include <utility>
 
 #include <spicy/rt/mime.h>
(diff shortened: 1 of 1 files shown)


rsmmr:
>  it seems to have no clear purpose this patch drops this validation.

It actually does have a purpose: it's there to reject the usage of `field: vector<T>` and tell people to use `field: T[]` instead.  I don't remember the details but I'm pretty sure the two aren't equivalent in the sense that if somebody ends up using the former,  parsing won't work as they would expect. So I'd prefer to keep/fix the check.

rsmmr:
> * `: vector<uint8>;`: rejected with `vector<T> syntax is no longer supported for parsing sequences; use T[] instead.`

Excellent, then I'm good.

Issue #1667, opened 2024-01-30
Parse error during regexp parsing could synchronize earlier


Mohan-Dhawan:
I modified the `synchronize-on-gap.test` btest to reproduce a behavior I am seeing with a pcap that has gaps in content. The modifications are as follows:

The synchronization token is `AB`.
```
public type Xs = unit {
        xs: (/AB/ &synchronize)[];
        on %synced { confirm; }
};
```
The input is as follows
```
int main() {
    hilti::rt::init();

    auto xs = hilti::rt::reference::make_value<__hlt::sync::Xs>();

    hilti::rt::ValueReference<hilti::rt::Stream> stream;
    stream->append("A");
    stream->append(nullptr, 1024); // Gap.
    stream->append("AB"); // should sync here, but is skipped
    stream->append("AB"); // instead gets synch'd here

    hlt::sync::Xs::parse2(xs, stream, {}, {});
    std::cout << xs << '\n';

    hilti::rt::done();
}
```

It generates a spicy-verbose log as below clearly indicating that the synchronization happens with the second token while the first is skipped.
```
  [spicy-verbose] - state: type=sync::Xs input="A<gap>..." stream=0x55e452d602e0 offsets=0/0/0/1029 chunks=4 frozen=no mode=default trim=yes lah=n/a lah_token="n/a" recovering=no
  [spicy-verbose] - parsing production: Unit: sync_Xs -> xs
  [spicy] sync::Xs
  [spicy-verbose]   - state: type=sync::Xs input="A<gap>..." stream=0x55e452d602e0 offsets=0/0/0/1029 chunks=4 frozen=no mode=default trim=yes lah=n/a lah_token="n/a" recovering=no
  [spicy-verbose]   - parsing production: While: xs  -> while(<look-ahead-found>): anon
  [spicy-verbose]     failed to parse list element, will try to synchronize at next possible element
  [spicy-verbose]     - state: type=sync::Xs input="A<gap>..." stream=0x55e452d602e0 offsets=0/0/0/1029 chunks=4 frozen=no mode=default trim=yes lah=n/a lah_token="n/a" recovering=yes
  [spicy-verbose]     - trimming input
  [spicy-verbose]     - state: type=sync::Xs input="ABAB" stream=0x55e452d602e0 offsets=1025/0/1025/1029 chunks=2 frozen=no mode=default trim=yes lah=n/a lah_token="n/a" recovering=yes
[...]

PR #1673, opened 2024-02-20
Allow user-specified synchronization function in `%synchronize-at`.

bbannier:


bbannier:
@J-Gras @Mohan-Dhawan this came out of a discussion with your team where the claim was that regex matching during recovery was too slow for your needs and you wanted to hook in faster custom code you had in mind. I have posted a fix for #1667, and I would be interested whether you could share numbers showing that this patch here is worthwhile at all on top of that.

doc/programming/parsing.rst
```diff
@@ -921,6 +921,19 @@ properties:
...
+    with the following semantics:
+
+    - ``int8`` value is ``1``: synchronization point *is* in the view at the
+      offset specified by the ``uint64`` value.
+    - ``int8`` value is ``0``: synchronization point *could be* in the view,
+      need at least the number of bytes specified in ``uint64`` value to
+      decide.
+    - ``int8`` value is ``-1``: synchronization point not in view and never
+      will, call function again with new data.
+
```
rsmmr, line 936: Couple thoughts on the API description:

- Suggest to add to the docs that case 0 means buffering: we'll hold on to the current input data until the synchronization functions declares success (1) or failure (0).

- A function could return -1 but still retain state about what it saw, right? So a way to avoid the buffering with case 0 would be a state machine that remembers partial matches, returns -1, and then continues matching with the next data chunk. Might be worth pointing that out in the docs: -1 isn't necessarily a total failure, it's just that the function won't need the data anymore.

rsmmr (commented):
> I would be interested whether you could share numbers showing that this patch here is worthwhile at all on top of that.

Yeah, some evidence that this is helping solve a problem would be good before considering merging this.

doc/programming/parsing.rst
rsmmr: >  So a way to avoid the buffering with case 0 would be a state machine that remembers partial matches, returns -1, and then continues matching with the next data chunk.

Well, I guess that's assuming that the actual synchronization point follows *after* whatever the function is looking for. If it's at the beginning, we need to buffer.

bbannier: This has the semantics identical to `%synchronize-at` with expression, the function needs to point at data which parses successfully. This means that we have to buffer for `0`, and `1`, and can trim input only for `-1`.

bbannier:
Closing this as there seems to be no interest by the folks who originally requested the feature.

PR #1679, opened 2024-02-26
Enable optimizations when running `spicy-build`

### LINKED ISSUE #1622: `spicy-build`: Uncaught exception of type hilti::rt::UnsetOptional if input does not align
```Spicy
module Test;

import spicy;

type MyPDU = unit {
        x1: uint16;
        x2: uint8;
};

public type PDUs = unit {
        : MyPDU[];
};
```
Using `spicy-build` I see a strange error if I am trying to process input that doesn't align in terms of the number of bytes to be parsed:
```
# spicy-build test.spicy
# printf '\x01\x02\x03' | ./a.out
# printf '\x01\x02' | ./a.out
[error] a.out: terminating with uncaught exception of type hilti::rt::UnsetOptional: unset optional value (test.spicy:5:14-8:2)
```

### LINKED ISSUE #1624: `spicy-build`: High memory consumption reading files.
While processing a file (`-f`) with an executable created by `spicy-build` the process' memory consumption is constantly growing up to the size of the input file. Looks like processed input is never released. Running the same parser using `spicy-driver`, the memory consumption is stable.

### LINKED ISSUE #1625: Consider removing `spicy-build`
We currently provide `spicy-build` as a way to build standalone binaries. I would argue that this provides little benefit over _always_ (and exclusively) using `spicyc` to compile a HLTO file and using it with e.g., `spicy-driver` or `spicy-dump`.

In addition to requiring lots of duplicate code for a `spicy-build`-specific driver or in general not being well exercised, another downside of `spicy-build` is that the way it is currently written it needs to disable optimizations which leads to issues like #1624. I believe we should strongly consider removing `spicy-build`.

bbannier:
Closes #1624.
Closes #1625.
Closes #1622.


tests/hilti/codegen/type-info.hlt
```diff
@@ -173,22 +173,23 @@ public function tuple<strong_ref<TestOptionals>, TypeInfo> makeTestOptionals() {
 
 using namespace hilti::rt;
 
-namespace __hlt::Test {
+namespace __hlt_::Test {
```
rsmmr, line 176: Now that I see this, I think it's worth tracking down where this changes comes from and fix it. The `__hlt` is actually configurable, and there shouldn't be anything added to it.

rsmmr (commented):
One request, rest looks good.

tests/hilti/codegen/type-info.hlt
bbannier: I added a commit addressing this.

rsmmr (approved):

PR #1686, opened 2024-03-06
Constness fixes and improvements

rsmmr:
- **Infer constness of `<unit>.context()` from that of `<unit>`.**
- **Fix constness validation for unit field assignments.**
- **Simplify constness check for struct field assignments.**
- **Ensure constness of result for const map/vector index operator.**

This is on top of #1684.


hilti/toolchain/src/ast/operators/map.cc
```diff
@@ -255,7 +255,7 @@ class IndexConst : public Operator {
     }
 
     QualifiedTypePtr result(Builder* builder, const Expressions& operands, const Meta& meta) const final {
-        return operands[0]->type()->type()->as<type::Map>()->valueType();
+        return operands[0]->type()->type()->as<type::Map>()->valueType()->recreateAsConst(builder->context());
```
bbannier, line 258: Nit: I wonder whether we need similar treatment for iterators into const collections which we do not seem to model.

bbannier (approved):

hilti/toolchain/src/ast/operators/map.cc
rsmmr: I'll check.

PR #1687, opened 2024-03-08
Introduce custom RTTI system for casting safely between node types.

rsmmr:
Our custom system is faster than C++'s `dynamic_cast<>`, resulting in
a noticeable speed-up for larger parsers.

We assign a unique integer tag to  each `Node`-derived class. Each
class' constructors pass an array through to the top-level `Node`
class that contains a series of these integers describing the
inheritance path from the derived class back up to `Node`. The `Node`
class stores this path for fast type checks. In addition, each
`Node`-derived class `T` gets a couple constants as members: (1) a
copy of its own tag (`T::NodeTag`), and (2) a level indicating its
distance from `Node` in the inheritance tree (`T::NodeLevel`). The
level is used as index into an instance's tag array when performing
type checks. The result is that we can do `isA<T>` operation with just
a single comparison between an array element and a constant value.

This system is optimized for our very simple Node hierarchy: not very
deep (max. 4 entries in the tag array), single-inheritance only,
derived class are always the same distance from `Node`, and
all `Node`-derived classes are known upfront.


rsmmr:
The patch is large, but most changes are mechanical across the AST classes.

hilti/toolchain/include/ast/node-tag.h
bbannier, line 1: There is a lot of knowledge about the node hierarchy encoded in these magic numbers which is duplicated in the `NodeTags` we create from it. Keeping this consistent seems like a lot of work.

How do you feel about using a single enum to assign the values? To do that we would move the current namespaces into the tag identifier (which would require some tweaking in the macros), e.g.,

```cpp
enum tag {
    tag_Node = 1,

    // Numbered consecutively.
    tag_ASTRoot,
    tag_Attribute,
    tag_AttributeSet,
    tag_Ctor,
    tag_Declaration,
    tag_Expression,
    tag_Function,
    tag_QualifiedType,
    tag_Statement,
    tag_UnqualifiedType,
```

To make translating values to tag names easier we could e.g., use x-macros.

The other benefits of using values in an enum would be that the enum values can be used in `constexpr` contexts so with`std::array`'s `constexpr` ctr more optimization opportunities come up (but we could also just make all the `const` `Tag`s here `constexpr`).

bbannier (dismissed):
This looks like a good direction, just wondering whether we could simplify maintenance of the data required for this.

It seems that the approach here doesn't make it too hard to add additional AST nodes, but I am unsure whether they could e.g., be added from external code (might be possible?).

hilti/toolchain/include/ast/node-tag.h
rsmmr: > How do you feel about using a single enum to assign the values?

I'd prefer to use an enum but not sure I see how to do that. Couple challenges:

1. The Spicy code needs to be able to extend the set of tags for its own nodes. As one can't add new enum tags to an already existing type, it'd need to be a new additional enum type of some kind, plus casting I guess. Can't really see how to make this work nicely?

2. The macros would probably need another parameter to build those `tag_` values as I don't think the current ones lend themselves to that scheme.

> To make translating values to tag names easier we could e.g., use x-macros.

Had to lookup x-macros, didn't know those (at least not by that name). But actually not sure this is really better: Unless I'm missing something that looks like it wouldn't be that good to optimize in our case, readability would suffer, and I don't see how to extend that on the Spicy side?

> The other benefits of using values in an enum would be that the enum values can be used in `constexpr` contexts so with`std::array`'s `constexpr` ctr more optimization opportunities come up (but we could also just make all the `const` `Tag`s here `constexpr`).

Good point. I was assuming that the compiler could already optimize the `const unsigned ints` aways but we can make them all `constexpr`.

> There is a lot of knowledge about the node hierarchy encoded in these magic numbers which is duplicated in the `NodeTags` we create from it. Keeping this consistent seems like a lot of work.

The numbers themselves have no magic meaning at all, they just need to be unique, that's all. They are arranged/spaced out that way in the source file just to make it easier to maintain, but in the end the values don't matter. What matters is the namespacing because the macros need that. However, the namespacing is exactly the same as what the AST classes use, so I believe that shouldn't be too surprising either.

rsmmr:
> It seems that the approach here doesn't make it too hard to add additional AST nodes, but I am unsure whether they could e.g., be added from external code (might be possible?).

Yeah, that's essentially what the Spicy-side is doing. Two parts to that: adding new `Tags` is easy as long as one doesn't conflict with existing values; so one can just reserve some so far unused integer range (which Spicy is essentially doing). Some of the macros are replicated on the Spicy side for change in the visitor class, but the common parts are factored out and that's probably something not needed by other external code.

bbannier (approved):

PR #1694, opened 2024-03-15
Add Cirrus tasks running non-brew builds and tests.

bbannier:


rsmmr (dismissed):

rsmmr (commented):
Actually, why not leave one Homebrew task in? Would be good to test both.

bbannier:
> Actually, why not leave one Homebrew task in? Would be good to test both.

I left in the Ventura Homebrew task.

PR #1697, opened 2024-03-19
Add control-flow based optimizer passes

bbannier:


3rdparty/CMakeLists.txt
```diff
@@ -49,3 +49,13 @@ check_cxx_compiler_flag("-Wno-changes-meaning" _has_no_changes_meaning_flag)
...
+
+include(FetchContent)
+FetchContent_Declare(
+    cxxgraph
+    GIT_REPOSITORY https://github.com/ZigRazor/CXXGraph.git
+    GIT_TAG v3.1.0
+)
+FetchContent_MakeAvailable(cxxgraph)
+add_library(CXXGraph INTERFACE)
+target_include_directories(CXXGraph SYSTEM INTERFACE ${cxxgraph_SOURCE_DIR}/include)
```
bbannier, line 61: This is a temporary solution until we decide on a proper graph library (or implement one ourselves); we cannot use this library as it is licensed under AGPL-3.0. It also seems to be of poor quality (e.g., major bugs around directed graphs, poor test suite).

bbannier: They have since switched to the Mozilla Public License so we could use their code. We probably should still migrate off it to e.g., a custom container since we do not need their features and using their API probably makes our code more complicated and computationally expensive.

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -0,0 +1,893 @@
...
+
+        else if ( stmt->isA<statement::Declaration>() )
+            // Names in declaration statements appear on the RHS.
+            transfer.use.insert(decl);
+
+        else if ( auto* global = stmt->tryAs<declaration::GlobalVariable>() ) {
+            // Names in the global declaration appear on the RHS.
+            transfer.use.insert(decl);
+
+            // FIXME(bbannier): handle local declarations as well?
```
evantypanski: Currently not aliasing here causes some bad removals, I think:

```
$ cat test.spicy
module Test;

function f() {
    local x = new 123;
    local y = x;
    *y = 456;
    print x;
    *y = 789;
    print x;
}

f();
$ spicyc -j test.spicy
123
123
$ spicyc -gj test.spicy
456
789
```

I'd kind of rather handle this now than deal with some gross bugs

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,142 @@
...
+struct Reachability {
+    std::unordered_set<Node*> in;
+    std::unordered_set<Node*> out;
+};
+
+struct Transfer {
+    std::unordered_set<Node*> use;
+    std::unordered_map<Node*, Node*> gen;
+    std::unordered_map<Node*, std::unordered_set<Node*>> kill;
+    std::unordered_set<Node*> aliases;
```
evantypanski: We'll never be 100% sure for all nodes `x, y` that `x` does/does not alias `y`.

So I'd prefer this be called `may_alias` or something similar

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -0,0 +1,893 @@
...
+    }
+    else
+        addEdge(parent, scope_end);
+
+    return scope_end;
+}
+
+Node* CFG::addCall(Node* parent, operator_::function::Call* call) {
+    auto* c = getOrAddNode(call);
+    addEdge(parent, c);
```
evantypanski: Something about calls creates two edges in the CFG, but I'm not sure what, or if that changes anything:

```
# test.hlt

module Foo {

function void f2(uint<64> x) {}

function void f() {
    local uint<64> x = 5;
    f2(x);
}

f();

}
```

cfg

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -0,0 +1,893 @@
...
+            return;
+
+        auto* stmt = root;
+        // If the statement was a simple `Expression` unwrap it to get the more specific node.
+        if ( stmt->isA<statement::Expression>() ) {
+            if ( auto* child = stmt->child(0) )
+                stmt = child;
+        }
+
+        if ( auto* assign = stmt->tryAs<expression::Assign>() ) {
```
evantypanski, line 589: Here's one of those places that assignment expressions will bite. There are TONS of complicated interactions that can happen here (I'm also thinking assignments on the right-hand-side of a shortcircuited operator like `||`). But, here's a simple case that gets broken:

```
module Foo {

import hilti;

function void f() {
    local uint<64> x;
    x = 1;
    local bool b = ((x = 5) == 5);
    hilti::print(x);
}

f();

}
```

Since the `local bool b` is removed, the `x = 5` is not output, so with optimizations this prints `1`, without it prints `5`.

I haven't dove further into this, but you may also hit problems if you just keep all declarations that have assignments in them if you blindly assume that the assignment shadows previous ones because of short circuiting. For example, have to make sure that both possible values of `x` are live after this:

```
x = 5;
local bool b = (y == 5) || ((x = 6) == 6);
```

Or else you might wrongly assume that `x == 6` after that point.

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -0,0 +1,893 @@
...
+        for ( auto&& [n, transfer] : _dataflow ) {
+            for ( const auto* alias : transfer.aliases ) {
+                auto* stmt = find_node(alias);
+                if ( ! stmt || ! _dataflow.count(stmt) )
+                    // Could not find node declaring aliased name.
+                    util::detail::internalError(
+                        util::fmt(R"(could not find CFG node for "%s" aliased in "%s")", alias->print(), n->print()));
+
+
+                _dataflow.at(stmt).aliases.insert(n);
```
evantypanski: Could this be done when inserting the alias in the first place, then just use a helper function there? Since aliases are symmetric I'd expect that to be encoded on the initial insertion if possible, this feels mixed up to me

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -0,0 +1,893 @@
...
+
+        else if ( auto&& throw_ = c->tryAs<statement::Throw>() )
+            parent = addThrow(parent, *throw_, scope_end);
+
+        else if ( auto&& return_ = c->tryAs<statement::Return>() )
+            parent = addReturn(parent, return_->expression());
+
+        else if ( c->isA<statement::Continue>() || c->isA<statement::Break>() )
+            // `continue`/`break` statements only add flow, but no data.
+            parent = addReturn(parent, nullptr);
```
evantypanski: I'm a little confused by this, since `expression` is `nullptr` this always returns `parent`, so this is just `parent = parent;`, right? Are there plans for more here?

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -0,0 +1,893 @@
...
+    auto last = std::find_if(stmts.begin(), stmts.end(), [](auto&& c) {
+        return c && (c->template isA<statement::Return>() || c->template isA<statement::Throw>() ||
+                     c->template isA<statement::Continue>() || c->template isA<statement::Break>());
+    });
+    const bool has_dead_flow = last != stmts.end();
+    if ( has_dead_flow )
+        last = std::next(last);
+
+    // Node this block will eventually flow into.
+    auto* scope_end = getOrAddNode(create_meta_node<ScopeEnd>(scope));
```
evantypanski: (not sure exactly where this comment applies)

Scoping for blocks gets into a similar problem to lambda captures, so this code will incorrectly remove a dead store of `x = 5;`:

```
function void f() {
    local uint<64> x = 1;
    { x = 5; local uint<64> x = 10; }
    hilti::print(x);
}
```

Generally the solution would be to add a counter "how many scopes out until I found `x`?" - I guess that'd be a de Bruijn index

Though some other stuff could work too, with symbol tables attached to the scopes or transfers or something (I haven't really thought much about it)

hilti/toolchain/include/base/graph.h
```diff
@@ -0,0 +1,115 @@
...
+    using EdgeId = EdgeId_;
+
+    std::unordered_map<NodeId, T> _nodes;
+    std::unordered_map<EdgeId, std::pair<NodeId, NodeId>> _edges;
+
+    const auto& nodes() const { return _nodes; }
+    const auto& edges() const { return _edges; }
+
+    std::optional<NodeId> getNodeId(const T& x) const {
+        if ( auto it = std::find_if(_nodes.begin(), _nodes.end(), [&](auto&& n) { return n.second == x; });
```
rsmmr, line 39: Linear search can become expensive for large graphs. Is that a problem? (Note that `addNode()` always calls this.)

hilti/toolchain/include/base/graph.h
```diff
@@ -0,0 +1,115 @@
...
+#include <algorithm>
+#include <cstdint>
+#include <optional>
+#include <set>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+namespace hilti::util::graph {
+enum class Direction : char { Downstream, Upstream };
```
rsmmr: I find these names a bit unintuitive. How about just `In` and `Out`?

hilti/toolchain/src/compiler/cfg.cc
bbannier: The important point is that we handle them, I'll replace this with an empty block.

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1575,28 +1707,37 @@ void detail::optimizer::optimize(Builder* builder, ASTRoot* root) {
...
 
     // If no user-specified passes are given enable all of them.
     if ( ! passes ) {
         passes = std::set<std::string>();
         for ( const auto& [pass, _] : creators )
-            passes->insert(pass);
+            if ( pass != "cfg" )
+                passes->insert(pass);
+
+        passes->insert("cfg");
```
rsmmr: Why this special-casing here? We control the order of `creators` anyway.

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,9 +1593,105 @@ struct MemberVisitor : OptimizerVisitor {
...
+
+            for ( auto&& x : cfg.unreachableStatements() )
+                modified |= remove_node(cfg, x, "statement result unused");
+
+            if ( modified )
+                break;
+
+            auto unreachable_nodes = cfg.unreachableNodes();
+            if ( unreachable_nodes.empty() )
+                break;
```
rsmmr: Can skip the empty check, just for consistency with the case before.

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,142 @@
...
+    [[nodiscard]]
+    Node* addIf(Node* parent, const statement::If& if_);
+    [[nodiscard]]
+    Node* addTryCatch(Node* parent, const statement::Try& try_);
+    [[nodiscard]]
+    Node* addReturn(Node* parent, const Node* expression);
+    [[nodiscard]]
+    Node* addThrow(Node* parent, statement::Throw& throw_, Node* scope_end);
+    [[nodiscard]]
+    Node* addCall(Node* parent, operator_::function::Call* call);
```
rsmmr: Can these all move into `private`? More generally, make clear what the public API is for this class.

rsmmr: Also, not necessarily for now, but we'll probably want methods at some point to access the reachability and transfer information from the outside (i.e. from the optimizer). 

There's a more general question what kind of logic running on top of the graph should be inside the class here, vs. done externally by whoever needs it. Example: `unreachableStatements()` could also be computed by the optimizer, based on graph that is has access to. 

I think I'm leaning towards moving more into the optimizer, and letting the CFG be just a graph providing information. Then that `std::unordered_map<Node*, Transfer>` would essentially become the main API for the class.

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -0,0 +1,893 @@
...
+            // `continue`/`break` statements only add flow, but no data.
+            parent = addReturn(parent, nullptr);
+
+        else if ( auto&& call = c->tryAs<operator_::function::Call>() )
+            parent = addCall(parent, call);
+
+        else if ( auto&& block = c->tryAs<statement::Block>() )
+            parent = addBlock(parent, block->statements(), block);
+
+        else {
```
rsmmr: Why not use a visitor here, then we also don't need to all those individual `add*` methods?

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,142 @@
...
+    // Some versions of GCC incorrectly diagnose maybe uninitialized members here.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpragmas"
+#pragma GCC diagnostic ignored "-Wunknown-warning-option"
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+    MetaNode(node::Tags node_tags) : Node(nullptr, node_tags, {}, {}) {}
+#pragma GCC diagnostic pop
+
+    uint64_t counter = instances++;
+    static uint64_t instances;
```
rsmmr: What's this needed for?

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,142 @@
...
+        auto* r = n.get();
+        _meta_nodes.insert(std::move(n));
+        return r;
+    }
+
+    Node* getOrAddNode(Node* n);
+    void addEdge(Node* from, Node* to);
+
+    // Add flow for globals if `root` corresponds to a global module block.
+    [[nodiscard]] Node* addGlobals(Node* parent, const Node& root);
```
rsmmr: Here and elsewhere: I find the name `parent` misleading because it implies a hierarchy. `predecessor`?

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,142 @@
...
+
+struct Reachability {
+    std::unordered_set<Node*> in;
+    std::unordered_set<Node*> out;
+};
+
+struct Transfer {
+    std::unordered_set<Node*> use;
+    std::unordered_map<Node*, Node*> gen;
+    std::unordered_map<Node*, std::unordered_set<Node*>> kill;
```
rsmmr: I'm not sure I fully understand `kill`. Example:

```
function void f(inout int<64> i) {
     i = 1;
     i = 2;
 }
```
f

Looking at node 0, why is it killing `i: [i = 2;]`?

rsmmr (commented):
I'm in the process of taking a high-level look. Just a flagging a few pieces I noticed so far (some pretty random). I'll continue looking. 

(We'll need more documentation/comments, and some tweaking of coding convention stuff, but I'll leave that for later. Btw, is `.clang-tidy.toolchain` still being used?)

hilti/toolchain/include/compiler/detail/cfg.h
rsmmr: One additional observation from reading over the code: the CFG is captured in terms of generic `Node` instances, both for the graph itself and the API functions. However, only certain types of `Node` can actually occur in a graph (statements and call expressions, I believe). It would be nice to capture that restriction more directly somehow, maybe through some joint parent class for those nodes that can be part of a CFG, and then working with pointers to that parent (though not sure if that works that way).  I think that would make the code quite a bit more readable, because it would always give you the context on what type of node is expected.

hilti/toolchain/include/compiler/detail/cfg.h
bbannier: Left-over code I forgot to remove, dropping it.

hilti/toolchain/src/compiler/cfg.cc
bbannier: That would look much cleaner, yes. The tricky bit is that unless we enumerate all possible nodes I would need a generic visitor for the `else` case which matches on generic `Statement*` nodes so e.g., a `statement::Return` would in a naive implementation be visited twice (once as `statement::Return*` and once as `Statement*`). Let me check whether I can restructure the implementation.

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -0,0 +1,893 @@
...
+
+    std::vector<Node*> result;
+    for ( auto&& [n, uses] : uses ) {
+        if ( uses > 0 )
+            continue;
+
+        if ( _dataflow.at(n).keep )
+            continue;
+
+        result.push_back(n);
```
evantypanski: (not sure where this applies)

Globals are a little weird here when combined with locals, for example:

```
module Test {

import hilti;

global uint<64> x = 0;

function void f() {
    x = x + 1;
    local uint<64> x = 5;
    hilti::print(x);
}

f();
hilti::print(x);
f();
hilti::print(x);
f();
hilti::print(x);
f();
hilti::print(x);

}
```

This should keep the increment of global `x` so it prints `5` then `1` then `5` then `2` etc.

But instead it prints `5` and `0` repeatedly, so it removes the `x = x + 1;`

Note that the case without optimizations fails to compile because of #2086 though... so it's probably not that big of a deal :)

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,9 +1593,105 @@ struct MemberVisitor : OptimizerVisitor {
...
+                modified |= remove_node(cfg, n, "unreachable code");
+
+            if ( ! modified )
+                break;
+        }
+    }
+
+    void operator()(declaration::Function* f) override {
+        if ( auto&& body = f->function()->body() )
+            visit_node(body);
```
evantypanski, line 1997: I think this covers everything but inline methods, this doesn't end up getting optimizations:

```
type S = struct {
    method void f() { } # No CFG!
};
```

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -0,0 +1,893 @@
...
+
+                    else if ( auto&& p = n_->tryAs<declaration::Parameter>();
+                              p && p->kind() == parameter::Kind::InOut ) {
+                        ++uses[in];
+                    }
+
+                    else if ( const auto* expr = n_->tryAs<declaration::Expression>() ) {
+                        if ( auto* keyword = expr->expression()->tryAs<expression::Keyword>();
+                             keyword && keyword->kind() == expression::keyword::Kind::Self )
+                            ++uses[in];
```
evantypanski: Is this run after the aliases are marked as uses? If so I think a case like this would fail (if aliases were good on locals, regardless it gets an incorrect result now):

```
module Test {

import hilti;

type S = struct {
    uint<64> i;

    method void print() {
        hilti::print(self.i);
    }

    method void f();
} &on-heap;

method void S::f() {
    self.print();
    local strong_ref<S> ref = self;
    ref.i = 1;
    self.print();
    ref.i = 2;
    self.print();
}

global S s = [$i = 0];
s.f();

}
```

should print `0 1 2`, prints `0 0 0`

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,142 @@
...
+    void populateReachableExpressions();
+    std::vector<Node*> unreachableStatements() const;
+
+    util::graph::DirectedGraph<Node*, uintptr_t> g;
+
+private:
+    std::unordered_set<std::unique_ptr<MetaNode>> _meta_nodes;
+    std::unordered_map<Node*, Transfer> _dataflow;
+    Node* _begin;
+    Node* _end;
```
evantypanski: Is `_end` the end of the *block*, the *function*, etc?

More of asking as a way to say that something is slightly broken, but may not currently manifest much. Given this code:

```
module Test {

import hilti;

function void f() {
    {
        hilti::print("hey");
        return;
    }
    hilti::print("bad!");
}

f();

}
```

The optimization pass doesn't remove `hilti::print` with the scope, but does without it. `return` should point to the end of the *function*, not block - then unreachable nodes need to account for that (the issue might be there and not with what this `_end` is?)

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,142 @@
...
+    const Node* scope;
+};
+
+struct Reachability {
+    std::unordered_set<Node*> in;
+    std::unordered_set<Node*> out;
+};
+
+struct Transfer {
+    std::unordered_set<Node*> use;
```
evantypanski: (this doesn't go anywhere in particular, just on `use` since it's related ish)

So with these changes, we start to run into problems: What if we remove code that would otherwise throw an exception? This is trivially easy to trigger:

```
module Test {

18446744073709551615 + 1;

}
```

this HILTI code throws an exception without optimizations, but not with optimizations.

Normally this is just hand-waved in various ways (undefined behavior, static safety checks, etc). Then the compiler is free to assume it won't happen. What is Spicy's policy on this sort of thing?

I don't think it's reasonable to say that we'll keep all arithmetic (and... probably all of everything) because of exceptions. I just want to know what line we can draw, if any, for cases like this.

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -0,0 +1,893 @@
...
+            else if ( meta->isA<End>() )
+                ss << util::fmt("    %s [label=end shape=Msquare %s];\n", id, xlabel ? *xlabel : "");
+
+            else if ( meta->isA<Flow>() )
+                ss << util::fmt("    %s [shape=point %s];\n", id, xlabel ? *xlabel : "");
+
+            else if ( auto* scope = meta->tryAs<ScopeEnd>() ) {
+                ss << util::fmt("    %s [label=\"scope_end %s\" shape=triangle %s];\n", id,
+                                scope->scope->meta().location(), xlabel ? *xlabel : "");
+            }
```
evantypanski: Not having a `ScopeBegin` is a readability nightmare for these CFGs with shadowed variables :(

```
module Test {

import hilti;

function void f() {
    local x = 5;
    hilti::print(x);
    {
        local x = 6;
        hilti::print(x);
        {
            local x = 7;
            hilti::print(x);

        }
    }
}

f();

}
```
cfg

evantypanski (commented):
OK I did round 2, this is all I have for now.

BTW for a lot of the edge cases, I'd be happy if this gets merged behind a feature flag then convert them into issues. That way we can split this up and get it in a bit sooner

hilti/toolchain/src/compiler/optimizer.cc
bbannier: While iterating on the code I had this pass run last (and all passes in a `vector`). This got dropped in the meantime, but I forgot to adjust this.

I'll introduce a feature flag so I'll use separate handling again now.

hilti/toolchain/src/compiler/cfg.cc
bbannier: I'll keep this as is for now to avoid having to deal with invocations of multiple handlers.

hilti/toolchain/include/compiler/detail/cfg.h
bbannier: I'd prefer if we could classify this as simply "dead code" since it has no side effect outside of possibly throwing; I actually fixed a couple test to have additional side effects, like printing something.

Since users have no way to influence program behavior once code has thrown (outside of printing the exception with an error hook) this seems fine. ~The whole situation becomes much more complicated if we every implemented catching, but thankfully we are not there yet.~ HILTI actually has explicit `try`/`catch` blocks, but I think the argument about missing side effects should still hold.

hilti/toolchain/include/compiler/detail/cfg.h
bbannier: Yeah, `_end` was intended as the final node for some global visited  `Node`. We store `_begin` and `_end` so we can trace flows through e.g., function bodies. I also connected nodes for `Return` to `_end` to capture early exit.

There were some inconsistencies in how it was treated as opposed to `ScopeEnd`, so I cleaned this up and used a single node for both (now just `End` which always has some scope). I also cleaned up the handling of `Return` without expression which previously would store nothing in the CFG; now we still store the return so we can remove it.

hilti/toolchain/src/compiler/cfg.cc
bbannier: It was missing handling for `statement::Expression` in `addBlock`. I added that and these extra edges now go away.

hilti/toolchain/include/compiler/detail/cfg.h
evantypanski: Yeah I think the smart thing here would be to not worry about exceptions - no `throw` should get explicitly removed, but we don't guarantee exception semantics with optimizations. Exceptions for something like integer overflows are probably better classified as an implementation detail to ensure safe code.

That might change how things parse in rare cases, but that's probably worth it.

hilti/toolchain/src/compiler/cfg.cc
bbannier: This should work now.

hilti/toolchain/src/compiler/cfg.cc
bbannier: Yeah, it is not ideal, but outside of these dot graphs there is not strong case for introducing an explicit scope begin yet, so I'd like to punt on that.

Note I have cleaned up the CFG so we do not emit these dead expressions anymore.

hilti/toolchain/src/compiler/cfg.cc
bbannier: This should be fixed now, in the existing framework.

Since we always uses by `Name` which in valid code always resolves to exactly one declaration there should be no issue with us confusing the outer and the inner `x`. I added a test to make sure things stay that way.

hilti/toolchain/include/compiler/detail/cfg.h
bbannier: `kill` isn't flow-aware at all, instead if a node generates an update to some value `i` it kills all other nodes which generate an update. From the implementation side this is not hard to do as we can just iterate all nodes in the graph, but it leads to artifacts (hopefully without ill effects) where a node would kill an update which only happens later; we could remove these artifacts by only keeping updates which flow into the current node, but this would require reachability analysis.

I'd like to punt on this for now, and would propose documenting this in a `TODO`.

bbannier: FTR, we discussed this more offline, and the issue is that what this currently has as "kill sets" isn't really a kill set. We should instead align this more with the classic meaning, i.e., for each node in the graph track reachable inputs, and kill sets come then naturally. Having that data available in each node might also make it unnecessary to inspect the full graph to gather related information.

hilti/toolchain/include/compiler/detail/cfg.h
bbannier: - I renamed the helpers and made them `private`
- I moved `unreachableStatements` and `unreachableNodes` into the optimizer pass, and added an accessor for `CFG::_dataflow` so it can implement them.
- I introduced a dedicated class to distinguish graph nodes from AST nodes. This class still just wraps a `Node`, but can only be constructed from a subset of AST nodes (currently seven). This isn't exactly what you suggested, but still makes the code more readable, and enforces more correct use (I actually fixed a number of issues along the way)

hilti/toolchain/include/compiler/detail/cfg.h
bbannier: I switched this over to classical `kill` sets, and now also compute and track this directly together with the `in` and `out` sets.

hilti/toolchain/include/base/graph.h
```diff
@@ -0,0 +1,115 @@
...
+#include <optional>
+#include <set>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+namespace hilti::util::graph {
+enum class Direction : char { Out, In };
+
+// FIXME(bbannier): remove customizable NodeId, EdgeId.
```
rsmmr: There are number of FIXMEs in this header. Are they easy to address before merge?

hilti/toolchain/include/base/graph.h
```diff
@@ -0,0 +1,115 @@
...
+template<typename T, typename NodeId_ = std::uint64_t, typename EdgeId_ = std::uint64_t>
+struct DirectedGraph {
```
rsmmr: It would be nice to comment the API in this class in our usual style.

hilti/toolchain/include/ast/node.h
```diff
@@ -358,7 +358,7 @@ class Node {
     std::string typename_() const { return _typename(); }
 
     /** Returns a globally unique numeric identifier for the node. */
-    uintptr_t identity() const { return reinterpret_cast<uintptr_t>(this); }
+    uint64_t identity() const { return count; }
```
rsmmr: Im curious: what is this change needed for? It adds an `uint64_t` to every node, and we have many of them. Not the end of the world, but still.

hilti/toolchain/include/ast/node.h
```diff
@@ -1051,6 +1051,9 @@ class Node {
 
     std::unique_ptr<Scope> _scope = nullptr; // scope associated with the node, or null if non (i.e., scope is empty)
     std::unique_ptr<std::vector<node::Error>> _errors; // errors associated with the node, or null if none
+
+    static uint64_t counter;
+    uint64_t count = counter++;
```
rsmmr: Prefix with underscore for private members.

rsmmr: Suggest to call this `_identity`.

doc/toolchain.rst
```diff
@@ -91,6 +91,10 @@ control the compilation process:
     ``HILTI_PRINT_SETTINGS``
         Set to see summary of compilation options.
 
+    ``HILTI_OPTIMIZER_ENABLE_CFG``
+        Set to `1` to enable experimental control-flow based optimizations.
+        This might break valid code.
```
rsmmr: So is the assumption that it's not yet working correctly? Then let's add a `.. todo` here and day there that it's work in progress.

tests/Baseline/spicy.optimization.default-parser-functions/opt.hlt
```diff
@@ -68,6 +68,7 @@ method tuple<const view<stream>, int<64>, const iterator<stream>, optional<hilti
 
     }
 
+    (*self).__error = __error;
```
rsmmr: There is a commit that sounds like it would address this now showing up, but this remains in the baseline here and for a number of other tests. That expected?

hilti/toolchain/include/base/graph.h
```diff
@@ -0,0 +1,115 @@
...
+
+    std::optional<std::pair<NodeId, NodeId>> getEdge(EdgeId id) const {
+        if ( auto it = _edges.find(id); it != _edges.end() )
+            return it->second;
+
+        return {};
+    }
+
+    // FIXME(bbannier): this should return a `Vec<(EdgeId, NodeId)`>.
+    std::vector<NodeId> neighbors(NodeId id, Direction dir) const {
```
rsmmr, line 156: Make this private given that we have the upstream/downstream versions? Or alternative: remove the upstream/downstream versions, no need to have both I'd say.

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,171 @@
...
+    GraphNode addFor(GraphNode predecessor, const statement::For& for_);
+    GraphNode addWhile(GraphNode predecessor, const statement::While& while_, GraphNode scope_end);
+    GraphNode addIf(GraphNode predecessor, const statement::If& if_);
+    GraphNode addTryCatch(GraphNode predecessor, const statement::Try& try_);
+    GraphNode addReturn(GraphNode predecessor, const statement::Return& return_);
+    GraphNode addThrow(GraphNode predecessor, statement::Throw& throw_, GraphNode scope_end);
+    GraphNode addCall(GraphNode predecessor, operator_::function::Call& call);
+
+    // Add flow for globals if `root` corresponds to a global module block.
+    GraphNode addGlobals(GraphNode predecessor, const Node& root);
```
rsmmr: Prefix private methods with underscore.

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,171 @@
...
+    GraphNode getOrAddNode(GraphNode n);
+    void addEdge(const GraphNode& from, const GraphNode& to);
+
+    std::string dot() const;
+
+    void populateDataflow();
+
+    const auto& dataflow() const { return _dataflow; }
+
+    util::graph::DirectedGraph<GraphNode, uintptr_t> g;
```
rsmmr: No public data members please.

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,171 @@
...
+
+    bool keep = false; // Whether this node should be kept.
+};
+
+class CFG {
+public:
+    CFG(const Node* root);
+
+    template<typename T, typename... Args>
+    MetaNode* create_meta_node(Args... args)
```
rsmmr: c
```suggestion
    MetaNode* createMetaNode(Args... args)
```

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,171 @@
...
+
+    std::unordered_map<Node*, GraphNode> gen;
+
+    std::unordered_set<Node*> read;
+    std::unordered_set<Node*> write;
+
+    bool keep = false; // Whether this node should be kept.
+};
+
+class CFG {
```
rsmmr, line 169: Please document the API, including semantics of any ordering dependencies: what are the constraints between adding nodes & edges and computing the dataflow? (Do I need to recompute after each change? Can I even add more that has been computed?)

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,171 @@
...
+        assert(scope); // Should always contain a valid scope.
+    }
+
+    HILTI_NODE_1(End, MetaNode, final);
+
+    const Node* scope;
+};
+
+// The data we hold in the control flow graph.
+class GraphNode {
```
rsmmr, line 82: Yeah, I like this.

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,171 @@
...
+namespace std {
+template<>
+struct hash<hilti::detail::cfg::GraphNode> {
+    auto operator()(const hilti::detail::cfg::GraphNode& n) const { return n.value() ? n->identity() : 0; }
+};
+} // namespace std
+
+namespace hilti::detail::cfg {
+
+struct Transfer {
```
rsmmr, line 132: Please document the semantics of the fields here pretty thoroughly. This is the main API to all the information provided by the CFG.

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,171 @@
...
+    std::unordered_set<GraphNode> kill;
+
+    std::unordered_set<Node*> maybe_alias;
+
+    std::unordered_map<Node*, GraphNode> gen;
+
+    std::unordered_set<Node*> read;
+    std::unordered_set<Node*> write;
+
+    bool keep = false; // Whether this node should be kept.
```
rsmmr: By whom?

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,171 @@
+// Copyright (c) 2020-now by the Zeek Project. See LICENSE for details.
```
rsmmr, line 1: I noted it explicitly at a couple of places in this file below, but more generally this is another file where adding more documentation to the various pieces would be quite helpful going forward.

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,171 @@
...
+    MetaNode* create_meta_node(Args... args)
+        requires(std::is_base_of_v<MetaNode, T>)
+    {
+        auto n = std::make_unique<T>(args...);
+        auto* r = n.get();
+        _meta_nodes.insert(std::move(n));
+        return r;
+    }
+
+    GraphNode getOrAddNode(GraphNode n);
```
rsmmr: Can this be private? Can't find a use outside of CFG.

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,171 @@
...
+    void addEdge(const GraphNode& from, const GraphNode& to);
```
rsmmr: I only see a single usage of this outside of CFG. Wondering if there's a different way to do that so this could become private too?

Essentially, I think it'd be nice if the CFG would just be constructed behind the scenes, and the public API is just access to it; would make semantics & dependencies much more clear.

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -37,10 +44,17 @@
 #include <hilti/base/logger.h>
 #include <hilti/base/timing.h>
 #include <hilti/base/util.h>
+#include <hilti/compiler/detail/cfg.h>
+
+#include "ast/expressions/assign.h"
+#include "ast/statements/expression.h"
```
rsmmr:
```suggestion
#include <hilti/ast/expressions/assign.h>
#include <hilti/ast/statements/expression.h>
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1858,9 +1906,284 @@ struct FunctionParamVisitor : OptimizerVisitor {
...
+
+        return result;
+    }
+
+    bool prune_uses(Node* node) override {
+        visitor::visit(*this, node);
+        return isModified();
+    }
+
+    bool remove_node(detail::cfg::CFG& cfg, Node* data, const std::string& msg = {}) {
```
rsmmr: Add some comment here on what that the method is doing, and maybe rename to `remove_dead_node` for clarity (if I understand it right).

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1858,9 +1906,284 @@ struct FunctionParamVisitor : OptimizerVisitor {
...
+
+    void operator()(declaration::Module* m) override {
+        OptimizerVisitor::operator()(m);
+
+        if ( auto&& body = m->statements() )
+            visit_node(body);
+    }
+};
+std::vector<Node*> FunctionBodyVisitor::unreachableStatements(const detail::cfg::CFG& cfg) const {
+    // This can only be called after dataflow information has been populated.
```
rsmmr, line 2009: Move this comment to the method prototype. Same for unreachableNodes.

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1858,9 +1906,284 @@ struct FunctionParamVisitor : OptimizerVisitor {
...
+            for ( auto&& [decl, stmts] : transfer.in ) {
+                // If the statement generated an update to the value ...
+                bool mark_used = false;
+
+                if ( decl->isA<declaration::GlobalVariable>() )
+                    mark_used = true;
+
+                else if ( auto&& p = decl->tryAs<declaration::Parameter>(); p && p->kind() == parameter::Kind::InOut ) {
+                    mark_used = true;
+                }
```
rsmmr: Style nit, here and elsewhere: no braces around single-line blocks.

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -0,0 +1,892 @@
...
+#include <hilti/hilti/ast/types/list.h>
+#include <hilti/hilti/ast/types/map.h>
+#include <hilti/hilti/ast/types/set.h>
+#include <hilti/hilti/ast/types/stream.h>
+#include <hilti/hilti/ast/types/vector.h>
+
+namespace hilti::detail::cfg {
+
+// Helper function to detect whether values of a type can alias their inputs.
+static bool is_aliasing_type(const UnqualifiedType& type) {
```
rsmmr: Naming style: 
```suggestion
static bool _isAliasingType(const UnqualifiedType& type) {
```

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -0,0 +1,892 @@
...
+         type.isA<type::vector::Iterator>() || //
+         type.isA<type::StrongReference>() ||  //
+         type.isA<type::WeakReference>() ||    //
+         type.isA<type::ValueReference>() )
+        return true;
+
+    return false;
+}
+
+static bool contains(const Node& outer, const Node& inner) {
```
rsmmr:
```suggestion
static bool _contains(const Node& outer, const Node& inner) {
```

rsmmr: Also add a comment what function does.

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -0,0 +1,892 @@
...
+        addEdge(last, _end);
+
+    // Clean up artifacts from CFG construction.
+    //
+    // - `End` nodes with no incoming edges. These can arise if blocks never
+    //   flow through their end node, e.g., due to early return.
+    while ( true ) {
+        std::set<uintptr_t> dead_ends;
+
+        for ( auto&& [id, n] : g.nodes() ) {
```
rsmmr: Style, here and elsewhere: use `const auto&` for iterations that don't modify the values. (Also in various `if` statements).

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,171 @@
...
+    std::unordered_map<Node*, std::unordered_set<GraphNode>> in;
+    std::unordered_map<Node*, std::unordered_set<GraphNode>> out;
+    std::unordered_set<GraphNode> kill;
+
+    std::unordered_set<Node*> maybe_alias;
+
+    std::unordered_map<Node*, GraphNode> gen;
+
+    std::unordered_set<Node*> read;
+    std::unordered_set<Node*> write;
```
rsmmr: Can we narrow what kind of Nodes are stored here? Specifically, I'm wondering if these may store Declarations only?

rsmmr (commented):
I didn't look too closely over the actual logic, I think you guys have combed through that already. However, I noticed a lot of style & code documentation issues, and I left various comments in that regards. We could leave those to later, after merge but before removing the feature flag. But honestly, my preference would be addressing them now, as this stuff tends to only get more difficult once the code is in already.

hilti/toolchain/include/base/graph.h
bbannier: I believe these are fine to drop for now.

hilti/toolchain/include/base/graph.h
bbannier: I added Doxygen documentation to the class and its API. I also slightly reorganized the code so data members, the internal `neighbors` function, and `Direction` become `private`.

hilti/toolchain/include/ast/node.h
bbannier: Renamed this to `_identity`, and `counter` to `_instances`.

hilti/toolchain/include/base/graph.h
bbannier: Made this private.

hilti/toolchain/include/compiler/detail/cfg.h
bbannier: Users want both read (for querying) and write access (to e.g., remove nodes or edges) to this, so I made this `public`. I now added both a `const` as well as a non-`const` method to provide that access.

hilti/toolchain/include/compiler/detail/cfg.h
bbannier: Done, also renamed it to `_getOrAddNode`.

hilti/toolchain/include/compiler/detail/cfg.h
bbannier: Yes, that seems possible, I'll switch them over since that makes the code much easier to follow.

hilti/toolchain/src/compiler/cfg.cc
bbannier: Did that, and now use variations of `auto&`, `const auto&` and `auto*`.

hilti/toolchain/include/ast/node.h
bbannier: At this point this is mostly used for deterministic node ordering when rendering dot graphs for tests. Making order depend on node creation order is stable, but by address is not.

In principle I'd love to sort by e.g., `Meta`, but it is not guaranteed to be set.

hilti/toolchain/include/compiler/detail/cfg.h
bbannier: Like discussed offline I added a function `removeNode` to `CFG` which performs the work I'd previously do in the optimizer pass. With that this function can become `private`, and we also do not need to expose mutable access to the graph in `CFG` anymore.

hilti/toolchain/src/compiler/optimizer.cc
bbannier: I changed the implementation so the CFG is automatically populated on construction, so this should not be relevant to users anymore. I also made `populateDataflow` `private` so even if they wanted they could do nothing.

hilti/toolchain/include/compiler/detail/cfg.h
bbannier: I heavily documented this file now.

hilti/toolchain/include/compiler/detail/cfg.h
bbannier: We basically extract whether a node has effects not modeled in the AST. This is hard to do for users since it requires visiting the full statement which we already do when computing `gen` and `use`. I expanded the comment.

hilti/toolchain/src/compiler/optimizer.cc
bbannier: This is really about removing any kind of node, but currently we just remove unreachable statements and unused writes. I added a comment.

doc/toolchain.rst
bbannier: Unfortunately yes. Since we want users to enable this flag we need to do better than "might break valid code", so I expanded the comment to

```rst
    ``HILTI_OPTIMIZER_ENABLE_CFG``
        Set to `1` to enable experimental control-flow based optimizations.
        These optimizations remove dead code both in user as well as in
        Spicy-generated code, so enabling these optimizations can improve
        runtime parser throughput as well as lead to faster faster C++
        compilation.

        Since this optimization does not have complete coverage of all possible
        Spicy statments yet it might incorrectly classify statements as having
        not visible side effects and remove them, thus breaking valid code. If
        some source code behaves as expected with these passes enable on under
        test it should however also work in general.

        .. todo:: Always enable these passes once we have full coverage.
```

tests/Baseline/spicy.optimization.default-parser-functions/opt.hlt
bbannier: The issue here is that we remove the dead store before we even arrive at the exact pattern by removing feature dependent code.

bbannier: This turns out to be a little tricky to do in the current framework. While I could detect empty statements like `default<void>();` this is insufficient as we could remove one of the assignments even if they before we have collapsed a dead block to that (think feature-dependent code in a `if (False)` block). One way around that might be to run the CFG and possibly `FunctionParamVisitor` pass after all others are done, in another big loop.

bbannier: I ended up splitting passes somewhat arbitrarily into different phases so the new pass can run after the other passes are finished. This seems to make this work.

hilti/toolchain/src/compiler/optimizer.cc
bbannier: This should be fixed now.

hilti/toolchain/src/compiler/cfg.cc
bbannier: This should be fixed now.

tests/Baseline/spicy.optimization.default-parser-functions/log
```diff
@@ -307,7 +300,7 @@
...
 [debug/optimizer] [default-parser-functions.spicy:10:1-21:2] declaration::Constant "const bool __feat%foo@@P2%uses_sync_advance = False;" -> disabled feature 'uses_sync_advance' of type 'foo::P2' since it is not used
-[debug/optimizer] [default-parser-functions.spicy:12:1-12:18] declaration::Type "type P0 = struct { optional<hilti::RecoverableFailure> __error &always-emit &internal; method tuple<const view<stream>, int<64>, const iterator<stream>, optional<hilti::RecoverableFailure>> __parse_foo__P0_stage2(); } &on-heap;" -> null (removing unused type)
+[debug/optimizer] [default-parser-functions.spicy:12:1-12:18] declaration::Type "type P0 = struct { optional<hilti::RecoverableFailure> __error &always-emit &internal; method tuple<const view<stream>, int<64>, const iterator<stream>, optional<hilti::RecoverableFailure>> __parse_foo__P0_stage2(inout value_ref<stream> __data, iterator<stream> __begin, copy view<stream> __cur, copy bool __trim, copy int<64> __lah, copy iterator<stream> __lahe, copy optional<hilti::RecoverableFailure> __error); } &on-heap;" -> null (removing unused type)
```
rsmmr, line 303: For my understanding, where is this difference for the stage2 parameters are coming from? Is this just due to a change in the order of the optimization passes?  Seeing many of among the baseline diffs, and also some related changes where parameters used to be removed but are no longer logged as such.

tests/Baseline/spicy.types.unit.params-reference/output
```diff
@@ -1,4 +1,4 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
 X, old
-Y, new
-struct_, [$x="xxx"]
+Y, old
+struct_, [$x=""]
```
rsmmr: This looks like an unexpected semantic change?

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,247 @@
...
+} // namespace std
+
+namespace hilti::detail::cfg {
+
+/**
+ * Dataflow facts about a node.
+ */
+struct Transfer {
+    /** Incoming edges, ordered by declaration they work on. */
+    std::unordered_map<Node*, std::unordered_set<GraphNode>> in;
```
rsmmr: Just to be sure: cannot use `Declaration*` here? (Comments sounds like we could.)

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,247 @@
...
+
+    /** Outgoing edges, ordered by declaration they work on. */
+    std::unordered_map<Node*, std::unordered_set<GraphNode>> out;
```
rsmmr: Same question.

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -0,0 +1,247 @@
...
+
+    /** The previous nodes killed by this node. */
+    std::unordered_set<GraphNode> kill;
+
+    /** Set of declarations this node may alias. */
+    std::unordered_set<Declaration*> maybe_alias;
+
+    /** Values this node updates, ordered by declaration. */
```
rsmmr: Can you clarify this comment? I don't quite get the meaning of key/value vs what is updating what.

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -0,0 +1,912 @@
...
+#include <hilti/hilti/ast/types/map.h>
+#include <hilti/hilti/ast/types/set.h>
+#include <hilti/hilti/ast/types/stream.h>
+#include <hilti/hilti/ast/types/vector.h>
+
+namespace hilti::detail::cfg {
+
+// Helper function to detect whether values of a type can alias their inputs.
+static bool _isAliasingType(const UnqualifiedType& type) {
+    // TODO(bbannier): Make this part of e.g., `UnqualifiedType` instead of a hardcoded list of types here?
```
rsmmr: (Just as a note, moving this into `UnqualifiedType` makes sense to me. But can do later.)

rsmmr (commented):
The APIs for the new classes, and their docs, look nice now, thanks!

My main remaining question is that one baseline change for `tests/Baseline/spicy.types.unit.params-reference/output`. Otherwise looks good!

tests/Baseline/spicy.optimization.default-parser-functions/log
bbannier: Yes I believe so, I only see this in the logging, but the generated code stays the same. This is due to now running the existing `FunctionParamVisitor` after all other existing passes so the previous phase has already removed the function before we can try to prune its parameters. If I move `FunctionParamVisitor` to the initial phase I see the previous log output.

tests/Baseline/spicy.types.unit.params-reference/output
bbannier: Definitely a bug, I'll take a look.

bbannier: This was due to the optimizer code which detects unreachable statements not handling unit parameters correctly (it only looked for `inout` parameters, but not parameter values which might be aliasing). I lifted `_isAliasingType` into `QualifiedType` and extended the parameter check.

hilti/toolchain/src/compiler/cfg.cc
bbannier: Done now to allow reusing this code in the optimizer pass.

hilti/toolchain/include/compiler/detail/cfg.h
bbannier: Sorry, I missed fixing the previous issue. We now use `Declaration*` as a key.

hilti/toolchain/include/compiler/detail/cfg.h
bbannier: Fixed.

hilti/toolchain/include/compiler/detail/cfg.h
bbannier: I changed this to the comment below, let me know whether is helpful enough.

```
    /**
     * Declarations this graph node generates updates for.
     *
     * For each updated declaration we return the graph node were this value
     * was last updated.
     */
```

hilti/toolchain/include/ast/type.h
```diff
@@ -303,6 +303,12 @@ class UnqualifiedType : public Node {
 
     hilti::node::Properties properties() const override;
 
+    /**
+     * Returns true if the data behind a value of this type could be aliased by
+     * another value.
+     */
+    bool isAliasingType() const;
```
rsmmr: Now that you moved this here, we should make this consistent with the other `isXXX()` methods: instead of hardcoding the list of types inside this method, make it virtual and let the relevant types override it:

```suggestion
  virtual bool isAliasingType() const { return false; }
```

Then also move the declaration up and sort it in with the other `is...()` methods.

bbannier: Sounds good, will do. What do you think about a default impl

```cpp
    virtual bool isAliasingType() const { return isReferenceType(); }
```

rsmmr: Would rather not. That makes it hard to follow when looking at a derived class. Better to have a well-defined default.

bbannier: ✅

rsmmr (approved):
Alright, I'm happy, don't have anything else!

PR #1698, opened 2024-03-21
Rework order of C++ module codegen.

rsmmr:
We used to potentially codegen modules multiple times. Now we cache a
module's generated C++ code inside the AST node the first time we
create it. From there, we can then easily reuse it later, in
particular when needing to import its declarations into another
module. Internally, we switch storage for `cxx::Unit` to shared
pointers, and tweak the debug logging a bit for better readability in
this new model.

### DIFF
--- a/hilti/toolchain/include/ast/ast-context.h
+++ b/hilti/toolchain/include/ast/ast-context.h
@@ -213,8 +213,7 @@ public:
      * dependent units, vs just direct dependencies of the specified unit
      * @return set of dependencies
      */
-    std::set<declaration::module::UID> dependencies(const declaration::module::UID& uid,
-                                                       bool recursive = false) const;
+    std::set<declaration::module::UID> dependencies(const declaration::module::UID& uid, bool recursive = false) const;
 
     /**
      * Updates an existing UID with new information.
--- a/hilti/toolchain/include/ast/ast-context.h
+++ b/hilti/toolchain/include/ast/ast-context.h
@@ -213,7 +213,7 @@ public:
      * dependent units, vs just direct dependencies of the specified unit
      * @return set of dependencies
      */
-    std::vector<declaration::module::UID> dependencies(const declaration::module::UID& uid,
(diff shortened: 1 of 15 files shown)


hilti/toolchain/src/ast/ast-context.cc
```diff
@@ -682,25 +682,24 @@ void ASTContext::_saveIterationAST(const Plugin& plugin, const std::string& pref
 }
 
 static void _recursiveDependencies(const ASTContext* ctx, declaration::Module* module,
-                                   std::vector<declaration::module::UID>* seen) {
-    if ( std::find(seen->begin(), seen->end(), module->uid()) != seen->end() )
+                                   std::set<declaration::module::UID>* seen) {
+    if ( seen->find(module->uid()) != seen->end() )
```
bbannier, line 686: No need to `find`, we do not need the element.

```suggestion
    if ( seen->count(module->uid()) )
```

hilti/toolchain/include/ast/ast-context.h
```diff
@@ -212,7 +212,7 @@ class ASTContext : public std::enable_shared_from_this<ASTContext> {
      * dependent units, vs just direct dependencies of the specified unit
      * @return set of dependencies
      */
-    std::vector<declaration::module::UID> dependencies(const declaration::module::UID& uid,
+    std::set<declaration::module::UID> dependencies(const declaration::module::UID& uid,
```
bbannier, line 215: We do not care about the order in the set, but pay for that feature. Let's instead use a `std::unordered_set` here and everywhere else.

bbannier (dismissed):

hilti/toolchain/include/ast/ast-context.h
rsmmr: Actually I think we do, at least for consistency/stability in our output (incl. debugging). And these sets are small so I think it shouldn't matter performance-wise.

PR #1699, opened 2024-03-22
Reimplement `IDBase` for better performance.

rsmmr:
Our codegen phase had quite some overhead due to repeated ID
operations recomputing the same information over and over again
(e.g., subpaths and namespaces). This reimplements the class to
compute everything once upfront, cutting codegen time into half.


rsmmr:
@bbannier  I might tweak/polish this a bit more but if you still have time to briefly review, take a look.

hilti/toolchain/tests/id-base.cc
```diff
@@ -7,10 +7,14 @@
 #include <hilti/autogen/config.h>
 #include <hilti/base/id-base.h>
 
-
 using namespace hilti;
 
-std::string normalizeID(std::string id) { return id; }
+std::optional<std::string> normalizeID(std::string_view id) {
+    if ( id.substr(0, 1) == "%" )
+        return std::string("XXX_") + std::string(id.substr(1));
```
bbannier, line 14: Yikes, honestly surprised that a `string_view` cannot be concatenated with a `std::string` and we need to create a throw-away temporary, https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2591r4.html 😞.

hilti/toolchain/src/compiler/cxx/elements.cc
```diff
@@ -21,134 +21,165 @@ static const unsigned int AddSeparatorAfter = (1U << 2U);  // Force adding a sep
...
+
+    if ( reserved_keywords.count(id) )
+        return std::string(id) + "_";
+
+    if ( std::all_of(id.begin(), id.end(), [](auto c) { return std::isalnum(c) || c == '_'; }) )
+        // Fast-path: no special-characters, no leading digits.
+        return std::nullopt;
+
+    char buffer[id.size() * 6 + 1]; // max possible size of modified string
+    char* p = buffer;
```
bbannier, line 151: It would be great to stay clear of VLAs since they are a non-standard extension, and require us to copy the data to compute the return value.

To do that you could have here

```cpp
    std::string buffer(id.size() * 6 + 1, '\0'); // max possible size of modified string
    char* p = buffer.data();
```

and then at the end

```cpp
    assert(p < buffer.data() + buffer.size());
    auto len = p - buffer.data();
    buffer.erase(len);
    buffer.shrink_to_fit();
    return buffer;
```

hilti/toolchain/src/compiler/cxx/elements.cc
```diff
@@ -669,10 +700,10 @@ cxx::Formatter& cxx::operator<<(cxx::Formatter& f, const cxx::Block& x) {
 cxx::Formatter& cxx::operator<<(cxx::Formatter& f, const cxx::Expression& x) { return f << std::string(x); }
 
 cxx::Formatter& cxx::operator<<(cxx::Formatter& f, const cxx::ID& x) {
-    if ( x.namespace_() == f.namespace_() )
+    if ( f.namespace_() && x.namespace_() == cxx::ID(*f.namespace_()) )
```
bbannier, line 703: If we made `Formatter::namespace_()` return a `cxx::ID` this would simply be

```suggestion
    if ( f.namespace_() == x.namespace_() )
```

hilti/toolchain/src/compiler/codegen/codegen.cc
```diff
@@ -703,7 +703,7 @@ cxx::declaration::Function CodeGen::compile(const ID& id, type::Function* ft, de
     if ( linkage == declaration::Linkage::Struct ) {
         // For method implementations, check if the ID is fully scoped with
         // the module name; if so, remove.
-        if ( id.sub(0) == _hilti_module->uid() )
+        if ( id.sub(0).str() == _hilti_module->uid().str() )
```
bbannier, line 706: We currently compute the module `UID::str` twice which allocates a `std::string` each time. Could you either change the return type of `UID::str` to `const std::string&`, or alternatively cache the value?

hilti/toolchain/include/compiler/detail/cxx/elements.h
```diff
@@ -48,6 +48,7 @@ class Element {
     bool isMultiLine() const { return _s.find('\n') != std::string::npos; }
 
     operator std::string() const { return _s; }
+    operator std::string_view() const { return _s; }
```
bbannier, line 51: With this overload present ideally we would never call the the implicit conversion to `std::string` anymore since it always creates a copy. Currently we cannot simply remove it since many functions expect e.g., `const std::string&` args, so I wonder whether it wouldn't make sense to get rid of both these implicit conversions and instead provide an explicit function

```cpp
    const std::string& str() const { return _s; }
```

hilti/toolchain/include/compiler/detail/cxx/elements.h
```diff
@@ -73,6 +74,7 @@ class Expression {
     bool isLhs() const { return _side == Side::LHS; }
 
     operator std::string() const { return _s; }
+    operator std::string_view() const { return _s; }
```
bbannier, line 77: See above.

hilti/toolchain/include/ast/node.h
```diff
@@ -127,7 +127,7 @@ inline std::string to_string(const PropertyValue& v) {
         auto operator()(int i) { return util::fmt("%d", i); }
         auto operator()(int64_t i) { return util::fmt("%" PRId64, i); }
         auto operator()(const std::string& s) { return util::escapeUTF8(s); }
-        auto operator()(const ID& id) { return id.str(); }
+        auto operator()(const ID& id) { return std::string(id.str()); }
```
bbannier, line 130: This seems unnecessary,

```suggestion
        auto operator()(const ID& id) { return id.str(); }
```

If some compiler complains about the differing return values from the visitors, maybe it would be clearer to instead make the return type explicit (e.g., `std::string operator()(const ID& id) { return id.str(); }` for all visitors).

hilti/toolchain/include/ast/id.h
```diff
@@ -33,7 +33,7 @@ class ID : public detail::IDBase<ID> {
      * @param s string to construct from
      * @param meta meta data to attach
      */
-    explicit ID(std::string s, Meta meta) : IDBase(std::move(s)), _meta(std::move(meta)) {}
+    explicit ID(std::string_view s, Meta meta) : IDBase(s), _meta(std::move(meta)) {}
```
bbannier, line 36: With this overload available we should be able to get rid of the ctr from `const char*` above which now seems unused.

bbannier (approved):

hilti/toolchain/src/compiler/cxx/elements.cc
rsmmr: That at-the-end part doesn't look to appealing to me:  very technical and I'm not sure the shrink-to-fit is better than the copy for the return value. But we can get rid of the VLA by using alloca(), I'll. make that change.

hilti/toolchain/include/compiler/detail/cxx/elements.h
rsmmr: And/or: make more methods take views. But either way that looks like a larger change not really that related to this PR, so I'll leave it for now.

hilti/toolchain/include/ast/node.h
rsmmr: This was just a left-over that's no longer needed.

hilti/toolchain/include/ast/id.h
rsmmr: That doesn't directly work unfortunately. For one, the `string_view` overload is explicit but the `const char *` isn't. That follows how it used to be and was on purposes (to allow passing string literals as IDs, but prevent string variables from implicit conversion). We could change that, however I just tried that and I'm still getting problems without the `const char*` overload where things don't convert over. Not sure why honestly. I'll leave as is for now.

PR #1702, opened 2024-03-27
Support/fix public type aliases to units.

### LINKED ISSUE #1493: Presence of type alias causes ambiguity in generated code
The mere presence of a type alias can lead to ambiguities in generated code which cause compilation to fail, e.g.,

```ruby
module foo;

public type A = unit {
    : uint8;
};

public type B = A;  # PRESENCE OF THIS LINE TRIGGERS THE ERROR.
```

```console
$ spicyc -dj foo.spicy
[error] foo.spicy:3:17-5:2: ID 'foo::A::parse1' is ambiguous
[error] foo.spicy:3:17-5:2: ID 'foo::A::parse2' is ambiguous
[error] foo.spicy:3:17-5:2: ID 'foo::A::parse3' is ambiguous
[error] foo.spicy:3:17-5:2: ID 'spicy_rt::waitForInput' is ambiguous
[error] foo.spicy:3:17-5:2: unknown ID 'self'
```

The only workaround for now seems to be to not use a type alias in case it causes such errors.

rsmmr:
An alias like `public type Unit1 = Unit2` used to lead to C++-side
compiler errors, which this fixes. We also fully support this now by
making both `Unit1` and `Unit2` available for parsing to host
applications. Internally, the `Unit1` parser is just a small facade
pointing to the parsing functions for `Unit2`.

Closes #1493.


hilti/toolchain/src/ast/ast-context.cc
```diff
@@ -387,10 +387,12 @@ struct VisitorCheckIDs : hilti::visitor::PreOrder {
...
-        if ( ! n->fullyQualifiedID() ) {
-            hilti::detail::ast_dumper::dump(std::cerr, n->parent()->as<Node>());
-            logger().internalError(util::fmt("declaration without fully qualified ID found: %s", n->id()));
-        }
+        /*
+         * if ( ! n->fullyQualifiedID() ) {
+         *     hilti::detail::ast_dumper::dump(std::cerr, n->parent()->as<Node>());
+         *     logger().internalError(util::fmt("declaration without fully qualified ID found: %s", n->id()));
+         * }
+         */
```
bbannier, line 395: Let's delete this so it does not go stale.

hilti/toolchain/include/ast/type.h
```diff
@@ -378,6 +378,9 @@ class QualifiedType : public Node {
     /** Returns the type's "sideness". */
     auto side() const { return _side; }
 
+    /** Shortcut to try-cast to `type::Name`. */
+    type::Name* alias() const;
```
bbannier, line 382: Returning a pointer to a mutable subverts the method being `const` and we seem to have a number of these since the recent refactor. Could you as a follow-up change this to be

```suggestion
    const type::Name* alias() const;
```

This requires adjusting the signature for `resolvedDeclaration` as well.

bbannier (approved):

PR #1704, opened 2024-04-02
Reduce memory usage

rsmmr:
- **Reuse `Meta` instances across nodes.**
- **Do not auto-allocate storage for errors with each AST node.**
- **Change storage for AstContext.**
- **Move inherit-scope information into virtual method.**
- **Compute an ID's internal views on demand.**
- **Use less memory for storing context IDs.**
- **Remove meta/location information from AST IDs.**
- **Reduce memory usage for operators with complex argument types.**
- **Remove unused class.**
- **Overhaul AST node memory management.**


hilti/toolchain/include/ast/node.h
```diff
@@ -172,6 +172,70 @@ using Properties = std::map<std::string, node::PropertyValue>;
...
+    void reset() {
+        unpin();
+        _node = nullptr;
+    }
+
+    T* operator->() const { return _node; }
+    T& operator*() const { return *_node; }
+    explicit operator bool() const { return _node != nullptr; }
+    operator T*() const { return _node; }
+    operator Node*() const { return _node; }
```
bbannier, line 219: In the usage we already expect `T` to derive from `Node` so this is not needed (and also seems slightly out of place since nothing else in `PinnedNodePtr` makes that requirement explicit).

```suggestion
```

hilti/toolchain/src/ast/node.cc
```diff
@@ -23,7 +23,10 @@ std::string node::to_string(const Tags& ti) {
     return util::join(util::transform(ti, [](auto i) { return std::to_string(i); }), ",");
 }
 
-Node::~Node() {}
+Node::~Node() {
+    clearChildren();
+    _ref_count = -1; // for debugging, mark as destroyed
```
bbannier, line 28: Suggest to drop this. The object with the updated `_ref_count` is going away so this update is shouldn't be observable outside of UB. We should also drop the checks for `_ref_count != -1` and could then make `_ref_count` unsigned.

```suggestion
```

hilti/toolchain/src/ast/node.cc
```diff
@@ -23,7 +23,10 @@ std::string node::to_string(const Tags& ti) {
     return util::join(util::transform(ti, [](auto i) { return std::to_string(i); }), ",");
 }
 
-Node::~Node() {}
+Node::~Node() {
+    clearChildren();
```
bbannier, line 27: Hmm, I guess this manual lifetime management wouldn't be needed if the children where `PinnedNodePtr` instead of simply `Nodes`. Is that impossible since it would break the use case of temporarily having unpinned nodes on the AST?

hilti/toolchain/src/ast/node.cc
```diff
@@ -115,8 +118,10 @@ void Node::replaceChildren(ASTContext* ctx, const Nodes& children) {
 
 void Node::clearChildren() {
     for ( auto& c : _children ) {
-        if ( c )
-            c->_clearParent();
+        if ( c ) {
+            c->_parent = nullptr;
```
bbannier, line 122: Do we need this update at all if we destroy all children after the loop anyway?

```suggestion
```

hilti/toolchain/src/ast/ast-context.cc
```diff
@@ -45,8 +45,36 @@ ASTContext::ASTContext(Context* context) : _context(context) {
...
+        clear();
+    } catch ( const std::exception& e ) {
+        logger().internalError(util::fmt("unexpected exception in ~ASTContext: %s", e.what()));
+    }
+}
+
+void ASTContext::clear() {
+    if ( ! _root )
+        return;
+
```
bbannier, line 58: The body does not depend on `_root` being set, and the early exit makes the code more fragile. WDYT about dropping this?

```suggestion
```

hilti/toolchain/src/ast/ast-context.cc
```diff
@@ -285,6 +353,8 @@ declaration::module::UID ASTContext::_addModuleToAST(declaration::Module* module
     assert(! module->hasParent()); // don't want to end up copying the whole AST
     auto uid = module->uid();
 
+    // We don't need to pin the module for tables because it's tied to the root
+    // node already.
```
bbannier, line 357: It isn't really clear to me what is meant here, could you rephrase?

hilti/toolchain/include/ast/node.h
```diff
@@ -441,6 +506,7 @@ class Node {
 
         if ( auto i = std::find(_children.begin(), _children.end(), n); i != _children.end() ) {
             (*i)->_parent = nullptr;
```
bbannier, line 508: It seems the update to `_parent` shouldn't matter since the node will be destroyed, suggest to drop.

Also below.

bbannier (dismissed):

hilti/toolchain/include/ast/meta.h
```diff
@@ -49,11 +51,31 @@ class Meta {
...
+
+    struct Hash {
+        size_t operator()(const Meta& meta) const {
+            size_t h = 0;
+            for ( const auto& c : meta._comments )
+                h = hilti::rt::hashCombine(h, std::hash<std::string>()(c));
+
+            return hilti::rt::hashCombine(h, std::hash<std::optional<Location>>()(meta._location));
+        };
+    };
```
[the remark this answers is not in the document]
rsmmr: Thanks. I actually had tried that originally but didn't find a version that would compile, so gave up. But yours indeed works.

hilti/toolchain/src/ast/node.cc
rsmmr: No, different reason: we have a few places where we expect `_children` to be a vector of `Node*`, so that we can access children, or ranges of children, to retrieve raw node pointers directly without having to dereference/convert them first for further usage elsewhere (we use raw pointers almost everywhere else).

hilti/toolchain/src/ast/node.cc
rsmmr: It's not guaranteed that we'll destroy after the loop: somebody might re-pin it in the meantime, e.g., by re-adding the node to the AST elsewhere. And having a null parent is the indicator that the node isn't part of the AST yet and doesn't need to be deep-copied when it's added.

hilti/toolchain/include/ast/node.h
rsmmr: See other comment.

hilti/toolchain/include/ast/node.h
```diff
@@ -695,17 +695,23 @@ class Node {
...
-        _errors.emplace_back(std::move(error));
+        if ( ! _errors )
+            _errors = std::make_unique<std::vector<node::Error>>();
+
+        _errors->emplace_back(std::move(error));
     }
 
     /** Returns true if there are any errors associated with the node. */
-    bool hasErrors() const { return _errors.size(); }
+    bool hasErrors() const { return _errors && _errors->size(); }
```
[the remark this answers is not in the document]
rsmmr: >  we seem to enforce that _errors is only set if there are errors

Now it's my time to say: having the empty check is less fragile. :-)

hilti/toolchain/src/ast/node.cc
rsmmr: I'd like to keep this, it's helpful for debugging. Because we're mostly passing around raw pointers, it's hard to catch dangling pointers in the new scheme. When looking at a `Node*` in a debugger, there's no way to tell if that instance actually still exists (ref_cnt==zero is legitimate). Setting the count to minus one says it's been destroyed (of course it's not guaranteed that the memory remains set like that, but it's quite likely). Similar for the asserts: it has a good chance of catching memory mgmt errors.

hilti/toolchain/include/ast/node.h
rsmmr: Interesting, I had added this because things didn't compile for me without it. But now they do, so removing.

hilti/toolchain/src/ast/ast-context.cc
rsmmr: That comment is actually stale, no longer applies, removing.

hilti/toolchain/include/ast/ast-context.h
```diff
@@ -58,7 +58,7 @@ class ContextIndex {
      * @param index value to store with the index, which should be larger than
      * zero for valid indices; zero is the default and reserved for `None`.
      */
-    explicit ContextIndex(uint64_t index = 0) : _value(index) {}
+    explicit ContextIndex(uint32_t index = 0) : _value(index) {}
```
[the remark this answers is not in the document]
rsmmr: Good idea, will do.

hilti/toolchain/include/ast/node.h
```diff
@@ -172,6 +172,70 @@ using Properties = std::map<std::string, node::PropertyValue>;
...
+    T* operator->() const { return _node; }
+    T& operator*() const { return *_node; }
+    explicit operator bool() const { return _node != nullptr; }
+    operator T*() const { return _node; }
+    operator Node*() const { return _node; }
+
+    T* get() const { return _node; }
+
+private:
+    void pin() {
```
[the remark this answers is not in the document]
rsmmr: I don't have that association but I'm fine renaming to `retain`/`release`. Also, I'll move the pointer class into the `node` namespace.

PR #1712, opened 2024-04-05
Fixes for references and `inout`

rsmmr:
Series of patches fixing/cleaning up/documenting Spicy references and `inout` parameters.


bbannier (approved):

spicy/toolchain/src/compiler/validator.cc
```diff
@@ -669,6 +669,13 @@ struct VisitorPost : visitor::PreOrder, hilti::validator::VisitorMixIn {
         if ( n->propertyItem("%synchronize-at") && n->propertyItem("%synchronize-after") )
             error("unit cannot specify both %synchronize-at and %synchronize-after", n);
 
+        for ( auto* p : n->parameters() ) {
+            if ( p->kind() == hilti::parameter::Kind::InOut ) {
+                auto t = p->type()->type();
+                if ( ! (t->isReferenceType() || t->isA<type::Unit>()) )
```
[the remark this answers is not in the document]
rsmmr: No, the tef-type really means one of our 3 reference types wrapping something else. But this is changing now in a subsequent commit to test only for Unit (probably should have amended that change here).

doc/programming/language/types.rst
```diff
@@ -308,6 +308,83 @@ This type supports the :ref:`pack/unpack operators <packing>`.
...
+
+Reference
+—--------
+
+A reference ``T&`` wraps a value of another type ``T``, allowing to
+pass it around without creating a copy. Multiple references can wrap
+the same value, and the value will stay around for as long as there's
+at least one reference to it.
+
+You can create a reference through the Spicy's ``new T`` operator,
```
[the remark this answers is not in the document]
rsmmr: `new` is an operator/expression so not sure it finds into "statements". We actually have `autogen/types/generic.rst` that documents `new` but looks like that isn't included anywhere. I'll see if I can fit that in.

doc/programming/language/types.rst
```diff
@@ -308,6 +308,83 @@ This type supports the :ref:`pack/unpack operators <packing>`.
...
+
+.. note::
+
+  Nothing prevents you from passing a reference as an ``inout``
+  parameter, like: ``function f(inout x: bytes&)``. However, doing so
+  isn't very useful: the ``inout`` refers to the reference itself
+  here, which isn't modifiable in the first place, so the ``inout``
+  doesn't buy you anything. You can still modify the wrapped value
+  through dereferentation, but that's independent of making the
+  reference ``inout``.
```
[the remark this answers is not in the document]
rsmmr: Yeah, like discussed in person, I think that would be too complex for Spicy. Making the wrapped type always mutable seems like nice sweet spot to make usage convenient.

rsmmr:
@bbannier ready to take another look. The previous state was still WIP. Unit parameters have changed more, and a couple of additional commits at the end.

doc/programming/language/types.rst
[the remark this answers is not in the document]
rsmmr: Ok, added a couple more commits at the end cleaning this up. Renamed the statement section to "statements & operators" and including the generic ones there now.

doc/scripts/spicy-doc-to-rst
```diff
@@ -296,6 +296,43 @@ class Method:
...
+
+            if self.fname.endswith("_"):
+                self.fname = self.fname[:-1]
+
+            self.fname = (
+                self.fname.lower().replace("::", "-").replace("_", "-") + ".rst"
+            )
+            self.fname = os.path.join(args.dir, self.fname)
+            self.fname_tmp = self.fname + ".tmp"
+            self.fd = open(self.fname_tmp, "w+")
```
bbannier, line 317: Let's constrain the encoding here,

```suggestion
            self.fd = open(self.fname_tmp, "w+", encoding='utf-8')
```

doc/scripts/spicy-doc-to-rst
```diff
@@ -296,6 +296,43 @@ class Method:
         return self.rst() < other.rst()
 
 
+# Manages a single output file.
+class OutputFile:
+    def __init__(self, ns, args):
```
bbannier, line 301: Nit: instead of passing `args` we could pass `args.dir` directly (probably as a parameter named `directory` to not hide the global `dir` function).

doc/scripts/spicy-doc-to-rst
```diff
@@ -296,6 +296,43 @@ class Method:
...
+            self.fname = (
+                self.fname.lower().replace("::", "-").replace("_", "-") + ".rst"
+            )
+            self.fname = os.path.join(args.dir, self.fname)
+            self.fname_tmp = self.fname + ".tmp"
+            self.fd = open(self.fname_tmp, "w+")
+        else:
+            self.fd = sys.stdout
+
+    def close(self):
```
bbannier, line 321: Let's instead implement [`__del__`](https://docs.python.org/3.9/reference/datamodel.html#object.__del__) so user do not have to worry about manual lifetime management,

```suggestion
    def __del__(self):
```

bbannier (dismissed):

doc/scripts/spicy-doc-to-rst
rsmmr: We actually need the manual close as we could open the same file multiple times, and need the previous one to be closed again already, which with `__del__` I don't think we could guarantee.

doc/scripts/spicy-doc-to-rst
rsmmr: Yeah, but seems fine either way, I'll leave it as is.

PR #1717, opened 2024-04-16
Bump baselines.

bbannier:
This is a fixup commit for 8531120819792b0a47449153266d64a5b3af68e2 where during review we changed the format of an error message, but missed updating the baselines.


rsmmr (approved):
Thanks!

PR #1722, opened 2024-04-18
Reject joint usage of filters and look-ahead.

### LINKED ISSUE #1655: Filter generates code with incompatible iterators

rsmmr:
- **Add test confirming that `&parse-{at,from}` don't interfere with outer look-ahead parsing.**
- **Reject joint usage of filters and look-ahead.**

Closes #1655.


spicy/toolchain/include/ast/types/unit.h
```diff
@@ -80,9 +80,22 @@ class Unit : public UnqualifiedType {
...
+     * from actual use of the type inside the global AST and will be valid only
+     * after the Spicy resolver has finished.
+     */
+    auto mayHaveFilter() const { return _may_have_filter; }
+
+    /**
+     * Sets the flag indicating whether the unit may see a filter getting
+     * attached. Should be called only by the resolver.
+     */
+    void setMayHaveFilter(bool may_have_filter) { _may_have_filter = may_have_filter; }
```
bbannier, line 97: Currently we never unset this (and I suspect it would be difficult since filters could be removed in a number of ways), so for now we could make this a one-way toggle, e.g.,

```suggestion
    void setMayHaveFilter() { _may_have_filter = true; }
```

bbannier (approved):

spicy/toolchain/include/ast/types/unit.h
rsmmr: As a general pattern I like to pass booleans to functions so that one can go both ways.

PR #1725, opened 2024-04-22
Fix skipping in size-contrained units.

### LINKED ISSUE #1724: `skip` fields trims too much data in `&size`-constrained unit
The following grammar fails to parse since field `S.b` consumes all available input:

```ruby
module skip;

type S = unit(size: uint64) {
    a       : bytes &size=2;
    b       : skip bytes &size=1;
    c       : bytes &size=1;
    d       : skip bytes &eod;

    on %done {
        print self;
    }
} &size=size;


public type X = unit {
    : S(6);
};
```

```console
$ printf "\x01\x02\x03\x04\x05\x06" | HILTI_DEBUG=spicy-verbose spicy-driver -d skip.spicy
[spicy-verbose] - state: type=skip::X input="\\x01\\x02\\x03\\x04\\x05\\x06" stream=0x6000000842b8 offsets=0/0/0/6 chunks=1 frozen=yes mode=default trim=yes lah=n/a lah_token="n/a" recovering=no
[spicy-verbose] - parsing production: Unit: skip__X -> Resolved_3
[spicy-verbose]   - state: type=skip::X input="\\x01\\x02\\x03\\x04\\x05\\x06" stream=0x6000000842b8 offsets=0/0/0/6 chunks=1 frozen=yes mode=default trim=yes lah=n/a lah_token="n/a" recovering=no
[spicy-verbose]   - parsing production: Unit: skip__S_2 -> a_2 b_2 c_2 d_2 (field '_anon', id n/a, parser/composer, args: (6))
[spicy-verbose]     - state: type=skip::S input="\\x01\\x02" stream=0x6000000842b8 offsets=0/0/0/6 chunks=1 frozen=yes mode=default trim=yes lah=n/a lah_token="n/a" recovering=no
[spicy-verbose]     - parsing production: Variable: a   -> bytes
[spicy-verbose]     - setting field 'a' to '\\x01\\x02'
[spicy-verbose]     - state: type=skip::S input="\\x03\\x04\\x05\\x06" stream=0x6000000842b8 offsets=2/0/2/6 chunks=1 frozen=yes mode=default trim=yes lah=n/a lah_token="n/a" recovering=no
[spicy-verbose]     - parsing production: Skip: b   -> skip: b
[spicy-verbose]       - skipped 1 bytes (0 left to skip)
[spicy-verbose]     - state: type=skip::S input="" stream=0x6000000842b8 offsets=3/0/6/6 chunks=1 frozen=yes mode=default trim=yes lah=n/a lah_token="n/a" recovering=no
[spicy-verbose]     - parsing production: Variable: c   -> bytes
[error] terminating with uncaught exception of type spicy::rt::ParseError: &size amount not consumed: expected 1 bytes, but got 0 bytes (skip.spicy:6:14-6:20)
```

This is due to the following code which incorrectly updates the limited view (from `&size`) of the unit with a position pointing after the unit (i.e., parsing the `skip` field consumes all available data). https://github.com/zeek/spicy/blob/1825bdbaea723ab184b45687821e9bee137d8d65/spicy/toolchain/src/compiler/codegen/parser-builder.cc#L661-L664

This has been broken since at least the introduction of this `skip` syntax in e134dc41e34bf339dd9827ce46cc652865f606bb, i.e., likely since the `skip` feature came about.

This was originally reported privately by @Mohan-Dhawan.

bbannier:
We previously could skip too much data if `skip` was used in a unit with
a global `&size`. This was due to the machinery moving the input forward
believing that `skip` units were that really field productions. While
this is true in a sense in this particular case it still lead to
incorrect behavior, in particular as far is input handling is concerned
`skip` productions largely behave like any other unit.

With this patch we store information about field in `Skip` productions.
We can use that additional information to in this case correctly handle
productions for `skip` as well.

Closes #1724.


spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -658,7 +658,7 @@ struct ProductionVisitor : public production::Visitor {
 
         else if ( ! meta.isFieldProduction() ) {
             // Need to move position ahead.
-            if ( state().ncur ) {
+            if ( auto field = meta.field(); field && ! field->isSkip() && state().ncur ) {
```
rsmmr, line 661: I'm not quite sure I follow why this is needed: what is difference between with and without skip that needs us ignore ncur in the former case but not the latter?

bbannier: Yeah, this is not great. The issue is that even with this patch I do not store enough information about the actual production of the equivalent non-`skip` field in the `Skip` production. I initially tried to instead compute the full `production::Meta` in [when processing the `Field`](https://github.com/zeek/spicy/blob/1825bdbaea723ab184b45687821e9bee137d8d65/spicy/toolchain/src/compiler/codegen/grammar-builder.cc#L125) (i.e., move the `isSkip` block after the production computation so it can pass its `production::Meta` into `Skip`), but this caused changes to production names as we'd potentially compute production names for the same `Field` twice.

rsmmr: I added one commit, see what you think.

bbannier: Thanks, that seems much more coherent. I left a few questions below, mostly for me to get a better understanding.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -1842,13 +1842,6 @@ struct ProductionVisitor : public production::Visitor {
         else
             hilti::logger().internalError("unexpected skip production");
 
-        if ( p->field()->emitHook() ) {
-            pb->beforeHook();
-            builder()->addMemberCall(state().self, ID(fmt("__on_%s", p->field()->id().local())), {},
-                                     p->field()->meta());
-            pb->afterHook();
-        }
-
```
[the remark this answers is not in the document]
rsmmr: I think the other path is now triggered because of making the top-level skip the field production.

spicy/toolchain/src/compiler/codegen/grammar-builder.cc
```diff
@@ -130,7 +130,7 @@ struct Visitor : public visitor::PreOrder {
             if ( const auto& ctor = n->ctor() ) {
                 auto prod = productionForCtor(ctor, n->id());
                 auto m = prod->meta();
-                m.setField(pf->currentField(), true);
+                m.setField(pf->currentField(), false);
```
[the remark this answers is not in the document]
rsmmr: I'm not sure we actually *require* this change (I didn't see difference in effect either), but conceptually it seems right. My fixup makes the Skip production the top-level production that's in charge of the field. With that, the ctor doesn't need (should) have the flag set anymore, because it's wrapped inside the Skip. That's also why I don't think we need to query the meta data in the ctor for this.

rsmmr (approved):

PR #1726, opened 2024-04-22
 Promote use of `-x` over `-c`, clean up `-P`.

rsmmr:
- **Remove generated, in-code linker JSON meta data.**
- **Promote use of `-x` over `-c`, clean up `-P`.**
- **Update NEWS.**


doc/host-applications.rst
```diff
@@ -197,24 +192,12 @@ Spicy arguments translate into C++ arguments, look at the C++
...
-
-    - If your Spicy code is importing any library modules (e.g., the
-      standard ``filter`` module), you'll need to compile those as
-      well in the same fashion.
+.. note::
 
+    When using ``spicyc -x`` you can specify any Spicy modules you
+    need/want. If any of your Spicy code is importing any library
+    modules (e.g., the standard ``filter`` module), you'll need to add
+    those to the command line as well.
```
bbannier, line 200: Is that true? I took the example from https://docs.zeek.org/projects/spicy/en/latest/programming/parsing.html#filters and the following succeeded, even though `foo.spicy` has `import filter`:

```console
spicyc -x foo foo.spicy
$(spicy-config --cxx --cxxflags --ldflags) foo_*cc -o foo
```

doc/examples/my_http-host-parse1.cc
bbannier, line 1: Nit: `#include <iostream>` unused.

bbannier (dismissed):

doc/host-applications.rst
```diff
@@ -22,16 +22,16 @@ We discuss both approaches in the following.
 
     Internally, Spicy is a layer on top of an intermediary framework
     called HILTI. It is the HILTI runtime library that implements most
-    of the functionality we'll look at in this section, so you'll see
+    of the functionality that we'll look at in this section, so you'll see
```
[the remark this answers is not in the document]
rsmmr: It *is* our style. :-)

doc/host-applications.rst
```diff
@@ -128,8 +125,8 @@ We can now use the standard C++ compiler to build all this into an
 executable, leveraging ``spicy-config`` to add the necessary flags
 for finding includes and libraries::
 
-    # clang++ -o my-http my-http-host.cc my-http.cc my-http-linker.cc $(spicy-config --cxxflags --ldflags)
-    # ./my-http $'GET index.html HTTP/1.0\n'
+    # clang++ -o my_http my_http-host.cc my_http___linker__.cc my_http_MyHTTP.cc $(spicy-config --cxxflags --ldflags)
```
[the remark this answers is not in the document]
rsmmr: Yeah, we were inconsistent and used this at a couple places. I decided against because I think it makes it harder to read/understand what's going on when looking at the docs, as one needs to know/anticipate what `spicy-config --cxx` and keep that in mind. But let me add a note pointing out the alternative.

doc/host-applications.rst
rsmmr: Good catch. I adapted that from the old text without trying, but we indeed don't need that when building everything at once.

PR #1728, opened 2024-04-24
Fix spelling of stored Spicy hook priorities.

bbannier:
While in Spicy code hook priorities are spelled `priority` in HILTI code and also in our handling of its AST the corresponding attribute is still called `&priority`. We also do not perform any translation of "Spicy attributes" to "HILTI attributes", so us using `priority` as attribute name for Spicy lead to Spicy priorities being ignored. This seems to have been broken since at least 2020.

This patch fixes Spicy AST generation to emit attributes named `&priority` instead of the previous `priority` to be consistent with the HILTI handling. We also add a test validating that Spicy hook priorities are handled correctly.


spicy/toolchain/src/compiler/parser/parser.yy
```diff
@@ -884,7 +884,7 @@ unit_hook_id: { driver->enableHookIDMode(); }
 
 unit_hook_attribute
               : FOREACH                          { $$ = builder->attribute("foreach", __loc__); }
-              | PRIORITY '=' expr                { $$ = builder->attribute("priority", std::move($3), __loc__); }
+              | PRIORITY '=' expr                { $$ = builder->attribute("&priority", std::move($3), __loc__); }
```
rsmmr, line 887: This makes sense for internal consistency. However, even previously we were seemingly adding `&priority` during codegen time; see https://github.com/zeek/spicy/blob/1825bdbaea723ab184b45687821e9bee137d8d65/spicy/toolchain/src/compiler/codegen/codegen.cc#L587.  Was that not working, and/or is now no longer needed?

rsmmr (commented):
One more nit: Spicy's `scanner.ll` includes `&priority` as valid property, which we should probably remove to avoid more confusion.

At a higher level, if were just starting I think I'd actually prefer the `&priority` syntax for Spicy, too, seems that would be more consistent. But I guess we don't want to change that now anymore.

spicy/toolchain/src/compiler/parser/parser.yy
bbannier: _I believe_ this is still needed. The issue fixed here is about a previously incorrectly persisted hook `priority` from Spicy user code which was set in a way that `Hook::priority` would not recognize,

https://github.com/zeek/spicy/blob/d33f06f9c44e20beeae695b1c29973396da288ca/spicy/toolchain/include/ast/declarations/hook.h#L51-L56

We use `Hook::priority` as input to `compileHook` so this is now consistent. I am not sure, but it could be the flow is more complicated than needed, but I didn't touch this in this PR.

bbannier: > I'll add a patch deprecating `priority` for `&priority` in Spicy.

Thinking about this more, switching this to attribute syntax could potentially allow surprising syntax (especially if we clean up the parser to use generic attribute parsing), e.g.,

```ruby
: bytes &eod &priority=4711 {}
```

which we could even allow to parse from

```ruby
: bytes &priority=4711 &eod {}
```

Allowing to blur what is a parsing attribute and what a hook attribute doesn't seem ideal.

In Spicy we already allow `%debug` as a hook property (stored internally as an attribute). What do you think about making syntax property syntax for `%priority` as well? (We also have a `foreach` attribute, but that syntax without `&` or `%` seems useful to me).

bbannier: I'll punt on deprecating `priority`, and will make the cleanup in the lexer.

bbannier:
I made another pass over this. It seems that this wasn't a bug after all, just an API which could be easily misused. All this should be helped by using the same attribute name in Spicy and HILTI. I removed the new test since we already had a test which showed that Spicy hook priorities worked correctly.

PR #1729, opened 2024-04-29
Fix `&convert` on bit range

rsmmr:
- **Suppress new clang-tidy warnings.**
- **Fix `&convert` typing issue with bit ranges.**

### DIFF
--- a/hilti/runtime/include/types/bitfield.h
+++ b/hilti/runtime/include/types/bitfield.h
@@ -23,7 +23,18 @@ struct isBitfield {};
 /// printing.
 template<typename... Ts>
 struct Bitfield : public trait::isBitfield {
-    std::tuple<std::optional<Ts>...> value;
+    using value_type = std::tuple<std::optional<Ts>...>;
+
+    Bitfield(value_type v = {}) : value(std::move(v)) {}
+
+    /**
+     * Support instantiation from another bitfield type as long as all types
+     * convert over.
+     */
+    template<typename... Us>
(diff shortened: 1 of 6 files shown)


hilti/runtime/include/types/bitfield.h
```diff
@@ -23,15 +23,26 @@ struct isBitfield {};
 /// printing.
 template<typename... Ts>
 struct Bitfield : public trait::isBitfield {
-    std::tuple<std::optional<Ts>...> value;
+    using value_type = std::tuple<std::optional<Ts>...>;
+
+    Bitfield(value_type v = {}) : value(std::move(v)) {}
```
bbannier, line 28: Its a bit unfortunate that we add a non-`explicit` ctr which will make it easier to subvert our type system, but it seems just replacing it with a direct ctr from bitfield field values does not work (another motivation for that: get rid of `make_bitfield` since we now have a proper class here):

```cpp
Bitfield(Ts... xs) : value(std::make_tuple(xs...)) {}
```

bbannier (approved):
Like discussed offline, I tried to make `Bitfield` a more "normal" class, but failed. I left a couple comments, mostly just documenting what I came across. Please feel free to merge as is.

hilti/runtime/include/types/bitfield.h
rsmmr: We indeed need this non-explicit, that's part of the original issue here.

PR #1734, opened 2024-05-06
Fix a Spicy scoping issue across imports.

rsmmr:
We could get a bogus "unknown ID" error for default arguments of
functions defined in an imported module if that default argument was
itself referring to an identifier inside yet another imported module.
The test case shows the exact situation that was broken.


tests/spicy/codegen/default-arg-aross-import.spicy
```diff
@@ -0,0 +1,20 @@
+# @TEST-EXEC: spicyc -j %INPUT >output
+# @TEST-EXEC: btest-diff output
+#
+# @TEST-DOC: Pull in a enum value defined in an indirect dsependency through a default argument.
+
+module Foo;
+
+import Bar;
+
+Bar::xxx();
```
bbannier, line 10: Just for the record: I manually checked that this patch does not insert the types referenced by the imported function into scope which seems right to me.

For that I added a function to `bar.spicy`

```ruby
public function xyz(protocol: spicy::Protocol = spicy::Protocol::TCP): spicy::Protocol {
    return protocol;
}
```

and bound a value to its return value here. This works as long as I do not name its type, i.e.,

```ruby
global a = Bar::xyz();  # WORKS.
global b: spicy::Protocol = Bar::xyz();  # FAILS: unknown ID 'spicy::Protocol'.
```

bbannier (approved):

PR #1736, opened 2024-05-08
Remove use of `std::function`

rsmmr:
This removes all usage of `std::function`, and all includes of the
corresponding `functional` header. That is large header that tends to
introduce quite a bit of overhead, and we don't actually need it: all
of our uses of `std::function` are easy to replace with other
mechanisms: often just raw pointers, sometimes template parameters.

This also includes some further cleanup of code no longer needed.


hilti/toolchain/include/compiler/plugin.h
```diff
@@ -101,7 +100,7 @@ struct Plugin {
      * @param arg3 file associated with the input stream
      * @return module AST if parsing succeeded
      */
-    Hook<Result<declaration::Module*>, hilti::Builder*, std::istream&, hilti::rt::filesystem::path> parse;
+    Hook<Result<declaration::Module*>, hilti::Builder*, std::istream&, const hilti::rt::filesystem::path&> parse;
```
bbannier, line 103: Interesting, looks like due to automatic arg type conversions for `std::function` a function pointer might be _more type-safe_ 🤯

spicy/toolchain/bin/spicy-dump/printer-text.h
```diff
@@ -40,7 +40,8 @@ class TextPrinter {
     void outputIndent() { out() << std::string(static_cast<std::basic_string<char>::size_type>(_level) * 2, ' '); }
 
     // Increase indentation level while executing callback function.
-    void indent(const std::function<void()>& func) {
+    template<typename Func>
+    void indent(const Func& func) {
```
bbannier, line 44:
```suggestion
    void indent(Func&& func) {
```

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -85,7 +85,8 @@ struct ProductionVisitor : public production::Visitor {
 
     auto builder() { return pb->builder(); }
     auto pushBuilder() { return pb->pushBuilder(); }
-    auto pushBuilder(std::shared_ptr<Builder> b, const std::function<void()>& func) {
+    template<typename Func>
+    auto pushBuilder(std::shared_ptr<Builder> b, const Func& func) {
```
bbannier, line 89:
```suggestion
    auto pushBuilder(std::shared_ptr<Builder> b, Func&& func) {
```

spicy/toolchain/src/compiler/codegen/parsers/literals.cc
```diff
@@ -50,9 +50,7 @@ struct Visitor : public visitor::PreOrder {
     auto context() { return pb()->context(); }
     auto pushBuilder(std::shared_ptr<Builder> b) { return pb()->pushBuilder(std::move(b)); }
     auto pushBuilder() { return pb()->pushBuilder(); }
-    auto pushBuilder(std::shared_ptr<Builder> b, const std::function<void()>& func) {
-        return pb()->pushBuilder(std::move(b), func);
-    }
+    auto pushBuilder(std::shared_ptr<Builder> b, void (*func)()) { return pb()->pushBuilder(std::move(b), func); }
```
bbannier, line 53: Are you deliberately ruling out callbacks with state here? I think passing a lambda (with `T&&`) would be more flexible and in line with similar APIs elsewhere.

spicy/toolchain/include/compiler/detail/codegen/parser-builder.h
```diff
@@ -516,8 +515,16 @@ class ParserBuilder {
      * @param features identifiers of the feature, will be combined with OR.
      * @param f callback building the feature-dependent code.
      */
-    void guardFeatureCode(const type::Unit* unit, const std::vector<std::string_view>& features,
-                          const std::function<void()>& f);
+    template<typename Func>
+    void guardFeatureCode(const type::Unit* unit, const std::vector<std::string_view>& features, const Func& f) {
```
bbannier, line 519:
```suggestion
    void guardFeatureCode(const type::Unit* unit, const std::vector<std::string_view>& features, Func&& f) {
```

spicy/toolchain/include/compiler/detail/codegen/parser-builder.h
```diff
@@ -263,7 +261,8 @@ class ParserBuilder {
     ScopeGuard makeScopeGuard() { return ScopeGuard(this); }
 
     /** Activates a statement builder for subsequent code. */
-    auto pushBuilder(std::shared_ptr<Builder> b, const std::function<void()>& func) {
+    template<typename Func>
+    auto pushBuilder(std::shared_ptr<Builder> b, const Func& func) {
```
bbannier, line 265:
```suggestion
    auto pushBuilder(std::shared_ptr<Builder> b, Func&& func) {
```

spicy/toolchain/src/compiler/codegen/parsers/types.cc
```diff
@@ -36,9 +36,7 @@ struct TypeParser {
     auto context() { return pb->context(); }
     auto pushBuilder(std::shared_ptr<Builder> b) { return pb->pushBuilder(std::move(b)); }
     auto pushBuilder() { return pb->pushBuilder(); }
-    auto pushBuilder(std::shared_ptr<Builder> b, const std::function<void()>& func) {
-        return pb->pushBuilder(std::move(b), func);
-    }
+    auto pushBuilder(std::shared_ptr<Builder> b, void (*func)()) { return pb->pushBuilder(std::move(b), func); }
```
bbannier, line 39: Stateful lambda deliberately ruled out?

spicy/toolchain/src/compiler/codegen/parsers/types.cc
```diff
@@ -115,7 +113,8 @@ struct Visitor : public visitor::PreOrder {
     auto context() { return pb()->context(); }
     auto pushBuilder(std::shared_ptr<Builder> b) { return pb()->pushBuilder(std::move(b)); }
     auto pushBuilder() { return pb()->pushBuilder(); }
-    auto pushBuilder(std::shared_ptr<Builder> b, const std::function<void()>& func) {
+    template<typename Function>
+    auto pushBuilder(std::shared_ptr<Builder> b, const Function& func) {
```
bbannier, line 117:
```suggestion
    auto pushBuilder(std::shared_ptr<Builder> b, Function&& func) {
```

bbannier (approved):
This looks fine to me, but I wonder whether it actually improves compile time, either for analyzers or Spicy itself since it touches both the runtime lib (filters, sinks, parse fcts) and the Spicy itself (I'd expect some of the work to instantiate `std::function` would also be needed for lambdas).

I had a quick look at the [list of clang-tidy checks](https://clang.llvm.org/extra/clang-tidy/checks/list.html) but didn't see anything we could configure to automatically reject uses of `std::function` in the future.

I left a couple of comments, a lot of them to keep the door open for stateful lambdas.

rsmmr:
> This looks fine to me, but I wonder whether it actually improves compile time, either for analyzers or Spicy itself since it touches both the runtime lib (filters, sinks, parse fcts) and Spicy itself; I'd expect some of the work to instantiate `std::function` would also be needed for lambdas.

Honestly I don't know either but I've been reading quite a bit about `std::function` and sheer size of the `<functional>` header for a while now, so I just went ahead because it doesn't look like we actually need it. 

> I had a quick look at the [list of clang-tidy checks](https://clang.llvm.org/extra/clang-tidy/checks/list.html) but didn't see anything we could configure to automatically reject uses of `std::function` in the future.

I'll keep an eye on it. :-)

hilti/toolchain/include/compiler/plugin.h
rsmmr: Yeah, I was thinking the same!

hilti/runtime/include/types/reference.h
```diff
@@ -716,6 +716,23 @@ class StrongReferenceGeneric {
...
+    ValueReference<T> derefAsValue() const {
+        if ( _ptr.empty() )
+            return {};
+
+        try {
+            return hilti::rt::any_cast<StrongReference<T>>(_ptr).derefAsValue();
+        } catch ( const hilti::rt::bad_any_cast& ) {
+            throw IllegalReference("invalid target type");
+        }
+    }
```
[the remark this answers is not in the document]
rsmmr: I'm adding one.

spicy/toolchain/src/compiler/codegen/parsers/literals.cc
rsmmr: Did it because we don't need stateful so far and it allows getting away without a template. But I agree with the consistency argument, so changing to `T&&`.

spicy/toolchain/src/compiler/codegen/parsers/types.cc
rsmmr: Same.

PR #1738, opened 2024-05-13
Bring back `DeferredExpression`

rsmmr:
This reverts and updates the previous change to remove
`rt::DeferredExpression`. Turns out we still need it for the Zeek
integration.

- **Revert "Remove support for deferred expressions."**
- **Remove `std::functional` from `DeferredExpression`.**

### DIFF
--- a/hilti/runtime/include/deferred-expression.h
+++ b/hilti/runtime/include/deferred-expression.h
@@ -1,0 +1,60 @@
+// Copyright (c) 2020-2023 by the Zeek Project. See LICENSE for details.
+
+#pragma once
+
+#include <functional>
+#include <string>
+#include <utility>
+
+#include <hilti/rt/extension-points.h>
+#include <hilti/rt/types/result.h>
+
+namespace hilti::rt {
+
+/**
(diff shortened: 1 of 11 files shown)


bbannier (approved):

hilti/runtime/include/deferred-expression.h
```diff
@@ -0,0 +1,64 @@
...
+ * The expression must be wrapped into a function call, and it's evaluation
+ * is requested through the wrapper's call operator.
+ *
+ * The function should be stateless as it might be invoked an unspecified
+ * number of times.
+ */
+template<typename Result, typename Expr>
+class DeferredExpression {
+public:
+    DeferredExpression(Expr&& expr) : _expr(std::move(expr)) {}
```
bbannier, line 24: I believe `Expr&&` is an universal reference here, not necessarily an r-value.

```suggestion
    DeferredExpression(Expr&& expr) : _expr(std::forward<Expr>(expr)) {}
```

rsmmr: Had that, but that actually didn't compile for me.

PR #1739, opened 2024-05-13
Add new doc section with guidelines and best practices.

rsmmr:
- **Docs: Update feedback section.**
- **Docs: Update Custom Extensions section.**
- **Docs: Add new section with guidelines and best practices.**


doc/programming/guidelines.rst
```diff
@@ -0,0 +1,291 @@
...
+could access the fields anyways.
+
+The most common pitfall here is top-level units that parse a sequence
+of PDUs:
+
+.. code::
+
+    public type MyPDUs = unit {
+        msgs: Message[]; # accumulates all messages until end of session
+        : Message[];     # anonymous field, no accumulation
```
bbannier, line 103: I would clearly call out antipatterns, e.g., here

```suggestion
        msgs: Message[]; # DANGEROUS: accumulates all messages until end of session
        : Message[];     # SUGGESTED: anonymous field, no accumulation
```

There might be more instances below where this might help people stay clear of copying the incorrect line.

doc/programming/guidelines.rst
```diff
@@ -0,0 +1,291 @@
...
+        unused4: skip bytes &eod;   # skip over remaining data
+    };
+
+.. rubric::  Avoid Spicy strings and string manipulation
+
+Use Spicy :ref:`strings <type_string>` sparsely in your analyzer, and
+stick to :ref:`bytes <type_bytes>` instead where you can. Typically,
+you would convert ``bytes`` to strings as late as possible just before
+presentation to the user (including passing to Zeek events). Strings
+are never needed for the parsing process itself.
```
bbannier, line 131: Re: passing to Zeek, Zeek automatically converts Spicy `bytes` to Zeek `string`s so no explicit conversion is needed (unless users want to cache the conversion results). Maybe this instead:

```suggestion
Use Spicy :ref:`strings <type_string>` sparsely in your analyzer, and
stick to :ref:`bytes <type_bytes>` instead where you can. Typically, Spicy code
would use ``bytes`` throughout and only convert to ``string`` if a ``string`` is
required as a function call argument, or operations only supported by ``string``
need to be performed. When passing ``bytes`` to Zeek in EVT files they will be
converted to Zeek ``string`` values automatically. Strings are never needed for
the parsing process itself.
```

doc/programming/guidelines.rst
```diff
@@ -0,0 +1,291 @@
...
+
+As a corollary, avoid Spicy string manipulation. Always
+manipulate/concatenate ``bytes`` and convert only the final result to
+Spicy strings, probably in a ``%done`` hook or an EVT event. In
+particular, format strings come with a significant overhead. In other
```
bbannier, line 136: Nit: personally triggered by "overhead", maybe this?

```suggestion
particular, format strings are not cheap to compute. In other
```

doc/programming/guidelines.rst
```diff
@@ -0,0 +1,291 @@
...
+words, avoid use of ``%s`` to generate strings from ``bytes``.
+
+.. rubric::  Don't use temporary variables of expensive types just for readability
+
+Don't use temporary variables just to improve readability. In
+particular, ``string`` and ``bytes`` (that are implemented as C++
+strings under the hood and can be expensive to use) need to be
+allocated and destroyed. This may introduce relevant overhead as it
+cannot be guaranteed that the C++ compiler will be able to optimize
```
bbannier, line 145:
```suggestion
cannot be guaranteed that the C++ compiler will be able to optimize away the temporary in
```

doc/programming/guidelines.rst
```diff
@@ -0,0 +1,291 @@
...
+the code generated by Spicy. To improve readability, comments are the
+tool of choice. In theory, benchmarking the impact of using a
+temporary variable is a valid approach as well. However, the benefit
+of slightly increased readability most likely does not justify the
+effort of benchmarking.
```
bbannier, line 150: Especially the last sentence is weird since we say on top that benchmarking is the ultimate guideline so it is unclear to me why we talk about trade offs between readability and benchmarking at all. We could probably replace all this with: _If in doubt, benchmark._

doc/programming/guidelines.rst
```diff
@@ -0,0 +1,291 @@
...
+.. rubric::  Avoid recursion
+
+The Spicy compiler allows declaration of recursive units with runtime
+conditions dictating when the recursion terminates. However,
+benchmarking has revealed that recursion introduces much higher
+overheads as compared to unrolled linear/iterative code performing
+similar functionality. This overhead is due to increased lifetimes of
+the units, data and their associated hooks; as well as less potential
+for compiler optimization due to additional internal machinery around
+the recursive calls.
```
bbannier, line 184: I would rephrase it to remove the "revelations" (which would seem weird for an in-tree doc),

```suggestion
The Spicy allows declaration of recursive units with runtime conditions
dictating when the recursion terminates. A recursive grammar dictates that
units higher up in the recursion chain to be kept around until the overall
recursion terminates. This incurs additional memory overhead and might reduce
potential for optimizations. A unrolled linear/iterative implementation does
not have these issues.
```

doc/programming/guidelines.rst
```diff
@@ -0,0 +1,291 @@
...
+
+.. rubric::  Inline small nested units
+
+The Spicy compiler is not smart enough yet to inline nested units.
+Since declaration of units incurs additional cost to maintain their
+associated state and hooks, it is advisable to manually inline small
+units where performance is critical.
```
bbannier, line 191: Let's not call ourselves dumb,

```suggestion
Spicy currently does not inline small nested units. Since setting up new unit
values is not free small inlining small units into their parent unit can
increase runtime performance, especially where it is an identified bottleneck.
```

doc/programming/guidelines.rst
```diff
@@ -0,0 +1,291 @@
...
+
+.. rubric::  Avoid small sized byte fields
+
+Avoid using the ``bytes`` type for fields that could be handled using
+integer types. As bytes will always allocate a C++ string under the
+hood, using integer types can improve performance.
+
+.. rubric::  Consider outsourcing into C++
+
+Consider outsourcing complex calculations required for your parsing
```
bbannier, line 229:
```suggestion
Consider outsourcing complex and performance-critical calculations required for your parsing
```

doc/programming/guidelines.rst
```diff
@@ -0,0 +1,291 @@
...
+
+.. rubric::  State-management in analyzers
+
+Using global maps to retain analyzer state can cause significant
+memory bloating over long(er) periods. The preferred way is to define
+a :ref:`%context <unit_context>` in top-level units and propagate that
+down to populate any maps or other state inside that context. When
+used with Zeek, Spicy ties the context state to individual connections
+that get teared down when the connection state is removed, thereby
+preventing any state space explosion.
```
bbannier, line 243: Even with a context a long-running connection could still lead to unbounded state growth. Maybe rephrase as

```suggestion
Using global maps to retain analyzer state can cause significant
memory bloating over long(er) periods if not managed correctly. If statekeeping cannot be avoided, the preferred way is to define
a :ref:`%context <unit_context>` in top-level units and propagate that
down to populate any maps or other state inside that context. When
used with Zeek, Spicy ties the context state to individual connections
that get teared down when the connection state is removed, thereby
preventing accidental state space explosion. Even when state tables are stored
in a ``%context`` one still needs to manually make sure that its size does not
grow unbounded (e.g., for long-running connections).
```

doc/programming/guidelines.rst
```diff
@@ -0,0 +1,291 @@
...
+Compilation Performance
+-----------------------
+
+Depending on the complexity of the Spicy code, it may take a bit (and
+sometimes quite a while) to compile your parsers. In the following, we
+collect some recommendations to speed up the compile process.
+
+.. note::
+
+    When processing Spicy code, generally the bulk of the time time
```
bbannier, line 256:
```suggestion
    When processing Spicy code, generally the bulk of the time
```

doc/programming/guidelines.rst
```diff
@@ -0,0 +1,291 @@
...
+
+During development of new parsers, it helps quite a bit to build
+non-optimized debug versions by adding ``--debug`` to the Spicy
+compiler's command-line. This emits almost identical code, but then
+compiles the generated code without ``-O2`` (i.e., not optimized),
+which avoids some work the C++ compiler would otherwise do. The
+produced HLTO will perform (much) less well so it is probably not
+useful for production.
+
+.. note::
```
bbannier, line 281: Let's call this out more strongly.

```suggestion
.. danger::
```

bbannier (dismissed):
Thanks for porting these note over. I left some comments, mostly around adjusting tone for in-tree docs.

doc/programming/guidelines.rst
```diff
@@ -0,0 +1,291 @@
...
+And if you aren't going for the handful of most common high-volume
+protocols (e.g., HTTP, DNS, TLS), then your parser will very likely
+end up processing only a tiny fraction of the overall traffic. At that
+point, its runtime performance is going to be dwarfed by everything
+else Zeek is doing. Hence, take the following with a grain of salt.
+Usually it's best to get you parser working first, then benchmark to
+see if you need to improve performance, and finally find the primary
+bottlenecks of needed.
+
+.. _performance_runtime:
```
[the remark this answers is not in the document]
rsmmr: Actually I prefer the explicit style because it means we can change the section titles without breaking links.

doc/programming/guidelines.rst
rsmmr: That new text doesn't quite parse for me. I also don't have problem calling myself dumb. :-) But I'll tweak it a bit to tone down the "it is advisable".

doc/programming/guidelines.rst
```diff
@@ -0,0 +1,291 @@
...
+protocol in Spicy, the resulting performance is a direct factor of the
+amount of traffic that parser will be processing. For example, in a
+typical Zeek setup, where you network carries a mix of various
+protocols, your parser will see only a subset of the overall traffic.
+And if you aren't going for the handful of most common high-volume
+protocols (e.g., HTTP, DNS, TLS), then your parser will very likely
+end up processing only a tiny fraction of the overall traffic. At that
+point, its runtime performance is going to be dwarfed by everything
+else Zeek is doing. Hence, take the following with a grain of salt.
+Usually it's best to get you parser working first, then benchmark to
```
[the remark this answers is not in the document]
rsmmr: That sounds to me like a separate section somewhere that we would just link to from over here. I won't tackle that within the scope of this PR.

doc/programming/guidelines.rst
rsmmr: While bytes convert to strings automatically for events, that doesn't take character sets into account. I'll try to clarify that.

doc/programming/guidelines.rst
rsmmr: Merging the two to stay with active voice. :-)

doc/programming/guidelines.rst
rsmmr: I'll try to tweak it a bit differently.

PR #1741, opened 2024-05-14
Bump to clang-18 in CI.

bbannier:


bbannier:
Looks like we will need to work around gulrak/filesystem#183 to make `lint` pass.

rsmmr (approved):

PR #1751, opened 2024-06-02
Fix a typo in pack/unpack docs

Kardbord:
Fix a typo in the `unpack` docs for `real` values where the template argument to `unpack` was `uintX|intX` instead of `real`.


bbannier (approved):
Thanks!

PR #1753, opened 2024-06-05
Update RTD config to latest version.

bbannier:
Since this version was first released readthedocs has revamped their approach to project configuration and also become much stricter in what they allow. Due to that the docs for the latest 1.8 release do not build anymore.

This copies the RTD config from `main` into this branch.


awelzel (approved):
I gather removal of `python.system_packages` is the crucial bit.

Screenshot from 2024-06-05 12-00-54

PR #1755, opened 2024-06-06
Add a couple new methods to data types

rsmmr:
- **Add `get_optional` method to maps.**
- **Add `to_real` method to `bytes`.**
- **Update some debug baselines.**


hilti/runtime/include/types/map.h
```diff
@@ -222,6 +222,19 @@ class Map : protected std::map<K, V> {
         }
     }
 
+    /**
+     * Attempts to get the value for a key.
+     *
+     * @param k key to retrieve
+     * @return the value, or an unset optional if the key is not set in the map
+     */
+    std::optional<V> get_optional(const K& k) const& {
```
bbannier, line 231: Let's add coverage for this in the unit tests, e.g.,

```cpp
TEST_CASE("get_optional") {
    {
        auto m = Map<int, int>{{1, 11}};

        CHECK_EQ(m.get_optional(1), 11);
        CHECK_EQ(m.get_optional(42), std::optional<int>{});
    }

    {
        auto m = Map<int, std::optional<int>>{{1, 11}};

        CHECK_EQ(m.get_optional(1), 11);

        auto x = m.get_optional(42);
        static_assert(std::is_same_v<decltype(x), std::optional<std::optional<int>>>);
        CHECK_EQ(x, std::optional<std::optional<int>>{});
    }
}
```

bbannier, line 231: Just wondering about the name, is maybe `try_get` more intuitive than `get_optional`? No strong preference though.

hilti/runtime/include/types/map.h
```diff
@@ -222,6 +222,19 @@ class Map : protected std::map<K, V> {
...
+        if ( auto it = this->find(k); it != static_cast<const M&>(*this).end() )
```
bbannier, line 232:
```suggestion
        if ( auto it = this->find(k); it != M::end() )
```

tests/hilti/types/bytes/to-real.hlt
```diff
@@ -0,0 +1,14 @@
+# @TEST-EXEC: hiltic -j %INPUT >output
+# @TEST-EXEC: btest-diff output
```
bbannier, line 2: Nothing in this baseline.
```suggestion
```

hilti/runtime/src/types/bytes.cc
```diff
@@ -205,6 +205,24 @@ uint64_t Bytes::toUInt(ByteOrder byte_order) const {
...
+        throw InvalidValue("cannot parse real value: null byte in data");
+
+    const char* cstr = Base::c_str();
+    char* endp = nullptr;
+
+    errno = 0;
+    auto d = strtod(cstr, &endp);
+    if ( endp == cstr || *endp != '\0' || (d == HUGE_VAL && errno == ERANGE) ) {
+        errno = 0;
+        throw InvalidValue(fmt("cannot parse real value: %s", cstr));
```
bbannier, line 220: Let's maybe put the value in quotes to make it more identifiable, e.g.,

```suggestion
        throw InvalidValue(fmt("cannot parse real value: '%s'", cstr));
```

hilti/runtime/src/types/bytes.cc
```diff
@@ -205,6 +205,24 @@ uint64_t Bytes::toUInt(ByteOrder byte_order) const {
     return i;
 }
 
+double Bytes::toReal() const {
```
bbannier, line 208: Can you add unit test coverage for this? This is what I checked, but there are probably more things we could encode

```cpp
TEST_CASE("toReal") {
    CHECK_EQ("100"_b.toReal(), 100);
    CHECK_EQ("0."_b.toReal(), 0.);

    CHECK_EQ("0.5"_b.toReal(), 0.5);
    CHECK_EQ("-0.5"_b.toReal(), -0.5);
    CHECK_EQ("+0.5"_b.toReal(), +0.5);
    CHECK_EQ(".5"_b.toReal(), 0.5);
    CHECK_EQ("-.5"_b.toReal(), -0.5);

    CHECK_EQ("1e42"_b.toReal(), 1e42);
    CHECK_EQ("+1e42"_b.toReal(), 1e42);
    CHECK_EQ("-1e42"_b.toReal(), -1e42);

    CHECK_EQ("1e+42"_b.toReal(), 1e42);
    CHECK_EQ("1e-42"_b.toReal(), 1e-42);

    CHECK_EQ("inf"_b.toReal(), std::numeric_limits<double>::infinity());
    CHECK_EQ("-inf"_b.toReal(), -std::numeric_limits<double>::infinity());

    CHECK(std::isnan("nan"_b.toReal()));
    CHECK(std::isnan("-nan"_b.toReal()));

    CHECK_THROWS_WITH_AS(""_b.toReal(), "cannot parse real value: ", const InvalidValue&);
    CHECK_THROWS_WITH_AS("abc"_b.toReal(), "cannot parse real value: abc", const InvalidValue&);
    CHECK_THROWS_WITH_AS("a.2"_b.toReal(), "cannot parse real value: a.2", const InvalidValue&);
    CHECK_THROWS_WITH_AS("2.a"_b.toReal(), "cannot parse real value: 2.a", const InvalidValue&);
}
```

hilti/toolchain/src/ast/operators/bytes.cc
```diff
@@ -876,6 +876,29 @@ byte order, and converts it into an unsigned integer.
...
+            .self = {parameter::Kind::In, builder->typeBytes()},
+            .member = "to_real",
+            .result = {Constness::Const, builder->typeReal()},
+            .ns = "bytes",
+            .doc =
+                R"(
+Interprets the ``bytes`` as representing an ASCII-encoded floating point number
+and converts that into a ``real``. The data can be in either decimal or
+hexadecimal format. If it cannot be parsed as either, throws an `InvalidValue`
+exception.
```
bbannier, line 893: All of this depends on the locale (decimal separator), could you add something calling that out?

hilti/runtime/src/types/bytes.cc
```diff
@@ -205,6 +205,24 @@ uint64_t Bytes::toUInt(ByteOrder byte_order) const {
...
+    // Ensure there are no null bytes inside our data, so that we can call strtod().
+    if ( Base::find('\0') != Base::npos )
+        throw InvalidValue("cannot parse real value: null byte in data");
+
+    const char* cstr = Base::c_str();
+    char* endp = nullptr;
+
+    errno = 0;
+    auto d = strtod(cstr, &endp);
```
bbannier, line 217: I was about to suggest `std::from_chars`, but float parsing is currently unsupported on libc++ 😞

bbannier (approved):

hilti/runtime/include/types/map.h
rsmmr: Actually I had considered `try_get` and also `get_try`, but ended up preferring `get_optional`. I'll leave it at that.

hilti/runtime/src/types/bytes.cc
rsmmr: Yeah, it's a pain. :-(

hilti/toolchain/src/ast/operators/bytes.cc
rsmmr: This is a good one. I don't think we want it to depend on locale, that could end up being very surprising. From a look it seems hard to come up with version that doesn't unfortunately, but I'll look around a bit more.

doc/autogen/types/bytes.rst
```diff
@@ -96,6 +96,13 @@
     Interprets the ``bytes`` as representing an binary number encoded with
     the given byte order, and converts it into signed integer.
 
+.. spicy:method:: bytes::to_real bytes to_real False real ()
+
+    Interprets the ``bytes`` as representing an ASCII-encoded floating
+    point number and converts that into a ``real``. The data can be in
```
bbannier, line 102: Let's call out that we assume a C locale.

bbannier (dismissed):

hilti/runtime/src/tests/bytes.cc
```diff
@@ -437,6 +437,18 @@ TEST_CASE("toReal") {
     CHECK_THROWS_WITH_AS("abc"_b.toReal(), "cannot parse real value: 'abc'", const InvalidValue&);
     CHECK_THROWS_WITH_AS("a.2"_b.toReal(), "cannot parse real value: 'a.2'", const InvalidValue&);
     CHECK_THROWS_WITH_AS("2.a"_b.toReal(), "cannot parse real value: '2.a'", const InvalidValue&);
+
+    // The next test should fail independent of the locale, so let's set one.
+
+    auto de_locale = newlocale(LC_ALL_MASK, "de_DE.UTF-8", nullptr);
```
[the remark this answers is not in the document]
rsmmr: I don't think that is actually well defined, the man pages differ for macOS vs BSD/Linux (`xlocale.h` vs `locale.h`). Given that it compiles on all platforms as it is, I'd rather not touch it ...

PR #1757, opened 2024-06-10
 Expose `result` and `spicy::Error` types to Spicy to facilitate error handling.

rsmmr:
Add `result` and `spicy::Error` types to Spicy to facilitate error handling.

The `result` and `error` types were already implemented internally
HILTI-side, but not yet available to Spicy users. This exposes them to
Spicy as well. To avoid name clashes with existing code, we don't
introduce `error` as a new type keyword, but instead make it available
as a library type ``spicy::Error``.

Typical usage is something like this:

```
function foo() : result<int64> {
     ...
    if ( everything_is_ok )
        return 42;
    else
        return error"Something went wrong.";
}

if ( local x = foo() )
    print "result: %d " % *x;
else
    print "error: %s " % x.error();
```

The documentation has more specifics.


rsmmr:
One change from the earlier draft: We now make the error type available as `spicy::Error`, that makes it accessible, but still avoids a new `error` keyword.

bbannier (approved):
Seems fine with me.

Do we want to use the opportunity to deprecate `error` as an `ID` in user code? That'd be preparation to in the future potentially also provide support for non-literal error strings with a ctr-like syntax (e.g., `error("The thing broke: %s" % concrete error)`).

rsmmr:
> Do we want to use the opportunity to deprecate `error` as an `ID` in user code? 

I don't think I'd want that become a reserved word at all, `error` just seems too common/intuitive to use in a Spicy program.

PR #1758, opened 2024-06-11
Bump centos-stream in CI.

bbannier:
centos8-stream became EOL 2024-05-31. The powertools-related changes to
the setup are borrowed from what is done in Zeek.


rsmmr (approved):
Thanks!

PR #1761, opened 2024-06-12
Fix generated code for huge `const` collections.

### LINKED ISSUE #1760: Unrolling of big collection constructors fails to compile for `const` variables
With #1744 we introduced code which unrolls constructors of huge containers. The code we now generate is incorrect for `const` variables which do not live in a scope (it does work for `global`s though which are managed more thightly by the module).

```ruby
module foo;

# Ctrs for collections of more than 10 elements get unrolled.
const CRC16_TABLE: vector<uint16> = [
  0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
];
```

```console
$ spicyc -dj foo.spicy
foo_198ff91772b68fab-628dbe1417976318.cc:12:84: error: non-local lambda expression cannot have a capture-default
    const ::hilti::rt::Vector<::hilti::rt::integer::safe<uint16_t>> CRC16_TABLE = [&]() { ...
                                                                                   ^
```

We probably need to adjust or even drop the capture for declarations emitted outside of a proper C++ scope. ~We still need to make sure that a constructor can reference other `const` variables though.~ A declaration of a `const` can only refer to other `const` variables; since `const` variables are emitted as non-local variables we do not need to capture them.

bbannier:
When unrolling constructors of collections we previously would always capture explicitly so that container elements could reference other variables in the scope. This breaks for `const` variables which are emitted at non-local scope (into a namespace); the lambda performing the initialization would also be a non-local and non-local lambdas are not allowed to capture anything (they do not live in a scope after all).

This means that we cannot capture for such lambdas at all which seems to be an issue -- after all with that it seems impossible to support constructors of collections which reference other `const`s. Fortunately C++ lambdas at non-local scope implicitly capture other non-locals already and they do not need to be captured explictly. This allows for a rather simple fix.

Closes #1760.

### DIFF
--- a/hilti/toolchain/src/compiler/codegen/ctors.cc
+++ b/hilti/toolchain/src/compiler/codegen/ctors.cc
@@ -116,7 +116,13 @@ struct Visitor : hilti::visitor::PreOrder {
                                                     }),
                                     " ");
 
-            result = fmt("[&]() { auto __xs = ::hilti::rt::Map<%s, %s>(); %s return __xs; }()", k, v, elems);
+            // If we are at block scope capture other variables so they can be
+            // used in the ctr. Outside of block scope we are emitting a
+            // non-local `const` (into a namespace) which can reference only
+            // other `const` variables which since they are non-locals as well
+            // can be referenced without capturing.
+            auto captures = (cg->cxxBlock() == nullptr) ? "" : "&";
+            result = fmt("[%s]() { auto __xs = ::hilti::rt::Map<%s, %s>(); %s return __xs; }()", captures, k, v, elems);
         }
 
         else
@@ -199,7 +205,13 @@ struct Visitor : hilti::visitor::PreOrder {
                                            [this](const auto& e) { return fmt("__xs.insert(%s);", cg->compile(e)); }),
                            " ");
 
-            result = fmt("[&]() { auto __xs = ::hilti::rt::Set<%s>(); %s return __xs; }()", k, elems);
+            // If we are at block scope capture other variables so they can be
+            // used in the ctr. Outside of block scope we are emitting a
+            // non-local `const` (into a namespace) which can reference only
+            // other `const` variables which since they are non-locals as well
+            // can be referenced without capturing.
+            auto captures = (cg->cxxBlock() == nullptr) ? "" : "&";
+            result = fmt("[%s]() { auto __xs = ::hilti::rt::Set<%s>(); %s return __xs; }()", captures, k, elems);
         }
 
         else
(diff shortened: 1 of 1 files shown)


hilti/toolchain/src/compiler/codegen/ctors.cc
```diff
@@ -116,7 +116,13 @@ struct Visitor : hilti::visitor::PreOrder {
                                                     }),
                                     " ");
 
-            result = fmt("[&]() { auto __xs = ::hilti::rt::Map<%s, %s>(); %s return __xs; }()", k, v, elems);
+            // If we are at block scope capture other variables so they can be
+            // used in the ctr. Outside of block scope we are emitting a
+            // non-local `const` (into a namespace) which can reference only
+            // other `const` variables which since they are non-locals as well
+            // can be referenced without capturing.
+            auto captures = cg->cxxBlock() == nullptr ? "" : "&";
```
rsmmr, line 124:
```suggestion
            auto captures = (cg->cxxBlock() == nullptr) ? "" : "&";
```

hilti/toolchain/src/compiler/codegen/ctors.cc
```diff
@@ -199,7 +205,13 @@ struct Visitor : hilti::visitor::PreOrder {
                                            [this](const auto& e) { return fmt("__xs.insert(%s);", cg->compile(e)); }),
                            " ");
 
-            result = fmt("[&]() { auto __xs = ::hilti::rt::Set<%s>(); %s return __xs; }()", k, elems);
+            // If we are at block scope capture other variables so they can be
+            // used in the ctr. Outside of block scope we are emitting a
+            // non-local `const` (into a namespace) which can reference only
+            // other `const` variables which since they are non-locals as well
+            // can be referenced without capturing.
+            auto captures = cg->cxxBlock() == nullptr ? "" : "&";
```
rsmmr, line 213:
```suggestion
            auto captures = (cg->cxxBlock() == nullptr) ? "" : "&";
```

hilti/toolchain/src/compiler/codegen/ctors.cc
```diff
@@ -287,8 +299,14 @@ struct Visitor : hilti::visitor::PreOrder {
...
                                     " ");
 
-            result = fmt("[&]() { auto __xs = ::hilti::rt::Vector<%s%s>(); __xs.reserve(%d); %s return __xs; }()", x,
-                         allocator, size, elems);
+            // If we are at block scope capture other variables so they can be
+            // used in the ctr. Outside of block scope we are emitting a
+            // non-local `const` (into a namespace) which can reference only
+            // other `const` variables which since they are non-locals as well
+            // can be referenced without capturing.
+            auto captures = cg->cxxBlock() == nullptr ? "" : "&";
```
rsmmr, line 307:
```suggestion
            auto captures = (cg->cxxBlock() == nullptr) ? "" : "&";
```

rsmmr (approved):

hilti/toolchain/src/compiler/codegen/ctors.cc
```diff
@@ -116,7 +116,13 @@ struct Visitor : hilti::visitor::PreOrder {
                                                     }),
                                     " ");
 
-            result = fmt("[&]() { auto __xs = ::hilti::rt::Map<%s, %s>(); %s return __xs; }()", k, v, elems);
+            // If we are at block scope capture other variables so they can be
```
bbannier, line 119: One caveat with not capturing is that we rely on any constants referenced by the declaration to have be declared before (since we emit C++ `const`: actually initialized), e.g., code like this would work as expected:

```ruby
const a = 1;
const b = vector(a, ...);
```

In general the expectation in Spicy/HILTI is that order of declarations does not matter, but this currently does not work for `const` variables, see #1763. That is unfortunate, but luckily in this case means that an empty capture `[]` should be sufficient.

PR #1764, opened 2024-06-13
Fix `if`-condition with `switch` parsing.

### LINKED ISSUE #1759: Trailing `if` on unit `switch` is ignored
We allow code with a unit `switch` with a trailing `if`, e.g.,

```ruby
public type X = unit {
    switch {
        -> a: b"A";
        -> b: b"B";
        -> c: b"C";
    } if(False) ;
};
```

The condition is however silently ignored; for above `False` I would expect `X` to consume no data but it does

```console
$ printf 'A' | spicy-dump -d foo.spicy
foo::X {
  a: A
}
```

We should either interpret such conditions correctly or reject them outright.

rsmmr:
The parser generator was ignoring `if` conditions attached to `switch`
constructs. While we actually had a test for this already, turns out
we had recorded a broken baseline. Plus, we were testing only one
variant of `switch` (expression-based, not look-ahead-based). This
implements and tests both variants now.

Closes #1759.


bbannier (approved):
Wow, that's indeed a tiny fix, thanks!

Issue #1133, opened 2022-02-21
Consider taking advantage of underlying data type during `LiteralMode::Search`


bbannier:
In the current implementation of literal search mode we simply`advance` the input by one byte until the given production can be parsed, e.g., [here](https://github.com/zeek/spicy/blob/ec00098525b99cd1dc524933c0beb7a8b94df6c3/spicy/toolchain/src/compiler/codegen/parser-builder.cc#L973).

While this works it very likely is not the most efficient search algorithm. When searching literals we could e.g., use a [Boyer-Moore search](https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore_string-search_algorithm) to find matches; similar improvements might be possible for regexp productions, but should probably be implemented in https://github.com/rsmmr/justrx/.

bbannier:
A user reported that they see the search during recovery take considerable amount of time (80-90%) if there are a lot of gaps (here: high packet loss in Zeek). They synchronize at a regex which is not complicated and matches an input of fixed width, and it seems like most of the time is spent starting off regex searches at each new byte in the input.

I am moving this issue into TODO state since it seems they could benefit from this (e.g., matching a regex for 10 characters with Boyer-Moore could slash time in recovery from 80-90% to <10%).

bbannier:
Dropping this from TODO again since the ultimate issue is high packet loss, and even a more optimized implementation could eventually become too slow.

JustinAzoff:
A problem here is that while packet loss may be the initial trigger for this issue, this issue then causes high cpu usage, which causes more packet loss and I don't think the process ever fully recovers.

rsmmr:
> A problem here is that while packet loss may be the initial trigger for this issue, this issue then causes high cpu usage, which causes more packet loss and I don't think the process ever fully recovers.

What do you think about [this idea](https://github.com/zeek/zeek/issues/3779#issuecomment-2165166375)? Basically, provide the tools for the analyzer to disable itself when things get crazy?

JustinAzoff:
If I'm understanding things correctly, when spicy is looking for a regex like "SMB" to resynchronize, it doesn't search for the regex inside buffered data, but instead tries to match it at each possible offset, turning what could be a constant time operation into quadratic.

Is what it is doing equivalent to these two functions in python?

```
import random
import re

fn = "smb_fast_10000.pcap"
with open(fn, 'rb') as f:
    data = f.read()

PKT_SIZE = 1500

def resync_match(pkt):
[...]

PR #1767, opened 2024-06-21
Collect stream statistics and offer new hook to check in on them

### LINKED ISSUE #1768: Add tooling or guidance on how a Spicy parser with error recovery should deal with huge packet loss
Spicy parser can provide [error recovery](https://docs.zeek.org/projects/spicy/en/latest/programming/parsing.html#error-recovery) so that they can resynchronize after parse errors (probably most common: after parse errors triggered by gaps due to packet loss). During error recovery the Spicy parser will enter a search mode where it searches the input stream for a pattern to synchronize on.

This process works, but is not free since the search might be (much) more expensive than parsing, especially if synchronization fails repeatedly due to more gaps before a match is found, and users reported that parsers spend most of their CPU time in recovery if there is high packet loss. We currently have the following ways to deal with that:

- use `%synchronize-at`/`%synchronize-after` to use a dedicated and possible more optimized search pattern
- zeek/spicy#1673: allow to use a custom function to find a synchronization points

If packet loss is high enough these will ultimately still fail (similar argument why even a more optimized search impl with zeek/spicy#1133 would at a certain point also be insufficient).

We should provide or at least document an approach for parsers to opt-out of further parsing of a connection if packet loss is too high. An automated approach could look at some packet loss metric (which one?) and then deactivate the parser if it exceeds a threshold (likely: configurable since I'd expect the trade-off between visibility and throughput to be site-specific). Alternatively we could also document an approach on how to implement such a mitigation (e.g., inspect stats counters and manually perform the deactivation). It could be that either of this requires additional Spicy-side hooks or metrics.

rsmmr:
This extends the `stream` type to track statistics about their data, and it adds
new `%sync_advance` hook that can be used to check statistics during
synchronization. For more information, see the individual commits.

Closes #1768.


tests/spicy/types/unit/sync-advance.spicy
```diff
@@ -0,0 +1,75 @@
...
+public type X1 = unit {
+    %port = 80/tcp;
+
+    xs: (/(A|B|C)/ &synchronize)[];
+
+    on %synced {
+        confirm;
+    }
+
+    on %sync_advance(offset: uint64) {
```
bbannier, line 21: Attaching general feedback here.

Since that the primary use case for this hook would be to raise a parse error if we do not synchronize fast enough, having this hook for every advance in sync mode in general seems like an expensive to me (even if we advanced less often by implementing #1133). That we do not emit it if it is unused does not help parsers which want to handle such scenarios. The original idea with providing a `%gap` hook might have been insufficient, but at least it likely would have led to us invoking the hook orders of magnitude less.

To get a feel for the cost of invoking this hook I ran the following parser:

```ruby
module foo;

public type X = unit {
    : (b"1" &synchronize)[];

    # Leave this in or comment it out to compare hook cost.
    # Also call `$stream.statistics()` in hook for 3rd benchmark.
    on %sync_advance(offset: uint64) { }

    on %synced {
        confirm;
    }
};
```

I ran this on an input file containing containing a `1` followed by 10,000,000 `0`, and then another `1`. I got the following timings:

```console
$ hyperfine -L h wo_hook,w_hook,w_hook_statistics 'spicy-driver {h}.hlto -f data'
Benchmark 1: spicy-driver wo_hook.hlto -f data
  Time (mean ± σ):     994.3 ms ±  17.2 ms    [User: 981.3 ms, System: 11.5 ms]
  Range (min … max):   960.9 ms … 1028.4 ms    10 runs

Benchmark 2: spicy-driver w_hook.hlto -f data
  Time (mean ± σ):      1.291 s ±  0.012 s    [User: 1.276 s, System: 0.013 s]
  Range (min … max):    1.276 s …  1.320 s    10 runs

Benchmark 3: spicy-driver w_hook_statistics.hlto -f data
  Time (mean ± σ):      1.399 s ±  0.019 s    [User: 1.382 s, System: 0.015 s]
  Range (min … max):    1.376 s …  1.430 s    10 runs

Summary
  spicy-driver wo_hook.hlto -f data ran
    1.30 ± 0.03 times faster than spicy-driver w_hook.hlto -f data
    1.41 ± 0.03 times faster than spicy-driver w_hook_statistics.hlto -f data
```

I.e. just having the hook invoked increased the parsing cost by 30%, and doing a minimal operation (just accesing `$stream.statistics()`) adds another 10%. This looks like a lot.

I am wondering whether we need a way to control how often such a hook gets invoked. We could introduce a notion of "scheduled" hook which gets triggered after a certain amount of data has passed through the parser, e.g.,

```ruby
public type X = unit {
   # ...
    on %init {
        # When entering search mode invoke `%sync_mode_scheduled` every 10,000 bytes.
        # Horrible name.
        spicy::sync_mode_schedule_every(10000);
    }

    # This is a hook so it has access to `self`.
    # Also horrible name.
    on %sync_mode_scheduled(offset: uint64) { $stream.statistcs(); }
};
```

I am not sure such a scheduled hook would be useful outside of sync mode.

rsmmr: Thanks, for the measurements, that's pretty bad. I was trying to do this without the need for configuration (like you do with the scheduling) nor additional runtime state. But maybe that's not viable. 

Not sure I like the dynamic schedule because that value would need to be stored inside the unit and then retrieved for each check. How about a compile-property property instead with some reasonable default:

```
public type X = unit {
    %sync_advance_block_size = 32768; # if absent, default to some reasonable value, like 4K

    # Executes during synchronization whenever the current offset 
    # crosses a `%sync_advance_block_size` boundary
    on %sync_advance_block(offset: uint64) { $stream.statistcs(); }
};
```

(Open to better names).

bbannier: I like the idea with a property since not a lot of this is dynamic anyway (properties can already take an `expr` so the exact value can be made easily customizable via e.g., Zeek config values). The only downside is that one cannot set property values outside of the unit like one can add hooks, but that is probably okay.

> `    %sync_advance_block_size = 32768; # if absent, default to some reasonable value, like 4K`

I would even go so far as requiring both property and hook if either is specified. We could set some default, but this is a very low-level tool and it might be worth forcing users to think about what they want.

hilti/toolchain/include/ast/expressions/keyword.h
```diff
@@ -19,14 +19,15 @@ enum class Kind {
     Self,         /**< `self` */
     DollarDollar, /**< `$$` */
     Captures,     /**< `$@` */
-    Scope         /**< `$scope`*/
+    Scope,        /**< `$scope`*/
+    Stream        /**< `$stream`*/
```
bbannier, line 23: I do understand where you want to take `$stream` yet. It would seem to me that it should be possible to implement this as a free-standing accessor function in `spicy`, e.g., `spicy::stream` (we currently have e.g., `spicy::current_time` as a similar construct). Do you have more functionality around `$ID`-style identifiers in mind or is this inspired by [EVT-style variables](https://docs.zeek.org/en/master/devel/spicy/reference.html#:~:text=likewise%20left%20unset.-,In%20addition%20to%20full,only%20with%20protocol%20analyzers.,-Note) -- I believe there we had a syntax need for them to make clear they were neither Spicy nor Zeek).

bbannier: Unless we want to provide direct access to the stream in the future we could probably get by with e.g., `spicy::stream_statistics()`.

tests/spicy/types/unit/sync-advance.spicy
rsmmr: >  The only downside is that one cannot set property values outside of the unit like one can add hooks, but that is probably okay.

Not quite like hooks but one can define it at the module level to set a different global default.

> I would even go so far as requiring both property and hook if either is specified. We could set some default, but this is a very low-level tool and it might be worth forcing users to think about what they want.

I think I prefer having a default. Assuming this is relevant only for large input stream anyways, the concrete value doesn't even matter that much. So we'd pick reasonable default (as in: unlikely to hurt performance) but allow fine-tuning.
 
> I think your suggestions work well together and make sense. The only potentially confusing part could be `block` since this is not about _chunk_, WDYT about `%sync_advance_check_interval` and `sync_advance_check`?

I don't like "interval" because it suggests something time-based to me (especially because we also have an `interval` type). Same goes for the earlier "schedule". I picked `block` so that's it different from "chunk", but yeah, I admit that subtle.

rsmmr:
Pushed a new version that introduces a property `%sync-advance-block-size` with the discussed semantics, and a default of 4K. I left the hook at `%sync-advance`. Staying with the `$stream` for now. Still open for more ideas on naming.

hilti/toolchain/include/ast/expressions/keyword.h
rsmmr: > I picked `$stream`

An in-between syntax would be making it a unit method: `self.stream()`.  That would make it more like `offset()`. How about that?

bbannier: > An in-between syntax would be making it a unit method: self.stream(). That would make it more like offset(). How about that?

I'd like that much more than a magic `$stream` since it would make it clear that this only returns a useful result in a unit's context (also: less magic syntax to pick up 😉).

rsmmr:
Ready for review now.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -2582,16 +2585,35 @@ void ParserBuilder::skip(Expression* size, const Meta& location) {
...
 
 void ParserBuilder::advanceToNextData() {
-    auto offset = builder()->memberCall(state().cur, "offset");
-    auto profiler = builder()->startProfiler(hilti::util::fmt("spicy/unit/%s/__gap__", state().unit_id), offset);
+    auto old_offset = builder()->addTmp("old_offset", builder()->memberCall(state().cur, "offset"));
+
+    auto profiler =
+        builder()->startProfiler(hilti::util::fmt("spicy/unit/%s/__sync_advance__", state().unit_id), old_offset);
 
     builder()->addAssign(state().cur, builder()->memberCall(state().cur, "advance_to_next_data"));
```
bbannier, line 2593: The caveat here is that we will not call the hook exactly after `%sync-advance-block-size` bytes if that would mean calling from inside a gap chunk (instead the hook is called only on data chunks). This might be surprising and pretty subtle, but is likely inconsequential since dealing with gap chunks is cheap.

If we wanted to deal with this on a more granular level, we could call the hook in data chunks, or on chunk boundaries (so that a parser can react to only seeing gap chunks). This would require hooking into `advance_to_next_data`.

spicy/toolchain/src/compiler/codegen/unit-builder.cc
```diff
@@ -219,6 +221,23 @@ UnqualifiedType* CodeGen::compileUnit(type::Unit* unit, bool declare_only) {
...
+                                                             hilti::Constness::Const),
+                                    builder()->attributeSet({builder()->attribute("&internal"), attr_uses_stream}));
+
+    v.addField(stream);
+
+    auto attr_sync_advance = builder()->attributeSet(
+        {builder()->attribute("&needed-by-feature", builder()->stringLiteral("uses_sync_advance"))});
+
+    add_hook("0x25_sync_advance", {builder()->parameter("offset", builder()->typeUnsignedInteger(64))},
+             attr_sync_advance);
```
bbannier, line 239: Not sure we need this optimization and associated feature constant at this point, but maybe worth leaving it in since we have it.

bbannier (approved):
I checked that `self.stream()` does not do something crazy for `&parse-from` (e.g., `__stream` becoming invalid in a subunit); since the code is so simple not sure we necessarily need dedicated tests for that (but might still be nice).

spicy/toolchain/src/compiler/codegen/parser-builder.cc
bbannier: Thinking about this more, it might be beneficial to document that we call the hook _earliest_ after `%sync-advance-block-size` bytes (i.e., current behavior, maybe just a docs tweak). With that we could then optimize advancing more freely in the future.

tests/spicy/types/unit/stream-statistics.spicy
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    on %done {
+        local stats = self.stream().statistics();
+        assert stats.num_data_bytes == 4;
+        assert stats.num_data_chunks == 2;
+        assert stats.num_gap_bytes == 1024;
+        assert stats.num_gap_chunks == 1;
+        print stats;
+    }
```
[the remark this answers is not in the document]
rsmmr: Yeah, I think it's fine. `%init` signals start of parsing, not construction of the object. So the stream may have some data at that point.

spicy/toolchain/src/compiler/codegen/unit-builder.cc
rsmmr: Not sure I follow. The optimization is to avoid calling the hook if we don't have an implementation (along with pre/post hook stuff). Do you mean something else?

spicy/toolchain/src/compiler/codegen/parser-builder.cc
rsmmr: Ok, I'll tweak the docs. I did mean to leave this somewhat fuzzy for easier/cheaper implementation, and because it shouldn't matter much.

spicy/toolchain/src/compiler/codegen/unit-builder.cc
bbannier: You are right, I got slightly confused. What we already do today is remove hooks which are not implemented, but the situation here is different since we want to call the hook depending on a feature constant. Resolving.

PR #1772, opened 2024-06-28
Improve doc wrt. %error hook in sync mode

J-Gras:


bbannier (approved):
Thanks for this!

I left a small comment which I will fix up during the merge.

PR #1773, opened 2024-06-28
Call `%sync_advance` hook also when entering sync mode.

bbannier:
This allows users to control the full synchronization lifecycle, from entering it, to regular checks during the search (both via `%sync_advance`, until it is left (via their explicit `confirm`, e.g., from a `%synced` hook).


spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -2587,13 +2587,21 @@ void ParserBuilder::skip(Expression* size, const Meta& location) {
 void ParserBuilder::advanceToNextData() {
     auto old_offset = builder()->addTmp("old_offset", builder()->memberCall(state().cur, "offset"));
 
+    // Call any `%sync_advance` hook as we enter sync mode.
+    guardFeatureCode(state().unit, {"uses_sync_advance"}, [&]() {
+        beforeHook();
+        builder()->addMemberCall(state().self, "__on_0x25_sync_advance", {old_offset});
+        afterHook();
+    });
+
```
rsmmr, line 2596: Is this really the right location for this code? `advanceToNextData()` is called from multiple places and I don't think all them are indicating the beginning of sync mode. With this code here, I think we essentially just run the hooks twice each time, once before we actually move `cur` and once before.

bbannier: Yes, this is currently not the right place. Let me look again (while avoiding calling it at the same offset multiple times).

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -1224,6 +1222,14 @@ struct ProductionVisitor : public production::Visitor {
         else if ( const auto& unit = p->tryAs<production::Unit>() )
             unit_type = unit->unitType();
 
+        // Call any `%sync_advance` hook as we enter sync mode.
```
bbannier, line 1225: This is still only WIP. We still call the hook twice when entering for some cases, e.g., the setups in `spicy.types.unit.sync-advance` where we invoke it both when entering and at the same offset since we now jumped across a gap. I tried adding special handling to suppress the second case by inspecting `search_start`, but we do not set this variable (in the same scope) for all synchronization scenarios.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
bbannier: Resolving this for now, opened a new issue for calling the hook multiple times at the same offset.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
bbannier: Fixed.

rsmmr (approved):
Looks good now, just one suggestion: factor out ` pb->guardFeatureCode() ... { ... }` into a separate method to avoid duplicating the code 3x, for easier maintenance in the future.

bbannier:
> Looks good now, just one suggestion: factor out ` pb->guardFeatureCode() ... { ... }` into a separate method to avoid duplicating the code 3x, for easier maintenance in the future.

I was thinking about introducing that helper, but was unsure whether that would make things much easier since the hook invocations during advancing inject an additional condition. I added a helper, let me know what you think.

PR #1779, opened 2024-07-04
Extend validation of synchronization properties.

### LINKED ISSUE #1775: Compiler segfault when using an non-literal with `%synchronize-at`
The current validation around at least `%synchronize-at` seems to miss cases where the given expression is not a literal. Before 28437292fcf01cb1602e0318924cd7c78f5429fe this would run forever, now it segfaults in the compiler.

```ruby
module foo;

type A = unit {
    %synchronize-at = A;
    : /AA/;
};

public type X = unit {
    : (A &synchronize)[] ;
};
```

```console
$ spicyc -dj foo.spicy
Assertion failed: (ctor_), function syncProduction, file parser-builder.cc, line 1215.
[1]    44368 abort      ./build/bin/spicyc -d foo.spicy -j
```

rsmmr:
Closes #1775.


tests/spicy/types/unit/synchronize-at-fail.spicy
```diff
@@ -7,5 +7,25 @@ module foo;
...
+    %synchronize-after = 1; # error: cannot have both properties
+};
+
+@TEST-START-NEXT
+
+module foo;
+
+global x: bytes;
+
+%synchronize-at = x; # error: must be a ctor
```
bbannier, line 19: Can you also add instances missing an argument at global and unit scope?

```ruby
%synchronize-at; # error: requires an argument.
%synchronize-after; # error: requires an argument.
```

bbannier (approved):
Thanks for this. I think when I looked at this I missed that we had separate validators for module and unit scope properties, and didn't look at the right location.

rsmmr:
> separate validators for module and unit scope properties, and didn't look at the right location.

Yeah, not great currently with all that duplication.

PR #1781, opened 2024-07-08
Apply `&on-heap` selectively

rsmmr:
In our generated C++ parsing code, so far we generally moved all units
to the heap. This PR changes that to do so only for cases where we
actually need it (primarily: if we wrap the unit into a reference
somewhere, or if the unit type is cyclic). The assumption is that this
should improve both performance and memory usage.

As a side effect, the new machinery we built for this also allows us
to improve how we output C++ type declarations: We can now compute
dependencies between types and output them in the correct order
automatically, whereas before we had to rely on some hackish manual
prioritization.

TODO: I have not done any performance measurement yet.

- **Let optional's deref operator pass through "LHSideness".**
- **Prepare for moving parser temporaries to the stack.**
- **Add API to retrieve a type's dependent types.**
- **Rework emitting of C++-side type declarations.**
- **Apply `&on-heap` selectively only where needed.**
- **Change a `set` to an `unordered_set`.**


rsmmr:
Rebased on top of #1833, now using our new AST dependency tracking.

rsmmr:
With #1833 and #1834 merged, this is now rebased on `main`. 

Still not sure if we want this or not due to runtime performance concerns.

rsmmr:
I ported this to current `main`, looking better now, see updated PR description.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -89,24 +89,53 @@ struct ProductionVisitor : public production::Visitor {
...
+    // etc.), we wrap it into an unset optional so that we won't need to
+    // default instantiate it (which sometimes isn't  even possible, like a
+    // struct taking parameters). In that case, we push a deref operation to
+    // the destination stack for reading the value, so that usage remains
+    // transparent.
+    void pushTmpDestination(const std::string& prefix, QualifiedType* t) {
+        if ( t->type()->isNameType() ) {
+            auto* dst = builder()->addTmp(prefix, builder()->typeOptional(t));
+            pushDestination(builder()->deref(dst), dst);
+        }
```
bbannier: This wraps many more types than need it, and looking at the patch adding this wrapping in isolation adds 5% or more to the `UnitVectorSize` benchmarks. Is there any reason we wouldn't just want to do this for types with parameters? That seems to add only about 1%.


```suggestion
        if ( ! t->type()->parameters().empty() ) {
            auto* dst = builder()->addTmp(prefix, builder()->typeOptional(t));
            pushDestination(builder()->deref(dst), dst);
        }
```

tests/spicy/doc/my_http-host-parse2.cc
```diff
@@ -34,7 +34,7 @@ int main(int argc, char** argv) {
     // Access fields.
     std::cout << "method : " << *request->method << std::endl;
     std::cout << "uri    : " << *request->uri << std::endl;
-    std::cout << "version: " << *(*request->version)->number << std::endl;
+    std::cout << "version: " << *(request->version)->number << std::endl;
```
bbannier: The way we put parens here around the deref is still misleading since what we want to deref is `number` which holds an `Optional`. Suggest to drop the parens.

While at it, we should also change the `std::endl` to `\n` since our own clang-tidy config would reject such code.

```suggestion
    std::cout << "uri    : " << *request->uri << '\n';
    std::cout << "version: " << *(*request->version)->number << '\n'';
    std::cout << "version: " << *request->version->number << '\n'';
```

spicy/toolchain/src/compiler/codegen/codegen.cc
```diff
@@ -154,6 +178,25 @@ struct VisitorPass1 : public visitor::MutatingPostOrder {
...
+            // Add &on-heap attribute to types that are wrapped into an
+            // explicit, Spicy-level reference anywhere.
+            add_on_heap = true;
+
+        auto* unit_decl = u->typeDeclaration();
+        const auto& decl = context()->dependentDeclarations(unit_decl);
+        if ( auto x = std::ranges::find(decl, unit_decl); x != decl.end() )
+            // Add &on-heap to types that are recursively self-referencing.
+            // Without, we couldn't express the type at the C++ level.
+            add_on_heap = true;
```
bbannier: Nit: This reads much more complicated to me than it is ("true for either in `units_with_references` or in `decl`).

WDYT about this?

```suggestion
        auto* unit_decl = u->typeDeclaration();
        const auto& dependent_decls = context()->dependentDeclarations(unit_decl);

        bool add_on_heap =
            // Add &on-heap attribute to types that are wrapped into an
            // explicit, Spicy-level reference anywhere.
            info->units_with_references.contains(n->canonicalID()) ||

            // Add &on-heap to types that are recursively self-referencing.
            // Without, we couldn't express the type at the C++ level.
            dependent_decls.contains(unit_decl);
```

I also used a plural for `decl` since it refers to a set and went with `contains` instead of `find`.

spicy/toolchain/include/compiler/detail/codegen/codegen.h
```diff
@@ -32,6 +32,7 @@ class ParserBuilder;
 struct ASTInfo {
     std::set<ID> uses_sync_advance; // type ID of units implementing %sync_advance
     std::set<uint64_t> look_aheads_in_use;
+    std::unordered_set<ID> units_with_references; // IDs of all unit types that are wrapped into a reference somewhere
```
bbannier: Nit: Would use a `std::set` here as well for consistency, or change all members.

spicy/toolchain/src/compiler/codegen/codegen.cc
bbannier: Nit: This reads much more complicated to me than it is ("true for either in `units_with_references` or in `decl`).

WDYT about this?

```suggestion
        auto* unit_decl = u->typeDeclaration();
        const auto& dependent_decls = context()->dependentDeclarations(unit_decl);

        bool add_on_heap =
            // Add &on-heap attribute to types that are wrapped into an
            // explicit, Spicy-level reference anywhere.
            info->units_with_references.contains(n->canonicalID()) ||

            // Add &on-heap to types that are recursively self-referencing.
            // Without, we couldn't express the type at the C++ level.
            dependent_decls.contains(unit_decl);
```

I also used a plural for `decl` since it refers to a set, and used `contains` instead of a `find`.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
rsmmr: > we wouldn't just want to do this for types with parameters

That's literally what I had originally, it worked but just felt a bit odd to store some structs differently than others. However, I didn't realize there's such a potential performance impact. I'll switch it back.

PR #1786, opened 2024-07-09
Prevent defining already forwarded constants.

### LINKED ISSUE #1785: Importing a module causes emitting all the type infos for the module
Since about #1462 we seem to emit the full type infos for all types in the imported module in each file which imports the type, where previously we would only forward declare them. This causes us to emit much more code which can reduce compilation speed for code making (very) heavy use of modules and imports.

bbannier:
Closes #1785.

### DIFF
--- a/hilti/toolchain/include/base/cache.h
+++ b/hilti/toolchain/include/base/cache.h
@@ -4,7 +4,6 @@
 
 #include <map>
 #include <optional>
-#include <string>
 #include <utility>
 
 namespace hilti::util {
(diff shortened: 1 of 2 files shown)


bbannier:
I checked this as part of Zeek for compiling a giant internal analyzer which took 2x the time before this patch. With this PR compilation time is down to roughly the previous level again.

I still need to check whether the way the compiler cannot inline anymore (the linker could though) has any negative performance impact with Spicy-based Zeek analyzers.

rsmmr (approved):

rsmmr:
> I still need to check whether the way the compiler cannot inline anymore (the linker could though) has any negative performance impact with Spicy-based Zeek analyzers.

Forgot to comment on this: I think it's fine. I believe we create constant forwards only for TypeInfos, so that's the only place where things should change (and for them, inlining doesn't matter).

PR #1794, opened 2024-07-12
Update developer's guide

rsmmr:
Includes:

- Rework and update developer documentation's architecture diagram.
This is now generated through Python's `diagrams` package (i.e.,
graphviz). Because the layout is automatically determined by `dot`,
it's not perfect; but seems good enough.

- Update debugging streams.

- Remove benchmarking section, as this was completely outdated.


doc/scripts/autogen-architecture-diagram
```diff
@@ -0,0 +1,342 @@
...
+
+            optimize = Pass(
+                "Optimize AST", xlabel=tt_small("hilti::detail::optimizer::optimize()")
+            )
+            (
+                resolve_ast_spicy.output
+                >> validate_post
+                >> optimize
+                >> (output := validate_post_opt)
+            )
```
bbannier, line 263:
```suggestion
            resolve_ast_spicy.output >> validate_post >> optimize >> (output := validate_post_opt)
```

doc/scripts/autogen-architecture-diagram
bbannier, line 1: I personally find this much more complicated to read and follow than plain dot or plantuml with light styling. If you'd start from scratch I'd probably suggest inline dot via e.g., https://www.sphinx-doc.org/en/master/usage/extensions/graphviz.html (which would also properly track dependencies so the graph is updated when its source is updated without having to write explicit build system code) 🤷

doc/installation.rst
```diff
@@ -243,7 +243,7 @@ Make sure you have Xcode installed, including its command-line tools
 If you are using `Homebrew <https://brew.sh>`_::
 
     # brew install bison flex cmake ninja python@3.8 sphinx-doc
-    # pip3 install btest sphinx_rtd_theme
+    # pip3 install btest sphinx_rtd_theme diagrams
```
bbannier, line 246: Nit: Since I have seen people repeatedly destroy their (in particular macos) Python installation with `pip install` outside of venvs I wonder whether we shouldn't just not mention doc dependencies at all here and move from `pip3 install` to `pipx install` to suggest a managed venv for btest ([pipx](https://pipx.pypa.io/stable/) only manages executables).

.pre-commit-config.yaml
```diff
@@ -18,7 +18,7 @@ repos:
     language: python
     stages: ["commit"]
     types: [text]
-    exclude: '^tests|NOTES|README|autogen|^(CHANGES|VERSION|scripts/ninja-build-stats|\.gitmodules|\.paths|.*\.(decl|json|rst|t2d|txt))$'
+    exclude: '^tests|NOTES|README|autogen|^(CHANGES|VERSION|scripts/ninja-build-stats|\.gitmodules|\.paths|setup.cfg|.*\.(decl|json|rst|t2d|txt|dot))$'
```
bbannier, line 21: Adding `dot` seems not needed. Can you just add a license header to `setup.cfg`?

```suggestion
    exclude: '^tests|NOTES|README|autogen|^(CHANGES|VERSION|scripts/ninja-build-stats|\.gitmodules|\.paths|.*\.(decl|json|rst|t2d|txt))$'
```

setup.cfg
bbannier, line 1: I'd store these flags in `.pydocstyle` since this is not really a Python project.

bbannier (dismissed):

rsmmr:
I'll make a quick cleanup pass but honestly, I don't it's a good use of time to dig too much into the details here. The content is the diagram, not the Python script. :-)

bbannier:
> I'll make a quick cleanup pass but honestly, I don't it's a good use of time to dig too much into the details here. The content is the diagram, not the Python script. :-)

The new diagram is great, I reviewed the content 😉

doc/scripts/autogen-architecture-diagram
rsmmr: I admit it didn't quite turn out the way I was hoping to, but I diagram came together and I learned something. Still think it's better than plain dot, though. Either way, not going to redo it. :-)

doc/scripts/autogen-architecture-diagram
```diff
@@ -0,0 +1,342 @@
...
+output = sys.argv[1]
+
+with Diagram(
+    "",
+    curvestyle="ortho",
+    show=False,
+    direction="TB",
+    filename=output,
+    graph_attr={"forcelabels": "true", "fontname": "Helvetica"},
+    outformat=["pdf", "dot", "svg"],
```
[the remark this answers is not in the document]
rsmmr: Pretty sure I got this out of the docs, it works. PDF is easier for preview during development, dot helps debugging, so prefer to keep.

doc/installation.rst
rsmmr: Let me split it out at least.

setup.cfg
rsmmr: `.pydocstyle` isn't read by the LSP server I'm using. Open to other ideas where to put it.

bbannier: > `.pydocstyle` isn't read by the LSP server I'm using. Open to other ideas where to put it.

Not a big issue, feel free to just keep as is 🤷 

I use https://github.com/microsoft/pyright and ALE with vim, and neither of them seems to look into `setup.cfg` but e.g., `.pydocstyle` with contents like this seems to work:

```yaml
ignore = E261,E301,E302,E501,W503
```

doc/scripts/autogen-architecture-diagram
```diff
@@ -0,0 +1,342 @@
...
+
+import diagrams
+from diagrams import Diagram, Edge
+
+
+def tt(text, pt="14"):  # font size 14 is default
+    return f'<font face="Courier" point-size="{pt}">{text}</font>' if text else ""
+
+
+def tt_small(text, pt="8"):
```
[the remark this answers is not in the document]
rsmmr: What are you using to look at it? It's SVG, so DPI shouldn't matter. It's a large diagram but I can zoom in just fine to see everything clearly.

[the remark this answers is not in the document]
bbannier: In my browser this looks like this:

Screenshot 2024-07-15 at 6 06 05 PM

I can zoom this, but I basically loose all context that way.

> It's SVG, so DPI shouldn't matter.

I believe even for SVG dot would compute a physical canvas size (in e.g., cm), and then use a graph's [`dpi` attribute](https://graphviz.org/docs/attrs/dpi/) to translate sizes given in points. I would have expected increased DPI to overall bigger fonts.

Again, not a blocking issue.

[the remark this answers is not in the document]
rsmmr: I tried some different DPI values, but they don't make it better.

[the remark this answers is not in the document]
bbannier: Thanks for looking, let's just go with what we have.

Issue #1795, opened 2024-07-12
Segfault with array-switch-size-array-try construction


awelzel:
I've worked some more on the LDAP and managed to trigger a segfault related to chunk handling.

This is the minimal reproducer (using latest Zeek master, but also crashes within the `zeek/zeek:6.2.1` container image.

```
$ echo CJ2RXS7YIm4Lb+s5Y3nb0v2MVYgrmvRzZPBPQNTS4zE= | base64 -d | spicy-driver --increment 11 ./x.spicy
Segmentation fault (core dumped)
```

```
module Test;

type Inner = unit {
  len: uint8;
  magic: bytes &size=self.len % 3;
};

type TryInner = unit {
  : Inner &try { }
};

type Message = unit {
  len: uint8;

  switch ( True ) {
    True -> :TryInner[] &eod;
    * -> skip bytes &eod;
  } &size=self.len;
};

public type Messages = unit {
  : Message[];
};
```

Wasn't planning on using such a construct, but the extensions to the existing LDAP parser ended like so. I haven't tried to further reduce it.

Test data was produced as follows:
```
c=0; while [ $c -eq 0 ]; do dd if=/dev/urandom of=test.data count=1 bs=32; spicy-driver -f test.data  --increment 11 ./x.spicy ; if [ $? -gt 100 ] ; then echo "SIGNAL"; break; fi; done
```

---

Backtrace
```
#0  hilti::rt::stream::detail::Chunk::offset (this=0x0) at /home/awelzel/corelight-oss/zeek/auxil/spicy/hilti/runtime/include/hilti/rt/types/stream.h:185
#1  hilti::rt::stream::View::extract (this=this@entry=0x7ffff6ffca40, dst=dst@entry=0x7ffff6ffc977 "", n=n@entry=1) at /home/awelzel/corelight-oss/zeek/auxil/spicy/hilti/runtime/include/hilti/rt/types/stream.h:1471
#2  0x00007ffff7e6cfd5 in hilti::rt::integer::unpack<unsigned char, hilti::rt::stream::View> (b=..., fmt=fmt@entry=...) at /home/awelzel/corelight-oss/zeek/auxil/spicy/hilti/runtime/include/hilti/rt/types/integer.h:161
[...]

Issue #35, opened 2020-04-06
Add support for composing binary data


bbannier:
This is Spicy's support for rewriting binary data streams. This is a larger task & very very low priority. We won't tackle this before all the other functionality has been brought back.

rsmmr:
Update: It's not *that* low in priority anymore, but remains something to do until other things stabilize.

PR #1796, opened 2024-07-13
Remove undocumented Spicy parser engine variations

bbannier:
We had support in the code for parsing unit fields like `bytes &size=1` (note the missing `:` for an anon field). This was neither documented nor tested.

Since code might have used this without issue one could consider this a breaking change. At the same time it was never properly documented as supported.

I came across this when looking at the reproducer in #1795.


rsmmr (approved):
> We had support in the code for parsing unit fields like `bytes &size=1` (note the missing `:` for an anon field)

For the record, this goes with #35. The original plan here was to support either `<` or `>` as alternative to `:` for selecting direction. Not sure why a missing `:` was accepted, probably a bug. 

Agree that we can remove this for now, doesn't look like we'll be getting to #35 any time soon.

PR #1799, opened 2024-07-16
Fix illegal stream content access through invalid view.

### LINKED ISSUE #1795: Segfault with array-switch-size-array-try construction

rsmmr:
When a view's beginning ended up outside of the stream's valid range
(e.g., because the stream has been trimmed in the meantime), we
wouldn't catch that on access and attempt to access data that was no
longer valid. This commit extends the safety checks accordingly.

Closes #1795.


tests/spicy/types/unit/size-inside-size.spicy
```diff
@@ -0,0 +1,27 @@
+# @TEST-EXEC-FAIL: echo CJ2RXS7YIm4Lb+s5Y3nb0v2MVYgrmvRzZPBPQNTS4zE= | base64 -d | HILTI_DEBUG=spicy spicy-driver -d %INPUT >output 2>&1
+# @TEST-EXEC: btest-diff output
+#
+# @TEST-DOC: Regression test for #1795: an inside &size is outside an outside &size.
+
+module Test;
+
+type Inner = unit {
+  len: uint8;
+  magic: bytes &size=self.len % 3; # This will eventuallu exceed bytes made available by the outer layer.
```
bbannier, line 10:
```suggestion
  magic: bytes &size=self.len % 3; # This will eventually exceed bytes made available by the outer layer.
```

hilti/runtime/include/types/stream.h
```diff
@@ -1568,11 +1568,11 @@ class View final {
...
     void _ensureValid() const {
-        //        if ( ! _begin.chain() )
-        //            throw InvalidIterator("view has invalid beginning");
+        if ( ! _begin.isValid() )
+            throw InvalidIterator("view has invalid beginning");
 
-        //        if ( ! _begin.isValid() )
-        //            throw InvalidIterator("view has invalid beginning");
+        if ( (! _begin.isUnset()) && _begin.offset() < _begin.chain()->offset() )
+            throw InvalidIterator("view starts before available range");
```
bbannier, line 1575: I am not a fan of the error message since for anybody but direct users of the C++ API they might have trouble figuring out what change they need to make to address the issue.

I am wondering whether it would make more sense to instead push this check down into `View::extract` where we would have a chance to raise an exception similar to the one we emit in the parser builder if less than expected `&size` bytes could be extracted.

hilti/runtime/include/types/stream.h
```diff
@@ -1568,11 +1568,11 @@ class View final {
     }
 
     void _ensureValid() const {
-        //        if ( ! _begin.chain() )
-        //            throw InvalidIterator("view has invalid beginning");
+        if ( ! _begin.isValid() )
+            throw InvalidIterator("view has invalid beginning");
```
bbannier, line 1572: Could you add a C++ unit test case for this?

hilti/runtime/include/types/stream.h
rsmmr: Agree in principle, but not sure how to make it much better. Even though `extract` is what the ticket is about, I think the problem doesn't just apply there but also to other places requiring a valid start iterator, hence adding it to `_ensureValid`. And we already have a quite a few other exceptions thrown by the runtime that are probably similarly challenging from a grammar perspective, which is a result of how our layering.

bbannier (approved):

PR #1801, opened 2024-07-16
Trim inline code emitted more than once

### LINKED ISSUE #1800: C++ codegen emits definitions where declarations would have been sufficient
When importing other modules we define full definitions (i.e., function bodies) where declarations would have been sufficient. In fact these definitions are marked `inline` so we end up having to compile the same body again and again. We do this transitively for all imported modules, so the issue gets worse the longer the import chain and the more types become visible.

```ruby
# @TEST-EXEC-FAIL: TMPDIR=$PWD spicyc -dj x?.spicy -T # Fail to keep generated files around.

# @TEST-START-FILE x1.spicy
module x1;

public type X1 = unit {
    x1: uint8;
};
# @TEST-END-FILE

# @TEST-START-FILE x2.spicy
module x2;

import x1;

public type X2 = unit {
    x2: uint8;
};
# @TEST-END-FILE
```

The C++ files emitted for each module contain

- full constructors for every defined type, e.g., in `x2_*.cc`

  ```cpp
  namespace __hlt::x1 {
    inline X1::X1() {
    }

    inline X1::X1(std::optional<std::optional<::hilti::rt::integer::safe<uint8_t>>> x1) : X1() {
        if ( x1 ) this->x1 = std::move(*x1);
    }
  }
  ```

- full streaming operators for each type, e.g., in `x2_*.cc`

  ```cpp
  static auto __hlt::x1::operator<<(std::ostream& o, const __hlt::x1::X1& x) -> std::ostream& {
      o << ::hilti::rt::to_string(x);
      return o;
  }
  ```

- there are also a lot of declarations of types, but these might typically also be found in a header file and seem like a separate issue

It should be possible to drop these definitions for the declarations we already emit.

bbannier:
Closes #1800.

### DIFF
--- a/hilti/toolchain/src/compiler/codegen/codegen.cc
+++ b/hilti/toolchain/src/compiler/codegen/codegen.cc
@@ -279,6 +279,24 @@ struct Visitor : hilti::visitor::PreOrder {
 
         module = n->id();
         unit->addInitialization(cg->compile(n->statements()));
+
+        // Non-recursively add imported types.
+        for ( auto* import : n->childrenOfType<declaration::ImportedModule>() ) {
+            auto uid = import->uid();
+            assert(uid);
+
+            auto* m = cg->context()->astContext()->module(*uid);
+
+            // Do not import types builtin modules. We use `%skip-implementation` as a proxy.
+            if ( m->skipImplementation() )
+                continue;
+
+            for ( auto* decl : m->declarations() ) {
+                // Only import declarations defined directly in the imported module.
+                if ( auto t = decl->tryAs<declaration::Type>(); t && t->typeID().namespace_() == m->id() )
+                    n->add(cg->context()->astContext(), decl);
+            }
+        }
     }
 
     void operator()(declaration::ImportedModule* n) final {
--- a/hilti/toolchain/src/compiler/codegen/types.cc
+++ b/hilti/toolchain/src/compiler/codegen/types.cc
@@ -748,7 +748,7 @@ struct VisitorStorage : hilti::visitor::PreOrder {
             sid, [&]() { return CxxTypes{.base_type = std::string(sid)}; },
             [&](auto& cxx_types) {
                 auto render_body = cxx::Block();
-                render_body.addStatement("o << ::hilti::rt::to_string(x); return o");
+                render_body.addStatement("return o << ::hilti::rt::to_string(x);");
 
                 auto render_decl =
                     cxx::declaration::Function{.result = "std::ostream&",
--- a/hilti/toolchain/src/compiler/codegen/types.cc
+++ b/hilti/toolchain/src/compiler/codegen/types.cc
@@ -279,7 +279,12 @@ struct VisitorDeclaration : hilti::visitor::PreOrder {
                                            .type_name = cxx::ID(id.local()),
                                            .ctor = std::move(ctor),
                                            .add_ctors = true};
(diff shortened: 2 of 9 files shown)


hilti/toolchain/src/compiler/codegen/types.cc
```diff
@@ -274,12 +274,18 @@ struct VisitorDeclaration : hilti::visitor::PreOrder {
...
 
+
                 auto t = cxx::type::Struct{.args = std::move(args),
                                            .members = std::move(fields),
                                            .type_name = cxx::ID(id.local()),
                                            .ctor = std::move(ctor),
                                            .add_ctors = true};
-                return cxx::declaration::Type{id, t, t.inlineCode()};
+
+                // Only emit inline code if we are emitting the module the struct was declared in.
```
bbannier, line 284: This is insufficient and leads to us never emitting bodies.

hilti/toolchain/src/compiler/codegen/codegen.cc
```diff
@@ -279,6 +279,19 @@ struct Visitor : hilti::visitor::PreOrder {
...
+        // Add imported types.
+        for ( auto* import : n->childrenOfType<declaration::ImportedModule>() ) {
+            auto uid = import->uid();
+            assert(uid);
+
+            auto* m = cg->context()->astContext()->module(*uid);
+
+            for ( auto* decl : m->declarations() ) {
+                if ( decl->isA<declaration::Type>() )
+                    n->add(cg->context()->astContext(), decl);
```
bbannier, line 292: With the changes we now emit many more type infos into C++ code (we emit a type info for each type _declared_ in a module). A naive fix would seem to be to add a check for same module so we only emit type infos into C++ files for modules declaring the type, https://github.com/zeek/spicy/blob/4a1b43ef07d1305a7e88a4f0866068dc49de9d06/hilti/toolchain/src/compiler/codegen/types.cc#L1198

This fails however since with that we have no C++ file to emit type infos of builtin types from e.g., `spicy_rt` or `hilti` to. For this to work we would either add artificial units for builtin libraries, or maybe even better build them into `libspicy` and `libhilti`.

hilti/toolchain/src/compiler/codegen/codegen.cc
```diff
@@ -31,6 +36,18 @@ inline const DebugStream Compiler("compiler");
...
 namespace {
 
+/// Helper to check whether a give module looks like a builtin module.
+bool isBuiltin(const declaration::Module& m, std::string_view parseExtension, Context* context) {
+    const auto parse_plugin = plugin::registry().pluginForExtension(parseExtension);
+    assert(parse_plugin);
+
+    auto fileDir = rt::filesystem::path(m.meta().location().file()).parent_path();
+
+    auto paths = (*parse_plugin).get().library_paths(context);
```
bbannier, line 46: This still fails when running as part of a Zeek build where the library paths point to install locations, not the files we actually use during the compile.

bbannier:
This works now. I tested this with a big internal Zeek analyzer, and due to more granular dependency tracking we end up generating type infos in more files than before which causes us to emit more code. I suspect type info constants could be expensive to compile so we should ideally trim this down. This is tricky since we might e.g., instantiate a value ref over a custom type in a certain module only, so it is not immediately clear where we should emit these (additionally we do not emit C++ unit for builtin modules).

bbannier:
Now part of #1822.

PR #1802, opened 2024-07-17
Do not search for Python in CMake setup.

bbannier:
We previously needed this since during the build we were generating files with Python. This is not the case anymore.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,6 @@ project(spicy LANGUAGES C CXX)
 
 set(flex_minimum_version "2.5.37")
 set(bison_minimum_version "3.0")
-set(python_minimum_version "3.2")
 set(macos_minimum_version "19.0.0") # macOS 10.15.0 (Catalina)
 
 ## Initialize defaults & global options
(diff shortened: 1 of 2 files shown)


awelzel (commented):
Given the that the test suite and docs generation requires Python, a better alternative would be to put `(for testing and documentation)` rather than completely excluding it? The `pip` commands following make this now implicit.

bbannier:
> Given the that the test suite and docs generation requires Python, a better alternative would be to put `(for testing and documentation)` rather than completely excluding it? The `pip` commands following make this now implicit.
> 
> EDIT: There's some more mentioning of python in the macOS section.

Yeah, I am wondering if I shouldn't just leave the docs as is since this is becoming pretty tricky: the only direct requirement for `python3` would be from pre-commit hooks, but to install them one would either need to install `pre-commit` with pip, or if using a system package it would pull in Python as a dependency. The situation is similar for building the docs which just needs Sphinx (which only requires Python indirectly), and BTest (probably impossible to install without `pip` which require Python).

Let me revert the docs change.

awelzel (approved):

PR #1809, opened 2024-07-26
Fix non-converging optimizer pass for used functions.

### LINKED ISSUE #1808: Compilation gets stuck if field named like type
Since b1f1d64 compilation gets stuck if a unit field has a name like an existing type, e.g.,

```ruby
module test;

type X = unit {
    # Field with name identical to type name below.
    foo: uint8;
};

type foo = unit {};
```

The following never terminates:
```console
$ spicyc -dj test.spicy
````

This is a blocker.

### PR DESCRIPTION
**Pull request by bbannier**
The optimizer's `FunctionVisitor` computes whether a function is used or required by any feature. Since recently we now hold all modules in a single AST. This means that we need at least one full pass through the AST to collect feature constants so we can decided whether a function is required by a feature and cannot be elided. The initial implementation of this used explicit control flow, but this is error-prone like seen in #1808.

This patch removes the explicit control flow and instead checks whether we have discovered any new feature constants. With that the optimizer pass converges again.

Closes #1808.

### DIFF
--- a/hilti/toolchain/src/compiler/optimizer.cc
+++ b/hilti/toolchain/src/compiler/optimizer.cc
@@ -2,6 +2,7 @@
 
 #include "hilti/compiler/detail/optimizer.h"
 
+#include <numeric>
 #include <optional>
 #include <string>
 #include <tuple>
@@ -100,8 +101,6 @@ struct FunctionVisitor : OptimizerVisitor {
         bool referenced = false;
     };
 
-    bool _collect_again = false;
-
     // Lookup table for feature name -> required.
     using Features = std::map<std::string, bool>;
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: hilti/toolchain/src/compiler/optimizer.cc [hunk 1]
```diff
@@ -174,6 +185,7 @@ struct FunctionVisitor : OptimizerVisitor {
 
         switch ( _stage ) {
             case Stage::COLLECT: {
+                // std::cerr << "NOPE COLLECT FIELD " << n->canonicalID() << '\n';
```
awelzel, line 188: stray? also below.

awelzel (dismissed):

hilti/toolchain/src/compiler/optimizer.cc
bbannier: Thanks, I missed them; they are debugging statements which shouldn't be committed (I used a identifiable prefix to catch them more easily but alas 🤷)

PR #1811, opened 2024-07-29
Fix nested look-ahead switches.

### LINKED ISSUE #1810: Unit `switch` with look-ahead cannot be nested
When branching on a value unit `switch` can be nested, e.g., the following works:

```ruby
module foo;

public type X = unit(typ: uint8) {
    switch (typ) {
        1 -> switch (typ) {
            1 -> a: b"a";
            2 -> b: b"b";
        };
        3 -> c: b"c";
    };
};
```

While this would seem contrived it allows collecting related information together which reduces the need to store values if other parsing depends on this.

The same pattern does not seem to work with unit `switch` with look-ahead, e.g.,

```ruby
module foo;

public type X = unit(typ: uint8) {
    switch {
        -> switch {
            -> a: b"a";
            -> b: b"b";
        };
        -> c: b"c";
    };

};
```

```console
$ spicyc -dj foo.spicy
[error] foo.spicy:5:12-5:17: syntax error, unexpected SWITCH
[error] spicyc: parse error
```

rsmmr:
Closes #1810.


bbannier (dismissed):
Awesome 🚀

PR #1816, opened 2024-08-01
Disallow expanding limited Views again with `limit`

### LINKED ISSUE #1815: &max-size will overread and only raise error afterwards
The `&max-size` attribute will happily allow inner units to over-read the data from the input stream till they are done, including raising hooks (or zeek events) - only raising an error after all parsing is finished.

Example:

```
module Test;

public type Top = unit {
  innerdata: Inner &max-size=2;
};

type Inner = unit {
   data: bytes &size=20;
   on %done { print self; }
};
```

```
$ printf '\x0b\x00\x00\x01\x00\x0e\xec\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00' | spicy-driver test2.spicy
[$data=b"\x0b\x00\x00\x01\x00\x0e\xec\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"]
[error] terminating with uncaught exception of type spicy::rt::ParseError: parsing not done within &max-size bytes (test2.spicy:4:20-4:30)
```

This obviously is a problem, especially in cases when one parses a protocol that specifies unit-lengths itself. 

To give a real-world example: This bug was found while examining a test-failure fo the TLS testsuite. In that case, an inner field specifies a much-too-large-length, exceeding the size of the outer fields by several kilobytes. As this overread seems to exceed the amount of data remaining in this connection, no error is ever raised.

bbannier:
The documented semantics of `View::limit` are that it creates a new view
with equal or smaller size. In contrast to that we would still have
allowed to expand views with more calls `limit` again as well.

This patch changes the implementation of `View::limit` so it can only
ever make a `View` smaller.

Closes #1815.


spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -842,7 +842,7 @@ struct ProductionVisitor : public production::Visitor {
             ncur_max_size = state().cur;
         }
 
-        else if ( auto a = field->attributes()->find("&size") )
+        else if ( auto a = field->attributes()->find("&size"); a && ! field->attributes()->find("&eod") )
```
rsmmr, line 845: I'm unclear why this (and below) is needed for the problem being fixed here. Mind extending the commit message a bit to explain?

tests/spicy/types/unit/size-inside-size.spicy
```diff
@@ -1,27 +1,14 @@
-# @TEST-EXEC-FAIL: echo CJ2RXS7YIm4Lb+s5Y3nb0v2MVYgrmvRzZPBPQNTS4zE= | base64 -d | HILTI_DEBUG=spicy spicy-driver -d %INPUT >output 2>&1
+# @TEST-EXEC-FAIL: echo XXXX | spicy-driver -d %INPUT >output 2>&1
 # @TEST-EXEC: btest-diff output
 #
 # @TEST-DOC: Regression test for #1795: an inside &size is outside an outside &size.
```
rsmmr, line 4: Is the simplified test still a regression test for #1795?

Personally, I wouldn't mind keeping the more complex test; exercising more of our machinery rarely hurts.

rsmmr (approved):

spicy/toolchain/src/compiler/codegen/parser-builder.cc
bbannier: Added the following to the commit message:

> We also tweak the implementation of the check for consumed `&size` when used together with `&eod`: if the `&size` was already nested in a limited view a larger `&size` value could previously extend the view so the `&eod` effectively was ignored. Since we now do not extend the `View` anymore we need to only activate the check for consumed `&size` if `&eod` was not specified since in this case the user communicated that they are fine with consuming less data.

The reason this worked correctly if a field with `&size` & `eod` was not nested is due to how we update `ncur`.

tests/spicy/types/unit/size-inside-size.spicy
bbannier: > Is the simplified test still a regression test for https://github.com/zeek/spicy/issues/1795?

Yes it does and still catches the incorrect behavior, even before the fixes to `View::limit`. After the fixes we do not trigger the exact error case anymore though since we change how the `View` is modified (more correctly I'd argue).

> Personally, I wouldn't mind keeping the more complex test; exercising more of our machinery rarely hurts.

While looking into why the baseline of this test changed figuring exactly what we parsed wasn't really clear to me (both due to the need to deobfuscate the input with `base64`, and also due to the changing lengths via `&size=self.len % 3`). I do not think there is anything fundamentally interesting going on here behind this complexity. If you want I can still role back the change.

rsmmr: It's ok.

rsmmr (approved):

bbannier:

PR #1821, opened 2024-08-05
to_int() ubsan fixes

awelzel:
* Reject empty byte ranges for `to_uint(ByteOrder)`.
* Fix shifting of 32bit value, causing `to_int(ByteOrder)` to return bogus values.

### DIFF
--- a/hilti/runtime/src/tests/bytes.cc
+++ b/hilti/runtime/src/tests/bytes.cc
@@ -375,6 +375,17 @@ TEST_CASE("toInt") {
         CHECK_EQ("100"_b.toInt(Enum(ByteOrder::Network)), 3223600);
         CHECK_EQ("100"_b.toInt(Enum(ByteOrder::Little)), 3158065);
 
+        CHECK_EQ("\x00\x00\x00\x01\x01"_b.toInt(Enum(ByteOrder::Big)), 257);
+        CHECK_EQ("\xff"_b.toInt(Enum(ByteOrder::Big)), -1);
+        CHECK_EQ("\xff\xff"_b.toInt(Enum(ByteOrder::Big)), -1);
+        CHECK_EQ("\xff\xff\xff\xff"_b.toInt(Enum(ByteOrder::Big)), -1);
+        CHECK_EQ("\xff\xff\xff\xff\xff\xff"_b.toInt(Enum(ByteOrder::Big)), -1);
+        CHECK_EQ("\xff\xff\xff\xff\xff\xff\xff\xff"_b.toInt(Enum(ByteOrder::Big)), -1);
+
+        // 2er complement according to Wikipedia: -(2**39) + 2**8 + 2**0 = -549755813631
+        CHECK_EQ("\x80\x00\x00\x01\x01"_b.toInt(Enum(ByteOrder::Big)), -549755813631);
+        CHECK_EQ("\x01\x01\x00\x00\x80"_b.toInt(Enum(ByteOrder::Little)), -549755813631);
+
         if ( systemByteOrder().value() == ByteOrder::Little )
             CHECK_EQ("100"_b.toInt(Enum(ByteOrder::Host)), 3158065);
         else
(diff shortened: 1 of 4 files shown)


hilti/runtime/src/tests/bytes.cc
```diff
@@ -391,8 +406,9 @@ TEST_CASE("toUInt") {
     SUBCASE("with base") {
         CHECK_EQ("100"_b.toUInt(), 100U);
         CHECK_EQ("100"_b.toUInt(2), 4U);
-        CHECK_THROWS_WITH_AS("-100"_b.toUInt(2), "integer overflow", const RuntimeError&);
 
+        CHECK_THROWS_WITH_AS(""_b.toInt(16), "cannot decode from empty range", const RuntimeError&);
```
bbannier, line 410: Duplicated from `toInt` test.

```suggestion
```

hilti/runtime/src/types/bytes.cc
```diff
@@ -193,6 +193,9 @@ uint64_t Bytes::toUInt(ByteOrder byte_order) const {
         case ByteOrder::Big: break;
     }
 
+    if ( isEmpty() )
+        throw RuntimeError("not enough bytes for conversion to integer");
```
bbannier, line 197: Let's throw the same error we throw for `atoi_n` unless there's a good reason to deviate.

```suggestion
        throw InvalidArgument("cannot decode from empty range");
```

hilti/runtime/src/types/bytes.cc
```diff
@@ -171,14 +171,14 @@ integer::safe<uint64_t> Bytes::toUInt(uint64_t base) const {
 }
 
 int64_t Bytes::toInt(ByteOrder byte_order) const {
-    auto i = toUInt(byte_order);
+    auto i = toUInt(byte_order); // throws on size == 0 or size > 8
     auto size_ = static_cast<uint64_t>(size());
 
-    if ( i & (1U << (size_ * 8 - 1)) ) {
+    if ( i & (1ULL << (size_ * 8 - 1)) ) {
```
bbannier, line 177: This should work now, but I still find the mixing of types with explicit (`uint64_t` and minimum sizes (`unsigned long long`) confusing. WDYT about about using sized types throughout, at least were we care, e.g.,

```suggestion
    if ( i & (UINT64_C(1) << (size_ * 8 - 1)) ) { // requires `#include <cstdint>`
```

bbannier (changes requested):
Throwing for empty ranges makes sense to me, but I think we should really call it out in the docs (we didn't even mention this for the already throwing operators). Could you do that? They are defined in `hilti/toolchain/src/ast/operators/bytes.cc`.

hilti/runtime/src/types/bytes.cc
awelzel: This was aligning to the RuntimeError just below (which seems in the same category). I'm also not quite sure I like the "decode" for interpreting 4 bytes as int (rather than decoding a ascii string to an integer).

If you confirm you prefer both, I'll change.

awelzel: > InvalidArgument

This one is documented as "Exception flagging invalid arguments passed to a function." - so it doesn't quite fit. I could see `InvalidValue` being more suitable, but would want to replace the below `throw` then, too.

awelzel:
> Could you do that? They are defined in hilti/toolchain/src/ast/operators/bytes.cc.

I've done that now. Still not sure what exactly to use as exceptions, but RuntimeError is at least not totally wrong in the docs...

hilti/runtime/src/types/bytes.cc
rsmmr: `InvalidValue` would sound good to me for both (and then actually also `InvalidArgumet` for the current `RuntimeError` thrown at the beginning for `Undef`).

That said, this all isn't clear cut because we aren't very consistent across the runtime library in what exactly to throw when. The important part is that both `InvalidArgument` and `InvalidValue` do derive from `RuntimeError`, so we get that  either way. `RuntimeError` is what Spicy's error handling keys on, so that's important. The specific derived class then just adds a bit of semantic value that however doesn't make a functional difference.

rsmmr (approved):

bbannier:
Backported to `release/1.11`.

PR #1822, opened 2024-08-05
Rework emission of C++ code.

rsmmr:
This reworks how we emit C++ code, in particular declarations. We now
determine what to emit (and when) through dependencies: at the AST
level, we compute how global declarations depend on each other. Then,
at the C++ level, we emit dependencies as needed, and no longer simply
bulk copy *everything* imported over into a unit. In some cases, this
cuts down emitted code by a lot (though not necessarily compile time
it turns out).

The main commits in this PR are `Add an AST pass collecting
dependencies between declarations.` and `Rework emission of C++
code.`. The latter is unfortunately a bit messy, mostly because our
C++ code emission is messy to begin with. However, overall, the change
simplifies things quite a bit, and gets rid of a number of low-level
pieces no longer needed. It might still be challenging to review
unfortunately.

This PR does also ports #1801 over. Some commits needed a bit of
tweaking to match the new code.

I checked analyzer compile time performance with these changes and
didn't see much of a difference either way. In fact, I saw no change
at all without the #1801 port. With #1801 included, I did see a small
increase in compile time for a large internal analyzer; but nothing to
worry about.

- **Fully qualify runtime identifiers in generated code.**
- **Fix `Node::tryAs_()`.**
- **Add an AST pass collecting dependencies between declarations.**
- **Remove usage of designated initializers for all `cxx::declaration::*` classes.**
- **Rework emission of C++ code.**
- **Remove unused methods to retrieve module dependencies.**
- **Simplify body of generated `operator<<`.**
- **Only emit full C++ body of functions in module of matching namespace.**
- **Tweak when we emit struct inline code.**
- **Rename `cxx::type::Struct`'s `inlineCode` to `code`.**
- **Properly set parse extension when importing by path.**

### DIFF
--- a/doc/autogen/types/bytes.rst
+++ b/doc/autogen/types/bytes.rst
@@ -90,11 +90,15 @@
     Interprets the data as representing an ASCII-encoded number and
     converts that into a signed integer, using a base of *base*. *base*
     must be between 2 and 36. If *base* is not given, the default is 10.
+    If the conversion fails, throws a `RuntimeError` exception, this
+    includes calling `to_int()` on empty ``bytes``.
 
 .. spicy:method:: bytes::to_int bytes to_int False int<64> (byte_order: spicy::ByteOrder)
 
     Interprets the ``bytes`` as representing an binary number encoded with
-    the given byte order, and converts it into signed integer.
+    the given byte order, and converts it into signed integer. If the
+    conversion fails, throws a `RuntimeError` exception, this can happen
+    when ``bytes`` is empty or its size is larger than 8 bytes.
 
 .. spicy:method:: bytes::to_real bytes to_real False real ()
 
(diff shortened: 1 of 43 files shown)


bbannier (approved):
FTR, I see big improvements compiling individual files for a bigger internal Zeek analyzer with this patch. This is primarily due to files Zeek's `spicyz` generates for hooks being much smaller now leading to >4x faster compile times with clang and >7x for GCC. Single big modules still have roughly the same size, and still compile "slowly", so with sufficient parallelism they limit the overall throughput; that most files now compile faster however should improve system load and RAM usage substantially (haven't measured exactly).

I left only minor comments around cleaning up some of the operator impls for cxx types.

rsmmr:
> so with sufficient parallelism they limit the overall throughput

Oh, nice observation, I missed that when I looked at runtimes.

PR #1825, opened 2024-08-05
Don't qualify magic linker symbols with C++ namespace.

### LINKED ISSUE #1823: spicyc -x is accidentally namespacing spicy version information incorrectly.
If I run `spicyc -x foo foo.spicy`, it generates `foo___linker__.cc` with the following two lines in it....

```
const char HILTI_EXPORT HILTI_WEAK * __hlt_foo_hlto_library_version = R"({"debug":false,"hilti_version":11100,"magic":"v1"})";
const char HILTI_EXPORT HILTI_WEAK * __hlt_foo_hlto_bind_to_version = spicy_version_1_11_0();
```

Those lines need to be generated without the `_foo` so that the resulting .hlto file will load correctly

rsmmr:
We need them at their original values because that's what the runtime
lbirary is hard-coded to expect.

Closes #1823.

### DIFF
--- a/tests/hilti/codegen/library-version.hlt
+++ b/tests/hilti/codegen/library-version.hlt
@@ -1,0 +1,11 @@
+# @TEST-EXEC: hiltic -x foo %INPUT
+# @TEST-EXEC: grep -q _hlt_hlto_library_version foo___linker__.cc
+# @TEST-EXEC: grep -q _hlt_hlto_bind_to_version foo___linker__.cc
+#
+# A hand-built HLTO file has all symbols needed to load correctly.
+# @TEST-EXEC: $(spicy-config --cxx --cxxflags-hlto --ldflags-hlto --debug) -o foo.hlto foo*cc
+# @TEST-EXEC: hiltic -j foo.hlto
+#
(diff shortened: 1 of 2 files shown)


tests/hilti/codegen/library-version.hlt
```diff
@@ -0,0 +1,7 @@
+# @TEST-EXEC: hiltic -x foo %INPUT
+# @TEST-EXEC: grep -q _hlt_hlto_library_version foo___linker__.cc
+# @TEST-EXEC: grep -q _hlt_hlto_bind_to_version foo___linker__.cc
```
bbannier, line 3: The original issue was that hand-built HLTO files would not load due to us rejecting them because of missing version info, maybe add that to the test as well since we didn't have an end-to-end test for that.

```suggestion
# @TEST-EXEC: hiltic -x foo %INPUT
# @TEST-EXEC: grep -q _hlt_hlto_library_version foo___linker__.cc
# @TEST-EXEC: grep -q _hlt_hlto_bind_to_version foo___linker__.cc
#
# A hand-built HLTO file has all symbols needed to load correctly.
# @TEST-EXEC: $(spicy-config --cxx --cxxflags-hlto --ldflags-hlto --debug) -o foo.hlto foo*cc
# @TEST-EXEC: hiltic -j foo.hlto
```

bbannier (approved):

bbannier:

PR #1826, opened 2024-08-06
Fix --enable-werror werror_flags.

awelzel:
Found while trying to test something with EXTRA_CXX_FLAGS. This got lost in e27cd5ac0443659e7e9fdae2f3bbe5920bd3564b.

### DIFF
--- a/hilti/runtime/include/types/integer.h
+++ b/hilti/runtime/include/types/integer.h
@@ -99,7 +99,8 @@ inline Bytes pack(integer::safe<T> i, ByteOrder fmt) {
     if ( fmt == ByteOrder::Host )
         return pack<T>(i, systemByteOrder());
 
-    uint8_t raw[sizeof(T)];
+    Bytes b(sizeof(T), '\0');
+    auto* raw = reinterpret_cast<uint8_t*>(b.data());
 
     switch ( fmt.value() ) {
         case ByteOrder::Big:
@@ -146,7 +147,7 @@ inline Bytes pack(integer::safe<T> i, ByteOrder fmt) {
         case ByteOrder::Undef: throw RuntimeError("attempt to pack value with undefined byte order");
     }
 
-    return Bytes(reinterpret_cast<Bytes::Base::value_type*>(raw), sizeof(raw));
+    return b;
 }
 
 template<typename T, typename D>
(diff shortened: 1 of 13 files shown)


awelzel:
@rsmmr / @bbannier - I'm not sure I'm capable to efficiently fix up the warnings. Mind taking a peak and/or possibly label this so it's running all the platforms?

bbannier:
> @rsmmr / @bbannier - I'm not sure I'm capable to efficiently fix up the warnings. Mind taking a peak and/or possibly label this so it's running all the platforms?

In contrast to the Zeek build this repo simply skips the jobs by default instead of disabling them based on labels. You can simply navigate to the skipped job in Cirrus and trigger it explicitly.

hilti/runtime/include/types/integer.h
```diff
@@ -99,7 +99,9 @@ inline Bytes pack(integer::safe<T> i, ByteOrder fmt) {
     if ( fmt == ByteOrder::Host )
         return pack<T>(i, systemByteOrder());
 
-    uint8_t raw[sizeof(T)];
+    std::string str;
+    str.resize(sizeof(T));
```
bbannier, line 103: You could allocate a string of the right size on declaration to avoid a potential realloc, and since we want `Bytes` which are just a small wrapper around `std::string` you could directly construct the `Bytes` we return (should remove the `std::move` on `return`),

```suggestion
    Bytes str(sizeof(T), '\0');
```

hilti/runtime/include/types/integer.h
```diff
@@ -146,7 +148,7 @@ inline Bytes pack(integer::safe<T> i, ByteOrder fmt) {
         case ByteOrder::Undef: throw RuntimeError("attempt to pack value with undefined byte order");
     }
 
-    return Bytes(reinterpret_cast<Bytes::Base::value_type*>(raw), sizeof(raw));
+    return Bytes(std::move(str));
```
bbannier, line 151: If you construct a `Bytes`:

```suggestion
    return str;
```

hilti/runtime/include/types/integer.h
awelzel: Nice, folded that in.

bbannier (dismissed):

Issue #1807, opened 2024-07-23
Convenience functions for string type


awelzel:
The `bytes` type has a few helpers like which `string` is missing.

* `starts_with()`
* `find()`
* `split()`
* `lower()` and `upper()`

I had a use for `starts_with()` and retreated to encoding the string to get access to `starts_with()`.

(The parser in question maybe should've used `bytes` instead of `string` throughout, but seems having these helpers may still be reasonable).

evantypanski:
I think the linked PR adds enough, namely these helpers were added:

- `starts_with`
- `split`
- `split1`
- `lower`
- `upper`

And not `find` because that's at least less trivial than porting over helpers that `bytes` already has (due to the `Iterator`). Maybe a new issue or this can be reopened if that's necessary

PR #1828, opened 2024-08-06
Add some easy string helpers

evantypanski:
This may or may not be enough for #1807 - but I'll let others decide that

This just adds a few easy methods for `string` that already had implementations for `bytes` (or, even, within the `string` namespace). Those are:

- `starts_with`
- `split`
- `split1`
- `lower`
- `upper`

There are a couple of points worth noting:

1) The implementation for `split1` is a little inconsistent (and `split`, but there are notes for that in code). I didn't want to change behavior for already defined functions here, so I just updated the docs to be more clear in my opinion. The main point is calling `split1` with an empty string `""` will immediately split, then put the whole string in the second value of the tuple. That may be entirely expected, but it's a little unintuitive at first
2) This doesn't implement `find` which was called out in the issue. That's just because it relies on the `Bytes` iterator and I wanted something in with the easy ones first. It may be just as easy, but this is enough for an initial bit :)

So from here we could just go down the list of `bytes` operators and try to bring them all into `string` - but would that be worthwhile? Is this enough? I have no idea because I don't know the relative utility


hilti/toolchain/src/ast/operators/bytes.cc
```diff
@@ -590,9 +590,11 @@ class Split1 : public BuiltInMemberCall {
...
 as a 2-tuple, with the separator removed. If the separator is not found, the
-returned tuple will have the whole bytes value as its first element and an empty value
-as its second element. If the separator is not given, or empty, the split will
-take place at the first sequence of white spaces.
+returned tuple will have the whole bytes value as its first element and an empty
+value as its second element. If the separator is empty, the returned tuple will
+have an empty first element and the whole bytes value as its second element. If
+the separator is not provided, the split will occur at the first sequence of
+white spaces.
 )",
```
bbannier, line 598: Ditto.

hilti/toolchain/src/compiler/codegen/operators.cc
```diff
@@ -815,6 +815,10 @@ struct Visitor : hilti::visitor::PreOrder {
         result = fmt("::hilti::rt::string::startsWith(%s, %s)", self, args[0]);
     }
 
+    void operator()(operator_::string::LowerCase* n) final { result = fmt("::hilti::rt::string::lower(%s)", op0(n)); }
```
bbannier, line 818: Tagentially related tech debt: would you mind adding a patch which changes the runtime impl of both `lower` and `upper` so instead of taking `const std::string&` they take `std::string_view`?

This should have no effect for user-provided values which are always owning strings, but could help should we use these functions with internal strings which are something non-owning.

bbannier (changes requested):
Thanks for the patch, this looks great. I noted down a few possible suggestions for streamlining.

hilti/toolchain/src/ast/operators/string.cc
```diff
@@ -126,6 +126,68 @@ class Encode : public BuiltInMemberCall {
...
+        return Signature{
+            .kind = Kind::MemberCall,
+            .self = {parameter::Kind::In, builder->typeString()},
+            .member = "split",
+            .param0 =
+                {
+                    .name = "sep",
+                    .type = {parameter::Kind::In, builder->typeString()},
+                    .optional = true,
+                },
```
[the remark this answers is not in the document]
evantypanski: I think it's reasonable to just use a default, but one useful part that would get lost with this suggestion is that the no-argument `split()` will use any whitespace as a separator:

https://github.com/zeek/spicy/blob/3a50444e255e07cb47238eae88efa05f8c969f14/hilti/runtime/src/util.cc#L135

https://github.com/zeek/spicy/blob/3a50444e255e07cb47238eae88efa05f8c969f14/hilti/runtime/include/util.h#L159

Whereas `split(" ")` would just split on spaces, and `split(detail::whitespace_chars)` would not split on the individual whitespaces:

https://github.com/zeek/spicy/blob/3a50444e255e07cb47238eae88efa05f8c969f14/hilti/runtime/src/util.cc#L115

Instead, how about `split` called with an empty string (`""`) `delim` just switches its behavior to the old `split()` behavior (since I don't think the empty string argument would be useful anyway) and then the default is `.default_ = builder->stringLiteral("")` (not `" "`)?

Then that would also change for `bytes` and fix all of the documentation confusion I pointed out, from what I can tell, so long as the empty string case is documented properly

Not too sure if that's what you intended, but the code suggestion has `" "` as the default so just making sure that seems reasonable

hilti/toolchain/src/compiler/codegen/operators.cc
evantypanski: scooped this one up into 7b748f914ef8b853dfc83e8ae830d030ad34f577

evantypanski:
I went through everything, but I kept the default param stuff unresolved since I'm not completely convinced I took the right path there, let me know what you think. Enough should be in the [commit message](https://github.com/zeek/spicy/pull/1828/commits/8213d05f529c8085a65961f77707d51a1f9fca86) and [initial comment](https://github.com/zeek/spicy/pull/1828#discussion_r1707132609), tl;dr is that I made the empty string default to the no-separator-provided behavior, I think that fixes the edge case of an empty string and simplifies

hilti/toolchain/src/ast/operators/string.cc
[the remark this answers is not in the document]
bbannier: Yeah, you got me convinced :)

bbannier (approved):
Thanks, this looks great! 🚀

PR #1834, opened 2024-08-09
Fix optimizer regression.

### LINKED ISSUE #1831: Optimizer leaves incorrect code
Something's off with the optimizer in 1.11:

```
# spicyc  -p tests/spicy/types/unit/underscores-in-ids.spicy
module Test {

import hilti;
import spicy_rt;

type Foo = struct {
    uint<32> x &optional;
    vector<Foo_> y &optional;
    optional<hilti::RecoverableFailure> __error &always-emit &internal;
    hook void __on_x(uint<32> __dd);
} &on-heap;

[... some Foo functions, but no further types ...]
}
```

It removes all the other `Foo_*` types, but leaves `Foo` in, and then `Foo` still refers to the now removed `Foo_`.

This is a regression, it works with 1.10, which outputs:

```
type Foo = struct {
    uint<32> x &optional;
    vector<Foo_> y &optional;
    optional<hilti::RecoverableFailure> __error &always-emit &internal;
    hook void __on_x(uint<32> __dd);
} &on-heap;
type Foo_ = struct {
    value_ref<Foo__> v &optional;
    optional<hilti::RecoverableFailure> __error &always-emit &internal;
} &on-heap;
type Foo__ = struct {
    value_ref<Foo___> v &optional;
    optional<hilti::RecoverableFailure> __error &always-emit &internal;
} &on-heap;
type Foo___ = struct {
    uint<8> v &optional;
    optional<hilti::RecoverableFailure> __error &always-emit &internal;
} &on-heap;
```
 I.e., it leaves the other types in, which at least lets the code remain valid.

rsmmr:
We were no longer marking types as used that are referenced through a
type name.

Closes #1831.

### DIFF
--- a/hilti/toolchain/src/compiler/optimizer.cc
+++ b/hilti/toolchain/src/compiler/optimizer.cc
@@ -583,6 +583,26 @@ struct TypeVisitor : OptimizerVisitor {
         }
     }
 
+    void operator()(type::Name* n) final {
+        auto t = n->resolvedType();
+        assert(t);
+
+        switch ( _stage ) {
+            case Stage::COLLECT: {
+                if ( const auto& type_id = t->typeID() )
+                    // Record this type as used.
+                    _used[type_id] = true;
+
+                break;
+            }
+
+            case Stage::PRUNE_USES:
+            case Stage::PRUNE_DECLS:
+                // Nothing.
+                break;
+        }
+    }
+
     void operator()(UnqualifiedType* n) final {
         if ( n->parent(2)->isA<declaration::Type>() )
             return;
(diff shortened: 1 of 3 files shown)


hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -583,6 +583,26 @@ struct TypeVisitor : OptimizerVisitor {
         }
     }
 
+    void operator()(type::Name* n) final {
+        auto t = n->resolvedType();
+        assert(t);
```
bbannier, line 588: Nit: I wonder whether we actually _need_ the `assert` here and couldn't instead return early like in many other places in this file. An unresolved `type::Name` shouldn't be "used" if we never needed to resolve it; it's likely some internal inconsistency, but we do not need to abort for it here.

bbannier (approved):

hilti/toolchain/src/compiler/optimizer.cc
rsmmr: At this point a name expression should never be left unresolved, the validator would have caught that and aborted. I was debating if we need the check at all, just felt better to have the assert just in case.

Issue #1824, opened 2024-08-05
Attach %error handlers to unit fields


sethhall:
I would like to be able to attach %error handler blocks to fields within a unit instead of only the entire unit.  I have a use case for this in a kerberos parser where I'm using `self.backtrack()` to deal with optional fields but I only want to call backtrack on particular field failures but there are other places that the unit could fail that I would like to have "normal" failures.

I would like to be able to write this code...

```
type Int32 = unit(tagged: bool=False, itag: uint8=255) {
   : ASN1::ExplicitTagHeader(itag) if (tagged) %error {
       self.backtrack();
   };

   : ASN1::Tag;
   len: ASN1::Length;
   value: bytes &size=self.len
                &convert=$$.to_int(spicy::ByteOrder::Network)
                &requires=(self.value >= -2147483648 && self.value <= 2147483647);
} &convert=self.value;
```

You can see how the &requires attribute can cause a failure in the unit but in that case I don't want to backtrack.  I only want the backtrack behavior if that first field fails.

rsmmr:
I like this, and I think it should be relatively easy to implement. We can track `%error` like we currently do with `%debug`: as a property on a hook that we query during parser generation. Then, when emitting a field that has an `%error` hook, we just need to wrap it into its own try/catch, like we already do for unit-level `%error` hooks.

sethhall:
Would this cause code generation to create a hook for every field in every unit?  Seems like that could end up ballooning the generated code quite a bit if so.

rsmmr:
No, we just add a try/catch to fields that do have an `%error` handler, with the `catch` executing that handler.

sethhall:
Ah, nice.

rsmmr:
Done by #1838

PR #1838, opened 2024-08-22
Add support for field-local `%error` handlers.

rsmmr:
Add support for field-local `%error` handlers.

We now support attaching an `%error` handler to an individual field:

        type Test = unit {
            a: b"A";
            b: b"B" %error { print "field B %error", self; }
            c: b"C";
        };

With input `AxC`, that handler will trigger, whereas with `ABx` it
won't. If the unit had a unit-wide `%error` handler as well, that one
would trigger in both cases (i.e., for `b`, in addition to its field
local handler).

The handler can also be provided separately from the field:

        on b %error { ... }

In that separate version, one can receive the error message as well by
declaring a corresponding string parameter:

        on b(msg: string) %error { ... }

This works externally, from outside the unit,  as well:

        on Test::b(msg: string) %error { ... }

This is debased on top of `topic/robin/optimize-type-parsing` so that
we get the peephole optimizer.

Addresses #1824.


rsmmr:
@sethhall another one to try

sethhall:
This doesn't seem to work if you have any attributes attached to the field...

```spicy
type Test1 = unit {
    a: b"A";
    b: bytes &size=1 %error { print "field B %error", self; }
    c: b"C";
};
```

rsmmr:
>     b: bytes &size=1 %error { print "field B %error", self; }

Turns out this gets parsed as `&size=(1 % error)` (i.e., a module expression) ... !

rsmmr:
> Turns out this gets parsed as `&size=(1 % error)` (i.e., a module expression) ... !

... and this is because in the past we have explicitly added support for parsing things like `l %k == 10;` as a valid expression.

sethhall:
> > Turns out this gets parsed as `&size=(1 % error)` (i.e., a module expression) ... !

haha, well I'm glad I bumped across that when u tried to use the feature!

spicy/toolchain/src/compiler/resolver.cc
```diff
@@ -131,9 +131,7 @@ struct VisitorPass2 : visitor::MutatingPostOrder {
     }
 
     void operator()(declaration::UnitHook* n) final {
-        if ( ! n->fullyQualifiedID() ) {
-            if ( auto m = n->parent<hilti::declaration::Module>() )
-                n->setFullyQualifiedID(m->id() + n->id()); // global scope
+        if ( n->hook()->hookType() == declaration::hook::Type::Error ) {
         }
```
bbannier, line 135: This seems dead.

spicy/toolchain/src/compiler/validator.cc
```diff
@@ -478,10 +479,22 @@ struct VisitorPost : visitor::PreOrder, hilti::validator::VisitorMixIn {
     }
 
     void operator()(spicy::declaration::Hook* n) final {
-        if ( auto field = n->parent<spicy::type::unit::item::Field>() ) {
-            if ( n->isForEach() && ! field->isContainer() )
-                error("'foreach' can only be used with containers", n);
+        int count = 0;
```
bbannier, line 482: An intermediary patch had rejection of attributes unsupported for hooks here which I think would be great to have (ideally with a test). With this I can write fields like this:

```ruby
    xs: uint8[] %error foreach &size=1 {}
```

spicy/toolchain/src/compiler/validator.cc
```diff
@@ -478,10 +479,22 @@ struct VisitorPost : visitor::PreOrder, hilti::validator::VisitorMixIn {
...
+                ++count;
+                break;
+
+            case declaration::hook::Type::Error: ++count; break;
+
+            default: break;
         }
+
+        if ( count > 1 )
+            error("hook cannot have be both types, 'foreach' and '%error'", n);
```
bbannier, line 497: This should be impossible to trigger since we do not loop above so we always have `count <= 1`.

tests/Baseline/spicy.types.unit.skip-field/skips.txt
```diff
@@ -1,9 +1,9 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
-    # Begin parsing production: Skip: a_2 -> skip: a
+        # Begin parsing production: Skip: a_2 -> skip: a
```
bbannier, line 2: What's going on here and below?

doc/programming/parsing.rst
```diff
@@ -2162,6 +2162,87 @@ Note that the units for the two sides of the connection need to
...
+
+Error Handling
+===============
+
+Whenever a parser encounters an unexpected situation during
+processing, it triggers a runtime error. This includes parsing errors
+due to input that does match the current unit; failing :ref:`&requires
+<attribute_requires>` conditions; and also any logic errors in hooks,
+such as attempting to read an unset unit field or accessing an invalid
+vector index.
```
bbannier, line 2175:
```suggestion
Whenever a parser encounters an unexpected situation during
processing it triggers a runtime error. This includes parsing errors
due to input that does not match the current unit, failing :ref:`&requires
<attribute_requires>` conditions, and also any logic errors in hooks
such as attempting to read an unset unit field or accessing an invalid
vector index.
```

doc/programming/parsing.rst
```diff
@@ -2162,6 +2162,87 @@ Note that the units for the two sides of the connection need to
...
+
+By default, any runtime error will cause the parsing to terminate
+immediately, with a corresponding error message reported back to the
+host application. The Spicy parser will not be able to continue
+processing afterwards. However, there are a couple of ways to catch
+*parsing errors* (but not other runtime errors) and potentially
+recover from them, which we discuss in the following.
```
bbannier, line 2182:
```suggestion
By default, any runtime error will cause parsing to terminate
immediately with a corresponding error message reported back to the
host application. The Spicy parser will not be able to continue
processing afterwards. However, there are a couple of ways to catch
*parsing errors* (but not other runtime errors) and potentially
recover from them, which we discuss in the following.
```

doc/programming/parsing.rst
```diff
@@ -2162,6 +2162,87 @@ Note that the units for the two sides of the connection need to
...
+
+The ``msg`` parameter is optional. If it's specified, it will contain
+an error message describing the issue.
+
+By default, even with an ``%error`` hook in place, the parser will
+still terminate after executing the hook. To change that, the hook may
+use :ref:`backtracking` to specify where to continue parsing after the
+error. Alternatively, if :ref:`automatic error recovery
+<error_recovery>` is in place, the parser will attempt recovery after
+the hooks has executed.
```
bbannier, line 2212:
```suggestion
the error hooks have executed.
```

spicy/toolchain/src/compiler/parser/parser.yy
```diff
@@ -898,9 +901,8 @@ unit_hook_id: { driver->enableHookIDMode(); }
 unit_hook_attribute
               : FOREACH                          { $$ = builder->attribute("foreach", __loc__); }
```
bbannier, line 902: I am wondering whether we should as a follow-up add support for `%foreach` using the existing `foreach` machinery and deprecate hook identifier. This would make things more uniform since all hooks but `foreach` start with `%...`.

doc/programming/parsing.rst
```diff
@@ -2162,6 +2162,87 @@ Note that the units for the two sides of the connection need to
...
+
+.. versionadded:: 1.12 Per-field ``%error`` handler
+
+Rather than defining a unit-wide ``%error`` hook, it is also possible
+to just have an individual field catch its own parsing errors. The
+easiest way to do this is to attach an ``%error`` attribute to an
```
bbannier, line 2218:
```suggestion
to handle parsing errors on a per-field basis. The
simplest way to do this is to attach an ``%error`` attribute to an
```

doc/programming/parsing.rst
```diff
@@ -2162,6 +2162,87 @@ Note that the units for the two sides of the connection need to
...
+
+    module My;
+
+    type MyType = unit {
+        magic: b"MAGIC" %error { # will run if magic cannot be parsed
+            print "magic not found";
+        }
+    };
+
+To get access to the error message as well, define it out of line like this:
```
bbannier, line 2231:
```suggestion
To get access to the error message define the hook out of line like this:
```

spicy/toolchain/src/compiler/parser/scanner.ll
```diff
@@ -231,7 +232,7 @@ Null                  return token::CNULL;
 <NEW_KEYWORD>skip     return token::SKIP;
 
 {attribute}           yylval->build(std::string(yytext)); return token::ATTRIBUTE;
-<INITIAL>{property}   yylval->build(std::string(yytext)); return token::PROPERTY;
+{property}            yylval->build(std::string(yytext)); return token::PROPERTY;
```
bbannier, line 235: This change could break existing code, e.g.,

```ruby
module foo;

const debug = 2;

public type X = unit {
    var did_run: bool = False;

    xs: uint8[] &size=3 %debug {
        self.did_run = True;
        assert |$$| == 1: "expected 1, got %d" % |$$| ;
    }

    on %done {
        assert self.did_run;
    }
};
```

Before this would behave identical for debug and non-debug builds since the constant was used, now we interpret this as a hook which should only run in debug mode.

I do not have a good idea whether it is possible to fix the parsing with error hooks starting with `%error`. Another option might be to at least switch the inline form to `error` (which would be consistent with `foreach` 😬).

spicy/toolchain/src/compiler/parser/parser.yy
```diff
@@ -898,9 +901,8 @@ unit_hook_id: { driver->enableHookIDMode(); }
...
               | PRIORITY '=' expr                { $$ = builder->attribute("&priority", std::move($3), __loc__); }
-              | PROPERTY                         { if ( $1 != "%debug" ) error(@$, "unexpected hook property, only %debug permitted");
-                                                   $$ = builder->attribute("%debug", __loc__);
-                                                 }
+              | PROPERTY                         { $$ = builder->attribute($1, __loc__); }
+              | attribute                        { $$ = std::move($1); }
```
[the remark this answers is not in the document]
rsmmr: I thought about that as wellbut we can't store parameters inside attributes, and I didn't want to extend attributes with that ability if there's only this single attribute that would make use of that. Alternatives would be storing this in some other way than attributes but that would require a number of changes across the board; whereas attributes mostly just work. Plus, not sure about parsing this reliably either. Plus, the implemented syntax does mimic the `%debug` syntax, so it's not completely off.

In short: I thought we can live with this syntax for a use-case that won't be very common in the end.

spicy/toolchain/src/compiler/validator.cc
rsmmr: `foreach` and `%error` should be caught (except for your comment below about `count`, will check). `ip&v4` goes back to where we allow parsing attributes,  no? Seems that's a larger topic.

doc/programming/parsing.rst
rsmmr: I think this is fine.

spicy/toolchain/src/compiler/parser/parser.yy
rsmmr: Had the same thought but would indeed rather not do as part of this.

doc/programming/parsing.rst
rsmmr: It's fine.

spicy/toolchain/src/compiler/parser/scanner.ll
rsmmr: Yes, but this is a more general problem, our (existing) syntax is simply ambiguous, and I don't see how we can fix it 100%. We had the same issue with attributes and have fixed it the same way as I do here. I can live with `&size=3 %debug` being interpreted the wrong way if somebody actually wants that as an expression there (i.e., (1) have an ID called `debug` that supports modulo and (2) skip the space when typing out that modulo expression).

bbannier: Sounds good, and should be okay if we call that out as breaking changes in the release notes.

spicy/toolchain/src/compiler/validator.cc
rsmmr: Yeah, this was a late-stage rework of the logic--which screwed it up.

bbannier (dismissed):
Thanks! 🚀

PR #1841, opened 2024-08-27
Add support for `if` to group unit items

### LINKED ISSUE #1839: Aggregate if statements
It would be great to be able to do aggregate if statements like this...

```
type X = unit {
    has_extra_fields: uint8;
    if ( self.has_extra_fields == 1 ) {
        field1: uint8;
        field2: uint8;
        field3: uint8;
    }
};
```

Currently this would be done with the if statement suffix that is currently supported but it's a bit harder to read, write, and comprehend so that same code would look like this...

```
type X = unit {
    has_extra_fields: uint8;
    field1: uint8 if ( self.has_extra_fields == 1 );
    field2: uint8 if ( self.has_extra_fields == 1 );
    field3: uint8 if ( self.has_extra_fields == 1 );
};
```

rsmmr:
We now support `if` around a block of unit items:

    type X = unit {
         x: uint8;

         if ( self.x == 1 ) {
             a1: bytes &size=2;
             a2: bytes &size=2;
         };
    };

One can also add an `else`-block:

    type X = unit {
         x: uint8;

         if ( self.x == 1 ) {
             a1: bytes &size=2;
             a2: bytes &size=2;
         }
         else {
             b1: bytes &size=2;
             b2: bytes &size=2;
         };
    };

Closes #1839.


rsmmr:
Updates this to align syntax with `switch`: Requiring semicolon at the end and supporting the same attributes (`&size`, `&parse-*`) on the whole block.  PR description updated accordingly.

I think the main question at this point is if we want/like this syntax. Feedback welcome.

rsmmr:
This is ready for review now. I ended up deciding against attributes and pure blocks; but for `else` support.

doc/programming/examples/_parse-if-block-with-else.spicy
```diff
@@ -0,0 +1,15 @@
+# Automatically generated; edit in Sphinx source code, not here.
+module Test;
+
+public type Foo = unit {
+    a: int8;
+
+    if ( self.a == 1 ) {
+        b: int8;
+    }
+    else {
```
bbannier, line 10: With my `spicy-format` hat on I feel we should format this on a single line `} else {`, if only since it makes the absence of a `;` after the `if` block less surprising.

spicy/toolchain/src/compiler/parser/parser.yy
```diff
@@ -921,6 +922,13 @@ unit_switch_case
               | ARROW unit_field                 { $$ = builder->typeUnitItemSwitchCase($2, __loc__); }
               | ARROW unit_switch                { $$ = builder->typeUnitItemSwitchCase($2, __loc__); }
 
+opt_unit_block_else_items
+              : ELSE '{' unit_items '}'          { $$ = std::move($3); }
+              | /* empty */                      { $$ = {}; }
+
+unit_block    : IF '(' expr ')' '{' unit_items '}' opt_unit_block_else_items ';'
```
bbannier, line 929: WDYT about switching to `opt_unit_items` here to allow empty blocks, e.g., for commenting easier commenting out during development? This would mirror what is allowed for procedural `if`.

tests/spicy/types/unit/block.spicy
```diff
@@ -0,0 +1,54 @@
...
+
+    y: uint8;
+
+    on %done {
+        print self;
+
+        if ( self?.b4 )
+            print self.b4.c1;
+    }
+
```
bbannier, line 53: Could you add something here which parses with ridiculous nesting, e.g.,

```ruby
    if (True) {
        if (True) {
            if (True) {
                if (True) {
                    # TODO: actually parse something down here.
                    : void;
                };
            };
        };
    };
```

tests/spicy/types/unit/block.spicy
```diff
@@ -0,0 +1,54 @@
...
+# @TEST-EXEC: ${SCRIPTS}/printf '\x02\x10b2b3\x10\x20\xff' | spicy-driver %INPUT.hlto >>output
+# @TEST-EXEC: ${SCRIPTS}/printf '\x02\x20\x0a\x10\x20\xff' | spicy-driver %INPUT.hlto >>output
+# @TEST-EXEC: ${SCRIPTS}/printf '\x03d1\xff' | spicy-driver %INPUT.hlto >>output
+# @TEST-EXEC: btest-diff output
+#
+# @TEST-DOC: Checks unit if-blocks.
+
+module Mini;
+
+public type X = unit {
```
bbannier, line 13: Interesting tidbit: the parser currently permits putting unit vars (which are also just items) in these new blocks which might allow to more cleanly scope code (only for folding, items in syvblock are still visible in whole unit). Should we add a test for that to make sure to not break it?

If one introduces the same var in multiple blocks the error message is currently not very helpful since we report the full generated struct range (and also talk about attributes). Incorrect code like

```ruby
    if (True) {
        var myvar: uint8;
    } else {
        var myvar: uint8;
    };
```
produces

```
block.spicy:13:1-63:2: duplicate attribute in struct type
```

bbannier (dismissed):

spicy/toolchain/include/ast/types/unit-items/block.h
```diff
@@ -0,0 +1,62 @@
...
+    Block(ASTContext* ctx, int else_start, Nodes children, Meta meta)
+        : unit::Item(ctx, NodeTags,
+                     node::flatten(QualifiedType::create(ctx, hilti::type::Void::create(ctx), hilti::Constness::Const),
+                                   std::move(children)),
+                     ID(), std::move(meta)),
+          _else_start(else_start) {}
+
+    SPICY_NODE_2(type::unit::item::Block, type::unit::Item, Declaration, final);
+
+    int _else_start;
```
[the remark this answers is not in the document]
rsmmr: initialized via actor

spicy/toolchain/src/ast/types/unit.cc
```diff
@@ -98,12 +104,18 @@ static std::pair<unit::item::Field*, hilti::type::bitfield::BitRange*> findRange
             if ( auto bits = t->bits(id) )
                 return std::make_pair(field, bits);
         }
+
         else if ( auto field = item->tryAs<type::unit::item::Switch>() ) {
-            for ( const auto& c : field->cases() ) {
-                if ( auto result = findRangeInAnonymousBitField(c->items(), id); result.first )
+            for ( const auto* c : field->cases() ) {
```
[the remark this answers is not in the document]
rsmmr: readability (we should indeed use that more regularly with the new AST code)

doc/programming/examples/_parse-if-block-with-else.spicy
rsmmr: Not sure I like that ...

tests/spicy/types/unit/block.spicy
rsmmr: I think I'd rather reject because initialization semantics become muddy. But now new, I suppose this is the same already with switches.

tests/spicy/types/unit/block.spicy
rsmmr: Well, turns out this is triggering some issue inside the optimizer. Looking into it.

spicy/toolchain/src/ast/types/unit.cc
```diff
@@ -98,12 +104,18 @@ static std::pair<unit::item::Field*, hilti::type::bitfield::BitRange*> findRange
...
+                if ( auto result = findRangeInAnonymousBitField({c->block()}, id); result.first )
                     return result;
             }
         }
+
+        else if ( auto field = item->tryAs<type::unit::item::Block>() ) {
+            if ( auto result = findRangeInAnonymousBitField(field->allItems(), id); result.first )
```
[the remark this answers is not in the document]
rsmmr: Added to `spicy.types.unit.block`.

tests/spicy/types/unit/block.spicy
rsmmr: Adding a validator check.

PR #1849, opened 2024-09-02
Do not by default skip CI in update-changes cfg. [skip CI]

bbannier:
We rely on CI to create releases, so having `[skip CI]` in our update-changes config was always a hassle which needed to be edited out when creating a release (just tagging would autocommit with the message template without a way to change the commit message before tagging).

Drop this part to make releases easier and less error prone.


bbannier:
This PR marked `[skip CI]` for added irony.

rsmmr (approved):

PR #1853, opened 2024-09-03
Fix `skip` with units.

### LINKED ISSUE #1852: `skip` field can't receive parameters
I have some variable length data that needs to be skipped, similar to the following example:
```
module Test;

type Ctx = tuple<is_large_pad: bool>;

public type Testing = unit {
    %context = Ctx;

    padding: skip Pad(self.context());
    data: bytes &eod;

    on %init { self.context().is_large_pad = True; }
    on %done { print self; }
};

type Pad = unit(c: Ctx&) {
    : skip uint8;
    : skip uint8[3] if ( c.is_large_pad );
};
```

But this doesn't work

```bash
printf '\01\02\03\04Foo' | spicy-driver test.spicy
[error] terminating with uncaught exception of type hilti::rt::NullReference: attempt to access null reference (test.spicy:17:5-17:42)
```

When I removed the `skip` keyword from `padding`, everything worked fine

```diff
-    padding: skip Pad(self.context());
+   : Pad(self.context());
```
```bash
printf '\01\02\03\04Foo' | spicy-driver test.spicy
[$padding=[], $data=b"Foo"]
```

rsmmr:
For unit parsing with `skip`, we would create a temporary instance
but wouldn't properly initialize it, meaning for example that
parameters weren't available. We now generally fully initialize any
destination, even if temporary.

Closes #1852.

### DIFF
--- a/tests/spicy/types/unit/skip-unit.spicy
+++ b/tests/spicy/types/unit/skip-unit.spicy
@@ -1,0 +1,13 @@
+# @TEST-EXEC: printf 'x' | spicy-driver -d %INPUT
+#
+# @TEST-DOC: Check that even with skip, a unit instance gets properly initialized for paresing; regression test for #1852
+
+module Test;
+
+public type Testing = unit {
+    padding: skip Pad(42);
+};
+
+type Pad = unit(x: uint8) {
+    on %init { assert(x == 42); } # before fix, x ended up being zero here
+};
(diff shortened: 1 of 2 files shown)


tests/spicy/types/unit/skip-unit.spicy
```diff
@@ -0,0 +1,13 @@
+# @TEST-EXEC: printf 'x' | spicy-driver %INPUT >output
```
bbannier, line 1: Suggest to default to debug mode for usually faster compile times. Also we can drop `output` since it is empty and we do not look at it.

```suggestion
# @TEST-EXEC: printf 'x' | spicy-driver -d %INPUT
```

bbannier (dismissed):

bbannier:
Let's also backport this to `release/1.11`.

PR #1854, opened 2024-09-03
Fail for vectors with bytes but no stop condition

### LINKED ISSUE #1832: Bytes vector compiles without termination condition for bytes field.
This toy parser is applying size to the total size of the vector but the bytes in the vector actually have no termination condition.

```spicy
module test;

public type X = unit {
    a: bytes[10] &size=5;
};
```

evantypanski:
Fixes #1832 

I went back and forth on this for a while, but ended up going with the naive solution here.

The issue is just that attributes are connected to a field, not a type. So, when the vector type gets validated, it just passes along its own attributes. The inner bytes just copy the vector's `&size`:

https://github.com/zeek/spicy/blob/64c0000c2a9385020e7f357711c0da3de4b03517/spicy/toolchain/src/compiler/validator.cc#L147-L152

That just seems like it shouldn't be allowed because it's just not useful - now everything will just get shoved in the first vector element since the first element of the vector can fit all of the bytes.

You can fix that by using parentheses around the `bytes`, which [makes an anonymous field](https://github.com/zeek/spicy/blob/64c0000c2a9385020e7f357711c0da3de4b03517/spicy/toolchain/src/compiler/parser/parser.yy#L816-L822) with its own attributes. Now the validator correctly grabs that field's attributes. You can also just redirect with a type definition. Either seems preferred to this

For the fix, since [a condition before](https://github.com/zeek/spicy/blob/64c0000c2a9385020e7f357711c0da3de4b03517/spicy/toolchain/src/compiler/validator.cc#L139-L145) checks for the `item` case, we'd only get the new error with a `bytes` type that doesn't have its own declaration nor its own field. This error will only get triggered if there is an inner `bytes` type that cannot have attributes.

Other solutions would mainly be done at parse time, so I don't like them and won't go deep into that. This seems like the most consistent fix to me


spicy/toolchain/src/compiler/validator.cc
```diff
@@ -145,6 +145,9 @@ hilti::Result<hilti::Nothing> isParseableType(QualifiedType* pt, const type::uni
     }
 
     else if ( const auto& x = pt->type()->tryAs<hilti::type::Vector>() ) {
+        if ( x->elementType()->type(false)->isA<hilti::type::Bytes>() )
+            return hilti::result::Error(
+                "vectors of bytes must use a separate declaration in order to provide attributes");
         if ( auto rc = isParseableType(x->elementType(), f); ! rc )
             return rc;
```
bbannier, line 152: This code is still nonsense as it would validate whether attributes on a vector would make some vector element type parseable (even though they do not apply to it at all); I believe nothing depends on it and we should just drop this. If anything we should pass in the attributes on the element type, but we do not have access to them after the grammar builder; if OTOH the code had the element type in parens we'd already trigger the usual validation (e.g., here whether a `bytes` would have some stop condition).

This makes me wonder whether we shouldn't just completely drop this `else if` case (nonsense before, and the added error message is not immediately obvious, and only triggers the diagnostics we are interested in after the user has added parens). Instead we could change the parser so it constructs the same initial Spicy AST so we can use the existing validation. We already now produce the same parsing code for this unit, regardless of whether the user spelled `uint8` or `(uint8)`,

```ruby
type X = unit {
    xs: (uint8)[] &size=1;
};
```

The parsing code for the element type with parens is https://github.com/zeek/spicy/blob/64c0000c2a9385020e7f357711c0da3de4b03517/spicy/toolchain/src/compiler/parser/parser.yy#L795-L796

With some juggling it might be possible to always use the `unit_field_in_container` whenever `[..]` is present (I suspect this requires switching from making `[..]` optional to requiring it to distinguish vector and non-vector instances).

WDYT?

evantypanski: Ah, I initially was going to explain more about why I chose this over changing parsing but decided that'd probably just get skimmed over, but here are my thoughts. I added a commit that does it the parsing way for reference, if that's preferred I can drop the other commit.

I'm not personally a fan of modifying the parser to add something that only really makes sense at a semantic level. Really at parsing it shouldn't matter whether it's in a field or a type, it should just get into whatever is most convenient for the parser. Theoretically a vector currently supports both items or types, I think this mostly (or entirely) gets rid of the possibility of a vector-of-type - though I may be wrong with that

So I definitely would prefer the check be in the validator. I think it's just a bit weird how the current state of the validator makes that look - the fix in the validator is pretty hacky. Maybe a bit of a rework could make that nicer. One benefit to doing it in the validator is it seems more future proof, so if another case adds behavior that makes this more likely it's caught at validation, rather than requiring another parser change.

Though there's also a different approach to the problem: the `&size` *is* correctly applied to bytes (well, kind of, at least in practice). It's just unintuitive that the vector's `&size` is basically also applied to the inner `bytes`. The reason for that is simply that types don't have any attributes, fields do. I think the preferred way to have a vector of bytes is to use a type declaration for the `bytes` rather than parentheses. The initial approach lets that be the solution given, not simply applying `&size` to bytes (which without parentheses is invalid code). At least that's my take

Both approaches seem valid to me, though. I'm happy to go with whichever you prefer or modify an error or something as well

bbannier: This parser change isn't really what I was after. What I had in mind was to explicitly construct an item like currently with `unit_field_in_container` for any field involving `[..]`. This could require getting rid of `opt_unit_field_repeat` in most places and instead use `unit_field_repeat`. With that we should construct identical final ASTs regardless of whether the element type is in parens or not, and the validation change would not be needed since we already reject e.g., `(bytes)[]` where `&size` is missing on `bytes`.

evantypanski: Discussed offline, the parser change does the small amount necessary to get all `base_type`s in fields so it should work without the validator change. There is still a fair bit of tech debt sitting around, namely with `opt_unit_field_repeat` still getting used for vectors (whereas maybe that can be entirely removed for cases like constructors etc.). Since this change also makes vectors always apply to items, there may be places that the AST node `Vector` type can be changed with that assumption (or validate that assumption is correct?). I'll make a separate issue capturing that then grab it later

I also pushed a fixup change we discussed to remove the `base_type_in_container` in favor of just using `base_type` and making the field. Note that it changes the lines since it uses the `Meta` from the Vector's field rather than the anonymous `bytes` field, I wanted to make it capture more accurate column numbers by using the `base_type`'s meta but that doesn't seem done anywhere else in the grammar. I'm not sure if there's a particular reason for that. If not then I can add that to narrow the line numbers down (can see in the fixup commit that the column numbers changed, which I think is bad), or just not include the fixup.

spicy/toolchain/src/compiler/parser/parser.yy
```diff
@@ -781,10 +781,14 @@ unit_property : PROPERTY opt_attributes ';'      { $$ = builder->typeUnitItemPro
               | PROPERTY '=' base_type_no_ref ';'       { $$ = builder->typeUnitItemProperty(ID(std::move($1)), builder->expressionType(builder->qualifiedType(std::move($3), hilti::Constness::Mutable)), {}, false, __loc__); };
 
 
-unit_field    : opt_unit_field_id ':' opt_skip base_type opt_unit_field_repeat opt_attributes opt_unit_field_condition opt_unit_field_sinks opt_unit_item_hooks
```
bbannier, line 784: This is nice and compact, but misses using the same item format for the following input :/

```
type MyBytes = bytes;

public type BytesVectors = unit {
    : MyBytes[10] &size=5;
};
```

evantypanski: This case too 🤦 

```
public type BytesVectors = unit {
    : (bytes[4] &size=6)[10] &size=5;
};
```

I think this PR will remain just fixing the issue then #1855 will be to try to clean up the grammar so that all of these could be more intuitive to catch, I think there's potential to make it a fair bit nicer but maybe not in this PR since I'd want to dig in a fair bit deeper

bbannier: The issue I mentioned offline was actually with

```ruby
type MyBytes = bytes;
```

This is accepted with the current patch, even though it would cause the issue described in #1832 (albeit through an alias). I believe this PR should address this so we can close the issue.

---

You are seeing an error about "missing `&size`" with

```ruby
type MyBytes = bytes &size=4;
```

since the `&size` is a parsing attribute and it is ignored in the declaration of the type alias. This came up before and I just filed https://github.com/zeek/spicy/issues/1861 to address this by adding a diagnostic.

evantypanski: OK so for the case I pointed out I just went in and modified the test `offset-parse-from-vector.spicy` to not use attributes in a type alias so it doesn't come up

bbannier (approved):
Thanks, I think this looks great now! 🚀

The last patch c3c57f091684506b2c66bcc468ec9f2a0df1325c makes sense to me, but changes no baselines, so we probably miss some coverage there (like for many other location setups).

bbannier:
Merged with f9bbf82b7cadbf31f0e606708c380f42022847cc.

PR #1862, opened 2024-09-09
Fix resynchronization issue with trimmed input.

rsmmr:
(Sneaking in a small improvement to the printer.)

- **Add missing printer support for exception values.**
- **Fix resynchronization issue with trimmed input.**


hilti/toolchain/src/compiler/printer.cc
```diff
@@ -204,6 +204,8 @@ struct Printer : visitor::PreOrder {
 
     void operator()(ctor::Error* n) final { _out << "error(\"" << n->value() << "\")"; }
 
+    void operator()(ctor::Exception* n) final { _out << "exception(" << n->value() << ")"; }
```
bbannier, line 207: What a subtle way to bubble up incorrect exception messages 🫠

bbannier (approved):

PR #1863, opened 2024-09-09
Fix parsing for vectors of literals.

rsmmr:
- **Fix error message.**
- **Fix parsing for vectors of literals.**


bbannier (approved):
Could we please backport this to `release/1.11`?

PR #1864, opened 2024-09-16
Cleanup some vector parsing nits

### LINKED ISSUE #1855: Cleanup vector fields in the spicy grammar
Continuation of #1854 

That fix changes parsing vectors with a `base_type` to create a separate field. That removed the use of `opt_unit_field_repeat`, instead having two separate options, one with `unit_field_repeat` and one without so that a field can wrap the type. Having a field then makes sure the attributes are correctly validated.

That does not, however, change other instances of `opt_unit_field_repeat` which may be better as their own separate clauses. So there may be a couple opportunities to make that grammar more intuitive by removing uses of the optional one, instead having one possibility where it's required and one where it's not. Then wrangle the grammar a bit to make it look nicer from there. That'd also apply to `unit_field_in_container`

Then whatever cleanups follow would be nice. One possibility could be that the [unresolved field factory methods](https://github.com/zeek/spicy/blob/d087366cbf81fabcefcee00d00644b89ab56f33d/spicy/toolchain/include/ast/builder/node-factory.h#L99-L126) lose the `repeat` parameter and other vector methods are added for that particular use case, though that may be too far, I don't really know without doing it.

It may also be worth making sure a `Vector` in other scenarios won't have a type (instead it's always a unit since types are wrapped in fields). Then remove some of the [type validation logic](https://github.com/zeek/spicy/blob/d087366cbf81fabcefcee00d00644b89ab56f33d/spicy/toolchain/src/compiler/validator.cc#L147-L152) - since that case may be impossible. I may be wrong with that, though, I'd have to try it to see

Those are just spitballing suggestions though, the core is just clean up the grammar WRT vector fields

evantypanski:
Closes #1855 

This was initially supposed to be getting certain cases using `unit_field_repeat` and applying the same fix as #1854 then making the grammar nicer. But turns out [other cases were needed](https://github.com/zeek/spicy/pull/1854#discussion_r1745734171) so work was done in that PR for that. This PR just adds some more cleanup.

I was going to do some fun stuff with the grammar and split `unit_field` into like `vector_field` and `single_field` or something, then never use `opt_unit_field_repeat`... but I really didn't like how it looked. I couldn't find a way that both seemed clean and easy to follow (being "clever" in a Bison grammar seems.. wrong). So I aborted and just touched up other areas.

One thing to note is that now there's an internal error for vectors-of-type (in a field, that is). They should always be vectors of a field so that parsing attributes would properly apply. Seems to work well that way


bbannier (approved):
This doesn't really address #1855 in that we still have fields where we allow parsing `opt_unit_field_repeat` which muddies the distinction between fields with and without vectors. Like discussed offline, this isn't that straightforward to disentangle and likely would require other changes, so lets close that issue with this PR.

PR #1869, opened 2024-09-24
Fix nested look-ahead parsing

rsmmr:
When parsing nested vectors all using look-ahead, we need to return
control back to upper level when an inner look-ahead isn't found.


spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -1888,18 +1888,15 @@ struct ProductionVisitor : public production::Visitor {
         }
 
         auto builder_default = switch_.addDefault();
```
bbannier, line 1890: Nit: could move this after the condition handling now.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -1908,6 +1905,10 @@ struct ProductionVisitor : public production::Visitor {
         pushBuilder(builder_alt2);
         parseProduction(*p->alternatives().second);
         popBuilder();
+
+        pushBuilder(builder_default);
+        pb->parseError("no expected look-ahead token found", p->location());
+        popBuilder();
```
bbannier, line 1911: Nit: this was just moved here, but maybe

```suggestion
        pushBuilder(builder_default, [&]() { pb->parseError("no expected look-ahead token found", p->location()); });
```

bbannier (dismissed):

PR #1876, opened 2024-09-27
Add new library function `spicy::bytes_to_mac`.

### LINKED ISSUE #1874: Add bytes_to_mac_address function to runtime library
Requesting an enhancement by adding a new function `bytes_to_mac_address` - in the runtime library - to convert byte sequences to MAC address strings.

[Link to Slack conversation](https://zeekorg.slack.com/archives/C012CQV1K29/p1727326108447479)

Cisco do display MAC addresses differently but aside for them (AFAIK) the default way of viewing them is Colon-separated: "XX:XX:XX:XX:XX:XX".

rsmmr:
```
## Returns a bytes value rendered as a MAC address string (i.e., colon-separated hex bytes).
public function bytes_to_mac(value: bytes) : string;
```

Closes #1874.

### DIFF
--- a/spicy/runtime/src/util.cc
+++ b/spicy/runtime/src/util.cc
@@ -20,23 +20,41 @@ std::string spicy::rt::version() {
 #endif
 }
 
-std::string spicy::rt::bytes_to_hexstring(const hilti::rt::Bytes& value) {
-    std::string result;
 
-    for ( auto x : value )
-        result += hilti::rt::fmt("%02X", x);
+static inline void byte_to_hex(unsigned char byte, char* hex_out) {
+    static constexpr char hex_chars[] = "0123456789ABCDEF";
+    hex_out[0] = hex_chars[(byte & 0xf0) >> 4];
+    hex_out[1] = hex_chars[byte & 0x0f];
+}
+
+std::string spicy::rt::bytes_to_hexstring(const hilti::rt::Bytes& value) {
+    const auto& data = value.str();
+
+    if ( data.empty() )
+        return "";
+
+    std::string result;
+    result.resize(data.size() * 2); // 2 digits per hex byte
+
+    for ( unsigned long i = 0; i < data.size(); i++ )
+        byte_to_hex(data[i], &result[i * 2]);
 
     return result;
 }
 
 std::string spicy::rt::bytes_to_mac(const hilti::rt::Bytes& value) {
-    std::vector<std::string> bytes;
-    bytes.reserve(value.size());
+    const auto& data = value.str();
 
-    for ( auto x : value )
-        bytes.emplace_back(hilti::rt::fmt("%02X", x));
+    if ( data.empty() )
+        return "";
 
-    return hilti::rt::join(bytes, ":");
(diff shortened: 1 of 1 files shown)


spicy/runtime/src/util.cc
```diff
@@ -29,6 +29,16 @@ std::string spicy::rt::bytes_to_hexstring(const hilti::rt::Bytes& value) {
...
 }
 
+std::string spicy::rt::bytes_to_mac(const hilti::rt::Bytes& value) {
+    std::vector<std::string> bytes;
+    bytes.reserve(value.size());
+
+    for ( auto x : value )
+        bytes.emplace_back(hilti::rt::fmt("%02X", x));
+
+    return hilti::rt::join(bytes, "::");
```
awelzel, line 39: Total flyby: Zeek formats `endpoint$l2_addr` with a single `:` (and also has special logic for discarding the last two bytes of 8byte mac's ending with `\x00\x00`). Can see being consistent or configurable with `::` vs `:` useful.

https://github.com/zeek/zeek/blob/96c091f7acac53ccf64a28c6221b3830fd2bf439/src/net_util.cc#L220

We've also previously found that calling `snprintf()/fmt()` for individual bytes results noticeable overhead (it's done for every new ConnVal twice by default). Unsure that matters all that much here, but think `rt::fmt` is pretty heavy.

spicy/runtime/src/util.cc
```diff
@@ -20,11 +20,44 @@ std::string spicy::rt::version() {
...
+    std::string result;
+    result.resize((data.size() * 2) +
+                  (data.size() - 1)); // 2 digits per hex byte, plus one colon per byte except the last
+
+    for ( unsigned long i = 0; i < data.size(); i++ ) {
+        if ( i > 0 )
+            result[(i * 3) - 1] = ':';
+
+        byte_to_hex(data[i], &result[i * 3]);
+    }
```
bbannier, line 60: Nit: we could allocate `result` with the correct size immediately, and also avoid the branch inserting `:` below if we initialize the full result with `:`s, e.g.,

```suggestion
    // Two digits per hex byte, plus one colon per byte except the last.
    std::string result((data.size() * 2) + (data.size() - 1), ':');

    for ( unsigned long i = 0; i < data.size(); i++ )
        byte_to_hex(data[i], &result[i * 3]);
```

bbannier (dismissed):

spicy/runtime/src/util.cc
rsmmr: Actually I think we want `reserve()`, not `resize()`, so that we don't initialize the data twice. I'll fix that.

bbannier: Since `reserve` only changes the capacity, but not the size that would mean replacing all current element updates via `[..]` with `push_back` which would probably be more expensive.

rsmmr: > Actually I think we want `reserve()`, not `resize()`, so that we don't initialize the data twice. I'll fix that.

Never mind, then we can't directly access the indices. I'll take your version then.

PR #1877, opened 2024-09-27
Fix when input redirection becomes visible.

### LINKED ISSUE #1842: Problem with bytes &size=self.offset()
I'm not sure if this is related to some expression parsing issue in &size attributes on bytes fields or something else, but I'm seeing a weird behavior...

This works as you might expect (parse 3 bytes then calculate the crc32 for those bytes)...
```spicy
module Test;
import spicy;

public type X = unit {
    a: uint8;
    b: uint8;
    c: uint8;
    : bytes &size=3 &parse-at=self.input() {
        self.crc32 = spicy::crc32_add(spicy::crc32_init(), $$);
    }
    var crc32: uint64;
};
```

But this doesn't work (even though `self.offset()` returns 3)...
```spicy
module Test;
import spicy;

public type X = unit {
    a: uint8;
    b: uint8;
    c: uint8;
    : bytes &size=self.offset() &parse-at=self.input() {
        self.crc32 = spicy::crc32_add(spicy::crc32_init(), $$);
    }
    var crc32: uint64;
};
```

It seems that self.offset() used in the &size field ends up returning zero?

rsmmr:
With `&parse-at/from` we were updating the internal state on our
current position immediately, meaning they were visible already when
evaluating other attributes on the same field afterwards, which is
unexpected.

Closes #1842.


tests/spicy/types/unit/size-with-offset.spicy
```diff
@@ -0,0 +1,22 @@
...
+import spicy;
+
+public type X = unit {
+    a: uint8;
+    b: uint8;
+    c: uint8;
+
+    x: bytes &size=self.offset()  # yields 3
+             &parse-at=self.input() {
+        print self.offset(); # yields 0
```
bbannier, line 17: Like the comment said this should print `0` (we are parsing from redirected input and the "underlying stream" is `self.input()` where we start parsing at position `0`). The baseline prints `3` OTOH.

rsmmr: There's something wrong here, but there are a couple of things:

- In the test, I think it's the comment that's wrong, it should be 3 because the underlying *stream* is still the original one and `input()` is at offset 3 inside it (`&parse-at` doesn't change the stream, just the view).

- That said, the same question applies to `&parse-from=b".."`, which does change the stream (just updated the test to include this). However,  here I am wondering if zero is actually what we want because that doesn't gives us any information: it would always be zero with `&parse-from`. I'm tempted to leave this at returning the current position inside the original stream instead and just update the documentation accordingly. What you you think?

bbannier: > I'm tempted to leave this at returning the current position inside the original stream instead and just update the documentation accordingly.

I agree, the current semantics of `offset` in a field hook when the field was `&parse-from` are probably not that useful. A use case which might be more important is that `offset` behaves identical regardless of whether a _unit_ was parsed from a stream or a value, e.g., code like this:

```ruby
public type X = unit {
    a: bytes &size=1;

    y1: Y &parse-at=self.input();
    y2: Y &parse-from=self.a; # Behaves identical.
};

type Y = unit {
    a: bytes &size=1 {
        assert self.offset() == 0 : "First element should always be at offset 0";
    }
};
```

With that we should be able to change the documentation to

> Returns the offset of the current location in the input stream relative to the unit’s start. ~If executed from inside a field hook, the offset will represent the first byte that the field has been parsed from.~

rsmmr: > code like this:

I'll add that as another test.

> If executed from inside a field hook, the offset will represent the first byte that the field has been parsed from.

The main point here is actually that the offset represents the beginning of the field, not the end. I'll tweak this a bit rather than delete.

bbannier (approved):

PR #1878, opened 2024-09-27
Fix bug with captures groups.

rsmmr:
- **Add missing trim after matching a regular expression.**
- **Fix bug with captures groups.**


tests/spicy/types/regexp/parse-captures-requires.spicy
```diff
@@ -0,0 +1,15 @@
...
+#
+# @TEST-DOC: Check that we can use capture group from inside attributes
+
+module Test;
+
+public type X = unit {
+    a: /b/;
+    b: /a(b+)c/ &requires=($1 == self.a)
+          {
+              assert($1 == self.a);
```
bbannier, line 12: Nit: this came from the bug report, but we could simplify this slightly, e.g.,

```suggestion
    b: /a(b)c/ &requires=($1 == b"b")
          {
              assert($1 == b"b");
```

bbannier (approved):

tests/spicy/types/regexp/parse-captures-requires.spicy
rsmmr: Keeping the original version doesn't seem to hurt, might catch something else some time.

PR #1879, opened 2024-09-30
Fix potential nullptr dereference when comparing streams.

### LINKED ISSUE #1875: Generated code segfaults when `%synchronize-after` appears at the end of input
The following produces a segfault when run:

```
module RESP;

public type Datas = unit {
    : (Data &synchronize)[];
};

type Data = unit {
    %synchronize-after=b"a";
    ty: uint8 &requires=($$ < 5);
    };
```

Input:

```
$ printf "ba" | spicy-driver test.spicy
[1]    36912 done                printf "ba" |
       36913 segmentation fault  spicy-driver test.spicy
```

The necessary conditions are just triggering the `synchronize-after` such that it will attempt to go after the input, it seems. If the `synchronize-after` condition is never matched it exits with an exception, which seems reasonable. For example this works as I would expect:

```
$ printf "bb" | spicy-driver test.spicy
[error] terminating with uncaught exception of type spicy::rt::ParseError: failed to synchronize: &requires failed: ($$ < 5) (test.spicy:9:15-9:32)
```

rsmmr:
Because we are operating on unsafe iterators, need to catch when one
goes out of bounds.

Closes #1875.


hilti/runtime/src/types/stream.cc
```diff
@@ -586,6 +589,9 @@ bool stream::View::operator==(const Bytes& other) const {
     auto j = other.begin();
 
     while ( i != unsafeEnd() ) {
+        if ( ! i.chunk() ) // out-of-bounds, cannot  match
```
bbannier, line 592:
```suggestion
        if ( ! i.chunk() ) // out-of-bounds, cannot match
```

.typos.toml
```diff
@@ -5,3 +5,5 @@ inout = "inout"
 ocur = "ocur"
 bigsur = "Big Sur"
 Smoot = "Smoot"
+ba = "ba"
+Datas = "Datas"
```
bbannier, line 9: Nit: TBH, since it is so unlikely we'd use the identifiers again I'd probably just use different data in the test and renamed the top-level unit to e.g., `Messages` than add these here.

bbannier (approved):

.typos.toml
rsmmr: Would also require changing input here.

PR #1881, opened 2024-10-03
Fix recursive type segfault in `hiltic`

### LINKED ISSUE #1867: Recursive type alias segfaults
The following code segfaults during compilation via `spicyc`:

```
module segfault;

type Data = Data;
```

Like:

```
etyp:/Users/etyp/src/tests/spicy $ spicyc -c segfault-recursive.spicy
[1]    16815 segmentation fault  spicyc -c segfault-recursive.spicy
```

[Offending line](https://github.com/zeek/spicy/blob/1feeb214098569ad46c555d9e5f75734179a6af4/hilti/toolchain/include/ast/types/name.h#L28) just loops infinitely trying to resolve (unsurprisingly)

I think there's merit in having this be possible, but it can also be a difficult problem sometimes, so just not segfaulting would work too.

EDIT: To be clear, that minimal case above doesn't seem useful, but this kind of type may be:

```
# This still segfaults
type Data = tuple<
    # Some random fields...
    optional<real>,
    optional<int64>,
    optional<bytes>,
    # Or a vector!
    optional<vector<Data>>,
>;
```

evantypanski:
Closes #1867

tl;dr: I went back and forth but went with the almost-too-simple solution for this one, but I'm open to suggestions :)

There are two different segfaults that can happen, one from something like `Data = Data`, the other from like `Data = Data&`. The first is solved by adding a `recursion_depth` to the resolved type (which is just a magic 1000 limit but having a way to configure that would be nice, I just didn't find any). The second is solved with a cycle detection in the unifier (approximating an ["occurs check"](https://en.wikipedia.org/wiki/Occurs_check))

Note that this just disallows both cases (so `type Data = Data&` will error, not be accepted and properly generate code). Adding support to allow containers with recursive types would be a fair bit more work that should probably come after this, in my opinion. The crux there is that type definitions just end up being "find and replace" aliases in codegen, which will just loop infinitely. I think the solution would be to create a new struct or something in the C++ generated code and use its incomplete type, but that seemed a bit outside the scope here. Was that behavior ever allowed? Because it seems pretty deeply engrained that a type would never reference itself.

FWIW, this solution is a bit mediocre. A lot of weird problems arose with other solutions. It would be nice to make `Name::resolvedType` return a result (since now there is a "fail state") but the number of changes for that is pretty substantial and probably not really worth it. So instead we get an iffy error message in `resolver.cc` because I don't want to assume the only fail state is hitting the recursion limit.

Then putting the cycle detection in the unifier is weird since there were two separate places that would segfault (initially the unifier, but if that's "fixed" codegen also segfaults). A separate validation would make sense, but since the fix slots in neatly to act like an occurs check, it seems fine there. Then we don't need to traverse the tree in the somewhat unique way required to get the infinite recursion - which the type unifier will do when following names.

but.. open to any other suggestion, this is mainly what worked in the easiest way with the fewest drawbacks, in my opinion.


hilti/toolchain/include/ast/types/name.h
```diff
@@ -19,13 +17,16 @@ class Name : public UnqualifiedType {
...
+        if ( recursion_depth > 1000 )
+            return nullptr;
+
         auto t = context()->lookup(_resolved_type_index);
         if ( auto n = t->tryAs<type::Name>() )
-            return n->resolvedType();
+            return n->resolvedType(recursion_depth + 1);
         else
             return t;
     }
```
bbannier, line 32: Is there a reason we wouldn't want to use a cycle detector here? It seems that would help avoid the arbitrary recursion limit. Something like e.g.,

```suggestion
    UnqualifiedType* resolvedType(node::CycleDetector* cd = nullptr) const {
        if ( ! _resolved_type_index )
            return nullptr;

        auto t = context()->lookup(_resolved_type_index);
        assert(t);

        if ( auto n = t->tryAs<type::Name>() ) {
            if ( cd && cd->haveSeen(n) )
                return nullptr;

            if ( ! cd ) {
                node::CycleDetector _cd;
                _cd.recordSeen(n);
                return n->resolvedType(&_cd);
            }

            cd->recordSeen(n);
            return n->resolvedType(cd);
        }
        else
            return t;
    }
```

hilti/toolchain/src/compiler/type-unifier.cc
```diff
@@ -203,9 +201,18 @@ class VisitorTypeUnifier : public visitor::MutatingPostOrder {
 } // namespace
 
 void type_unifier::Unifier::add(UnqualifiedType* t) {
+    // Occurs check: We cannot handle recursive types. Error out if we see the same
+    // node twice.
+    if ( _cd.haveSeen(t) ) {
+        t->addError("Invalid cycle detected in type");
```
bbannier, line 207: Suggest to drop "invalid" since we do not give a suggestion on how to do "valid" cycles. Logging the type might also be helpful, e.g.,

```suggestion
        t->addError(util::fmt("cycle detected in definition of type '%s'", t->typeID()));
```

hilti/toolchain/src/compiler/resolver.cc
```diff
@@ -66,7 +66,10 @@ struct VisitorPass1 : visitor::MutatingPostOrder {
         }
 
         if ( n->resolvedTypeIndex() ) {
-            if ( auto resolved = n->resolvedType(); resolved->isOnHeap() ) {
+            auto resolved = n->resolvedType();
+            if ( ! resolved )
+                n->addError(util::fmt("type %s cannot be resolved by its name", n->id()));
```
bbannier, line 71: Nit: let's put the type into quotes,

```suggestion
                n->addError(util::fmt("type '%s' cannot be resolved by its name", n->id()));
```

bbannier (changes requested):
Thanks, I do not think that this looks particularly bad.

My main question would be whether we'd want the arbitrary recursion limit in the first place.

hilti/toolchain/include/ast/types/name.h
evantypanski: I actually prefer the recursion limit. This particular problem comes up with just deeply nested names. For example, I wrote a little script that makes a bunch of types which refer to each other (and the last one is `uint8`):

```
# make-test.rb
f = File.new('deeplynested.spicy', 'w')
f.write("module Test;\n")
num_iterations = 10000
(0..num_iterations).each do |i|
  f.write("type Data%d = Data%d;\n" % [i, i + 1])
end

f.write("type Data%d = uint8;\n" % [num_iterations + 1])

f.close()
```

So running `spicyc` on the generated `deeplynested.spicy` file will segfault in main but not in this branch (though the errors are a little annoying). I figured killing two birds with one stone would be better than keeping around an obvious segfault, even if it is unlikely.

Then it's slightly less overhead to increment a number than add a value to an unordered set and check it, even if it's inconsequential it could theoretically add up

Though, the random `recursion_depth > 1000` limit is pretty rough.

bbannier: Ah I see, that make sense and is indeed unrelated to cycle. Could you add a test which generates such a long chain of type aliases and then captures the new `spicyc` failure so we have it in a baseline?

evantypanski: [Added](https://github.com/zeek/spicy/blob/0b1f3214559c7eec894b8d3f906fed86577f2cd4/tests/hilti/types/id/recursion-limit.sh) - though it's a bit of a weird test :)

evantypanski: I'll hold off on this for the day, but the CI jobs for this were taking a lot longer than normal, I cancelled them and the `id.recursion-limit` test wasn't run on them (well, I suspect was currently running for a very, very long time). The CI jobs that passed (like centos stream 9) had that test looking proper.

I'm not too sure about that test :/

bbannier (approved):

hilti/toolchain/include/ast/types/name.h
evantypanski: Oops, yeah that looks good, my bad

PR #1882, opened 2024-10-06
Stop using deprecated pre-commit stage names.

bbannier:
The names we used have been deprecated since some time, see https://github.com/pre-commit/pre-commit/issues/2732.


evantypanski (approved):

bbannier:

PR #1883, opened 2024-10-08
Fix "it's" used as a possessive

evantypanski:
🤷 

(I didn't go through every case, I just grepped for "it's type")


bbannier (approved):
Thanks!

PR #1885, opened 2024-10-14
Associate source code locations with current fiber instead of current thread.

### LINKED ISSUE #1868: Switch to per-fiber location information
A user [reported on Slack](https://zeekorg.slack.com/archives/C01T41ULYKG/p1725485068421499?thread_ts=1725466191.193019&cid=C01T41ULYKG) that in Zeek they see error messages from an unrelated parser, e.g., Zeek's PostgreSQL parser reporting a _data missing_ error with a location in the QUIC analyzer. This seems to be due to different analyzers being active at the same time, and picking up incorrect location information which is stored globally via `hilti::rt::debug::location`.

We should check whether we could move location information into e.g., the fiber context, and change `debug::location` to access that information. This could have a performance impact even though location information should only rarely be surfaced to the user.

I was not able to create a standalone Spicy batch setup to reproduce this, but we might be able to see it in Zeek.

rsmmr:
This fixes potential location mix-ups when switching between fibers.
Note that we still need a context-wide fallback location as well
because we're not always running inside a fiber.

I ran a performance comparison before/after and couldn't measure a
difference. Looks like using TLS storage was a case of premature
optimization.

Closes #1868.

### DIFF
--- a/hilti/runtime/include/logging.h
+++ b/hilti/runtime/include/logging.h
@@ -36,7 +36,7 @@ void warning(std::string_view msg);
     }
 
 /** Shortcut to `hilti::rt::debug::setLocation`. */
-#define __location__(x) ::hilti::rt::debug::detail::tls_location = x;
+#define __location__(x) ::hilti::rt::debug::setLocation(x);
 
 namespace debug {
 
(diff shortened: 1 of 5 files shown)


hilti/runtime/include/logging.h
```diff
@@ -79,23 +79,32 @@ inline void dedent(const std::string_view stream) {
...
+    }
+    else
+        return nullptr;
+}
 
 /**
- * Sets the current source code location or unsets it if no argument.
- * *loc* must point to a static string that won't go out of scope.
+ * Sets the current source code location, or unsets it if argument is null.
+ * @param l pointer to a statically allocated string that won't go out of scope.
```
bbannier, line 98:
```suggestion
 *
 * @param l pointer to a statically allocated string that won't go out of scope.
```

bbannier (approved):

PR #1888, opened 2024-10-15
Fix `&max-size` on unit containing a `switch`.

### LINKED ISSUE #1871: `&max-size` incorrectly evaluated for units containing `switch`
If a unit contains a `switch` `&max-size` checks always seem to fail.

```ruby
module foo;

public type X = unit {
    switch (0) {
        * -> : void;
    };
} &max-size=4711; # Should never fail as unit consumes zero bytes.
```
```console
$ echo -n | spicy-driver foo.spicy
[error] terminating with uncaught exception of type spicy::rt::ParseError: parsing not done within &max-size bytes (foo.spicy:7:3-7:16)
```

This seems to be broken back to at least spicy-1.6.0.

rsmmr:
We would advance our input too early, letting the subsequent
`&max-size` check fail.

Closes #1871.


rsmmr:
Actually this isn't ready yet, need to check.

rsmmr:
Ready now.

bbannier (approved):
Thanks! Could you backport this to `release/1.11`?

PR #1890, opened 2024-10-15
Disallow ignored attributes on type decls

### LINKED ISSUE #1861: Missing warning on ignored parsing attributes on type alias
The parser currently accepts parsing attributes on type alias declarations, e.g., this is alias is accepted:

```
type Addr = addr &ipv4;
```

Using this alias triggers an error though:

```ruby
type X = unit {
    : Addr; # ERROR: address field must come with either &ipv4 or &ipv6 attribute
};
```

This is due to us accepting the attribute on the alias, but ignoring it (correctly since here `&ipv4` is not an proper attribute for a type, but only a field). In this case this is just confusing, but could make it to silently introduce bugs if attributes like e.g., `&requires` or `&max-size`. See e.g., https://github.com/zeek/spicy/issues/1845#issuecomment-2316899404 where this cause confused a user.

I feel we should outright reject such attributes with a diagnostic.

evantypanski:
This may make previously fine code invalid. Since that code would likely be hiding a bug, that seems okay to me.

I searched for a while and couldn't find any attributes that apply to type decls other than `&cxxname` and `&bit-order`, but maybe my grepping wasn't creative enough

Closes #1861


evantypanski:
As a note, using the `AttributeList` makes the locations off:

https://github.com/zeek/spicy/blob/818155348f621ba962bfd71de5d1f19246f67e07/tests/Baseline/spicy.attributes.invalid-type-attributes/output#L2-L5

I took a trivial look at making `Node::addChild` do some location shenanigans but that might make other stuff less intuitive. And it's a little hard to "combine" the cached `Meta` instances at first glance. So I think that should be a separate issue that I can look into

bbannier:
> As a note, using the `AttributeList` makes the locations off:
> 
> https://github.com/zeek/spicy/blob/818155348f621ba962bfd71de5d1f19246f67e07/tests/Baseline/spicy.attributes.invalid-type-attributes/output#L2-L5
> 
> I took a trivial look at making `Node::addChild` do some location shenanigans but that might make other stuff less intuitive (one `&requires=` diagnoses the whole attribute, another switches to just the inner part? not exactly sure why) And it's a little hard to "combine" the cached `Meta` instances at first glance. So I think that should be a separate issue that I can look into

Good catch. I'd suggest you instead go with the location of the decl `n` for now so we don't report bogus source ranges, and add a TODO to switch to the attr location once issue #HIJKL is fixed.

bbannier (approved):

PR #1892, opened 2024-10-16
Fix usage of `&convert` with unit's requiring parameters.

rsmmr:
Fix usage of `&convert` with unit's requiring parameters.

The generated code needs to create a temporary instance of the type,
but doesn't have any arguments to provide to it. But that's ok, and we
now let the validation pass and just instantiate a default-constructed
instance.

This is a bit hackish due to needing to shift things around to the
right places, but couldn't think of a better fix.

- **Factor out logic to validate arguments given to a type.**
- **Fix usage of `&convert` with unit's requiring parameters.**

### DIFF
--- a/hilti/toolchain/include/compiler/validator.h
+++ b/hilti/toolchain/include/compiler/validator.h
@@ -38,6 +38,10 @@ public:
     /** Returns the number of errors reported so far. */
     auto errors() const { return _errors; }
 
+    /** Validates if provided type arguments match a type's expectation. */
+    void checkTypeArguments(const node::Range<Expression>& have, const node::Set<type::function::Parameter>& want,
+                            Node* n, bool allow_no_arguments = false, bool do_not_check_types = false);
+
 private:
     Builder* _builder;
     int _errors = 0;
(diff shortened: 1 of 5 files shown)


hilti/toolchain/include/compiler/validator.h
```diff
@@ -38,6 +38,10 @@ class VisitorMixIn {
     /** Returns the number of errors reported so far. */
     auto errors() const { return _errors; }
 
+    /** Validates if provided type arguments match a type's expectation. */
+    void checkTypeArguments(const node::Range<Expression>& have, const node::Set<type::function::Parameter>& want,
+                            Node* n, bool allow_no_arguments = false, bool do_not_check_types = false);
```
bbannier, line 43: Nit: I'd rename these to get rid of negatives, e.g., here and also in the impl.

```suggestion
                            Node* n, bool allow_empty_args = false, bool check_types = true);
```

The `allow_empty_args` does not change semantics, but `do_not_check_types` does and would require flipping the callside and the check in the impl.

bbannier (approved):

hilti/toolchain/include/compiler/validator.h
rsmmr: I don't know but I prefer the negative versions to specify the "unusual" case.

PR #1894, opened 2024-10-16
Fix spicy typo in configure `Spict`->`Spicy`.

evantypanski:


bbannier (approved):

configure
```diff
@@ -83,7 +83,7 @@ if [ ! -e "$source_dir/3rdparty/doctest/CMakeLists.txt" ] && [ -d "$source_dir/.
 You seem to be missing the content of the 3rdparty/doctest directory.
```
bbannier, line 83: This is outdated as we have more that `3rdpartydoctest` now, probably should be

```suggestion
You seem to be missing the contents of the 3rdparty/ directory.
```

PR #1897, opened 2024-10-17
Prevent SafeInt from throwing exceptions on bitshift overflows.

### LINKED ISSUE #1829: Assertion `(U)bits < (int)safeint_internal::int_traits< T >::bitCount' failed.
Running the following code through `spicyc -d -j` triggers an assertion. Haven't looked further, but seems the expression should fit 64bit, or a reasonable RuntimeError raised.

```
$ cat ./issues/sh.spicy 
module sh;

global first = uint8(167);
global rem = vector<uint8>(84, 144, 87);

global saslLen: uint64;

saslLen = (first << 24) + (rem[0] << 16) + (rem[1] << 8) + rem[2];

print "saslLen", saslLen;

$ ./build/bin/spicyc -j ./issues/sh.spicy  -d
spicyc: /home/awelzel/corelight-oss/spicy/hilti/runtime/include/hilti/rt/3rdparty/SafeInt/SafeInt.hpp:6007: constexpr SafeInt<T, E> SafeInt<T, E>::operator<<(SafeInt<U, E>) const [with U = long unsigned int; T = unsigned char; E = hilti::rt::integer::detail::SafeIntException]: Assertion `(U)bits < (int)safeint_internal::int_traits< T >::bitCount' failed.
Aborted (core dumped)
```

Without `-d`, it prints `saslLen, 87` which seems to indicate only `rem[2]` is taken into account? The other uint8 values are all shifted to 0? Is this something one needs to worry about, or should shifting promote they type? A better result comes with:
```
(uint32(first) << 24) + (uint32(rem[0]) << 16) + (uint32(rem[1]) << 8) + uint32(rem[2]);
```

---

This was found while fuzzing Zeek's LDAP analyzer locally. It employs such a shift [here](https://github.com/zeek/zeek/blob/d52ec28c8922ec2f9f42f08462a49b0094d44805/src/analyzer/protocol/ldap/ldap.spicy#L198-L199):
```
  rem: uint8[3] if ( ctx.messageMode == MessageMode::ENCRYPTED && (|self.mech| == 0 || self.mech.starts_with(b"GSS")) ) {
    self.saslLen = (self.first << 24) + ($$[0] << 16) + ($$[1] << 8) + $$[2];
  }
```

Zeek's fuzzers are build with `assert()` enabled.

bbannier:
Closes #1829.


tests/Baseline/hilti.types.integer.shift-overflow/output
```diff
@@ -0,0 +1,2 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+uncaught exception hilti::rt::Overflow: <no error>
```
bbannier, line 2: This is not ideal as it does not include a location. I tried to make use of `hilti::rt::debug::location`, but sadly `hilti/runtime/include/logging.h` includes this header already.

bbannier: Well, I could just forward-declare it.

PR #1899, opened 2024-10-17
Encompass child node's location in parent.

### LINKED ISSUE #1893: `AttributeSet` report invalid locations
Followup from #1890 

Using the `AttributeSet` in a diagnostic gets the following output:

```
 [error] <...>/invalid-type-attributes.spicy:11:17-11:16: attributes are not allow on type aliases 
 [error] <...>/invalid-type-attributes.spicy:12:15-12:14: attributes are not allow on type aliases 
 [error] <...>/invalid-type-attributes.spicy:15:28-15:27: attributes are not allow on type aliases 
 [error] <...>/invalid-type-attributes.spicy:16:20-16:19: attributes are not allow on type aliases 
```

Where the locations are wrong (see: `11:17-11:16`). The location of the `AttributeSet` should encompass all of its children, I think. I'd ideally put this in `addChild` for the node, so any children expand a node's location, but that may have some other consequences.

evantypanski:
Closes #1893

When a child is added with `addChild`, the parent's location should (generally) span over that child as well. This primarily helps for cases where a node doesn't have much of a location until it gets children added - like `AttributeSet`. The locations for those should encompass all of the attributes within the set. That logic applies for any node with a child: if it's the child, then its location should reflect that.

There is an edge case that needed special handling: later in `node::detail::deepcopy`, including the child's location there caused issues with resolved names. A field used in an attribute, for example, would get added as a child to that attribute. Then the location would weirdly go up to encompass the field, then whatever is in the middle, then the attribute. There may be other cases that a child is added that is generated or something, but I didn't catch any and it doesn't seem super pressing if that's the case.


hilti/toolchain/src/ast/meta.cc
```diff
@@ -5,3 +5,46 @@
...
+    int from_character = -1;
+    int to_character = -1;
+
+    // These calculations are weird. The logic:
+    //
+    // 1) If this starts on a line before other, then choose this for the first point
+    // 2) If this and other are same line, choose this if it is a lower column
+    // 3) Otherwise, choose other
+    //
+    // Then the same logic applies to 'to' as well
```
evantypanski, line 25: Please check this logic, this is a gosh darn mess but I dunno how to do it better. :/

rsmmr (commented):
Haven't looked very closely but quick comment: That location merging is pretty complex, and I'd be reluctant to run that on every `addChild()`. Is the issue we are addressing here primarily relevant to just `AttributeSet`, or do we see it more generally? If the former, I'm wondering if could fix it more narrowly too. For example, `AttributeSet::add()` could update locations itself; or we could have the resolver look at `AttributeSets` and recompute their locations if necessary. 

One other idea: we could merge locations only when we actually retrieve them. Either the validator does it just before printing an error; or built into `location()` to always do it on access. 

But overall I think the node-specific solution (e.g., do it in `AttributeSet::add()` would be my preference if we can identify the (small number of) cases where it matters. Could still have a `combineLocations()` in `Node`, but would call it only on demand from child classes.

evantypanski:
> Is the issue we are addressing here primarily relevant to just AttributeSet, or do we see it more generally?

So the two cases in the debug logs in the PR are related to attributes, namely:

1) `AttributeSet`
2) `Attribute` like `&cxxname`

no others seemed to pop up, hard to tell if the locs would be bogus but we never use them in diagnostics or something else.

It seems to me like this is generally applicable, but adding this logic on every added child may be a bit much, I agree. That was the intention of [this check](https://github.com/zeek/spicy/blob/14845ec88b54b43c198582220686e6c8638162b2/hilti/toolchain/src/ast/meta.cc#L10-L11), but then you still go through the trouble of accessing the cache every time a child is added that's just unneeded work, too.

I did initially put this in `Location` but got tripped up with the `Meta` caching stuff. I understand it better now though, so it can probably go back there :) I'll work on that.

hilti/toolchain/src/ast/node.cc
```diff
@@ -149,7 +149,8 @@ Node* node::detail::deepcopy(ASTContext* ctx, Node* n, bool force) {
     auto clone = n->_clone(ctx);
 
     for ( const auto& c : n->children() )
-        clone->addChild(ctx, c); // this will copy the children recursively (because they have a parent already)
+        clone->addChild(ctx, c, /*update_meta=*/false); // this will copy the children recursively
```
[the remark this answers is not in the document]
evantypanski: no longer needed since per @rsmmr's suggestion, the logic was taken out of `addChild`

bbannier (dismissed):
Thanks, this looks good to me though I wonder how much more expensive aggressively merging locations on `addChild` would have been. @rsmmr, are we good to go here?

hilti/toolchain/include/ast/attribute.h
```diff
@@ -136,7 +136,10 @@ class AttributeSet : public Node {
     bool has(std::string_view tag) const { return find(tag) != nullptr; }
 
     /** Adds an attribute to the set. */
-    void add(ASTContext* ctx, Attribute* a) { addChild(ctx, a); }
+    void add(ASTContext* ctx, Attribute* a) {
+        addChild(ctx, a);
+        combineLocations(a);
```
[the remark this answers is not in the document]
evantypanski: I prefer the separate method for clarity but this works too :)

hilti/toolchain/include/ast/attribute.h
```diff
@@ -136,7 +136,12 @@ class AttributeSet : public Node {
     bool has(std::string_view tag) const { return find(tag) != nullptr; }
 
     /** Adds an attribute to the set. */
-    void add(ASTContext* ctx, Attribute* a) { addChild(ctx, a); }
+    void add(ASTContext* ctx, Attribute* a) {
+        addChild(ctx, a);
+        // Combine this location with the attribute's location so this spans the range
+        auto new_meta = Meta(meta().location().mergedRange(a->location()), meta().comments());
+        setMeta(std::move(new_meta));
+    }
```
rsmmr, line 144: How about one more intermediary: add `mergeLocation()` to `Meta` so that this can become:

```suggestion
    void add(ASTContext* ctx, Attribute* a) {
        addChild(ctx, a);
        // Combine this location with the attribute's location so this spans the range
        setMeta(meta()->mergeLocation(a->location()));
    }
```

Then it's easier to reuse elsewhere if other similar cases come up.

hilti/toolchain/include/ast/location.h
```diff
@@ -44,6 +44,12 @@ class Location {
     auto from() const { return _from_line; }
     auto to() const { return _to_line; }
 
+    /**
+     * Merges this Location with the provided location. Returns new a location
+     * which encompasses the entire span.
+     */
+    Location mergedRange(const Location& loc) const;
```
rsmmr, line 51: Suggest to simply call this `merge`:

```suggestion
    Location merge(const Location& loc) const;
```

rsmmr (commented):
Yeah, I like this, just a couple further tiny suggestions.

rsmmr (approved):

PR #1902, opened 2024-10-18
Fix namespacing of `hilti` IDs in Spicy-side diagnostic output.

rsmmr:
Fix namespacing of `hilti` IDs in Spicy-side diagnostic output.

We now show them with a `spicy` prefix, which makes more sense for
users.

This is a bit messy because our printer code (1) isn't aware of the
HILTI/Spicy distinction; (2) isn't differentiating between output
that's user-visible vs not; and (3) manages its current
parameterization a bit awkwardly through global state. So this
shoehorns a notion of user-visibility in there, and enables the Spicy
plugin to replace `hilti::` with `spicy::` when it deems that
necessary. Eventually we should overhaul the printer code to be more
modular and contextual.

- **Add printer plugin hook to customize ID printing.**
- **Fix namespacing of `hilti` IDs in Spicy-side diagnostic output.**


hilti/toolchain/include/compiler/printer.h
```diff
@@ -13,26 +13,32 @@
...
  * `print_ast` hooks.
+ *
+ * @param out output stream
+ * @param root top-level node of the AST to print (which does not need to be an `ASTRoot`)
+ * @param compact if true, create a one-line representation
+ * @param user_visible if true, signal to the printer that the output is
+ * intended for user consumption, permitting it to do some visual polishing
  */
-void print(Stream& stream, Node* root); // NOLINT
+void print(std::ostream& out, Node* root, bool compact = false, bool user_visible = true);
```
bbannier, line 34: This is now pretty context-sensitive while before `compact` was more about looks. WDYT about making all these parameters required? The only call site not passing all args is `CodeGen::_getOrCreateTypeInfo` which IMO should be passing `user_visible=false` even though flipping that seems to make no difference for the test suite.

hilti/toolchain/src/compiler/printer.cc
```diff
@@ -1111,20 +1118,26 @@ void printer::print(std::ostream& out, Node* root, bool compact) {
...
 }
 
-void printer::print(printer::Stream& stream, Node* root) {
+void printer::Stream::_print(Node* root) {
     util::timing::Collector _("hilti/printer");
 
+    struct _ {
+        ~_() { detail::State::current->current_plugin = nullptr; }
+    } __;
+
```
bbannier, line 1131: We do almost the same in `printer::print` already but for the check for `detail::State::depth`, and if I remove this code here the test suite still passes, so not sure why we need it.

In any case I think not having this management code here in `_print` but in `print` would be much easier to follow since `current_plugin` gets set up there as well (otherwise calling `_print` would always reset `current_plugin` as a side-effect).

spicy/toolchain/src/compiler/printer.cc
```diff
@@ -16,6 +16,26 @@
...
+    // have a corresponding Spicy type. The alternative would be to explicitly
+    // identify valid mappings somehow (e.g., through a shared `&cxxname`).
+    // However, that's neither easy nor is it clear that that's worth it. For
+    // one, we currently do indeed maintain only such 1:1 mappings (i.e., we
+    // don't rename IDs existing at both layers other than the namespace). And
+    // second, when displaying Spicy code to users, there should really never
+    // be any HILTI identifier showing up anyways; so if we still end up with
+    // any, printing them with a `spicy` prefix is probably still a better
+    // solution than just printing them as-is.
+    if ( out.state().user_visible && id.namespace_() && id.sub(0) == ID("hilti") ) {
```
bbannier, line 31: Entering nit territory:

```suggestion
    if ( out.state().user_visible && id.length() > 1 && id.sub(0) == ID("hilti") ) {
```

The semantics of `namespace_()` are always slightly surprising to me since it returns everything up to the local part, e.g., in `spicy::Charset::UTF8` the "namespace" is `spicy::Charset`. What we actually want here IDs with `length > 1`.

This would also avoid allocating a `string` for `namespace` when all we want to check is presence.

bbannier (approved):

spicy/toolchain/src/compiler/printer.cc
rsmmr: I read the original versions "does it have a namespace?", which sounds right, doesn't matter how long it is, that's what the next check is for.

hilti/toolchain/include/ast/node.h
```diff
@@ -710,17 +710,34 @@ class Node {
...
+     * its children. This always renders the code as *not*
+     * "user-visible", per the flag in the extended version of Zeek.
+     *
+     * Note that this can be called from inside a debugger.
+     */
+    std::string printRaw() const;
+
+    /**
+     * Renders the node as HILTI source code, using the same semantics
+     * as `print()`.
```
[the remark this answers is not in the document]
rsmmr: I don't understand why it would be better to skip this?

hilti/toolchain/include/compiler/printer.h
```diff
@@ -13,26 +13,32 @@
...
+struct Plugin;
 
-/**
- * Prints an AST as HILTI source code. This consults any installed plugin
- * `print_ast` hooks.
- */
-void print(std::ostream& out, Node* root, bool compact = false);
+namespace printer {
+
+class Stream;
```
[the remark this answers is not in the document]
rsmmr: Ah, this can go away now indeed.

hilti/toolchain/src/compiler/printer.cc
rsmmr: It's technically probably not necessary, but seems cleaner not leave the state set to something that's not longer valid/relevant.

> In any case I think not having this management code here in _print but in print would be much easier to follow since current_plugin gets set up there as well 

`current_plugin` gets set only in `_print` actually, just before we execute the plugin's hook, so I'd leave the clearing there as well.

Btw, I think we should add a utility class sometime that's like this: https://github.com/zeek/zeek-agent-v2/blob/7e07589bc81d7ee6f49b703ffad13239c1ab7b8e/src/util/helpers.h#L36-L49

bbannier: I got confused :confused:; `print` sets up the current _state_ via `current` (so it can e.g., `setCompact` which is in `State`, and `print_` sets _`current_plugin`_ which lives in the `current` state. Sorry for the noise.

The way the cleanup works here we still leak `current_plugin` from the loop, maybe like this:

```cpp
void printer::Stream::_print(Node* root) {
    util::timing::Collector _("hilti/printer");

    for ( auto& p : plugin::registry().plugins() ) {
        if ( ! p.ast_print )
            continue;

        auto* prev = std::exchange(detail::State::current->current_plugin, &p);
        auto _ = scope_exit([&]() { detail::State::current->current_plugin = prev; });

        if ( (*p.ast_print)(root, *this) )
            return;
        else {
            // If the print hook did not succeed defer to default printer.
            // This might still make use of the currently selected plugin.
            Printer(*this).dispatch(root);
            return;
        }
    }

    // Defer to the default printer with the current plugin (which might be unset).
    //
    // TODO: We should be able to replace this with `util::cannotBeReached()`
    // if the HILTI plugin provides `ast_print` (e.g., which could just return `false`).
    Printer(*this).dispatch(root);
}
```

This uses a hypothetical scope guard a la [`std::experimental::scope_exit`](https://en.cppreference.com/w/cpp/experimental/scope_exit)

```cpp

template<typename EF>
struct scope_exit {
    scope_exit(EF f) noexcept : _f(std::forward<EF>(f)) {}

    scope_exit(const scope_exit&) = delete;
    scope_exit(scope_exit&&) = delete;

    ~scope_exit() { _f(); }

    EF _f;
};
```

hilti/toolchain/include/ast/node.h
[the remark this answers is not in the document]
bbannier: I got confused by the overloads where one `print` previously had some args defaulted 🤦. This _does have_ the semantics of `print()` (the argless overload).

hilti/toolchain/src/compiler/printer.cc
rsmmr: >  we still leak current_plugin from the loop

Oh, thanks for that observation! Turns out that demonstrates something unexpected: if I change it to not leak, the PR doesn't work anymore. The thing is that the `dispatch()` still needs to get that new ID hook to execute when we're printing Spicy code. But when executing, it checks the current plugin and if that's now unset, the hook won't do anything. Need to restructure that a bit once more.

bbannier (approved):

PR #1904, opened 2024-10-21
Support `&requires`for individual vector items.

### LINKED ISSUE #1857: `&requires` attribute on vector element ignored
We currently do not evaluate `&requires` attributes on vector elements, e.g., I would expect the following to fail with a `&requires` error, but instead reach `"unreachable"`.

```ruby
module foo;

public type X = unit {
    : (uint8 &requires=$$ == 0)[] foreach {
        assert $$ == 0 : "unreachable";
    }
};
```

```console
$ printf '\x00\x01' | spicy-driver -d foo.spicy
[fatal error] terminating with uncaught exception of type hilti::rt::AssertionFailure: unreachable (foo.spicy:5:9-5:39)
```

This looks similar to #1860 but was broken before the use case reported there, at least since v1.1.0-dev-22-g0bfdd358c.

rsmmr:
Closes #1857.


spicy/toolchain/include/compiler/detail/codegen/parser-builder.h
```diff
@@ -474,10 +474,10 @@ class ParserBuilder {
 
     /**
      * Called when a container item has been parsed. Returns a boolean
-     * expression that true if container parsing is to continue.
+     * expression that is true if container parsing is to continue.
      */
-    Expression* newContainerItem(const type::unit::item::Field& field, Expression* self, Expression* item,
```
evantypanski, line 479: just curious, why were these (`container`, that is) swapped from const reference to const pointers?

evantypanski (approved):
👍 from me

spicy/toolchain/include/compiler/detail/codegen/parser-builder.h
rsmmr: It's clean-up. I pretty much rewrote the whole AST code last year. Before the rewrite, AST nodes generally would be passed by value (or const ref) because of how their memory was managed (it was using type erasure and copy-by-value, so the pointers stained internal and weren't visible to users of the AST node). Turned out that whole approach wasn't such a good idea for multiple reasons, so with the rewrite I switched to a more classic representation of plain AST nodes with inheritance and virtual methods, and generally passing raw pointers around, which made everything much more efficient. 

So that's the style now, but not all places have been fully adapted, and this was one of them. So in the PR  I took the opportunity to switch to passing a pointer to increase consistency (one could argue here that the pointer isn't really needed because `Field` is a leaf class, so no polymorhism needed that the pointer would facilityate; but I think it's still a good rule of thumb to generally pass all AST nodes by pointer now, no matter what). Second piece to this is that callees may generally modify nodes passed, hence passing nodes as non-const is the second piece for that rule of thumb. Again, here, it could have stayed const due not the method not making changes, but doesn't really hurt either.

PR #1907, opened 2024-10-23
Do no longer escape backslashes when printing strings or bytes.

rsmmr:
Escaping backslashes in `print` output seems both unnecessary and
confusing.

- **Introduce style flags for functions rendering values into strings.**
- **Rename `expandEscapes()` to `expandUTF8Escapes`.**
- **Do no longer escape backslashes when printing strings or bytes.**


hilti/runtime/include/util.h
```diff
@@ -285,7 +286,48 @@ constexpr auto enumerate(T&& iterable) {
...
+ * Flags specifying escaping style when rendering raw data for printing. The
+ * default style renders all non-printable characters as hex escapes (`\xNN`)
+ * and escapes backslashes with a second backslash. Any specified flags modify
+ * the default style accordingly.
+ */
+enum class Bytes {
+    Default = 0,                    /**< name for unmodified default style */
+    EscapeQuotes = (1U << 1U),      /**< escape double quotes with backslashes */
+    UseOctal = (1U << 2U),          /**< escape non-printables with `\NNN` instead of `\xNN` */
+    NoEscapeBackslash = (1U << 3U), /**< do not escape backslashes */
```
evantypanski, line 303: It's a little confusing to me that some of these are "do X" but this one is "Don't do X" - I know there's some simplicity you get from this, but I'd prefer them all be one or the other.

My preference would be you pass in flags for what you want escaped, so this becomes `EscapeBackslash` and all logic flipped (and each time it is not passed in it gets passed in).

hilti/runtime/include/util.h
```diff
@@ -285,7 +286,48 @@ constexpr auto enumerate(T&& iterable) {
...
+ * style accordingly. If not otherwise noted, any escapings are reversible
+ * through `expandUTF8Escapes()`.
+ */
+enum class UTF8 {
+    Default = 0,                    /**< name for unmodified default style */
+    EscapeQuotes = (1U << 1U),      /**< escape double quotes with backslashes */
+    NoEscapeBackslash = (1U << 2U), /**< do not escape backslashes; this may leave the result non-reversible */
+    NoEscapeControl = (1U << 3U),   /**< do not escape control characters and null bytes */
+    NoEscapeHex =
+        (1U << 4U), /**< do not escape already existing `\xNN` escape codes; this may leave the result non-reversible */
```
evantypanski, line 320: same comment as before, I'd prefer these flags be changed to `EscapeBackslash`, `EscapeControl`, and `EscapeHex` - escaping features that you opt into rather than a mix.

hilti/runtime/include/util.h
```diff
@@ -295,10 +337,8 @@ std::string expandEscapes(std::string s);
  * @param escape_quotes if true, also escapes quotes characters
  * @param use_octal use `\NNN` instead of `\XX` (needed for C++)
  * @return escaped string
- *
- * \todo This is getting messy; should use enums instead of booleans.
  */
-std::string escapeBytes(std::string_view s, bool escape_quotes = false, bool use_octal = false);
+std::string escapeBytes(std::string_view s, bitmask<render_style::Bytes> style = render_style::Bytes::Default);
```
evantypanski, line 341: If swapping to only using positive "escape this" type queries, maybe this default gets changed to some constant that's defined above? Then users can modify from that constant

evantypanski (approved):
I like this, it seems a lot more clear. My only preference would be each `enum` value be in the affirmative ("escape X") rather than a mix ("escape X" + "don't escape Y") but I understand how it makes passing in those values simpler

rsmmr:
> I like this, it seems a lot more clear. My only preference would be each `enum` value be in the affirmative ("escape X") rather than a mix ("escape X" + "don't escape Y") but I understand how it makes passing in those values simpler

At first, I actually had it exactly the way you're suggesting: All affirmative rather than negative, plus a suitable default argument--but I didn't like it. 

The problem with that is that once you need to change one flag from the default atgument, you now need list them all, meaning you also need to know what the default combination actually is and make sure to include all of that as well. And if we later add another flag that then needs to be included into the default, we would need to update all the callers accordingly to ensure they are not changing behavior. So instead, I opted for default styles that offer the most common case, from which one can select individual behaviour to change one by one, even if that needs these `No...`-style flags. That ended up simplifying many call sites and if we add more flags later, call sites won't need up updating either (indeed that final commit is already an example of that).

evantypanski:
> The problem with that is that once you need to change one flag from the default atgument, you now need list them all

Hmmm... maybe I'm [overengineering this slightly](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIAGykrgAyeAyYAHI%2BAEaYxBKBAA6oCoRODB7evgGkyamOAqHhUSyx8VyBdpgO6UIETMQEmT5%2BlZj2BQx1DQRFkTFxCbb1jc3ZlSO9Yf2lgxUAlLaoXsTI7BwA9BsA1AAqAJ6JmNv7y8TbaFjbCHGYpNsk27SoTOjbhtuYqqyJ9AB0Jg0AEFASCzABmMLIbxXEzgtxOBQEYiYVhw7Cg0FI4heBzbCz7AiYBTbEwAdiswO21M%2BDB82wAamIvMcQNswgRSRTQTTedtsApREcAIpeVBEklwgAi2wgXDkpPhcLc23liz52w2ACotcrPoKmEdtuhltF6NsAI5iiXbADuhAQ22iTGQAGsFAYFDcSVqNjyNdtFJgAPIOMSKmVyhXKvVmOTqgPU7W6%2BH6oXHBgCAC0iWIHKYZuJdodpP8GhMAFY3BEayYy%2ByGEjUW9UFRS%2BWq6pa/Xff6NQL0xYXe7PY7pbL5Yq3HrwfH7kmdXriemncOPUwvUXe1SaeSpXDKSCd9Smd5jmFwsQD5jjw3aNMLgIm6pc/jCUXUEdiEwiFfLBACQlEAQFPFkHgIG5iHmLlD0TFECBWBgYO2P4LziCNtiRH88GQAB9UQkWVUDMHRCAsMcPCCIIZUOVIiC8AULN0TQqCuRVcicPwjdqPhWjwWwCBxUg%2BZoL3a9bzE4E%2BzvB80EbIkX3OQCPy/H8SCzAD3wUYDiPA4SYOk3l4MQ5DUIYS8MI4yjuKI5kSP4sj6gorjCN4wQ6IQBimP4ljRLMfxtgAPyslyeIRdyHKEuIRK5fdwVg3cyX3KTb2iVBPG2KhUQQlFNKAkC7My7LRO5W8jMwHKkJY0l/KKn9xKBXlJMa3dbysmTwkfeSvlfZSSSwKgmC8WgCFwiASoSjVjOIJDyQsFDqvHYjgIHQ1MCHN11y9WKGqapLMX2%2BKbyBAA3VA8DeKg8qLLKf0lcEZT64CBqGkaxpiubpKRdBgLQLxORjVNzDMVajStcViThAAxGqzCnPVboIBQ/kRlZMGu7SQFBzBRQhhRRKVVNvuA1xaF2mliZQZYAcJlVgaDB4wzJ8EYeB%2BHU0R5HUdyp6QCDUN6loAnpyJggfpAUnyepSm/ppkW6bMEGDSNZ1NtHaHYfZlVOZR7K0YxlblfWtdR2FvVKclo62rF37qa1zCbYlhh0GZw9mo5bYWCYMJxoM28rvmKXMoNkAXuG0bfesN98ux3GJUDq2WupK7ebDt7fazaPiUNwcTY3BAE7dpKOEWWhOArXg/G4XhUE4adLCjhQzjWGrwR4UgCE0EvFldEAKw0fROEkXgWAkDQB6rrRSFrjheG0gfO44LRFjgWAkC%2Bap/pIchKAaYAFGUQx2iEBBUFtKv27QFhEjobCBEP8JaBPs/J94K%2Bb/oeJoUMYAuAATi4BuJQjRSDvzoHECIrB1hv1QNfcBxBgz/WfufLuQRVDVCBMQfenBeAb2QHUfAVdeD8EECIMQ7ApAyEEEGNQqDdBcH0D/FA1hrD6DwNEbSkBFifk6NpWe08TpxDzFgTh41SA4kEHgNgux0q0FEYsJuKw1h6G%2BmEB%2Bx9T4oOrh3FE7AB62m/IkTgPBS7l0rqgme2B0HIC3ucVQAAOfwWZ/CSAuAYIwqo/5/C4H8DQsogFxF4dBCA9crCWHuLgQgjxzBt3mLwRey8e59wHmXDgw9SCv2njg2wIAF5dxEoPDgZhzFLxrtkhJ3dSCCOIKkZwkggA%3D) ;)

That does, however, seem to fix those issues

But I'm ok either way.

rsmmr:
> That does, however, seem to fix those issues. 

Nice idea with the `default_()`. However, I still don't like that everybody needs to remember calling that (`f(Bytes::default_() + Bytes::EscapeQuotes);`, so in the end I still prefer my original approach.

> If the default changes I'd expect to go through each call site to make sure that's okay anyway, personally

With the approach in this PR, behavior (including default) will never change for existing call sites: any new flags would be opt-in.

Anyways, I think we're all collectively overthinking this at this point. :-) It's a couple of small internal functions and the PR is primarily about something else, so I'll go ahead and merge.

PR #1909, opened 2024-10-25
Fix performance regression when parsing bytes.

### LINKED ISSUE #1908: Optimize parsing for bytes &size=N regressed QUIC performance
As reported in the Zeek repo, a recent bump of Spicy regressed performance of pcap-quic-16-50mb.

https://github.com/zeek/zeek/pull/3953#issuecomment-2431443070

I bisected the range of the bump and it is: ca5de306a7c397c1d37aa7eb662dabafaa8ca26b

```
$ git bisect log
git bisect start
# status: waiting for both good and bad commits
# good: [64c0000c2a9385020e7f357711c0da3de4b03517] Merge branch 'topic/bbannier/update-update-changes' [skip CI]
git bisect good 64c0000c2a9385020e7f357711c0da3de4b03517
# status: waiting for bad commit, 1 good commit known
# bad: [7347a2be4188a7ec2b3b71cf620fd3e3267c78af] Merge remote-tracking branch 'origin/topic/robin/simplify-ast-tests'
git bisect bad 7347a2be4188a7ec2b3b71cf620fd3e3267c78af
# good: [0af29c57e43f1c1c17f888469d67849ff9200b5b] Add a new unit item `Block` that stores a sequence of subitems.
git bisect good 0af29c57e43f1c1c17f888469d67849ff9200b5b
# bad: [8bb969e1c4c7ca416c27d26fb1caa0c6e4d92dec] Simplify parsing of literals.
git bisect bad 8bb969e1c4c7ca416c27d26fb1caa0c6e4d92dec
# good: [a1f8334e85a9b7c843e2820542b8f82e94b34548] Merge remote-tracking branch 'origin/topic/robin/gh-1839-aggregate-if'
git bisect good a1f8334e85a9b7c843e2820542b8f82e94b34548
# skip: [a0b18dca4a426f972e0a0060a6a1f072b6fd00de] Refactor type parsing modes.
git bisect skip a0b18dca4a426f972e0a0060a6a1f072b6fd00de
# good: [96d6cb47a3c06cae81f23f32f986a9808f042841] Add mode for optimize types parsing.
git bisect good 96d6cb47a3c06cae81f23f32f986a9808f042841
# bad: [ca5de306a7c397c1d37aa7eb662dabafaa8ca26b] Optimize parsing for `bytes &size=N`.
git bisect bad ca5de306a7c397c1d37aa7eb662dabafaa8ca26b
# first bad commit: [ca5de306a7c397c1d37aa7eb662dabafaa8ca26b] Optimize parsing for `bytes &size=N`.
```

The timings for a `hyperfine -r 3 'taskset -c 1 zeek -C -r ./quic-16-50mb-transfers.pcap Log::default_writer=Log::WRITER_NONE'` run are as follows locally:

```
000c2a9385020e7f357711c0da3de4b03517 4.035 4.075

0af29c57e43f1c1c17f888469d67849ff9200b5b 4.078 4.054

8bb969e1c4c7ca416c27d26fb1caa0c6e4d92dec 4.318 4.344

a1f8334e85a9b7c843e2820542b8f82e94b34548 4.045 4.038

a0b18dca4a426f972e0a0060a6a1f072b6fd00de failed to build

96d6cb47a3c06cae81f23f32f986a9808f042841 4.033 4.021

ca5de306a7c397c1d37aa7eb662dabafaa8ca26b 4.269 4.221

7347a2be4188a7ec2b3b71cf620fd3e3267c78af 4.239
```

PCAP is available on the os-perf-1 host `/mnt/data/test_data/quic-16-50mb-transfers.pcap`.

rsmmr:
Turns out our improved error messages were adding additional overhead
because we were now constructing them through `fmt()` each time we
needed more data, independent of whether there was actually going to
be an error reported.

This adds a second version of `waitForInput()` that doesn't receive an
already prepared error message, but just returns false on error so
that the caller can throw itself.

Closes #1908.


evantypanski (approved):
Would it make sense to expand this more in the future so it doesn't come up again? There's a doc comment now, which is good, but more guarantees to prevent us from shooting ourselves in the foot could be nice in the future (not here though, probably)

Something like a lazy-evaluated `fmt` would be really cool. But probably not necessary :)

rsmmr:
> Would it make sense to expand this more in the future so it doesn't come up again? 

Yeah, good point. In fact I was trying to come up with something that would enforce correct usage, but couldn't think of anything viable that wouldn't beed larger changes. My two ideas were: (1) leveraging the type system by passing a `const char*` so that the result of `fmt()` wouldn't be accepted, and (2) passing a lambda instead of the error message, similar to your idea of a lazy-evaluated `fmt`. But (1) broke other stuff with generated code passing string_views; and (2) I think would itself be too expensive because it would now need to capture each time. Let me know if you have other ideas!

spicy/runtime/src/parser.cc
```diff
@@ -194,19 +208,24 @@ hilti::rt::Bytes detail::extractBytes(hilti::rt::ValueReference<hilti::rt::Strea
...
+    else if ( ! detail::waitForInputNoThrow(data, cur, size, filters) ) {
+        auto msg = hilti::rt::fmt("expected %" PRIu64 " bytes (%" PRIu64 " available)", size, cur.size());
+        throw ParseError(msg, std::string(location));
+    }
 
     return cur.sub(cur.begin() + size).data();
 }
 
 hilti::rt::Bytes detail::expectBytesLiteral(
     hilti::rt::ValueReference<hilti::rt::Stream>& data, const hilti::rt::stream::View& cur, hilti::rt::Bytes literal,
```
awelzel, line 220: Fly-by as I had mentioned this on the original issue: If IIUC this results in a string allocation for something that in generated code is just a constant. If the literal is large enough and short-string optimization not in effect, this results in a malloc() and copy (might be visible generating a parser with a longer and longer literal in use).

rsmmr: I saw the note but I don't think this has changed (didn't check but I thought we were creating an instance in the original code too); and it didn't seem to be the culprit, so didn't look further into it. I think this would fall into the "optimization" category: the general machinery expects a `Bytes` at runtime because that's how these values are represented, but yes, we should be able to optimize this specific case. If you think it'd be worth it, yes, file a new ticket and I'll look into it.

PR #1911, opened 2024-10-29
Optimize parsing of literal bytes.

rsmmr:
- **Introduce `UnsafeConstIterator` for bytes instances.**
- **Optimize parsing of literal bytes.**


rsmmr:
@awelzel How does this look to you? It gets the runtime for your examples down quite a bit. There's still a difference between the cases, and I'm actually not fully sure where exactly that is coming from (hard to pin down), but it can't be the literal instantiation anymore.

awelzel:
> I'm actually not fully sure where exactly that is coming from (hard to pin down), but it can't be the literal instantiation anymore.

Staring at flamegraphs, I believe It's from the return value of `expectBytesLiteral`. Now that this passes in a reference, the return `literal` needs to create a copy, including a potential allocation, while before it probably just moved out the `literal` that was provided.

Does expectBytesLiteral()` need a return value? I'm looking at the generated code and the return value isn't used, so maybe can just void it?

rsmmr:
> Does expectBytesLiteral()` need a return value? I'm looking at the generated code and the return value isn't used, so maybe can just void it?

Ah, good one, I actually did mean to get rid of that, but forgot. Let me try that.

awelzel:
I sent you one more patch idea in DM, but not sure it's applicable: It looked like there's a non-productive assignment within the generated code.


```
--- a/spicy/toolchain/src/compiler/codegen/parsers/literals.cc
+++ b/spicy/toolchain/src/compiler/codegen/parsers/literals.cc
@@ -104,8 +104,6 @@ struct Visitor : public visitor::PreOrder {
                     pb()->parseError("unexpected data when consuming token", n->meta());
                     popBuilder();
 
-                    builder()->addAssign(lp->destination(n->type()->type()), literal);
-
                     pb()->consumeLookAhead();
                     popBuilder();
```

I continue to see a difference between the 3 cases, but I'll open a separate ticket. It's kind of weird.

rsmmr:
> It looked like there's a non-productive assignment within the generated code.

That's non-productive only with `skip` I think, I'll add a condition to leave it out when skipping.

hilti/runtime/src/types/bytes.cc
```diff
@@ -40,7 +40,7 @@ std::tuple<bool, Bytes::const_iterator> Bytes::find(const Bytes& v, const const_
             if ( *x++ != *y++ )
                 break;
 
-            if ( y == v.end() )
+            if ( y == v.unsafeEnd() )
```
evantypanski, line 43: The variable names weren't changed here, but I spent way more time than I'm willing to admit just figuring out what this function is doing. If I read `View::_findBackward` first that would've been nice since those names are slightly more descriptive. I think it'd be good to touch that up at some point :)

spicy/toolchain/src/compiler/codegen/parsers/literals.cc
```diff
@@ -75,6 +75,14 @@ struct Visitor : public visitor::PreOrder {
     void operator()(hilti::ctor::Bytes* n) final {
         auto len = builder()->integer(static_cast<uint64_t>(n->value().size()));
 
+        auto literal = builder()->id(
+            hilti::ID(fmt("__bytes_%" PRIu64, hilti::util::uitoa_n(hilti::util::hash(n->value()), 16, 5))));
+
+        if ( ! pb()->cg()->haveAddedDeclaration(literal->id()) ) {
+            auto d = builder()->constant(literal->id(), builder()->expression(n));
+            pb()->cg()->addDeclaration(d);
+        }
```
evantypanski, line 84: This feels quite close to the plumbing for string interning without quite getting there, would it make sense to formalize that in an intern pool or something, then just look up declarations in that for this specific case? :)

or a dedicated `symbol` type? But that is out of scope here for sure

obviously I'm lacking some context, not sure if string interning is or has happened (or is happening and I can't find it formally), so let me know if so

evantypanski (approved):
I mostly have minor nits, I'll defer to @awelzel for the more in depth analysis with microbenchmarks since it seems like you're on that ;)

awelzel:
> I'll defer to @awelzel for the more in depth analysis with microbenchmarks since it seems like you're on that ;)

I had no further input other than what was provided in #1913 :-)

hilti/runtime/include/types/bytes.h
```diff
@@ -144,13 +156,83 @@ class Iterator {
...
+    friend bool operator!=(const UnsafeConstIterator& a, const UnsafeConstIterator& b) { return ! (a == b); }
+    friend auto operator<(const UnsafeConstIterator& a, const UnsafeConstIterator& b) { return a._i < b._i; }
+    friend auto operator<=(const UnsafeConstIterator& a, const UnsafeConstIterator& b) { return a._i <= b._i; }
+    friend auto operator>(const UnsafeConstIterator& a, const UnsafeConstIterator& b) { return a._i > b._i; }
+    friend auto operator>=(const UnsafeConstIterator& a, const UnsafeConstIterator& b) { return a._i >= b._i; }
+    friend auto operator-(const UnsafeConstIterator& a, const UnsafeConstIterator& b) { return a._i - b._i; }
+};
+
+inline std::string to_string(const UnsafeConstIterator& /* i */, rt::detail::adl::tag /*unused*/) {
+    return "<bytes iterator>";
```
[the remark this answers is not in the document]
rsmmr: This is user-facing, where "unsafe iterators" aren't a thing, so better to leave as is I think.

hilti/runtime/src/types/bytes.cc
rsmmr: Agree, this could use some cleanup (and some comments). I'll rename the function parameters to be a more descriptive, but will leave the rest to some other time.

spicy/toolchain/src/compiler/codegen/parsers/literals.cc
rsmmr: Admittably, I copied and adapted this from what we already do for regular expressions. Agree that it would make sense to add some infrastructure here, let me try that.

rsmmr: Done, @evantypanski have a look.

evantypanski (commented):
Hm, I'm sorry I missed this the first time, but I can actually break this pretty easily. I'll use a silly example just to prove the point, though. When I run this I will only ever get 256 constants, then some get reused, which ends up meaning that some constants are used in places they should not be (probably due to hash size limits or something).

Here's how I came to that conclusion:

This file just creates a big `Message` unit that skips fields with 1 to `num_iterations` `0`s:

```
# bigconsts.rb
f = File.new('a-lot-of-bytes.spicy', 'w')
f.write("module Test;\n")
f.write("type Message = unit {\n")

num_iterations = 500
(1..num_iterations).each do |i|
  f.write(": skip b\"%s\";\n" % ["0" * i])
end
f.write("on %done { print \"DONE\"; }")
f.write("};\n")
f.write("public type Messages = unit { : Message[]; };")

f.close()


# Make data file, covering each case exactly
dat = File.new('test.dat', 'w')
dat.write("0" * (1..num_iterations).sum)
f.close();
```

You can then roughly get the number of constants:

```
$ spicyc -c a-lot-of-bytes.spicy | grep ctor__bytes_....\ = | wc -l
     256
```

I'd expect that to equal `num_iterations`, and sure enough if I run this with the created `test.dat` file it gets an exception:

```
$ spicy-driver -f test.dat a-lot-of-bytes.spicy
DONE
[error] terminating with uncaught exception of type spicy::rt::ParseError: expected 61 bytes for bytes literal "0000000000000000000000000000000000000000000000000000000000000" (40 available)) (a-lot-of-bytes.spicy:63:8-63:71)
```

Same exact file, switching to `main`:

```
$ spicy-driver -f test.dat a-lot-of-bytes.spicy
DONE
```

So this causes a regression versus main for a program with 500 bytes constants.

Is that common? Not really, and this example is a little far fetched, but I think 256 is too few.

spicy/toolchain/src/compiler/codegen/parsers/literals.cc
evantypanski: I definitely prefer that, I think interning would be useful if we were comparing constants a lot and had a symbol type, but I don't think that'll ever be a common enough use case, thanks!

evantypanski (approved):
looks solid to me :)

PR #1912, opened 2024-10-30
Improve error message when a unit parameter is used as a field.

### LINKED ISSUE #1859: Cannot use unit parameters to specify a literal to parse
When parsing a field one can use e.g., `const` values to specify literals to expect, but using unit parameters in the same place does not resolve, e.g.,

```ruby
module foo;

type X = unit {
    : Y(0);
};

type Y = unit(expected: uint8) {
    : expected;
};
```

```console
$ spicyc -dj foo.spicy
[error] foo.spicy:8:5-8:15: unknown ID 'expected'
```

rsmmr:
Closes #1859.


evantypanski (dismissed):

spicy/toolchain/src/compiler/validator.cc
```diff
@@ -890,11 +891,18 @@ struct VisitorPost : visitor::PreOrder, hilti::validator::VisitorMixIn {
...
+            // Re-lookup ID to see if it exists at all.
+            if ( auto resolved = hilti::scope::lookupID<hilti::Declaration>(id, n, "field"); ! resolved )
+                error(resolved.error(), n, node::ErrorPriority::High);
+
+            if ( n->hasErrors() )
+                // Report existing error, probably from the resolver.
+                return;
+        }
+
+        // I don't think this can actually happen ...
```
evantypanski, line 904: Little nit: Would an assertion here make sense, just so a debug build would show if this case needs a bit more care? Now it's relying on `n->hasErrors()` as well, so other changes could technically cause this to have weird behavior that'd be nice for an assertion to catch IMO

evantypanski (approved):

spicy/toolchain/src/compiler/validator.cc
rsmmr: Let me make it an internal error in both release/debug, then we'll at least see for sure if/when it happens unexpectedly.

evantypanski (approved):

Issue #1901, opened 2024-10-18
Implement consistent attribute validation


rsmmr:
Our validation of which attributes are allowed where is currently a bit messy/spotty. I think it would be helpful to centralize this. Couple ideas:

- Right now attributes are identified as strings throughout the code, making it easy to typo them in validation code and elsewhere (like checking for `&byte_order` instead of `&byte-order`). It would be good to switch to enums instead: we'd translate the name of the attribute into an enum label right at parsing time and then carry that around. (Note that the Spicy lexer has a list of valid attributes already).

- I think we could define a central table specifying for each attribute in which contexts it's allowed to be used. Like: `&eod` is a "parse attribute", `&cxxname` is a "type attribute", etc.. Then the validation code could just ask at the corresponding locations: is this attribute I'm seeing here a type attribute as I would expect? It might take a bit of trial-and-error to find the right context categories (e.g., is it "parse attribute", or instead a combination "field attribute" & "unit attribute"? Also: Would we have separate categories for type-specific parse attributes, like `&ipv4` only applies to address fields, or would be continue to hardcode that distinction directly in the validator?) Also note that an attribute may apply to multiple context categories. But independent of the details, the idea would be centralize the knowledge about where an attribute it expected to been.

(Much of this applies to properties, too, btw).

evantypanski:
So looking at part 2 of this (with a [WIP commit](https://github.com/zeek/spicy/commit/a405f36f425e0362917f0eedf9871d23642e79fb)), I think minimally we could use a node's tag as some sort of key (or in a `switch` or something) in order to look up which attributes are allowed on.

I talked with both Robin and Benjamin and it seems like that's safe. We may lose some information for derived types - if that ends up having a noticeable impact then switching to using the whole `_node_tags` seems reasonable. I'll wait on that until there's a pressing need, though, because I'm not sure if there is one.

Doing this mapping isn't really complete. Some attributes may only be valid in some contexts, and we can't validate that just with a node's tag. I don't think there's a very easy way to get that full context though, since it's dependent on what node you're asking about. There are two important points here:

1) Fields are complex, there are a bunch of attributes that they accept. Then most of those attributes are context dependent. Just validating fields have the "field attributes" seems a bit crude. Maybe there will be another table that uses the field's attributes for the node, so that way you can have one table to verify questions like "Is this `&ipv4` attribute on an `addr` type within a field?" - separate from the question "Is this `&ipv4` attribute valid on a field?"
[...]

PR #1916, opened 2024-11-05
Move type alias attribute validation to parser

### LINKED ISSUE #1898: Disallow attributes on "type aliases" in parser
With #1890 we added a validator to reject any attributes on "type alias". To detect such type aliases we used a pretty nasty hack,

https://github.com/zeek/spicy/blob/500f2852f22e6dc01cd84ed717c0441c8483f47e/spicy/toolchain/src/compiler/validator.cc#L360-L363

We should clean this up by instead disallowing attributes on any declaration which is not a enum or unit (not sure there are more entities on the RHS where even with #1890 we support attributes). A way to do that would be to push the parsing of attributes from `type_decl` into e.g., `unit_type`; in fact the parser already has special handling for units, 

https://github.com/zeek/spicy/blob/500f2852f22e6dc01cd84ed717c0441c8483f47e/spicy/toolchain/src/compiler/parser/parser.yy#L381-L385

With that the added validator would become redundant.

evantypanski:
Closes #1898

This just moves the logic from #1890 to an allow list of types that can have attributes in the parser so that we avoid a hacky check. I don't really like checks in the parser like this very much, but since this fixes a weird validation I think it's ok.

The rest of this is just on what I chose *not* to do, which may be interesting depending what you think. That's all on the [`topic/etyp/cleanup-parsing-attributes`](https://github.com/zeek/spicy/tree/topic/etyp/cleanup-parsing-attributes) branch now.

My initial impression was that attributes should apply to types, then remove `opt_attributes` from the `type_decl`. That technically seems possible, but I'd hardly call it a cleanup. It essentially requires duplicating the attributes logic into many `UnqualifiedType` implementations (struct, enum, etc. but also like `exception`).

This goes into hilti world, so now the hilti parser needs changed, and once you get there all bets are off. You can use the `&on-heap` attribute on a lot of things, which *is* a declaration attribute. That's because what that means for a type is just that it becomes a `value_ref`. So now we think about these attributes on declarations as type attributes, which just doesn't intuitively make sense. I guess you could try to have both, but since there isn't a great distinction from syntax alone, it's hard.

There's a fair bit more, but I think the attribute changes should try to stay minimal *unless* we're going to really rework them, which I kind of advocated for in the issue this is closing... but that's a big change :)

As an added point: The branch I did that on had some positive consequences - tests like [`unsupported_attribute.spicy`](https://github.com/zeek/spicy/blob/main/tests/spicy/types/unit/unsupported-attribute.spicy) flag an error on an attribute, but the range given does not even encompass the attribute. I added a continuation of #1899 for unit's `setAttributes` for that case just as an added bonus.


spicy/toolchain/src/compiler/parser/parser.yy
```diff
@@ -383,6 +383,12 @@ type_decl     : opt_linkage TYPE scoped_id '='   { _docs.emplace_back(driver->do
                                                        u->setAttributes(builder->context(), $7);
                                                        $7 = {}; // don't associate with declaration
                                                        }
+                                                   // Type decls can have attributes, but only for certain types
```
bbannier, line 386: This is one way to do it and it gets rid of the hacky stringification which had to use in the parser before since we didn't have enough context. It was hoping for a way to have this encoded explicitly in the parsing rules, but I can see how that is a bigger and more tricky change due to tech debt in the parser and think this is still an improvement.

Nit: you matched the indention in this rule, but it seems it was already inconsistent with everything else we have in this file. WDYT about the following instead?

```cpp
                qtype opt_attributes ';'         { if ( auto u = $6->type()->tryAs<type::Unit>(); u && $7 && *$7 ) {
                                                      u->setAttributes(builder->context(), $7);
                                                      $7 = {}; // don't associate with declaration
                                                   }

                                                   // Type decls can have attributes, but only for certain types
                                                   if ( $7 && !$7->attributes().empty() ) {
                                                      auto ty = $6->type();
                                                      if ( ! (ty->isA<hilti::type::Struct>() || ty->isA<hilti::type::Enum>() || ty->isA<hilti::type::Bitfield>()) )
                                                          error(@7, "attributes are not allowed on type aliases");
                                                   }

                                                   $$ = builder->declarationType(std::move($3), std::move($6), std::move($7), std::move($1), __loc__);
                                                   $$->setDocumentation(_docs.back());
                                                   _docs.pop_back();
                                                 }
```

bbannier (approved):

spicy/toolchain/src/compiler/parser/parser.yy
evantypanski: yeah good call, honestly these Bison parsers go so far beyond the width of my editor that at all looks like a mess when programming anyway so it's tough to tell.

And don't worry, I was also hoping for that. I think depending on how #1901 gets implemented, it could become easier/better to make that change. I was going to delay until after that, but figured this solution is easy enough to put in before while it's on the top of my mind.

My other attempt at cleaning that up didn't exactly go smoothly :)

PR #1921, opened 2024-11-11
Fix some inconsistencies in random access docs.

### LINKED ISSUE #1920: Cannot subtract number from iterators
In the ["Changing input"](https://docs.zeek.org/projects/spicy/en/latest/programming/parsing.html#changing-input) section in the docs, there is a reference that seems like it implies you can subtract some number from an iterator:

> For random access, you’d typically get the current position through input(), subtract from it the desired number of bytes you want to back, and then use set_input to establish that new position.

So I tried `self.set_input(self.input() - 1);` and got:

```
[error] /Users/etyp/src/spicy-redis/analyzer/resp.spicy:25:28-25:43: unsupported operator: <const iterator<stream>> - <uint<64>
```

which seems exactly like the docs were telling me to do, so I got a little upset.

I don't see any `iterator - uint` ish operators anywhere - was this ever supported, or intended to be supported? I'd hardly call it "typical" if it's not supported :) so either that should be added or the docs changed.

But, I don't have the history of this to know which one it should be. I'd just prefer if you could subtract the number of bytes, that seems nice.

evantypanski:
Closes #1920


doc/programming/parsing.rst
```diff
@@ -1828,13 +1828,10 @@ control the current position:
...
 
-For random access, you'd typically get the current position through
-``input()``, subtract from it the desired number of bytes you want to
-back, and then use ``set_input`` to establish that new position. By
-further storing iterators as unit variables you can decouple these
-steps and, e.g., remember a position to later come back to.
-
-Here's an example that parses input data twice with different sub units:
+You can achieve random access by saving an iterator from ``input()``
+in a unit variable, then later return to that position by calling
```
rsmmr, line 1832: Yeah, this is good. The only thing is that now it seems to suggest that one *only* return to the position that `input()` returns. But moving the iterator forward (i.e., further into the unit), is fine and often more useful. How about this tweak to point that out:

```suggestion
in a unit variable, then later return to that position (or one derived from it) by calling
```

rsmmr (approved):

PR #1922, opened 2024-11-13
Bump Fedora versions in CI.

bbannier:


bbannier:
This seems to fail currently due to a [GCC false positive](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107532).

awelzel (approved):
In Zeek we build with werror disabled for Spicy, so don't think this needs a backport?

bbannier:
> In Zeek we build with werror disabled for Spicy, so don't think this needs a backport?

Nothing here changes functionality, and unless the expectation is that we build warning-free on newer distros this shouldn't be needed now. If it becomes an issue down the line we could still backport it to `release/1.11`.

PR #1923, opened 2024-11-19
Fix potential segfault with stream iterators.

### LINKED ISSUE #1918: Generated code segfaults when parsing `bytes` with `&until`
I can't actually provide a reproducer - the one I have is with production data and I haven't been able to narrow it down. But I do have *some* info.

For context, I'm getting this when running the [Redis analyzer on all TCP ports](https://github.com/evantypanski/spicy-redis/tree/topic/etyp/all-ports) 1-10000 for some big pcaps. So not an every day use case, but I wouldn't expect a segfault :)

The segfault occurs here:

https://github.com/zeek/spicy/blob/99049d80d29049b4142f356bc571c574f908a76d/hilti/runtime/src/types/stream.cc#L342

where that function is called when parsing [`RedisBytes`](https://github.com/evantypanski/spicy-redis/blob/14f851d3f08571eaa0fabf8e80751fed0e498dbb/analyzer/resp.spicy#L90-L93), namely in the `__parse_RESP__RedisBytes_6_stage2` function.

The segfault is [here](https://github.com/zeek/spicy/blob/99049d80d29049b4142f356bc571c574f908a76d/hilti/runtime/include/types/stream.h#L1010) when dereferencing `i`. Note that it *does* check `! byte` but the value is not `nullptr`. In fact, when I run `i->_chunk->_data`, it also is invalid memory in the debugger ("error: Cannot access memory at address 0xXX")

`first` in this case is `\r` - which is the first byte of the `&until` clause. I removed synchronization and still get the segfault.

This was from a `find` call in parser code (when searching for the `__until_bytes`)

That's basically all I have for here. Maybe it's a use-after-free of some sort?

I'll continue to try to narrow this down a bit, but for now it may be useful to have input. I've had weird issues with spicy segfaults on rebuilds in the past that get magically fixed when I rebuild, but this seems to be different (I've deleted & reinstalled many times and rebuilt Zeek in debug).

rsmmr:
When trimming off the beginning of a stream, an existing iterator
could end up dereferencing its internal chunk pointer even if the
chunk now no longer existed. The issue was inside the
increment/decrement operations, which didn't check if the *current*
iterator offset was still valid (because only then the current chunk
is guaranteed to be valid too).

Closes #1918.

### DIFF
--- a/hilti/runtime/include/types/stream.h
+++ b/hilti/runtime/include/types/stream.h
@@ -717,13 +717,22 @@ private:
         if ( ! n )
             return;
 
+        if ( _chain->isValid() ) {
+            const Chunk* hint = nullptr;
+
+            if ( _chain->inRange(_offset) )
+                hint = _chunk; // current chunk is still valid
+            else
+                hint = _chain->head(); // previous chunk was likely trimmed off, try new head
+
+            _chunk = _chain->findChunk(_offset + n, hint); // null if we're pointing beyond the end now
+        }
+        else
+            // Invalid chain will trigger exception when dereferenced, but
+            // invalidate chunk to be safe.
+            _chunk = nullptr;
+
         _offset += n;
-
-        if ( ! (_chain && _chain->isValid()) )
-            return; // will be caught when dereferenced
-
-        _chunk = _chain->findChunk(_offset, chunk());
-        // chunk will be null if we're pointing beyond the end.
     }
 
     void _decrement(const integer::safe<uint64_t>& n) {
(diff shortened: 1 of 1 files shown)


rsmmr:
@evantypanski can you see if this fixes the issue from #1918? It's now raining an exception (`view starts before available range`) though I'm not quite sure if that's semantically correct.

evantypanski:
Sure I'll take a look on my reproducers tomorrow, I'll be driving for the rest of the day :)

Thanks for looking at this, I figured it'd take me a little longer just to figure out the generated code

evantypanski:
@rsmmr this indeed fixes the use-after-free in the reproducers I have

rsmmr:
> @rsmmr this indeed fixes the use-after-free in the reproducers I have

Great, thanks for confirming. Is it now reporting an exception (`view starts before available range`) for you too? If so, are you able to tell if that's correct/expected behviour?

rsmmr:
> it's pretty tough for me to tell, but at least this doesn't seem to do anything bad and fixes the segfault so I see that as correct.

That sounds good enough to me. I'll get this ready for review & merge then. Thanks!

PR #1924, opened 2024-11-20
Validate that sets are sortable.

### LINKED ISSUE #1919: Missing validation that type in `set` is sortable
For maps we validate that its key type is sortable.

```ruby
module foo;

type X = struct {};

global m = map<X, uint8>(); # ERROR: type cannot be used as key type for maps (because type 'X' is not sortable)
```

We do not seem to perform the same validation of the `set` element type, e.g., the following fails only in C++ compilation

```ruby
module foo;

type X = struct {};

global s = set<X>();
```

```console
$ spicyc -j foo.spicy
...
/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/usr/include/c++/v1/__functional/operations.h:403:21: error: invalid operands to binary expression ('const __hlt::foo::X' and 'const __hlt::foo::X')
  403 |         {return __x < __y;}
      |                 ~~~ ^ ~~~
/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/usr/include/c++/v1/__tree:2031:28: note: in instantiation of member function 'std::less<__hlt::foo::X>::operator()' requested here
 2031 |     if (__hint == end() || value_comp()(__v, *__hint))  // check before
      |                            ^
/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/usr/include/c++/v1/__tree:2122:36: note: in instantiation of function template specialization 'std::__tree<__hlt::foo::X, std::less<__hlt::foo::X>, std::all
ocator<__hlt::foo::X>>::__find_equal<__hlt::foo::X>' requested here
 2122 |     __node_base_pointer& __child = __find_equal(__p, __parent, __dummy, __k);
      |                                    ^
/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/usr/include/c++/v1/__tree:1274:16: note: in instantiation of function template specialization 'std::__tree<__hlt::foo::X, std::less<__hlt::foo::X>, std::all
ocator<__hlt::foo::X>>::__emplace_hint_unique_key_args<__hlt::foo::X, const __hlt::foo::X &>' requested here
 1274 |         return __emplace_hint_unique_key_args(__p, _NodeTypes::__get_key(__v), __v).first;
      |                ^
/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/usr/include/c++/v1/set:806:25: note: in instantiation of member function 'std::__tree<__hlt::foo::X, std::less<__hlt::foo::X>, std::allocator<__hlt::foo::X>
>::__insert_unique' requested here
  806 |                 __tree_.__insert_unique(__e, *__f);
      |                         ^
/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/usr/include/c++/v1/set:706:13: note: in instantiation of function template specialization 'std::set<__hlt::foo::X>::insert<const __hlt::foo::X *>' requested
 here
  706 |             insert(__il.begin(), __il.end());
      |             ^
/opt/homebrew/Cellar/zeek/HEAD-e4dbba2/include/hilti/rt/types/set.h:130:39: note: in instantiation of member function 'std::set<__hlt::foo::X>::set' requested here
  130 |     Set(std::initializer_list<T> l) : std::set<T>(std::move(l)) {}
...
```

evantypanski:
Fixes #1919

~The baseline change for `ctor-fail` isn't ideal, and `map` indeed has a similar test that has a different error ("inconsistent key types in map"). It'd be nice if the error was similar. But since that check for sets is in `typeForExpressions` I think that duplicating that logic isn't necessary. It'd be trivial to just duplicate a check in the resolver for inconsistent types like [what map does](https://github.com/zeek/spicy/blob/d9c025d19a6f8a494500c6b0a732d8ae0ffc50d5/hilti/toolchain/src/compiler/resolver.cc#L612-L626), but there's not as pressing a need in my opinion.~


hilti/toolchain/src/compiler/validator.cc
```diff
@@ -631,6 +631,11 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
             error(fmt("integer type's width must be one of 8/16/32/64, but is %d", n->width()), n);
     }
 
+    void operator()(type::Set* n) final {
+        if ( auto rc = isSortable(n->elementType()); ! rc )
+            error(fmt("type cannot be used as element type for sets (because %s)", rc.error()), n);
```
bbannier, line 636: We trigger the change in the baseline of `hilti.types.set.ctor-fail` since for unresolved types `n->elementType()` is `type::Unknown` which is not sortable; the diagnostic is however not interesting to users since original error is elsewhere and lead to us not being able to compute an element type. The fix is probably to only report the error if we have a valid type.

```suggestion
        if ( ! n->elementType()->type()->isA<type::Unknown>() ) { // unknown will be reported elsewhere
            if ( auto rc = isSortable(n->elementType()); ! rc )
                error(fmt("type cannot be used as element type for sets (because %s)", rc.error()), n);
        }
```

While at it we should make a similar adjustment for the check on `type::Map` above.

bbannier (changes requested):

hilti/toolchain/src/compiler/validator.cc
evantypanski: I mainly just assumed that we'd want to catch unknown if it gets there without a ctor, but now I don't think I can make a case where that happens. So - agreed :)

if it does later it should be a separate error anyway

bbannier (dismissed):
Thanks! 🚀 

I'll add the suggested copypasta comments while merging.

evantypanski:
> I'll add the suggested copypasta comments while merging.

oh, oops

PR #1925, opened 2024-11-20
Fix inconsistency with `$$` in hooks.

### LINKED ISSUE #1914: `$$` is different in block versus in `&requires`
Found via https://github.com/bbannier/spicy-lint/pull/54

Given this program:

```
module Test;

public type Data = unit {
    x: uint8 &convert=($$ + 5) &requires=($$ < 50) {
        print $$ < 50;
    }
};
```

I would expect in to print `True` every time since the `&requires` succeeded, but instead it may print `False`:

```
$ echo '0' | spicy-driver test.spicy
False
```

I think `$$` should continue to refer to the pre-conversion value, personally, but instead it refers to the pre-conversion value in `&requires`, but after-conversion value in the block.

evantypanski:
Fixes #1914

`$$` says it refers to the pre-conversion value, but that's not true within a hook. This fixes that so that `$$` always refers to the pre-conversion value and type.

I have this as a draft because this breaks two tests - [`switch-convert-mixed.spicy`](https://github.com/zeek/spicy/blob/main/tests/spicy/types/unit/switch-convert-mixed.spicy) and [`switch-shared-id-constness.spicy`](https://github.com/zeek/spicy/blob/main/tests/spicy/types/unit/switch-shared-id-constness.spicy). Any time you have the same field name now, they must be the same parsed type, or compiling the hook will fail.

My personal opinion is it's more important to have `$$` be consistent than have a niche case covered. But, I'll defer to others, I'd be ok if this is just closed with no action because we're not willing to do that. I don't really see another good way to achieve consistency without breaking other stuff (maybe duplicating the hook decl for each field with different `$$` type and calling the correct one based on its type? that's probably possible).

If this continues then there should be nicer validation, right now the error is pretty unclear. Spicy errors when it can't find a matching overload to the hook call when compiling the second field's hook call.

Maybe @rsmmr any thoughts?


evantypanski:
Note, I think either way this needs a docs pass to be precise with the language. I'm going through with a fine-tooth comb, but that's because I think that imprecise language here could cause some really subtle bugs for people. For example:

> You can however still get to the parsed value inside a corresponding field hook using the reserved $$ identifier

This currently feels somewhat correct, it's specifying `$$` as a way to get to the parsed value, in contrast to using `self.x` for a field name. So, you can use `$$` as a drop-in replacement for `self.x`, which is true, but it is not necessarily what got parsed, it's what would have been stored.

> Inside field attributes and hooks, $$ refers to the just parsed value, even if it’s not going to be directly stored in the field. The value of $$ is writable and may be modified.

This is currently false, `$$` refers to the just parsed value in attributes but not hooks - putting them together in the same sentence makes it seem they act the same.

> The parsed value is accessible through the $$, potentially with any relevant type conversion applied. The same will also have been assigned to the field already.

This is currently true - and the exact inconsistency this PR fixes

> Inside [a `foreach`] hook, $$ refers to the just parsed element

It refers to the after-conversion value of what was just parsed.

-------------------------

Basically, whether this goes in or not, I'd really like to go through docs and be very clear what is pre-conversion and what is post-conversion. The language in the docs is fluid with what "parsed" means sometimes but not others. With this fix it's easy since `$$` is always pre-conversion, but without it, the docs need to be *very* precise to make sure it's clear... in my opinion :)

rsmmr:
I definitely see the confusion here and agree that either way we need to clarify the docs.

Also, if we started from a clean slate, I would agree with your proposal: making `$$` consistent makes sense of course, and it would actually fit together nicely to generally use `$$` for pre-converted and `self.field` for post-converted (except that with anonymous hooks, one now wouldn't be able to get to converted value at all anymore; but combining anonymous with `&convert` is probably questionable in the first place, so not a big deal).

That said, two things give me pause:

1. This would break backwards-compatibility, and I'm not sure how large the blast radius is there. It would be helpful if we could check out existing analyzers to see how often `$$` is used with fields that have a `&convert`; but not sure how easy that is to do.
2. I'm reluctant to break that `switch` use case of multiple fields having the same name but different parse types; I think it's a nice way to unify different parse types into a single joint representation. But as you say, we may be able to fix that even with the new semantics for `$$`; would need some investigation.

If we could figure out those two and conclude it's not a big deal, I'd be fine making the change.

evantypanski:
@rsmmr what do you think of something like this to answer those questions (before I go too far into something if it would never work):

1) Make a new keyword for hooks (say `__value` but I haven't thought about the name much). This will replace `__dd` as the first argument for a hook and always be present. It will keep the semantics as `__dd` is now.
2) Any cases in the hook have `$$` translated to `__value` instead of `__dd` - this will later be changed
3) Add different syntax to access `__value` (like `$*` or something, I really have no clue)
4) Add 2 warnings: one where `__dd` and `__value` differ in type, and one where `$$` is used with `&convert` in a hook. Suggest switching those to `$*` to keep semantics

let that sit for a while to see if there are tons of cases this causes issues. Then:

5) Revert point 2 - Stop translating `$$` to `__value` - it goes back to `__dd`
6) Only pass `__dd` if the hook uses it so that different parse types can share a hook if `$$` is never used

This answers question 2 by just not passing `__dd` unless it's used (so those cases just can't use it and get extra validation) and question 1 with nice deprecation warnings. This change could also just be aborted if those deprecations are too big of an issue. I really don't know how else to make it convenient, I don't think we can search all spicy parsers in any reasonable way. Maybe even make a script to translate `$$` to the new `$*` syntax in hooks for them.

So it's a lot more work for something pretty simple. Another option is to just start deprecate `$$` in hooks in favor of some new syntax, so it's clear that it's not the same as `$$` in attributes. That's certainly easier. I can't say I have any good use cases for accessing the pre-conversion value in a hook.

Any thoughts? Sorry about overblowing a small point, it's just really unintuitive for me so I wonder if others hit it and just don't say anything.

rsmmr:
Good thoughts. I need to mull over this a bit more, not quite sure yet about how to proceed, but I do see a couple of quick concerns with that proposal:

> 3. Add different syntax to access `__value` (like `$*` or something, I really have no clue)

If I understand correctly, this would stay, right? I'm reluctant to add more `$` identifiers, and particular if their semantics are close to something we already have (here: "value of parsed field", with slight difference in semantics). Worried about the user experience there, it reminds me too much of Perl. :-) 

> let that sit for a while to see if there are tons of cases this causes issues. 

Not sure whether we'd get this feedback before cutting a release. We can ask some folks to try it out, but my guess is that we'd run into more cases later once it's finalized, potentially even only quite a bit down the road.

> 6. Only pass `__dd` if the hook uses it so that different parse types can share a hook if `$$` is never used

This is unfortunately somewhere between tricky and really hard to implement. In the past, it would even have been an "impossible" because we used to allow compiling Spicy source files separately from each other and only later link them together. In that old model, the call site for the hook wouldn't have been able to tell if any of the hook implementations was using `$$`. These days, we require seeing all Spicy code at once (through the new single AST structure), so in principle we could make that decision if any of the hook implementations is making use of `$$`. However, the code isn't prepared for that kind of a late decisions and would need some changes (not sure how much, therefore that range from tricky to really hard :-)). Plus, generally I prefer to leaving late, global decisions to the final optimizer stage, and not sure if that would actually be a good place for this.

> Another option is to just start deprecate $$ in hooks in favor of some new syntax, so it's clear that it's not the same as $$ in attributes. That's certainly easier.

Yeah, certainly easier, but goes back to adding more special `$..`. Plus, use of `$$` is so common that I don't think deprecating it altogether is warranted for this fixing this inconsistency.

>  I can't say I have any good use cases for accessing the pre-conversion value in a hook.

Debugging is one use case, just printing it out for example. Or some maybe some additional error checking.

> Oh, and another option is abort this entirely and just make it very very clear in the docs what they mean. I am fine with that, but I like it the least - languages should really aim to have as little surprising subtle behavior as possible IMO

No worries, I totally get it, and I don't like the inconsistency myself. I appreciate all the thinking how we could improve this, open to more ideas or refinements!

As said above, not sure yet what's the best way to proceed. I think we could solve (2) through multiple hook implementations (could break existing code, too; but I'd be surprised if that's used much at all anyways). (1) is what continues to worry me.

evantypanski:
Got it, thanks for the feedback. I'm really glad this discussion is enshrined in Github for reference :)

> Not sure whether we'd get this feedback before cutting a release.

My intention was to have a release where the behavior change of `$$` is just a warning that it will happen, go a few of releases, then make the actual swap. Since I don't think there is any way to actually judge the impact of breaking backwards compatibility, that's the closest I got to approximating it.

> (1) is what continues to worry me.

Me too, I'm intentionally being a bit bullish about breaking changes here just to see if there's a compromise that can be made, but it's pretty tough.

I think for now I'll just close this PR and make the docs changes, then we can revisit this if we either come up with something or (long term) have some reason to make a big breaking changes release with the language.

PR #1926, opened 2024-11-22
Make `$$` documentation more precise.

### LINKED ISSUE #1914: `$$` is different in block versus in `&requires`

evantypanski:
Closes #1914

See #1925 for more context

The documentation would currently refer to `$$` in an attribute and hook with similar language (the "parsed" value). But, in an attribute, it refers to the value before any conversions. In a hook, it refers to the value after any conversions.

Now, "parsed" is used to refer to the pre-conversion value, and "final" is used to refer to the post-conversion value. Those terms were chosen because they make sense when no conversions are applied (or when `$$` is used *in* a conversion). But, "final" can still be a little confusing, since you can write to `$$` - any other ideas for another term for "final"?

### DIFF
--- a/doc/programming/parsing.rst
+++ b/doc/programming/parsing.rst
@@ -278,8 +278,8 @@ Anonymous Fields
 Field names are optional. If skipped, the field becomes an *anonymous*
 field. These still participate in parsing as any other field, but they
 won't store any value, nor is there a way to get access to them from
-outside. You can however still get to the parsed value inside a
-corresponding field hook (see :ref:`unit_hooks`) using the reserved
+outside. You can, however, still get to the field's final value inside
+a corresponding field hook (see :ref:`unit_hooks`) using the reserved
 ``$$`` identifier (see :ref:`id_dollardollar`).
 
 .. spicy-code:: anonymous-field.spicy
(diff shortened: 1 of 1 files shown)


doc/programming/parsing.rst
```diff
@@ -551,10 +561,10 @@ The most commonly used hooks are:
 ``on <field name> foreach { ... }`` (container hook)
     Assuming the specified field is a container (e.g., a vector), this
     executes each time a new container element has been parsed, and
-    just before it's been added to the container. The parsed element
-    is accessible through the ``$$`` identifier, and can be modified
-    before it's stored. The hook implementation may also use the
-    :ref:`statement_stop` statement to abort container parsing,
+    just before it's been added to the container. The element's final
+    value is accessible through the ``$$`` identifier, and can be
+    modified before it's stored. The hook implementation may also use
```
evantypanski, line 566: This is the main place that referring to `$$` as the "final" value is confusing - it says it's the element's final value, then says it can be modified before it's stored. Just a slight bit of confusion

doc/programming/parsing.rst
```diff
@@ -359,9 +359,19 @@ currently being parsed:
...
+    :ref:`attribute_convert`). This applies even if the value is not
+    going to be directly stored in the field. The value of ``$$`` is
+    writable and may be modified.
+
+.. note::
+
+   ``$$`` has slightly different semantics in a field attribute and
+   in a hook. In an attribute, ``$$`` refers to the parsed value
+   *before* any conversions. In a hook, ``$$`` refers to the final
+   value *after* any conversions.
```
evantypanski, line 374: This note is slightly redundant with the explanation in `$$`, but I think it's worth it to hammer the point home. It also introduces the difference in terms from "parsed" and "final" value.

doc/programming/parsing.rst
rsmmr, line 566: What do you think about this:

```suggestion
    value is accessible through the ``$$`` identifier, although it can be
    further modified before it's stored. The hook implementation may also use
```

rsmmr (approved):
This is good, thanks!

PR #1927, opened 2024-11-25
Put attributes in an enum

evantypanski:
This is the first bullet for #1901 - but there's a bit more to get the second bullet and this can stand alone. Figured I'd get comments on this before moving on.

Also sorry about the sloppy commits - they were WIP commits as I was changing a bunch of stuff then it became hard to make them pretty.

This PR just turns checking for attributes from a string to an `Attribute::Kind` enum. It should have no visible consequences on its own, but there were a few slight cleanups that had test changes:

1) Some validation was cleaned up (`spicy/.../validator.cc`). I grouped some validations that were the same but weirdly disjointed, as well as one that had a slightly different error (`&byte-order` without an expression). Grouping the errors made a slight behavior change in some cases where an attribute was invalid, but also had no expression - now it will yell about having no expression without yelling about being invalid (see `switch-attributes-fail.spicy`).
2) Attribute checks were moved from the parser to `Attribute` because we need it there anyway - now attributes are just a simple regex check (`\&[a-z][a-z0-9_\-]*[a-z0-9]`) when parsing. Then when creating `Attribute` objects, we just see if it's in the big map.
3) A couple of unused attributes were removed (`hilti_type` and `length`) - seems like this has no consequence?

Other than that, this looks a lot cleaner IMO. Not sure if I'm breaking some coding style nits you have though (namely the `static inline` map and an inner `enum` without `enum class`)

Also I'm a little confused about the hooks being used as attributes - like `%debug`, `%error`, and `foreach` - let me know if I did something silly there


hilti/toolchain/include/ast/attribute.h
```diff
@@ -16,9 +16,60 @@ namespace hilti {
 /** AST node for an attribute. */
 class Attribute : public Node {
 public:
-    /** Returns the name of the attribute, including the leading `&`. */
+    enum Kind {
```
bbannier, line 19: I think we definitely want to use an enum class here even though it is more wordy. We not only prefer them due to scoping which prevents namespace pollution, but also since an enum class plays nicer with the type system (classic enums are just integers).

hilti/toolchain/include/ast/attribute.h
```diff
@@ -96,15 +163,59 @@ class Attribute : public Node {
...
-        : Node(ctx, NodeTags, std::move(children), std::move(m)), _tag(std::move(tag)) {}
+    Attribute(ASTContext* ctx, Nodes children, std::string tag, Kind kind, Meta m = Meta())
+        : Node(ctx, NodeTags, std::move(children), std::move(m)), _tag(std::move(tag)), _kind(kind) {}
 
     std::string _dump() const override;
 
     HILTI_NODE_0(Attribute, final);
 
 private:
     std::string _tag;
```
bbannier, line 174: This should be redundant for any successfully parsing `Attribute` since we can recompute it from `_kind`.

hilti/toolchain/include/ast/attribute.h
```diff
@@ -68,6 +119,22 @@ class Attribute : public Node {
      */
     Result<bool> coerceValueTo(Builder* builder, QualifiedType* dst);
 
+    static bool isOneOf(Kind kind, const std::unordered_set<Kind>& kinds) { return kinds.count(kind) > 0; }
```
bbannier, line 122: nit: all users seem to be fine with passing a `std::initializer_list` (by value) which should be cheaper to construct.

hilti/toolchain/include/ast/attribute.h
```diff
@@ -96,15 +163,59 @@ class Attribute : public Node {
     }
 
 protected:
-    Attribute(ASTContext* ctx, Nodes children, std::string tag, Meta m = Meta())
-        : Node(ctx, NodeTags, std::move(children), std::move(m)), _tag(std::move(tag)) {}
+    Attribute(ASTContext* ctx, Nodes children, std::string tag, Kind kind, Meta m = Meta())
```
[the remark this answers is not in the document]
evantypanski: > I guess the part you might be least excited about would be that this requires pushing some validation into the parser again.

You know me well

Well it's at parse time either way, just not in the parser code. I'm happier with that suggestion than how it was (which was when lexing)

evantypanski:
[CI failures](https://cirrus-ci.com/task/5292620820250624?logs=build#L195-L224) seem like a [GCC issue](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=102921) from using the `constexpr initializer_list` in some capacity

I pushed a fixup commit for now that goes back to `map` but I'm open to other suggestions

hilti/toolchain/include/ast/attribute.h
```diff
@@ -68,43 +114,103 @@ class Attribute : public Node {
...
+        {"&bit-order", Attribute::Kind::BIT_ORDER},
+        {"&chunked", Attribute::Kind::CHUNKED},
+        {"&originator", Attribute::Kind::ORIGINATOR},
+        {"&responder", Attribute::Kind::RESPONDER},
+        {"&try", Attribute::Kind::TRY},
+        {"&needed-by-feature", Attribute::Kind::NEEDED_BY_FEATURE},
+        {"&requires-type-feature", Attribute::Kind::REQUIRES_TYPE_FEATURE},
+        {"&always-emit", Attribute::Kind::ALWAYS_EMIT},
+        {"&transient", Attribute::Kind::TRANSIENT},
+        {"&anchor", Attribute::Kind::ANCHOR},
```
bbannier, line 209: As an aside, with this PR we can now see that `&anchor` does nothing, but its use doesn't even trigger a warning. I filed #1928 to clean this up.

bbannier (dismissed):

rsmmr:
> 2. Attribute checks were moved from the parser to `Attribute` because we need it there anyway - now attributes are just a simple regex check (`\&[a-z][a-z0-9_\-]*[a-z0-9]`) when parsing. Then when creating `Attribute` objects, we just see if it's in the big map.

This is the way it is to deal with an unfortunate (if rare) parsing ambiguity. Assume a unit with a parameter `foo: uint64` and a field `f: &size=42 &foo`. This is a valid `&size` expression (`42 & foo`) but if we parse attributes as anything-with-a-leading-&, this will fail to parse as an expression. Hardcoding the valid attributes in the lexer solves this for the most part (we still have an issue if the parameter is called, say, `eod`; but that's even more rare and less confusing). I thought we actually had a test for this, but maybe we don't. 

Note that `%..` properties have the same issue, that came up recently when we introduced `%error`.

bbannier:
> > 2. Attribute checks were moved from the parser to `Attribute` because we need it there anyway - now attributes are just a simple regex check (`\&[a-z][a-z0-9_\-]*[a-z0-9]`) when parsing. Then when creating `Attribute` objects, we just see if it's in the big map.
> 
> ... I thought we actually had a test for this, but maybe we don't.

Yeah, was wondering about this as well. I checked this with the following constructs and they both seemed to work as expected.

```ruby
global convert = True;

public type X = unit {
    f: uint8[] &size=4&convert=False;
};
```
```ruby
public type X = unit {
    data: bytes &eod;
    switch {
        -> f: uint8[];
    } &size=4&parse-from=self.data; # parses successfully, but `&parse-from` is ignored.
};
```

evantypanski:
> This is the way it is to deal with an unfortunate (if rare) parsing ambiguity.

I thought of that but didn't have a concrete example. I think that if it was just a parsing ambiguity I'd try to push this through, but it's lexing (which I see as worse). So any time `&` is followed by some identifier that could be an attribute, it's invalid. That means this code is invalid after this change, but not before:

```
module Test;

function f() {
    local foo = 9;
    42&foo;
}
```

If it only applied within another attribute, I think that's ok since you can just put spaces around it. But this case feels weird to a user and gives an unhelpful error (`syntax error, unexpected attribute, expecting ';'`)

It'll still be weird with the list of attributes in the lexer, since this also fails if that local is `eod` rather than `foo` (or any valid attribute) as you mention.

For now, I'll add that back in the lexer for that ambiguity, but I think that we should try to come up with a way to move that out, then also allow attribute names as variables in that context. Small issue - I don't really know how :)

tests/Baseline/spicy.types.vector.legacy-syntax-fail/output
```diff
@@ -1,6 +1,10 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
 [error] <...>/legacy-syntax-fail.spicy:7:8-7:20: vector<T> syntax is no longer supported for parsing sequences; use T[] instead.
+[error] invalid attribute '&length'
```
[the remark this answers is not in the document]
evantypanski: well less so since we had to add back validation in the lexer :( hopefully that can be resolved another way so that attributes can be validated. The reason this error will be staying is that the lexer allows it but `Attribute` doesn't have a `Kind` for it - which is probably how deprecated attributes should act.

Unfortunately the error for trying to use an attribute that doesn't exist like `: uint8 &omg` is really unhelpful still, since it doesn't even count it as an attribute:

```
[error] test.spicy:4:13-4:13: syntax error, unexpected '&'
```

So a minor win now (it's nice that we can accept some attributes in the lexer then reject it with a nicer error if we know about them). It could be a bigger win if we find a better way to solve the attribute vs. bitwise `&` problem.

hilti/toolchain/include/ast/attribute.h
evantypanski: do you have reference on when this was valid or its utility removed? I didn't really understand what it was, it was one of those attributes that never got caught in tests except for using it internally a few times (I think like some of those `needed-by-feature` like ones, but those are more obvious)

bbannier: No, I do not recall what it was originally used for. For above issue I actually removed all mentions of `&anchor` from the code base, and the test suite still passes with that. ☹️

Nothing for you to do here, so resolving.

bbannier (approved):
This looks good to me, thanks for sticking along. Could you squash together the changes?

In you last PR you undid you change to `spicy/toolchain/src/compiler/parser/scanner.ll` and added back parser support for the `&hilti_type` and `&length`; since they are unused let's drop them either with a separate cleanup patch here or as a follow-up.

evantypanski:
rebased&squashed it

I like the nicer error with `&length` saying it's an invalid attribute - that's a consequence of it lexing the attribute but then it's not in `Attribute::Kind` so it gets rejected. That seems nice to me for deprecated or invalid attributes for now to me - it's just an extra case in the lexer so it's not really tech debt IMO. If that's completely off base we can change it though, it'll just make the `&length` test error a little worse

PR #1929, opened 2024-11-26
Allow custom error messages with `&requires`

rsmmr:
This is experimental, both in terms of (1) do we want this, and (2)
what about the syntax? Open for discussion here.

The PR creates a new operator `COND !? ERROR` ("or-error") where
`COND` is a boolean expression, and `ERROR` is an expression of type
`error`. When evaluated, the new operator yields a value of type
`result<bool>`, which will be set to true if the `COND` is true, and
set to `ERROR` if `COND` is false. In other words, this is a short-cut
to both test a condition and provide an error message in case it
fails. The PR then adapts `&requires` to accept such or-error
expressions so that one can carry a custom error message through
there.

- **Add new "or-error" operator.**
- **Teach `assert` to accept an `result<bool>` value.**
- **Teach `&requires` to accept an or-error expression.**


evantypanski:
I like this a lot, I think it's really useful to have a message for these. So for me, my answer to (1) is a resounding yes. It being syntactic sugar for `result<bool>` is also clever.

I have a couple overarching points, since I probably have too many opinions about syntax.

1) I don't like the syntax `!?` - it feels arbitrary and if I saw that in code there's no way I'd know what it does without looking it up. What about using a keyword? I went in and replaced it with `else` and really like the look, and it doesn't seem to cause any new problems:

```
    y: uint8 &convert=E($$) &requires=($$ == 1 else error"trouble!'");
```

```
assert 1 == 0 else error"my error";
```

if overloading `else` is an issue, maybe `orelse` makes sense too. It just feels more natural (to me) when reading. The main benefit here is that anyone reading it has a better shot of understanding without having to look up each syntax element in the documentation.

2) Naming this `OrError` expression seems tough, so I won't claim to have a better name. But, a couple of other possibilities: `ConditionalResult` or maybe `BoolErrorUnion`?

Overall this seems like a nice step, since to the user `&requires` feels a lot like an assertion. It's a shame that syntax like the `: "my error message"` that `assert` has doesn't seem feasible.

awelzel:
> It's a shame that syntax like the : "my error message" that assert has doesn't seem feasible.

For a very naive outsider, why does `:` not work? Or even `,` ? (Mostly with Python's `assert` as motivation).

```
&requires=($$ >= 5, "expected at least 5, got %s" % $$)
```

Even if `,` or `:` would force users to add `()` in certain cases `requires=((expr...), "complicated requires failed")` that seems to look less cryptic.

bbannier:
Like Evan and Arne I am not a fan of the new syntax, and think sticking with something similar to what we have for `assert` would likely be less surprising. If this needs extra parentheses it would also help readability.

This all seems related to `result`, but the difference is that we do not care about the contained value (and we do not have an `Nothing` type to model this), so I don't think adding extra syntax to construct such `result` values is worth it. I picture this more as having an "error check expression" `expr [ ':' msg ];` where at runtime we construct an `optional<Error> failure_msg` (`Error` is not exposed in Spicy) and check that it is empty or else throw an exception with its contents (for `assert` an `AssertionFailure`, for `&requires` a `ParseError`). Both `assert` and `&requires` currently take an expression which can convert to a `bool`, but that `failure` value would convert the wrong way. We might be able to add overloads to `&requires` and for cleanup `assert` which take an `optional<Error>` though since users cannot write `Error` (we also do not have `auto` in Spicy) and there should be no trouble with existing code returning `optional<Error>`; we could make that overload convert the way we want.

rsmmr:
Thanks for the feedback, very helpful.

As you're all advocating for reusing `:`, I gave that another try--it's my preferred solution too. Couple things: 

- The main problem is that our syntax uses `:` inside/near expressions elsewhere, which makes parsing tricky; and I originally thought that's not easily fixable. However, I think I found a solution now that works; it's bit of a hack in `parser.yy`, but that might be worth it.

- The existing `assert COND : STRING` doesn't directly work anymore because what's a STRING  so far now needs to be of a type `error`. But we can make that work through automatic coercion: if the RHS is a string, coerce into an error. I was originally reluctant to add that coercion because I like the explicit `error"XXX"`  syntax, that feels more readable/clear; but on second thought, I think I can live with the coercion if it allows us to reuse `:`.

So I just added a commit making these changes, see what you guys think

> I went in and replaced it with else and really like the look

Nice thought, I like `else` too, with one catch: `else` sounds to me like it should support any type on the RHS, not just `error`--which would then lead to some more complications I believe.

> Naming this OrError expression seems tough, so I won't claim to have a better name. But, a couple of other possibilities: ConditionalResult or maybe BoolErrorUnion

`CheckedBool`, maybe?  But not sure. Left this as is for now.

> I picture this more as having an "error check expression" expr [ ':' msg ]; where at runtime we construct an optional<Error> failure 

I made the changes before seeing this comment. Need to think about it, but gut feel is that I like `result<bool>` more than `optional<Error>` because it seems semantically to "go the right way" (we expect the condition to be true).

>  : void &requires=result(False); # Should not pass.

This is I think we can make work correctly in the current approach, too, need to try,

bbannier:
> > I made the changes before seeing this comment. Need to think about it, but gut feel is that I like result more than optional because it seems semantically to "go the right way" (we expect the condition to be true).
> 
> Oh, maybe `result<void>` is what we really want?

And we [have that already](https://github.com/zeek/spicy/blob/14fa7dc4efdcb9ec2529763c4e772302974ea7e4/CHANGES#L1441) which I totally missed.

tests/spicy/types/result/ops.spicy
```diff
@@ -31,3 +31,9 @@ import spicy;
 global err: spicy::Error;
 err = error"abc";
 assert err.description() == "abc";
+
+global x = True : error"my-error";
+assert x;
```
bbannier, line 36: I like how this makes checks with dedicated error messages reusable, and the parser currently even accepts a plain `string` via implicit conversions which I like even better,

```ruby
global reality_check = True: "oh no!";
assert reality_check;
```

This works with string literals, but fails with runtime strings, do you think we could make this work?
```ruby
global oh_my = "oh no!";
global reality_check = True: oh_my; # ERROR: type of error expression must be 'error', but is string
assert reality_check;
```

By not allowing this we break existing code which we should continue to support (it will likely be even better with the new potential for reuse):

```ruby
global x = 42;
assert x % 2 == 0 : ("%d expected to be even" % x);
```

evantypanski:
> For a very naive outsider, why does : not work? 

Well I figured it would work, but cause some ambiguities for the user/grammar. I was thinking the same thing Robin mentioned:

> The main problem is that our syntax uses : inside/near expressions elsewhere, which makes parsing tricky

I think this could work like how Robin did it, though. I'd prefer it the opposite direction: instead of `expr_no_or_error` there's an `expr_or_error` that is used in places it's expected (maybe assert and in attributes). That way the ambiguity is at least less likely. Then you don't need a special case for `map_elem` etc.

I still worry a bit about using a symbol `:` that appears in/near multiple other expressions in such a general context (that is, in `expr` rather than its own production). Narrowing down the specific contexts that the `OrError` expression may appear in would be useful just to minimize the impact down the road. Concretely, that'd just mean making an `expr_or_error` production that is only allowed in assertions and attributes

I think I prefer the workarounds with `:` to some very specific syntax to Spicy like `!?`. I worry a bit about maintaining the parser - but that's hypothetical.

The benefit of a keyword like `else` or `orelse` that isn't already in expressions is just that it's completely unambiguous - `else` is used in exactly 2 other places in the spicy grammar, neither of which will have an ambiguity with an expression before a semicolon. So that's my preference, but the `:` route seems reasonable to be aligned more with user's expectations. Using `:` may warrant a bit more caution IMO

> `CheckedBool`, maybe? But not sure. Left this as is for now.

I like that a bit better, but it's pretty tough to narrow this down to a good name :/ but I may be the only one that cares about that name so it's not a big deal

Overall using `:` seems a bit nicer than I imagined in the parser. But, I can still picture times down the road that it could cause a headache if it's not limited.

tests/spicy/types/result/ops.spicy
rsmmr: > This works with string literals, but fails with runtime strings, do you think we could make this work?

Yes, will do. I actually left it out on purpose but didn't think about `%`.

rsmmr:
> an `expr_or_error` that is used in places it's expected (maybe assert and in attributes).

That makes its usage quite a bit less consistent, though. In general, there's nothing really wrong with using that new `:` thingy wherever an expression is expected (think just `print` for example). Limiting it to just a couple places feels surprising to me. In contrast, for the few cases where I am currently using `expr_no_or_error`, it wouldn't semantically fit there anyways. 

> I think I prefer the workarounds with `:` to some very specific syntax to Spicy like `!?`. I worry a bit about maintaining the parser - but that's hypothetical.

I  worry about maintaining the parser in general  ... It has accumulated a number of shortcomings these days, in particular how one cannot (easily) narrow down some syntax/tokens to a certain context. The real solution here is a rewrite, which is overdue, but also a lot of work for not that much immediate benefit. If we did that, I'm pretty sure we'd abandon Bison/Flex. We could either just write a new parser manually (which I'm tempted to do do), or go the treesitter route (which Benjamin would prefer, and he's probably right :-). Either way, bottom-line is that I wouldn't worry too much about the current parser becoming worse.

rsmmr:
So I switched to `result<void>` and cleaned up the commits. It's still called `OrError` internally, still mulling over that (`ConditionTest` is my current favorite ...). Docs are missing still too. But sounds like this is conceptually all agreeable?

bbannier:
> So I switched to `result<void>` and cleaned up the commits. It's still called `OrError` internally, still mulling over that (`ConditionTest` is my current favorite ...). Docs are missing still too. But sounds like this is conceptually all agreeable?

I like how this looks in code, and have no strong feelings around naming this `OrError`. The test suite currently uses error literals instead of just strings (literal or not), and I still do not believe `error"..."`  will be the default way this will be used or should be documented (I'd strongly prefer just using strings).

For codegen we currently seem to expand whatever was in the `result<void>` into the code generated for e.g., `assert`. I haven't looked carefully what this means for runtime strings, but it definitely might increase the size of the code if this gets used a lot; we might be able to instead have `assert` at runtime convert the `OrError` value  to `bool` and if needed pull out the error via `error()` (not exactly sure how this would look like for temporary `OrError`).

rsmmr:
> we might be able to instead have `assert` at runtime convert the `OrError` value (which is just a `result<void>`) to `bool` and if needed pull out the error via `error()` (not exactly sure how this would look like for temporary `OrError`; it probably would also require constructing a `result` value for instances without user message).

Not quite sure what you mean. This is the code I'm currently getting for an assert with error, which looks like what you're describing? (It is longer than before, though, mostly because of the result machinery):

``` 
if ( auto result = (::hilti::rt::integer::safe<std::uint64_t>{4U} == ::hilti::rt::integer::safe<std::uint64_t>{5U} ? ::hilti::rt::make_result(::hilti::rt::Nothing{}) : ::hilti::rt::Result<::hilti::rt::Nothing>(::hilti::rt::result::Error("XXX"))); ! result ) {
        throw ::hilti::rt::AssertionFailure(::hilti::rt::to_string_for_print(result.error().description()), "a.spicy:3:1-3:22");
    }
```

For an assert without error, I'm getting identical code than before.

bbannier:
> Not quite sure what you mean. This is the code I'm currently getting for an assert with error, which looks like what you're describing? (It is longer than before, though, mostly because of the result machinery):

You are right, I think I got confused when looking at the AST.

rsmmr:
Renamed to ConditionTest, added docs, ready for review.

tests/spicy/types/unit/requires-attribute.spicy
```diff
@@ -17,7 +17,7 @@ type E = enum { X = 1, Y = 2 };
 
 public type Foo = unit {
     x: uint8 &requires=($$ == 1);
-    y: uint8 &convert=E($$) &requires=($$ == 1);
+    y: uint8 &convert=E($$) &requires=($$ == 1 : error"trouble!'");
```
bbannier, line 20: nit:

```suggestion
    y: uint8 &convert=E($$) &requires=($$ == 1 : error"trouble!");
```

hilti/toolchain/src/compiler/parser/parser.yy
```diff
@@ -508,10 +508,10 @@ stmt          : stmt_expr ';'                    { $$ = std::move($1); }
               | TRY block try_catches
                                                  { $$ = builder->statementTry(std::move($2), std::move($3), __loc__); }
               | ASSERT expr ';'                  { $$ = builder->statementAssert(std::move($2), {}, __loc__); }
-              | ASSERT expr ':' expr ';'         { $$ = builder->statementAssert(std::move($2), std::move($4), __loc__); }
-              | ASSERT_EXCEPTION expr ';'        { $$ = builder->statementAssert(hilti::statement::assert::Exception(), std::move($2), {}, {}, __loc__); }
-              | ASSERT_EXCEPTION expr ':' expr ';'
+              | ASSERT_EXCEPTION expr_no_or_error ':' expr ';'
                                                  { $$ = builder->statementAssert(hilti::statement::assert::Exception(), std::move($2), {}, std::move($4), __loc__); }
+              | ASSERT_EXCEPTION expr_no_or_error ';'
```
bbannier, line 513: We could still have `expr_no_or_error` contain a `result<void>` and by using this constructor we would report an incorrect error, e.g.,

```ruby
global x = 0 : "foo";
assert x;
```

```console
$ spicyc -j foo.spicy
uncaught exception hilti::rt::AssertionFailure: failed expression 'x' (foo.spicy:15:1-15:9)
```

Part of this comes from the fact that `Assert` can hold either an expression and a message, or a `result<void>` and no message. To use `Assert::message` one needs to check the type of the expression, but this would also be error prone like e.g., with this patch

```ruby
global x = result("foo");
# Should work since a result did convert implicitly to a bool.
assert x : "but I initialized it!"; # ERROR: cannot coerce expression 'x' of type 'string' to type 'bool'
```

I think this needs cleanup in `Assert`.

hilti/toolchain/src/compiler/validator.cc
```diff
@@ -396,6 +396,14 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
...
     }
 
+    void operator()(expression::ConditionTest* n) final {
+        if ( ! n->condition()->type()->type()->isA<type::Bool>() )
+            error(fmt("type of condition must be boolean, but is %s", *n->condition()->type()), n);
+
+        if ( ! n->error()->type()->type()->isA<type::Error>() )
+            error(fmt("type of error expression must be 'error', but is %s", *n->error()->type()), n);
+    }
+
```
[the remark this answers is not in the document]
rsmmr: Good point, that can go.

hilti/toolchain/src/compiler/parser/parser.yy
rsmmr: > global x = 0 : "foo";
> assert x;
> 
> $ spicyc -j foo.spicy
> uncaught exception hilti::rt::AssertionFailure: failed expression 'x' (foo.spicy:15:1-15:9)

This seems correct, no? `0 : "foo"` gets the error condition because 0 coerces to false. And then the assert does a boolean check on its `result<void>` argument and finds it false, hence it triggers.

> global x = result("foo");
> assert x : "but I initialized it!"; # ERROR: cannot coerce expression 'x' of type 'string' to type 'bool'

The bug here seems elsewhere: `x` ends up having a type of string. If I change the first line to `global x: result<string> = result("foo");` it works. Will look into that.

rsmmr: > This seems correct, no?

Or are you saying it should print the message?

rsmmr: > Or are you saying it should print the message?

Ah, and if so, it's actually the same issue as the 2nd. This works:

```
global x: result<void> = 0 : "foo";
assert x;
...
uncaught exception hilti::rt::AssertionFailure: foo (b.spicy:3:1-3:9)
```

spicy/toolchain/src/compiler/resolver.cc
```diff
@@ -128,6 +128,35 @@ struct VisitorPass2 : visitor::MutatingPostOrder {
...
+                                       builder()->qualifiedType(builder()->typeBool(), hilti::Constness::Const));
+            if ( ! ne.coerced ) {
+                n->addError(ne.coerced.error());
+                return;
+            }
+
+            // Implicitly create an error message from the condition itself.
+            std::stringstream ss;
+            ss << *cond;
+            auto msg = hilti::util::fmt("&requires failed: %s", hilti::util::replace(ss.str(), "__dd", "$$"));
```
bbannier, line 154: We can directly stringify the node and avoid the `stringstream`,

```suggestion
             auto msg = hilti::util::fmt("&requires failed: %s", hilti::util::replace(cond->print(), "__dd", "$$"));
```

spicy/toolchain/src/compiler/validator.cc
```diff
@@ -623,9 +623,16 @@ struct VisitorPost : visitor::PreOrder, hilti::validator::VisitorMixIn {
...
-                      e && ! hilti::type::same((*e)->type()->type(), builder.typeBool()) )
-                error(fmt("&requires expression must be of type bool, but is of type %d ", *(*e)->type()), n);
+            else {
+                auto e = *n->valueAsExpression();
+                assert(e);
+
+                if ( auto result = e->type()->type()->tryAs<hilti::type::Result>();
+                     ! result || ! result->dereferencedType()->type()->isA<hilti::type::Void>() )
+                    error(fmt("&requires expression must be of type bool or result<void>, but is of type %d ",
+                              *e->type()),
```
bbannier, line 633: Nit: we don't really care about the value, but its stringification.

```suggestion
                              e->type()->print()),
```

spicy/toolchain/src/compiler/validator.cc
```diff
@@ -623,9 +623,16 @@ struct VisitorPost : visitor::PreOrder, hilti::validator::VisitorMixIn {
...
-            else if ( auto e = n->valueAsExpression();
-                      e && ! hilti::type::same((*e)->type()->type(), builder.typeBool()) )
-                error(fmt("&requires expression must be of type bool, but is of type %d ", *(*e)->type()), n);
+            else {
+                auto e = *n->valueAsExpression();
+                assert(e);
+
+                if ( auto result = e->type()->type()->tryAs<hilti::type::Result>();
+                     ! result || ! result->dereferencedType()->type()->isA<hilti::type::Void>() )
+                    error(fmt("&requires expression must be of type bool or result<void>, but is of type %d ",
```
bbannier, line 632:
```suggestion
                    error(fmt("&requires expression must be of type bool or result<void>, but is of type %d",
```

spicy/toolchain/src/compiler/validator.cc
```diff
@@ -691,8 +698,10 @@ struct VisitorPost : visitor::PreOrder, hilti::validator::VisitorMixIn {
                     if ( ! e )
                         error(e.error(), n);
                     else {
-                        if ( ! hilti::type::same((*e)->type()->type(), builder()->typeBool()) )
-                            error(fmt("&requires expression must be of type bool, but is of type %s ", *(*e)->type()),
+                        if ( auto result = (*e)->type()->type()->tryAs<hilti::type::Result>();
+                             ! result || ! result->dereferencedType()->type()->isA<hilti::type::Void>() )
+                            error(fmt("&requires expression must be of type bool or result<void>, but is of type %s",
+                                      *(*e)->type()),
```
bbannier, line 704: Nit, similar to above.

```suggestion
                                      e.value()->type()->print()),
```

bbannier (dismissed):
Thanks, I think this looks great. 🚀

I left mostly smaller nit comments and suggestions for additional tests.

hilti/toolchain/src/compiler/codegen/coercions.cc
```diff
@@ -134,6 +133,11 @@ struct Visitor : public hilti::visitor::PreOrder {
                 fmt("codegen: unexpected type coercion from %s to %s", *n, dst->type()->typename_()));
     }
 
+    void operator()(type::String* n) final {
+        if ( dst->type()->isA<type::Error>() )
+            result = fmt("::hilti::rt::result::Error(%s)", expr);
```
bbannier, line 138: Nit: consistency.

```suggestion
            result = fmt("%s(%s)", cg->compile(dst, codegen::TypeUsage::Storage), expr);
```

hilti/toolchain/src/compiler/codegen/expressions.cc
```diff
@@ -215,6 +215,13 @@ struct Visitor : hilti::visitor::PreOrder {
         result = {cxx::ID(n->id()), Side::LHS};
     }
 
+    void operator()(expression::ConditionTest* n) final {
+        auto type = cg->compile(n->type()->type()->as<type::Result>()->dereferencedType(), codegen::TypeUsage::Storage);
+        result = fmt(
+            "(%s ? ::hilti::rt::make_result(::hilti::rt::Nothing{}) : ::hilti::rt::Result<::hilti::rt::Nothing>(%s))",
```
bbannier, line 221: Nit: we want an `error` on the `else` branch and could directly get it out of `n` (`error()` already returns a correctly coerced result).

```suggestion
            "(%s ? ::hilti::rt::make_result(::hilti::rt::Nothing{}) : :%s)",
```

bbannier (commented):
Sorry, forgot the actual nits 😬

evantypanski (approved):
I'm a fan of this :)

I still slightly worry about the syntax colliding here even if the parser could get rewritten, just because it'd be hard to replicate this behavior in all cases without a GLR parser (and I always prefer handwritten parsers). But, I think practically that'll be a nonissue since I don't see this being used outside of specific contexts that it's unambiguous.

spicy/toolchain/src/compiler/validator.cc
rsmmr: The `*`-style is how it's done almost everywhere (mostly for historical reasons) so staying with that.

spicy/toolchain/src/compiler/validator.cc
rsmmr: Same as above.

PR #1931, opened 2024-11-28
Spell all `Attribute::Kind` values in Pascal case.

bbannier:
We are not 100% consistent, but we seem to prefer PascalCase over SNAKE_UPPERCASE for enum values. Rename the recently introduced `Attribute::Kind` values to that style.

An intended side effect of this is that we now avoid collisions with a possibly defined `DEBUG` macro (e.g., Zeek defines it).


evantypanski (approved):
Oh oops

Also has a side effect of removing the trailing semicolon from `BYTE_ORDER_` which was there for a similar reason

PR #1933, opened 2024-12-03
Fix `HILTI_CXX_FLAGS` for when multiple flags are passed.

bbannier:


evantypanski (commented):
One thing to note: This breaks a previously working example of arguments with a space, like:

```
HILTI_CXX_FLAGS="-DRETURN_VALUE=\"Hi there\""
```

If multiple arguments are allowed it has to tokenize to actually work "properly"

I'll leave that up to others to say if it's a deal breaker or not, though. I'm just pointing out that behavior change (that example works in current `main`)

bbannier:
> One thing to note: This breaks a previously working example of arguments with a space, like:

Good point, I did not think about that. I personally would be fine with breaking users, but then the implementation should work consistently, which it currently does not. To work it probably needs to be aware of shell escaping (e.g., POSIX), e.g., all of these should produce the same results `{"-I/foo", "-I/bar"}`,

```
-I/foo -I/bar
-I/'foo' -I/bar
-I/"foo" -I/bar
```

Using a different delimiter like `:` likely has similar issue in how it could appear in quoted portions.

rsmmr (commented):
Not sure what to do here. I see the issue but breaking arguments with spaces isn't great either. Wondering if we could find some code somewhere that implements the splitting in an sh-aware manner?

bbannier:
I tried to search for this, but could not come up with much (maybe because search engines seem to prefer to show me trivial string splitting results). I looked in CMake source code and they [implement this themselves](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Source/kwsys/System.c?ref_type=heads#L94) (in C). We could shell out and [let `sh` do the splitting](https://stackoverflow.com/a/19932085/176922), but this seems ridiculous.

evantypanski:
I've seen this exact issue a few times, the easiest solution always ends up being just coding in the splitting logic by hand.

I think it gets a bit hairier with Windows and response files, but I'd imagine dealing with that isn't a priority :)

bbannier:
I have reworked this to be on top of [`wordexp`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/wordexp.html) which is part of POSIX. This function passes the input to a shell (also in POSIX) for splitting and I believe it provides the behavior we need.

The `wordexp` documentation on e.g., macos cautions against using it with untrusted input which is exactly our use case. The rationale is that since `wordexp` directly passes its input to a shell process it could be used to interact with this shell in an uncontrollable way (e.g., gain shell access); I believe using it in a developer API is still fine.

evantypanski (approved):
Looks great, I couldn't break it

PR #1934, opened 2024-12-03
topic/robin/gh 1829 shift validation

### LINKED ISSUE #1829: Assertion `(U)bits < (int)safeint_internal::int_traits< T >::bitCount' failed.

rsmmr:
For constants, we now reject such shifts at compile time. At runtime,
we catch them by having the forked version of `SafeInt` throw an
overflow exception.

Closes #1829.

- **Change `SafeInt` submodule to a fork including a fix.**
- **Catch integer shifts exceeding the width of the operand.**


.gitmodules
bbannier, line 1: WDYT about instead forking this into the Zeek org? Given how silent upstream got fear we might need to fix more issues ourselves in the future.

hilti/toolchain/src/ast/operators/unsigned-integer.cc
```diff
@@ -37,6 +37,19 @@ inline UnqualifiedType* widestTypeUnsigned(Builder* builder, const Expressions&
...
 
+inline void validateShiftAmount(expression::ResolvedOperator* n) {
+    if ( auto expr = n->op1()->tryAs<expression::Ctor>() ) {
+        auto ctor = expr->ctor();
+        if ( auto coerced = ctor->tryAs<ctor::Coerced>() )
+            ctor = coerced->coercedCtor();
+
+        if ( auto i = ctor->tryAs<ctor::UnsignedInteger>() ) {
+            if ( i->value() >= n->op0()->type()->type()->as<type::UnsignedInteger>()->width() )
+                n->addError("shift amount cannot exceed operand's width", n->location());
```
bbannier, line 48:
```suggestion
                n->addError("shift amount cannot be equal or exceed operand's width", n->location());
```

tests/hilti/types/integer/shifting-too-much.hlt
```diff
@@ -0,0 +1,53 @@
+# @TEST-EXEC-FAIL: hiltic -jd %INPUT 2>>output
+# @TEST-EXEC: btest-diff output
+#
+# @TEST-EXEC: hiltic -jd runtime-fail-1.hlt 2>>stderr; grep -q hilti::rt::Overflow stderr
+# @TEST-EXEC: hiltic -jd runtime-fail-2.hlt 2>>stderr; grep -q hilti::rt::Overflow stderr
```
bbannier, line 5: I think it would be valuable to capture the location of the runtime failure. WDYT about dropping these explicit invocations here and switching `@TEST-START-FILE`/`@TEST-END-FILE` to `@TEST-START-NEXT`?

bbannier (dismissed):

.gitmodules
rsmmr: I'm planning to file another PR upstream, let's see how that goes. We can move it over later if needed.

bbannier (approved):

PR #1939, opened 2024-12-04
Add some consistency to attribute validation

evantypanski:
Closes (?) #1901 (that issue also mentions properties and slightly different attribute validation, but I count this as point 2)

The first commit is the chunk of the work, the second just goes in and touches up some stuff with attribute validation for readability - somewhat related

This only actually answers the question "does attribute X go on node Y?" in one place (and for types on fields because that's so prevalent). I think that doing any more requires more context than you can get generally. Ideas for how to get more centralized info on attributes could be useful in the future

As for implementation, this uses bitmasks because I got sick of writing loops or stuff like `std::find`. I actually find it pretty clean, though the interface there could be ironed out, open to suggestions there. A cool benefit is it lets an `AttributeSet` hold a mask of what attributes it contains. That just makes some calculations easier than looping through or finding the actual attribute node for simple queries ("Is X attribute present?")

The validation logic cleanup was my attempt to make the validator a little easier to read with these changes.. It wasn't as impactful as I had hoped, though.

The giant tables that map a node's tag to its attributes is really ugly - but that's by necessity, I think. I really don't want a node to know what attributes apply to it, I like nodes being as close to plain ol' data as possible. But, it could also live outside of the validators, I just don't know where else it'd go :)

Also finally: I went through all of the mentioned attributes in all of the documentation and tried to replicate the intended behavior, but I may have gotten some nodes wrong... really hard to tell.


hilti/toolchain/include/ast/node.h
```diff
@@ -240,7 +240,15 @@ class Node {
     virtual ~Node();
 
     /** Returns the node tag associated with the instance's class. */
-    auto nodeTag() const { return _node_tags.back(); }
+    node::Tag nodeTag() const {
+        // Get the last non-zero tag. The last element(s) may be unset
```
rsmmr, line 244: This surprises me, when did you see unset tags at the end?

rsmmr (commented):
I haven't looked very closely yet but one request in advance: can you elaborate why the masking is helpful, compared to just normal sets of some kind? From a quick look, the masks seek to be just used just like sets, but they add quite a bit complexity to the attribute code, including to external places using attributes (like `validateFieldTypeAttributes()` doing mask operations). So I'm wondering if it's worth it.

hilti/toolchain/include/ast/node.h
evantypanski: since `Tags` [is an array](https://github.com/zeek/spicy/blob/b1e305b92975bf9d1ab52049fc5c7472526fc1f2/hilti/toolchain/include/ast/node-tag.h#L17), anything that had less than 2 other bases would just return `0`. Which turns out to be most nodes.

evantypanski:
> I haven't looked very closely yet but one request in advance: can you elaborate why the masking is helpful, compared to just normal sets of some kind?

Ah, that's one of my strong opinions that I don't think everyone holds: anything that can be done at compile time easily should be done at compile time. Since sets aren't able to be `constexpr` I hardly even considered them. I can add a commit to swap over to sets, which would also remove a lot of the compile time niceties, but it certainly would make it cleaner.

I can make a commit to swap it over and see which you'd prefer today, it's not a big swap. I'm a bit more lenient on making it slightly more bit-operation-y if it means doing more of the work at compile time.

hilti/toolchain/include/ast/node.h
rsmmr: Oh, I see: it's an array *and* the method isn't used anywhere currently, meaning it wasn't obvious that it was broken. :-)  May have been a `vector` originally and the method is just a left-over.

rsmmr:
> anything that can be done at compile time easily should be done at compile time.

I guess this depends on the definition of "easily". :-) 

Yeah, please give it a try without the mask and then we can compare. Given that performance shouldn't really matter here, I would prefer a cleaner API/usage over compile-timeability (if that's a word...)

rsmmr:
> @rsmmr I moved it to an `unordered_set` in the last commit, let me know what you think.

Yeah, thanks, I definitely like this more.

> It was worth a shot with the `constexpr` bitmasks :)

No problem. :-)

hilti/toolchain/include/compiler/validator.h
```diff
@@ -42,9 +43,23 @@ class VisitorMixIn {
     void checkTypeArguments(const node::Range<Expression>& have, const node::Set<type::function::Parameter>& want,
                             Node* n, bool allow_no_arguments = false, bool do_not_check_types = false);
 
+    /**
+     * Ensures that the node represented by tag can have all of the provided
```
rsmmr, line 47:
```suggestion
     * Ensures that the node represented by tag is allowed to have all of the provided
```

hilti/toolchain/src/compiler/validator.cc
```diff
@@ -60,6 +60,27 @@
 using namespace hilti;
 using util::fmt;
 
+void validator::VisitorMixIn::checkNodeAttributes(node::Tag tag, AttributeSet* attributes,
+                                                  const std::string_view& where) {
+    auto it = allowed_attrs.find(tag);
+
+    if ( ! attributes )
+        return;
```
rsmmr, line 68: Suggest to switch these two around.

hilti/toolchain/src/compiler/validator.cc
```diff
@@ -172,6 +193,7 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
     }
 
     void operator()(Function* n) final {
+        checkNodeAttributes(n->nodeTag(), n->attributes(), "function");
```
rsmmr, line 196: I'm wondering if could move this into a generic `operator()(Node *n)` method so that it would execute for any nodes. Main catch is probably creating the `where` for the error message ...

hilti/toolchain/src/compiler/validator.cc
```diff
@@ -281,13 +305,10 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
 
         if ( auto attrs = n->attributes() )
             for ( const auto& attr : attrs->attributes() ) {
-                if ( attr->kind() != hilti::Attribute::Kind::RequiresTypeFeature )
-                    error(fmt("invalid attribute '%s' for function parameter", attr->attributeName()), n);
-
-                else {
+                // Already validated all attributes are RequiresTypeFeature
```
rsmmr, line 308: The comment seems misleading. Isn't it more like "already flagged attributes other than RequiresTypeFeature as errors"?

spicy/toolchain/src/compiler/validator.cc
```diff
@@ -36,8 +38,8 @@ bool supportsLiterals(QualifiedType* t) {
 // Helper to make sure a field's attributes are consistent. This is type-independent.
 hilti::Result<hilti::Nothing> checkFieldAttributes(type::unit::item::Field* f) {
     // Can't combine ipv4 and ipv6
-    auto v4 = f->attributes()->find(hilti::Attribute::Kind::IPv4);
-    auto v6 = f->attributes()->find(hilti::Attribute::Kind::IPv6);
+    bool v4 = f->attributes()->has(hilti::Attribute::Kind::IPv4);
+    bool v6 = f->attributes()->has(hilti::Attribute::Kind::IPv6);
```
rsmmr, line 42: Personally, I'd stay with `auto` here but matter of preference.

EDIT: Below it still uses `auto` with `has()`.

spicy/toolchain/src/compiler/validator.cc
```diff
@@ -368,13 +416,17 @@ struct VisitorPost : visitor::PreOrder, hilti::validator::VisitorMixIn {
     }
 
     void operator()(hilti::declaration::Type* n) final {
+        checkNodeAttributes(n->nodeTag(), n->attributes(), "type declaration");
```
rsmmr, line 419: Similar comment as above regarding using a more generic visitor.

spicy/toolchain/src/compiler/validator.cc
```diff
@@ -710,8 +772,8 @@ struct VisitorPost : visitor::PreOrder, hilti::validator::VisitorMixIn {
                     if ( ! a->hasValue() )
                         error("&convert must provide an expression", n);
                 }
-                else
-                    error(fmt("attribute %s not supported for unit types", a->attributeName()), n);
+
+                // Otherwise it has already been reported as an invalid attribute
```
rsmmr, line 776: I'd skip this comment as that's essentially just the new model.

hilti/toolchain/include/compiler/validator.h
```diff
@@ -42,9 +43,23 @@ class VisitorMixIn {
...
+     * attributes. This does not use any context, if more information is needed,
+     * then do the check elsewhere.
+     */
+    void checkNodeAttributes(node::Tag tag, AttributeSet* attributes, const std::string_view& where);
+
 private:
     Builder* _builder;
     int _errors = 0;
+
+    static inline std::unordered_map<node::Tag, std::unordered_set<Attribute::Kind>> allowed_attrs{
```
rsmmr, line 57: This feels like a bit of an awkward location for the table. How about moving this to the top of `validator.cc` as a static global (not class member)?

spicy/toolchain/src/compiler/validator.cc
```diff
@@ -198,6 +194,58 @@ struct VisitorPre : visitor::PreOrder, hilti::validator::VisitorMixIn {
...
+     * to the field but not its type, so this gives a bit more context-sensitive
+     * validation for a common case.
+     */
+    void validateFieldTypeAttributes(node::Tag typeTag, AttributeSet* attributes, const std::string_view& clazz) {
+        auto it = field_type_allowed_attrs.find(typeTag);
+
+        if ( ! attributes )
+            return;
+
+        std::unordered_set<hilti::Attribute::Kind> allowed =
```
rsmmr, line 234: Similar as above, would move the able out of there to the top of `validator.cc`.

spicy/toolchain/src/compiler/validator.cc
```diff
@@ -1124,6 +1177,63 @@ struct VisitorPost : visitor::PreOrder, hilti::validator::VisitorMixIn {
         if ( auto x = n->op0()->type()->type()->tryAs<type::Unit>(); x && ! x->isFilter() )
             error("unit type cannot be a filter, %filter missing", n);
     }
+
+    static inline std::unordered_map<node::Tag, std::unordered_set<hilti::Attribute::Kind>> allowed_attrs{
```
rsmmr, line 1181: Once more, move to top. Then we also have all tables together, which is nice of future changes/maintainance.

rsmmr: Also then add few comments to the tables describing their meaning. 

Basically, my goal would be to have single location to go to and change when we modify/add attributes, so that one doesn't need look around any further (unless more context-sensitivity is needed obviously)

hilti/toolchain/src/compiler/validator.cc
evantypanski: Yeah, I tried and the catch was that `where`. I looked for a way to get a string saying what a node is, but I couldn't find one for all nodes - is there one that exists? I could only find some for debugging.

If not, maybe this could be done in a followup? I think having a "print node for diagnostics" would be valuable for this, but that would probably apply to every node.

Ditching the `where` would make me very sad. It was absolutely crucial for debugging, I can imagine the same would apply to users.

hilti/toolchain/src/compiler/validator.cc
evantypanski: I'll just remove it since it's just assumed with the new model anyway, but you're right it was misleading

spicy/toolchain/src/compiler/validator.cc
evantypanski: I moved the map to the top for both `validator.cc` files, and combined the type attributes in.

I don't want to combine the hilti/spicy tables since they serve a bit different functions, then there isn't a really good place to put spicy attribute mappings that would also be used by the hilti validator. Does one table at the top for hilti and one for spicy seem ok? or is there somewhere else that would make more sense?

hilti/toolchain/src/compiler/validator.cc
rsmmr: > is there one that exists? I could only find some for debugging.

Yeah, nothing that would yield something nice for user display, just internal class names.

Ok to leave as is.

spicy/toolchain/src/compiler/validator.cc
```diff
@@ -198,6 +243,58 @@ struct VisitorPre : visitor::PreOrder, hilti::validator::VisitorMixIn {
...
+
+        auto it = allowed_attrs.find(typeTag);
+
+        std::unordered_set<hilti::Attribute::Kind> allowed =
+            {hilti::Attribute::Kind::Synchronize, hilti::Attribute::Kind::Convert, hilti::Attribute::Kind::Requires,
+             hilti::Attribute::Kind::Default,     hilti::Attribute::Kind::Size,    hilti::Attribute::Kind::MaxSize,
+             hilti::Attribute::Kind::Try,         hilti::Attribute::Kind::ParseAt, hilti::Attribute::Kind::ParseFrom};
+
+        if ( it != allowed_attrs.end() )
+            allowed.insert(it->second.begin(), it->second.end());
```
rsmmr, line 289: This part I still find a bit unintuitive, plus the table is still hidden inside the method. Suggest to pull it out to the top  as well (`allowed_attributes_for_any_field` or such) and just loop over it separately.

hilti/toolchain/src/compiler/validator.cc
rsmmr: > I think having a "print node for diagnostics" would be valuable for this, but that would probably apply to every node.

Maybe we could generalize `displayName()` to all nodes. (But indeed for later.)

spicy/toolchain/src/compiler/validator.cc
evantypanski: I moved it out and removed the `insert`, instead looping over 2 sets

rsmmr (approved):
Thanks, looks good now, will merge.

PR #1940, opened 2024-12-05
Deprecate &count attribute.

### LINKED ISSUE #1938: Deprecate undocumented `&count`
It seems like vectors can use `&count` (and it's in [a](https://github.com/zeek/spicy/blob/main/tests/spicy/types/unit/count.spicy) [few](https://github.com/zeek/spicy/blob/main/tests/spicy/types/unit/count-fail.spicy) [tests](https://github.com/zeek/spicy/blob/main/tests/spicy/types/unit/switch-blocks-with-lists.spicy)) but it's no where in the parsing documentation. The only reference is in release notes:

> When parsing vectors/lists of integers of a given length, use &count instead of &length.

Since `&length` is deprecated and there seems to be a better syntax with `[N]`, it seems reasonable to deprecate.

bbannier:
The preferred way to indicate how many elements should be parsed for a
vector has been vector syntax over `&count` for a long time, e.g.,

```
type X = unit {
    : uint8[] &count=42; # AVOID.
    : uint8[42];         # PREFER.
};
```

With this patch `&count` is now deprecated and we emit a warning. We
will remove support for `&count` in a future release.

Closes #1938.


spicy/toolchain/src/compiler/validator.cc
```diff
@@ -799,14 +799,17 @@ struct VisitorPost : visitor::PreOrder, hilti::validator::VisitorMixIn {
         }
 
         if ( count_attr && (repeat && ! repeat->type()->type()->isA<hilti::type::Null>()) )
-            error("cannot have both `[..]` and &count", n);
+            error("cannot have both '[..]' and &count", n);
+
+        if ( count_attr )
+            deprecated("&count is deprecated, prefer '[..]' syntax", count_attr->meta().location());
```
rsmmr, line 805: How about for slightly more clarity:

```suggestion
            deprecated("&count=N is deprecated, prefer '[N]' syntax", count_attr->meta().location());
```

rsmmr (approved):

PR #1941, opened 2024-12-05
Integrate Attribute::Kind into enum stringification framework.

bbannier:


bbannier:
This should be rebased on the final version of #1939.

hilti/toolchain/src/ast/attribute.cc
```diff
@@ -16,28 +16,28 @@ using namespace hilti;
 
 Result<Expression*> Attribute::valueAsExpression() const {
     if ( ! hasValue() )
-        return result::Error(hilti::util::fmt("attribute '%s' requires an expression", attributeName()));
+        return result::Error(hilti::util::fmt("attribute '%s' requires an expression", to_string(kind())));
```
rsmmr, line 19: I would think that these should work without the explicit `to_string()`?

rsmmr (approved):

hilti/toolchain/src/ast/attribute.cc
bbannier: At least the examples of `function::CallingConvention` I looked at needed this as well, see e.g., https://github.com/zeek/spicy/blob/c62d573bbf84ebcf6178e28f54d7b59ec209bf00/hilti/toolchain/src/ast/declarations/field.cc#L12

This is due to them not adding explicit `operator<<`, and the enum stringification machinery does not add it either.

PR #1942, opened 2024-12-06
Deprecate `&anchor` with regular expression constructors.

### LINKED ISSUE #1928: `&anchor` attribute is not used
While reviewing #1927 I noticed that we support parsing the `&anchor` attribute for regexps, but do not use it for anything. We also still document it implying that its use changes behavior.

We should probably remove or at least deprecate it.

rsmmr:
Closes #1928.


hilti/toolchain/src/compiler/validator.cc
```diff
@@ -354,6 +354,14 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
 
     void operator()(ctor::Null* n) final {}
 
+    void operator()(ctor::RegExp* n) final {
+        if ( n->attributes()->has(hilti::Attribute::Kind::Anchor) )
+            // This can end up reporting the same location multiple times,
+            // which seems fine. Otherwise we'd need to explicitly track what's
+            // reported already.
+            deprecated("&anchor is deprecated, it's the default now and can just be removed", n->meta().location());
```
bbannier, line 362: This does not seem to be the case, it seems currently regexp do not anchor anywhere, e.g.,

```c
module Foo {

import hilti;

global auto re1 = /23.*/;
hilti::print(re1.find(b"xxxxxx23xxxxxx")); # Should not match, but does.
hilti::print(re1.find(b"23"));

}
```

```console
$ hiltic -dj foo.hlt
(1, b"23xxxxxx")
(1, b"23")
```

Tangentially related, If I add an explicit anchor to the regex

```c
global auto re1 = /^23.*/;
```

I get the same printout which at the very least is confusing, but probably something which should be rejected as well.

hilti/toolchain/src/compiler/validator.cc
```diff
@@ -354,6 +354,14 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
 
     void operator()(ctor::Null* n) final {}
 
+    void operator()(ctor::RegExp* n) final {
+        if ( n->attributes()->has(hilti::Attribute::Kind::Anchor) )
+            // This can end up reporting the same location multiple times,
+            // which seems fine. Otherwise we'd need to explicitly track what's
+            // reported already.
```
bbannier, line 361: Yeah, this is due to `deprecated` directly emitting output while `error` implicitly deduplicates. The current impl also has the behavior that deprecations are reported before errors which seems backwards. I am not convinced it is worth pulling `deprecation` into `Node` like we do for `error` given the amount of deprecation warnings we currently emit.

hilti/toolchain/src/compiler/validator.cc
rsmmr: > This does not seem to be the case, it seems currently regexp do not anchor anywhere, e.g.,

I got that out of an older commit message, which seemed to say so, but admittedly didn't try it. 

> I get the same printout which at the very least is confusing, but probably something which should be rejected as well.

It's hard to test for that (because the `^` could be further down inside the regex). Also, the `find()` docs kind of explain what's going on ("will find matches at arbitrary starting positions").

The thing is that in the end `^` isn't really doing much at all (`match()` implicitly anchors, whereas `find()` essentially ignore it). But not really a problem I think.

PR #1943, opened 2024-12-06
Upgrade Python to 3.9 with pyupgrade.

bbannier:


tests/Scripts/license-header.py
```diff
@@ -6,10 +6,8 @@
 for f in sys.argv[1:]:
     with open(f) as input:
         if not any(
```
timwoj, line 8: Huh, I'd never seen that before.

tests/Scripts/stray_baselines.py
```diff
@@ -10,14 +10,12 @@
...
-        available_tests = set(
-            map(
-                lambda x: x.decode("utf-8"),
-                subprocess.check_output(
-                    ["btest", "-l", "-c", TEST_DIR + "/btest.cfg"]
-                ).splitlines(),
-            )
-        )
+        available_tests = {
+            x.decode("utf-8")
```
timwoj, line 14: What exactly is happening with this `x.decode` here? `x` isn't set to anything, so it's not just decoding a string.

timwoj (commented):
Huh, I'd never seen that `if not any` construct. Looks fine to me.

tests/Scripts/stray_baselines.py
bbannier: The part in `{..}` is a set comprehension, `x` is bound via `for x in ...` on the line below.

tests/Scripts/license-header.py
bbannier: Already before this was more functional style; the arg to `any` is new a generator expression here. I personally find this still pretty readable, but I could rewrite it e.g., using a temporary list if you could that would help.

timwoj (approved):

PR #1945, opened 2024-12-10
Update auto-generated architecture diagram.

rsmmr:
Looks like some dependency upgrade changed the content slightly.


bbannier (approved):
*Looks* good to me 😅 (differences seem to be mostly in label placement).

PR #1946, opened 2024-12-10
Add `bytes` decoding from UTF-16

### LINKED ISSUE #1788: decode() on bytes should support UTF-16
It looks like the current implementation only supports ASCII and UTF-8 to decode into a string and the current library being used is strictly for UTF-8.  In order to support anything with Windows roots, it would be nice to support UTF-16.

I poked around for a few minutes and found a potential small library that might work for the use case to decode UTF-16 into a string type.... https://github.com/nemtrif/utfcpp

bbannier:
Closes #1788.


hilti/runtime/src/types/string.cc
```diff
@@ -132,35 +137,87 @@ std::tuple<std::string, std::string> string::split1(const std::string& s, const
...
+                        }
                         case unicode::DecodeErrorStrategy::STRICT:
                             throw RuntimeError("illegal UTF8 sequence in string");
                     }
 
-                    p += 1;
-                    continue;
+                    p = std::next(p);
                 }
+            }
```
rsmmr: This first part is just the UTF8 conversion/validation again, right? Better to factor that out than duplicating the code.

hilti/runtime/src/types/bytes.cc
```diff
@@ -47,34 +106,86 @@ std::tuple<bool, Bytes::const_iterator> Bytes::find(const Bytes& needle, const c
...
 
             while ( p < e ) {
-                utf8proc_int32_t cp;
-                auto n = utf8proc_iterate(p, e - p, &cp);
-
-                if ( n < 0 ) {
+                try {
+                    auto cp = utf8::next(p, e);
+                    utf8::append(cp, t);
+                } catch ( const utf8::exception& ) {
```
rsmmr, line 123: Same question here as in `string.cc`.

hilti/runtime/src/types/bytes.cc
```diff
@@ -47,34 +106,86 @@ std::tuple<bool, Bytes::const_iterator> Bytes::find(const Bytes& needle, const c
...
+                            utf8::append(unicode::REPLACEMENT_CHARACTER, t);
+                            break;
+                        }
                         case unicode::DecodeErrorStrategy::STRICT:
                             throw RuntimeError("illegal UTF8 sequence in string");
                     }
 
-                    p += 1;
-                    continue;
+                    p = std::next(p);
```
rsmmr, line 134: ``++p``, right?

hilti/runtime/src/types/bytes.cc
```diff
@@ -47,34 +106,86 @@ std::tuple<bool, Bytes::const_iterator> Bytes::find(const Bytes& needle, const c
...
+                }
+            }
+
+            return t;
+        }
+
+        case unicode::Charset::UTF16BE: [[fallthrough]];
+        case unicode::Charset::UTF16LE: {
+            if ( Base::size() % 2 != 0 ) {
```
rsmmr, line 143: Are there tests for this special error handling?

hilti/runtime/src/types/bytes.cc
```diff
@@ -47,34 +106,86 @@ std::tuple<bool, Bytes::const_iterator> Bytes::find(const Bytes& needle, const c
...
+            // We can assume an even number of bytes.
 
-                t += std::string(reinterpret_cast<const char*>(p), n);
-                p += n;
+            std::u16string t;
+
+            // utfcpp expects to iterate a `u16string` or `u16string_view`.
+            auto v16 = std::u16string_view{reinterpret_cast<const char16_t*>(Base::data()), Base::size() / 2};
+
+            const auto order = cs.value() == unicode::Charset::UTF16LE ? U16Iterator::LE : U16Iterator::BE;
```
rsmmr, line 166: For readability:

`- const auto order = cs.value() == unicode::Charset::UTF16LE ? U16Iterator::LE : U16Iterator::BE;`
`+ const auto order = (cs.value() == unicode::Charset::UTF16LE ? U16Iterator::LE : U16Iterator::BE);`

hilti/runtime/src/types/string.cc
```diff
@@ -132,35 +137,87 @@ std::tuple<std::string, std::string> string::split1(const std::string& s, const
...
+            auto p = s.begin();
+            auto e = s.end();
 
             while ( p < e ) {
-                utf8proc_int32_t cp;
-                auto n = utf8proc_iterate(p, e - p, &cp);
+                try {
+                    auto cp = utf8::next(p, e);
+                    utf8::append(cp, t);
+                } catch ( const utf8::exception& ) {
```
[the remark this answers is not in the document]
bbannier: I now cleaned this up to use a more specific exceptions `utf8::invalid_utf[8|16]`. Overall this library does not do a great job to catch usage errors (e.g., I ended up filing https://github.com/nemtrif/utfcpp/issues/126 which caused me a lot of confusion).

hilti/runtime/src/types/bytes.cc
bbannier: See above.

hilti/runtime/src/types/bytes.cc
bbannier: I added tests to cover this. What I could not cover was the failure branches below for input `Bytes` with an even number of bytes -- while e.g., Python seems to reject certain char16 values (e.g., unpaired surrogates) the UTF16 library we use seems to normalize these to e.g., empty strings. I added a TODO for now.

hilti/runtime/src/types/string.cc
bbannier: Good idea, this can just call `string::encode` to UTF8.

rsmmr (approved):
The one thing I'm still missing is HILTI-level, i.e., use UTF16 in `tests/hilti/types/bytes/decode.hlt` and `tests/hilti/types/string/encode.hlt`. Otherwise looks good.

hilti/runtime/src/tests/bytes.cc
```diff
@@ -39,46 +40,47 @@ TEST_CASE("at") {
...
-    CHECK_EQ("\xc3\x28"_b.decode(bytes::Charset::UTF8, bytes::DecodeErrorStrategy::REPLACE), "\ufffd(");
-    CHECK_EQ("\xc3\x28"_b.decode(bytes::Charset::UTF8, bytes::DecodeErrorStrategy::IGNORE), "(");
-    CHECK_THROWS_WITH_AS("\xc3\x28"_b.decode(bytes::Charset::UTF8, bytes::DecodeErrorStrategy::STRICT),
+    CHECK_EQ("\xc3\x28"_b.decode(unicode::Charset::UTF8, unicode::DecodeErrorStrategy::REPLACE), "\ufffd(");
+    CHECK_EQ("\xc3\x28"_b.decode(unicode::Charset::UTF8, unicode::DecodeErrorStrategy::IGNORE), "(");
+    CHECK_THROWS_WITH_AS("\xc3\x28"_b.decode(unicode::Charset::UTF8, unicode::DecodeErrorStrategy::STRICT),
                          "illegal UTF8 sequence in string", const RuntimeError&);
 
-    CHECK_THROWS_WITH_AS("123"_b.decode(bytes::Charset::Undef), "unknown character set for decoding",
+    CHECK_EQ(Bytes("\0a\0b\0c"s).decode(unicode::Charset::UTF16BE, unicode::DecodeErrorStrategy::STRICT), "abc");
```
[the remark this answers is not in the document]
bbannier: I added BOM detection with fallback to the passed byte order, and additional tests.

rsmmr (approved):

PR #1947, opened 2024-12-11
Make sure `autogen-docs` pre-commit hook can always run in CI

bbannier:
We would previously exit early from this script if we could not generate the architecture diagram for which we needed the Python `diagrams` package (which implicitly depends on `dot` as well). This setup made it impossible to automatically update generated documentation unless one had both `diagrams` and `dot` installed. Additionally the generated `architecture.svg` was not stable across different versions of `dot`.

This patch rewrites `./doc/scripts/autogen-docs` so it can update documentation even if the dependencies for diagram generation are present. Since the `diagrams` output depends on dot (and its version) I currently see no stable, platform-indepdendent way to generate the diagram in a pre-commit hook, so we do not even attempt to make `diagrams` available to the pre-commit hook.

We also fix a few other issues in the generator script:

- make sure it works when `gsed` is not present
- make sorting order of `reserved-keywords.txt` determistic across GNU and BSD toolchains

### DIFF
--- a/doc/scripts/autogen-docs
+++ b/doc/scripts/autogen-docs
@@ -8,6 +8,17 @@
 set -o errexit
 set -o nounset
 
+# GNU `sort` sorts case-insensitive while macOS `sed` sorts case-sensitive which
+# we want. Force GNU sed into case-sensitive mode.
+export LC_COLLATE=C
+
+# We expect a GNU sed. If we find `gsed` use that instead since we are likely
+# on macOS which by default provides a BSD sed.
+SED="sed"
+if command -v gsed >/dev/null 2>&1; then
+    SED=gsed
+fi
+
 ROOTDIR="$(cd "$( dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)/../.."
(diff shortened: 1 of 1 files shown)


rsmmr (dismissed):

bbannier:
@rsmmr, would you mind taking another look? I needed to rewrite this since `diagrams` implicitly depends on the version of `dot` and I couldn't make this stable between our dep envs and CI (and I didn't want to force us to install the `dot` version in our CI image).

doc/scripts/autogen-docs
```diff
@@ -8,24 +8,50 @@
 set -o errexit
 set -o nounset
 
+# GNU `sort` sorts case-insensitive while MacOS `sed` sorts case-sensitive which
+# we want. Force GNU sed into case-sensitive mode.
+export LC_COLLATE=C
+
+# We expect a GNU sed. If we find `gsed` use that instead since we are likely
+# on macos which by default provides a BSD set.
```
rsmmr, line 16:
```suggestion
# on macOS which by default provides a BSD sed.
```

rsmmr (approved):

PR #1948, opened 2024-12-17
[WIP] Extend synchronization points to leverage fixed-size units/fields

rsmmr:
This is still experimental and WIP. The idea is letting error recovery
directly jump forward to specific offsets in the input for cases where
fixed-size unit/fields allow us to compute where to continue.

- **Fix `ParserBuilder::advanceInput()`.**
- **Remove `ParserBuilder::setInput()`.**
- **Enable productions to provide the number of bytes they parse, if known.**
- **Support synchronization points at fixed offsets, without pattern search.**


rsmmr:
Closing this for now, ended up only partially implemented.

PR #1950, opened 2024-12-17
Fix codegen for string literals containing null bytes.

### LINKED ISSUE #1949: Incorrect code generation for string literals containing literal zero bytes
The following always fails:

```ruby
module foo;

global x = "\x00ABC";
assert |x| == 4 : "'%s'" % x;
```

```console
$ spicyc -dj foo.spicy
uncaught exception hilti::rt::AssertionFailure: '' (foo.spicy:4:1-4:29)
```

This is due to us generating the following C++ for the initialization for `x`:

```cpp
::__hlt::foo::x = std::string("\0ABC");
```

This constructor takes a `const char*` which is null-terminated, i.e., `ABC` after the terminator is ignored.

A fix for that would be to use C++ string literals, e.g.,

```cpp
using namespace std::string_literals;
// ...
::__hlt::foo::x = "\0ABC"s;
```

This triggers direct construction of a `std::string` from the bytes inside quotes.

bbannier:
Our strings are UTF8 which can contain literal null bytes. We were previously generating incorrect C++ for such literals. This was due to us constructing C++ value via constructors taking `const char*` which is expected to be null-terminated, i.e., everything after the termiantor was ignored even though it was emitted.

With this patch we switch to C++ literals for generating literal and runtime strings.

Closes #1949.


bbannier:
I am considering backporting this to at least 1.12 (for zeek-7.1), but maybe also 1.11 (for zeek-7.0).

bbannier:
> However, I'm wondering about performance if this ends up in some hot path that previously didn't need the string, but now ends up recreating it each time. Do you think that could be an issue?

This fixes code where we already were emitting temporaries; unless I am missing something in terms of allocations this should behave identical to the previous code. We talked about possibly interning string literals previously (similar to what we do e.g., for regexps), but this should be orthogonal.

rsmmr (approved):

rsmmr:
Alles klar, scheint ich hab den Patch falsch gelesen.

PR #1951, opened 2024-12-18
Modernize clang-tidy setup

bbannier:
The idea of this PR is to bring us to date so that no `clang-tidy` diagnostics are raised in my editor via `clangd`, at least for my version of clang-tidy on macos-14.7.1. The way I went about this was to execute `clang-tidy` as part of the build, in `build/`

```console
$ cmake -DCMAKE_CXX_CLANG_TIDY=/opt/homebrew/opt/llvm/bin/clang-tidy \
        -DCMAKE_C_CLANG_TIDY=/opt/homebrew/opt/llvm/bin/clang-tidy ..
```

Suppressions are now be consistently configured via `.clang-tidy` configurations so they should work with the above setup. I fixed or suppressed some warnings I saw (often: only raised for newer `clang-tidy`).


hilti/runtime/src/fiber.cc
```diff
@@ -652,8 +652,8 @@ void detail::trackStack() {
         return;
 
     if ( fiber->type() == Fiber::Type::IndividualStack || fiber->type() == Fiber::Type::SharedStack ) {
-        if ( auto size = fiber->stackBuffer().activeSize(); size > detail::Fiber::_max_stack_size )
-            detail::Fiber::_max_stack_size = size;
+        detail::Fiber::_max_stack_size =
+            std::max<uint64_t>(fiber->stackBuffer().activeSize(), detail::Fiber::_max_stack_size);
```
rsmmr, line 656: Note there is a difference between these two: the new version performs a write operation every time, whereas the old version just reads most of the time. I don't know if that makes an actual difference in performance, but this code is executed a lot, so I wouldn't mind staying with the original version.

This applies to the other changes in this commit as well, but is probably not important for those.

rsmmr (dismissed):

PR #1952, opened 2025-01-06
Keep test `spicy.rt.time` working for the next 200 years.

bbannier:
This test still compares the current time against hardcoded times to check ordering. Extend the upper value so this test passes for the next ~200 years.


bbannier:
Also backported to `release/1.12`.

bbannier:
Backported to `release/1.11`.

PR #1953, opened 2025-01-06
Address a couple coverity findings

bbannier:


hilti/toolchain/include/ast/types/function.h
```diff
@@ -91,8 +91,8 @@ class Function : public UnqualifiedType {
         : UnqualifiedType(ctx, NodeTags, {}, std::move(children), std::move(meta)), _flavor(flavor) {}
 
     Function(ASTContext* ctx, Wildcard _, Nodes children, Meta meta)
-        : UnqualifiedType(ctx, NodeTags, Wildcard(), {"function(*)"}, std::move(children), std::move(meta)) {}
-
+        : UnqualifiedType(ctx, NodeTags, Wildcard(), {"function(*)"}, std::move(children), std::move(meta)),
+          _flavor(function::Flavor::Standard) {}
```
rsmmr, line 94: Any reason not to init during declaration: `_flavor = function::Flavor::Standard`?

rsmmr (approved):

hilti/toolchain/include/ast/types/function.h
bbannier: I was wondering about this as well since init at decl would IMO in general be preferable.

My primary reason for doing it here was that we had not "unset" flavor, and doing it here directly corresponds to this particular constructor which suggests a "standard" flavor since it passes `function(*)`. Does that make sense?

rsmmr: I see. Given that it's called "standard", I'd say that's a reasonable default generally; and the other constructor sets it explicitly anyways. But this is really minor, so either way is fine.

bbannier: Okay, will leave as is.

PR #1957, opened 2025-01-09
Remove copyright years [skip CI].

evantypanski:
It may be worth going through the files to change it to something similar as well, but for now I'll just put this in as a minimal amount. This covers the most forward-facing parts, at least :)

what docs copyright looks like now:

Screenshot 2025-01-09 at 11 25 18 AM


bbannier (approved):
In Spicy we also have copyright notices with time range in each source file; we have a script to automatically update these.

https://github.com/zeek/spicy/blob/2a8d74708fc5e5f0b123e60b0493365013418ac1/scripts/update-license-headers#L7

I'll bump that script and regenerate sources while merging.

PR #1959, opened 2025-01-14
Fix Coverity reports in generated code

bbannier:


hilti/runtime/src/init.cc
```diff
@@ -108,21 +107,18 @@ void hilti::rt::detail::registerModule(HiltiModule module) {
...
 }
 
-static std::unique_ptr<std::vector<void (*)()>> _registered_preinit_functions;
-
-RegisterManualPreInit::RegisterManualPreInit(void (*f)()) {
-    if ( ! _registered_preinit_functions )
-        _registered_preinit_functions = std::make_unique<std::vector<void (*)()>>();
-
-    _registered_preinit_functions->emplace_back(f);
+static auto& registered_preinit_functions() {
```
rsmmr, line 110: Style tweak: return a pointer to the vector, as a sign that it's writable.

rsmmr (dismissed):

PR #1961, opened 2025-01-22
Clean up some host application API pieces

rsmmr:
- **Fix generated C++ prototype header.**
- **Map C++ types corresponding to public structs into public namespace in generated C++ prototypes.**
- **Do not include CXX flags into linker flags.**
- **Make `type` a type.**


tests/spicy/types/type/type-info.spicy
```diff
@@ -0,0 +1,17 @@
...
+module X;
+
+type S = struct {
+    a: int32;
+    b: int32;
+};
+
+global s1: type = S;
+global s2: type = typeinfo(S);
+global s3: type = typeinfo(bool);
```
bbannier, line 15: The LHS is not really a type, but instead a type info, so I am not a fan of calling its type `type` (also easy to confuse with other uses of `type` which currently always introduce new types, either in a type declaration or type alias declaration). Could we call this e.g., `type_info` instead?

Could you add a test for the using `typeinfo` with a value instead of a type, e.g.,

```ruby
print typeinfo(true);
```

I think this will be very useful to users during debugging.

doc/programming/language/examples/_optional-check.spicy
```diff
@@ -1,9 +1,11 @@
...
+};
 
-if ( x )
-    print "'x' was set";     # Never runs.
-if ( ! x )
-    print "'x' was unset";   # Always runs.
\ No newline at end of file
+global s1: type = typeinfo(bool);
+global s2: type = typeinfo(S);
+global s3: type = S;
```
bbannier, line 9: Not a huge fan of  the implicit conversion in `global s3: type = S;` (without any explicit `typeinfo`); could we remove support for this, or is it unavoidable to support `typeinfo` taking either a value or a type?

hilti/toolchain/include/ast/types/library.h
```diff
@@ -26,17 +28,19 @@ class Library : public UnqualifiedType {
...
+        auto constness = (_constness == Constness::Const ? "true" : "false");
+        auto p = node::Properties{{"const", constness}, {"cxx_name", _cxx_name}};
         return UnqualifiedType::properties() + std::move(p);
     }
 
-    static auto create(ASTContext* ctx, const std::string& cxx_name, Meta meta = {}) {
-        return ctx->make<Library>(ctx, cxx_name, std::move(meta));
+    static auto create(ASTContext* ctx, Constness const_, const std::string& cxx_name, Meta meta = {}) {
+        return ctx->make<Library>(ctx, const_, cxx_name, std::move(meta));
     }
```
bbannier, line 38: The constructor takes `cxx_name` by value, let's do the same here and `move` the value,

```suggestion
    static auto create(ASTContext* ctx, Constness const_, std::string cxx_name, Meta meta = {}) {
        return ctx->make<Library>(ctx, const_, std::move(cxx_name), std::move(meta));
    }
```

With that we should also adjust the signature in `node-factory.h`.

hilti/runtime/include/type-info.h
```diff
@@ -518,7 +518,17 @@ class Function : public detail::NotImplementedType {};
...
+class Library : public detail::AtomicType<hilti::rt::TypeInfo*> {
+public:
+    /**
+     * Constructor.
+     *
+     * @param cxx_name C++-side name of the type
+     */
+    Library(std::string cxx_name) : cxx_name(std::move(cxx_name)) {}
+
+    const std::string cxx_name; /**< C++-side name of type */
```
bbannier, line 530: Let's make this member `private` and add an accessor.

tests/spicy/types/type/type-info.spicy
rsmmr: > The LHS is not really a type, but instead a type info

Don't agree. The way I see it, is that it *is* a type, it just happens to be represented internally as a `typeinfo`. And from that perspective, the first assignment (s1) is actually what I want: it's assigning a type to a type.

Ideally, we'd then not have the other two versions (s2/s3) and instead directly write:

```
global s3: type = bool;
```

But that's something our parser can't (easily) support, hence the kludge to have type-infos coerce to `type`. 

Note that this PR also documents it that way: `typeinfo(...)` returns a `type`. Maybe `typeinfo` isn't a good name for the operator, but not sure what else to use (`type` would be odd, and probably cause Bison trouble again as well).

hilti/runtime/include/type-info.h
rsmmr: Ok, will do. We're a bit inconsistent how type-infos handle visibility, but after another look, accessor seems to be more typical.

rsmmr:
I added 3 more commits to this PR.

evantypanski (commented):
This looks good to me. The `type` type does break C++ compilation when used as a function parameter, though:

```
module Test;

type S = struct {};

function f(ty: type) {
}

function calls_f() {
    f(S);
}

public type Unit = unit {
    on %init() { calls_f(); }
};
```

```
$ spicy-driver test.spicy
/private/var/folders/w6/h7tg594x0yng0_yzvcsn6mv00000gn/T/Test_71066fede46823da-3da77eaabaafa86a.cc:74:25: warning: duplicate 'const' declaration specifier [-Wduplicate-decl-specifier]
   74 |     static void f(const const ::hilti::rt::TypeInfo*& ty);
      |                         ^~~~~
/private/var/folders/w6/h7tg594x0yng0_yzvcsn6mv00000gn/T/Test_71066fede46823da-3da77eaabaafa86a.cc:225:5: error: no matching function for call to 'f'
  225 |     f(&type_info::__ti_Test__S_namex2aTest__Sx2b);
      |     ^
/private/var/folders/w6/h7tg594x0yng0_yzvcsn6mv00000gn/T/Test_71066fede46823da-3da77eaabaafa86a.cc:74:17: note: candidate function not viable: expects an lvalue for 1st argument
   74 |     static void f(const const ::hilti::rt::TypeInfo*& ty);
      |                 ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/private/var/folders/w6/h7tg594x0yng0_yzvcsn6mv00000gn/T/Test_71066fede46823da-3da77eaabaafa86a.cc:228:34: warning: duplicate 'const' declaration specifier [-Wduplicate-decl-specifier]
  228 | static void __hlt::Test::f(const const ::hilti::rt::TypeInfo*& ty) { ::hilti::rt::detail::checkStack(); }
      |                                  ^~~~~
2 warnings and 1 error generated.
[error] spicy-driver: JIT compilation failed
```

Forgive me, if I see a `type` type I immediately try to abuse it ;) fortunately (or not?) I'm already not allowed to do `function f(ty: type, my_param: ty) {}`

rsmmr:
> Forgive me, if I see a `type` type I immediately try to abuse it ;)

You're totally right to expect passing it to a function to work. I pushed a fix.

hilti/runtime/include/type-info.h
```diff
@@ -1260,6 +1277,10 @@ struct TypeInfo {
     // Actual storage for the held type.
     std::unique_ptr<const char, void (*)(const char*)> _storage = {nullptr, [](const char*) {}};
 
+    // Callback rendering a value of the type into a string. User-code should
+    // use `Value::to_string()` instead of calling this directly.
+    std::string (*const _to_string)(const void* const) = nullptr;
```
bbannier, line 1282: Just wondering, using a function ptr here is pretty restrictive, e.g., `_to_string` impl cannot be stateful lambda, we always are required to go through a `void*` which forces the nasty `reinterpret_cast` in `TO_STRING`/`TO_STRING_CONST`.

WDYT about instead using a `std::function<std::string(const void* const)>`? That would enable stateful functions and also give us covariance for free, i.e., we could implement functions for more pointers to more specific types, potentially allowing us to avoid `reinterpret_cast` at all (haven't tried this out completely). We'd instantiate one `std::function` per type.

bbannier (dismissed):

hilti/runtime/include/type-info.h
```diff
@@ -1207,7 +1224,7 @@ namespace hilti::rt {
  */
 struct TypeInfo {
```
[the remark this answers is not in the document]
rsmmr: I agree that this would be nice, but I'll leave it for some other time until need comes up. The reason is that it's not straight-forward what exactly "equal" should mean. It would be easy to compare the typeinfo pointers but I'm not sure if that's always the desired result. For example: should two structs with the same fields be considered equal in terms of comparing their `types`?  What if they do/don't have type IDs?

hilti/toolchain/src/compiler/codegen/coercions.cc
```diff
@@ -203,7 +203,12 @@ struct Visitor : public hilti::visitor::PreOrder {
                 fmt("codegen: unexpected type coercion from view<stream> to %s", dst->type()->typename_()));
     }
 
-    void operator()(type::Type_* n) final { result = cg->coerce(expr, n->typeValue(), dst); }
+    void operator()(type::Type_* n) final {
+        if ( auto lt = dst->type()->tryAs<type::Library>(); lt && lt->cxxName() == "::hilti::rt::TypeInfo*" )
```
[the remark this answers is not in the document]
rsmmr: Yeah, given that this is an internal thing, I think it's ok.

doc/programming/language/examples/_optional-check.spicy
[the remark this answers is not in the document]
rsmmr: I'm not sure what you mean? What's the issue?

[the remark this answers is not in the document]
rsmmr: Oh, I see, it just wasn't committed.

PR #1969, opened 2025-01-30
Fix coercion of function call arguments.

### LINKED ISSUE #1968: Overload resolution picks wrong function
If I implement the same function multiple times for integer types which are coercible into each other, the wrong overload is picked, e.g.,

```ruby
module foo;

function fn(x: uint8) {
    print "uint8";
}

function fn(x: uint64) {
    print "uint64";
}

global x: uint8 = 0;
fn(x);
```

```console
$ spicyc -dj foo.spicy
uint64
```

I would have expected this to print `uint8`.

Adding a `fn` taking a `string` has no effect one these.

It seems we consider overloads with coercible args before we check for explicitly matching ones. We should resolve the correct overload, and also check whether this behaves reasonable for functions taking multiple arguments (where non/some/all require coercions).

rsmmr:
Before, we could unnecessarily coerce integer types when a direct
match was available. The fix takes out the `TryCoercionWithinSameType`
style for non-coercing function call arguments, which solves it. I
don't recall why that was in there in the first place, but it doesn't
seem to break anything to skip it ...

Closes #1968.


hilti/toolchain/include/compiler/coercer.h
```diff
@@ -74,31 +74,32 @@ enum class CoercionStyle {
...
      */
-    TryAllForFunctionCall = (1U << 2U) | (1U << 3U) | (1U << 4U) | (1U << 5U) | (1U << 6U) | (1U << 9U),
+    TryAllForFunctionCall = TryDirectMatchForFunctionCall | TryCoercion | TryCoercionWithinSameType | TryDeref,
 
     /**
      * Shortcut style allowing for direct matches only in the context of
      * operator resolution.
      */
-    TryDirectForMatching = (1U << 1U) | (1U << 3U) | (1U << 4U) | (1U << 6U)
+    TryDirectForMatching = TryExactMatch | TryConstPromotion | TryCoercionWithinSameType
```
bbannier, line 102: I don't think we need `TryCoercionWithinSameType` here (else the bit should probably not named "Direct"), can you check why we added it here? Nothing in the test suite seems to depend on it.

bbannier (dismissed):
Thanks, looks good to me and only left a single issue for one further investigation.

Like discussed offline I added some further cleanup patches, please feel to squash/massage/drop how you like.

hilti/toolchain/include/compiler/coercer.h
rsmmr: Turns out it looks like we don't need `TryCoercionWithinSameType` for any of the short-cut styles anymore (mostly because `TryCoercion` already implies it). I've cleaned that up.

Issue #1966, opened 2025-01-27
Assertion using an anonymous unit


evantypanski:
This code asserts when compiling:

```
module Test;

public type Data = unit {
    : unit {};
};
```

```
$ spicyc -c test.spicy
Assertion failed: (id), function operator(), file grammar-builder.cc, line 297.
```

My guess is we don't want to enable me trying to be too clever, so I'll just say this should be an error instead

PR #1970, opened 2025-01-30
Fix assertion in anonymous units

evantypanski:
Just ensuring there is a `typeID` on unit types.

Two things I'm unsure about that I'll leave as review comments


spicy/toolchain/src/compiler/validator.cc
```diff
@@ -313,6 +313,11 @@ struct VisitorPost : visitor::PreOrder, hilti::validator::VisitorMixIn {
                     error("declaration needs a concrete struct type", n, node::ErrorPriority::High);
             }
         }
+
+        if ( auto unit_ty = n->type()->type()->template tryAs<type::Unit>() ) {
+            if ( ! unit_ty->typeID() )
+                error("unit types must be named", n->type());
```
evantypanski, line 319: why is this check (in `checkVariable`) necessary, even though there is a check in the `operator()(spicy::type::Unit*)` method for this too? Is it not resolved at this point? The `ast-orig` does not show the unit type for the local/global fields either

spicy/toolchain/src/compiler/validator.cc
```diff
@@ -832,6 +837,8 @@ struct VisitorPost : visitor::PreOrder, hilti::validator::VisitorMixIn {
                     error(fmt("field name '%s' cannot have name identical to owning unit '%s'", field->id(), type_id),
                           n);
         }
+        else
+            error("unit types must be named", n);
```
evantypanski, line 841: Were anonymous units allowed at some point? I'm confused why this condition existed in the first place

bbannier: Unsure, but I believe currently declaring types at variable declarations is not a thing in Spicy, e.g., the following should all be rejected (else stuff fails down the line like for units in #1966:

```ruby
module foo;

global x: bitfield(8) {
};

global y: struct {
    a : string;
};

global z: unit {
    a : /abc/;
};
```

Catching whether a type was declared inline like this on the variable declaration seems weird to me (and like currently written in this patch also is painful to scale); I feel rejecting this in the parser would be much more natural. The reason we currently allow this in the parser is that `local_decl`, `global_decl` and similarly `unit_field` allow user-defined types since they take a `qtype`/`base_type` for the type where they in reality we only expect a a `scoped_id` (references to user-defined type, can go to `QualifiedType` via `builder->qualifiedType(builder->typeName($1), hilti::Constness::Mutable)` or a builtin type (things like integers etc which are not resolved, but hardcoded in the parser). We should probably clean this up by introducing a parser rule which can be either a builtin type ("resolved" while parsing) or a `scoped_id` (resolved after parsing); while we do that we might realize that `base_type_no_ref` should only contain builtin types.

evantypanski: I think the parser makes sense, but I have two concerns with it:

1) The error message is worse (and it'll stop parsing when it hits the issue), eg your case just says "syntax error, unexpected BITFIELD" which I guess makes sense, I'd just like it a little nicer.
2) I don't really see it as "painful to scale" - it's adding in specific checks when there are issues. I think a linear relationship between the amount of checks and amount of things that should error is reasonable - you're adding one check in each place that would error.

I'll mess around with the parser and see if it can get a little cleaner there. Even still I personally prefer it in verification since I see new nonterminal indirection as tech debt. But I'll see what I can do.

Also: when does the bitfield fail? Seems like those are certainly [allowed to be anonymous](https://docs.zeek.org/projects/spicy/en/latest/programming/parsing.html#bitfield)

evantypanski:
I put it in the parser instead, it's OK I think. Also added fix for other cases that segfaulted like `struct`. Not for `bitfield` - I didn't see that failing because of anonymous types at least.

rsmmr (commented):
Sorry for being late, and contradicting, but I think this would be better in the validator because it's a semantic property of the AST, not just a syntax issue. Generally the validator is preferable for that because ASTs can be assembled in other ways than just parsing (main example: turning Spicy into HILTI). 

Not sure what the problem was with the original approach, I don't think I can see the code anymore.

evantypanski:
@bbannier any other thoughts on validator vs. parser? I personally prefer the validator too. I initially didn't have handling for structs/etc. so I'd add that too, just wanting to double check. It'll have a little bit of messiness to make sure locals, globals, unit variables, and fields types are all checked for "named" types, but I think it's a more natural place

bbannier:
> @bbannier any other thoughts on validator vs. parser? I personally prefer the validator too.

Since you both agree let's go with the validator approach.

The reason I was bringing up the parser it didn't distinguish between type declarations and type references like we would need for variable declarations, but if we are careful we can catch them later. A case like `local x: unit { a: bytes; };` looks completely weird to me, but Robin is right that rejecting this during parsing is insufficient (we'd should probably also adjust the AST 🤷). Sorry for the useless tangent.

hilti/toolchain/src/compiler/validator.cc
```diff
@@ -177,6 +177,17 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
         return Nothing();
     }
 
+    // Returns an error if the given type is not named but must be named when used.
+    Result<Nothing> checkIfProperlyNamed(QualifiedType* t) {
```
evantypanski, line 181: I dislike this function name but apparently I can't think because I can't figure out something better that's short

rsmmr: I believe we could simplify the call sites by moving the error reporting into the helper function. How about something like this:

```
void checkDeclarationType(Declaration* d, QualifiedType* t) {
    if ( t->type()->isA<hilti::type::Struct>() || t->type()->isA<hilti::type::Enum>() ||
         t->type()->isA<hilti::type::Union>() ) {
        if ( ! t->type()->typeID() )
            error(fmt("%s types must be named in declarations", t->type()->typeClass()), d);
    }
}
```

And then just call it from the declaration visitors: `checkDeclarationType(n, n->type())`.

hilti/toolchain/src/compiler/validator.cc
```diff
@@ -725,6 +744,8 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
         std::set<ID> seen;
 
         for ( const auto& f : n->fields() ) {
+            if ( auto rc = checkIfProperlyNamed(f->type()); ! rc )
+                error(fmt("type cannot be used as field type (because %s)", rc.error()), f->type());
```
rsmmr, line 748: I believe it should work to move this into its own visitor instead. Using the function from above:

```
void operator()(declaration::Field* n) final { checkDeclarationType(n, n->type()); }
```

spicy/toolchain/src/compiler/validator.cc
rsmmr: > I'm confused why this condition existed in the first place

Yeah, not sure either. As cleanup, I would suggest to move the typeID() check to the beginning and just early-exit if it's not named. Then assume it's set in the rest of the function body.

spicy/toolchain/src/compiler/validator.cc
rsmmr: Actually, I'm not sure either. I would also expect this to be unnecessary. And yes, looking at the AST output, that does look weird, not sure right now what's going on there.

tests/Baseline/spicy.types.enum.anonymous-enum-fail/output
```diff
@@ -0,0 +1,8 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+[error] <...>/anonymous-enum-fail.spicy:9:17-9:27: cannot coerce expression '<anon-enum>::A' of type 'enum { A = 0, B = 1 }' to type 'enum { A = 0, B = 1 }'
```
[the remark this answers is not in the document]
evantypanski: Thanks for nudging that, seems if I went one step further initially I would've gotten it. That check needs to go on the enum ctor, not the type. fixed

spicy/toolchain/src/compiler/validator.cc
rsmmr: Ok, I figured this one out. It's hard to see in the AST output but the qtype is an *external* qtype, meaning that rather having a type as its child, it's just pointing to type that already exists elsewhere. The assumption is that that existing type is part of the AST as well and validated there at its original location. However, turns out that assumption is not right here: the fact that it's an external type is just an artifact of how the parser creates the qtype, with the type we're pointing to not ending up in the AST itself. 

I'll add a commit at the beginning of the branch that fixes this; it's unnecessary & inefficient anyways. And I'll tie remove this block.

rsmmr:
I pushed an update to this branch. One new commits at the beginning, plus a new visitor for `declaration::Constant` that allows to remove the enum ctor checks. I think this looks pretty clean now after all.

PR #1979, opened 2025-02-28
Implement block statements (`{...}`)

### LINKED ISSUE #1538: Add support for explicit, nested scopes
It currently seems impossible to introduce nested scopes.

```ruby
module foo;

function f() {
    {
        local x = 1;
    }
}
```

```console
$ spicyc -dj foo.spicy
[error] foo.spicy:4:5: syntax error, unexpected '{'
[error] spicyc: parse error
```

We should consider introducing support for this as it can e.g., help avoid name pollution by limiting variable scope.

The current workaround is to use e.g., an `if` with a block.

evantypanski:
Closes #1538

I'm a little confused where the two extra reduce-reduce conflicts come from in the parsers, but it seems to function. I was surprised this wasn't possible before and figured it would be easy to implement.

There's a small thing in docs that I thought about changing: there are a few times that statements refer to `BLOCK` (ie for loops, `if`, etc.) but really it can be a single statement, then it turns into a block. I think with this it may be a little weird to expose that difference to the user, since they may get confused about `{}` braces in those cases, but with the snippets it's not that hard. And it's not a big deal for someone to use more `{}` than necessary in those cases.


doc/programming/language/statements.rst
```diff
@@ -41,6 +41,19 @@ will be carried along with the exception.
...
+---------
+
+::
+
+    { STATEMENTS }
+
+Defines a scope that limits the visibility of local variables to within
+the block. This also allows variable shadowing, meaning inner variables
+can temporarily override outer variables or constants with the same name.
+
```
bbannier, line 56: Documenting this at all feel confusing to me, at least here, e.g., one might wonder whether Spicy blocks are statements like for [Rust's block expressions](https://doc.rust-lang.org/reference/expressions/block-expr.html)). We say somewhere that Spicy has a C-like syntax, and these are just scopes.

We could probably just not document this at all.

doc/programming/language/statements.rst
```diff
@@ -41,6 +41,19 @@ will be carried along with the exception.
 
 .. include:: /autogen/types/generic-begin.rst
 
+.. _statement_block:
+
+``block``
+---------
```
rsmmr, line 47: This seem misleading as `block` is not a keyword. Two ideas:

1. Say `BLOCK` instead, to match how we refer to it elsewhere.
2. Add an explanation at the beginning of the section inside the preface (between 5.2.6 and 5.2.6.1 here: https://docs.zeek.org/projects/spicy/en/latest/programming/language/statements.html)
3. Given that this is primarily about variable scope, add it to https://docs.zeek.org/projects/spicy/en/latest/programming/language/variables.html

I think I'd go for (3).

rsmmr (dismissed):

doc/programming/language/statements.rst
evantypanski: I agree, 3 seems nicer. I changed it to go there. Feel free to wordsmith it as much as you want

rsmmr (approved):

PR #1980, opened 2025-02-28
Fix unit `&max-size` not returning to proper location

### LINKED ISSUE #1976: Vectors stop parsing units with `&max-size` immediately
Units right now [accept `&max-size` attributes](https://github.com/zeek/spicy/blob/2f46bd6c5d41e2eb93d88dda6468b966ab07bce5/spicy/toolchain/src/compiler/validator.cc#L40-L42). This seems to work without vectors, so a single field has its size properly limited. When used in vectors, I noticed two instances of surprising behavior, but there may be more (and these are probably the same underlying issue).

1) `&size` is ignored:

```
module Test;

public type Foo = unit {
    data: Data[] &size=2;

    on %done { print self; }
};

type Data = unit {
    b: uint8;
} &max-size=1024;
```

```
$ printf '\01\02' | spicy-driver test.spicy
[$data=[[$b=1]]]
```

Notice that only one byte is consumed.

2) Lookahead fails:

```
module Test;

public type Foo = unit {
    data: Data[];
        : /EOD/;
    x   : int8;
};

type Data = unit {
    b: uint8;
} &max-size=1024;
```

Same input as from documentation on lookahead:

```
$ printf '\01\02\03EOD\04' | spicy-driver test.spicy
[error] terminating with uncaught exception of type spicy::rt::ParseError: expected look-ahead token, but reached end-of-data (test.spicy:4:5-4:17)
```

Since it's allowed and tested I would expect these to work.

evantypanski:
Fixes #1976

`&max-size` was treated the same as `&size` - so it was going to the end of the limited view rather than to the proper location based on what was parsed.


spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -879,6 +879,20 @@ struct ProductionVisitor : public production::Visitor {
         return result;
     }
 
+    /**
+     * Computes how far to advance for `&max-size` parsing where we operate on a limited view,
+     * but do not necessarily consume it fully. The stream for `ncur_max_size` must point
+     * into the `cur` stream.
+     */
+    Expression* computeCurMaxSize(Expression* cur, Expression* ncur_max_size) {
```
bbannier, line 887: Nit: This is already in the API of `postParseField`, but isolated like this `ncur_max_size` reads like an expression holding a size where it really is a limited view. Maybe we should call it e.g., `limited` here and elsewhere, and could then call this helper e.g., `advanceLimited` since it has nothing to do with `&max-size` anymore.

tests/spicy/types/unit/unit-max-size-as-field.spicy
```diff
@@ -0,0 +1,47 @@
+# @TEST-DOC: Tests &max-size on units when not fully consumed from a field.
+#
+# @TEST-EXEC: printf '123' | spicy-driver -d %INPUT >output
+# @TEST-EXEC: btest-diff output
+
+module Test;
+
+public type Outer = unit {
+    dat: Inner;
+    b  : uint8[] &size=2;
```
bbannier, line 10: Nit: IMO the test baselines become more readable if we parsed `bytes &size=1` instead of `uint8`. Here and below.

bbannier (approved):

PR #1985, opened 2025-03-04
Reject function overloads which only differ in default'd parameters.

### LINKED ISSUE #1981: Overloading with default arguments is slightly surprising
We allow overloading functions if their arguments differ. This is roughly similar to how overload resolution behaves in C++, but deviates in the presence of default argument values, e.g.,

```ruby
module foo;

function fun(a: uint8): uint8 {
    return 1;
}

function fun(a: uint8, b: uint8 = 0): uint16 {
    return 2;
}

print fun(1); # Prints '2'.
```

Similar code in C++ would be rejected since the call on the last line is ambiguous, but compiles in Spicy since the default argument value is not in the declaration, but instead injected at the use,

```cpp
// ...
static auto fun(const ::hilti::rt::integer::safe<uint8_t> &a)
    -> ::hilti::rt::integer::safe<uint8_t>;
static auto fun(const ::hilti::rt::integer::safe<uint8_t> &a,
                const ::hilti::rt::integer::safe<uint8_t> &b)
    -> ::hilti::rt::integer::safe<uint16_t>;
// ...
  ::hilti::rt::print(fun(::hilti::rt::integer::safe<std::uint8_t>{1U},
                         ::hilti::rt::integer::safe<std::uint8_t>{0U}),
                     ::hilti::rt::Bool(true));
// ...
```

I am not sure whether this is a deliberate feature or something which appeared organically. In any case it is undocumented and I would argue probably hard to explain and likely too subtle, so I wonder whether we should just disallow this pattern altogether and align with e.g., C++'s behavior (in presence of defaulted arguments non-defaulted arguments need to differ in type). Alternatively we should document Spicy's overload behavior, e.g., in the doc section on functions.

bbannier:
Resolution of calls to functions who's signatures differ only in defaulted parameters behaved non-intuitively in that when given with less than the maximum possible number of parameters we resolved to the overload with most parameters.

With this patch we reject such overloads where the only difference is in defaulted parameters.

Closes #1981.


evantypanski (approved):
in the future it may be worthwhile to make the error message describe why the overload is invalid with a note or something too, since there are now multiple reasons an overload can be rejected. But this looks good.

PR #1991, opened 2025-03-06
Fix a few more overload C++ errors

### LINKED ISSUE #1982: Equivalent function definitions fail at C++ compilation
Found during #1978 

These functions are equivalent and have a body so they should be rejected, instead they fail at C++ compilation:

```
module Test;

function equivalent(): int32 { return 1; }
function equivalent(): int32 { return 2; }

equivalent();
```

This is a little more complicated since I believe this should still be allowed:

```
function equivalent(): void;
function equivalent(): void {}
```

But a naive implementation just checking if the first name declaration has a body misses this case:

```
function equivalent(): int32;
function equivalent(): int32 { return 1; }
function equivalent(): int32 { return 2; }
```

error for reference:

```
$ spicyc -j test.spicy
/private/var/folders/w6/h7tg594x0yng0_yzvcsn6mv00000gn/T/Test_a9eb71556dd6d426-ede2ac17643f4e9d.cc:33:26: error: redefinition of 'equivalent'
   33 | static auto __hlt::Test::equivalent() -> ::hilti::rt::integer::safe<int32_t> {
      |                          ^
/private/var/folders/w6/h7tg594x0yng0_yzvcsn6mv00000gn/T/Test_a9eb71556dd6d426-ede2ac17643f4e9d.cc:27:26: note: previous definition is here
   27 | static auto __hlt::Test::equivalent() -> ::hilti::rt::integer::safe<int32_t> {
      |                          ^
1 error generated.
```

evantypanski:
Fixes #1982

~I think I'll inevitably find another case to cover so this is starting as a draft~

This started just fixing the equivalent functions causing C++ errors, but turned into more:

- Disallowing function prototypes without `&cxxname` (technically breaks valid code, but easy fix to remove them)
- Add reason to why an overload was rejected
- make overloads check if it's valid with all (previous) decls rather than just the first
- Fix a case with defaulted overload issues not erroring with the same return type


hilti/toolchain/src/ast/types/function.cc
```diff
@@ -27,16 +27,28 @@ bool type::Function::isResolved(node::CycleDetector* cd) const {
     return true;
 }
 
-bool type::isValidOverload(Function* f1, Function* f2) {
+hilti::Result<Nothing> type::isValidOverload(Function* f1, Function* f2) {
+    // Equivalent functions that aren't hooks
+    if ( f1->flavor() != function::Flavor::Hook && areEquivalent(f1, f2) )
```
bbannier, line 32: The check for hook flavor doesn't seem right here and we might still work on hooks below. I'd suggest to instead move this into the caller, e.g., there

```cpp
if ( current_fn_ty->flavor() == type::function::Flavor::Hook &&
     current_fn_ty->flavor() == previous_fn_ty->flavor() )
    continue;
```

and here
```suggestion
    if ( areEquivalent(f1, f2) )
```

tests/hilti/ast/basic-module.hlt
```diff
@@ -7,6 +7,6 @@ module Foo {
 
 type X = bool;
 
-declare string foo(real bar);
+function string foo(real bar) {}
```
bbannier, line 10: Could you add valid bodies for this one and the ones below?

```suggestion
function string foo(real bar) { return ""; }
```

We are currently not rejecting these, but I filed #1992 to eventually taking care of these.

tests/hilti/validation/duplicate.hlt
```diff
@@ -63,10 +68,33 @@ function uint<8> fun(uint<8> a) {
...
+
+module Foo {
+
+# Equivalent functions signatures should not be allowed.
+function void fail1() { return; }
+function void fail1() { return; }
+
+function string fail2(uint<8> x, string y) { return "hi!"; }
+function string fail2(uint<8> different, string names) { return "hi!"; }
+
```
bbannier, line 99: Could you add a variant where the signatures are identical, but the bodies differ? We do not look at bodies when checking for overloads, so we'd expect them to be rejected as well, e.g.,

```suggestion
function uint<8> fail3() { return 1; }
function uint<8> fail3() { return 2; }
```

hilti/toolchain/src/compiler/validator.cc
```diff
@@ -196,27 +196,33 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
...
+            auto sorted_nodes = std::vector<Declaration*>(nodes.begin(), nodes.end());
+            std::sort(sorted_nodes.begin(), sorted_nodes.end(),
+                      [](const auto& a, const auto& b) { return a->location() < b->location(); });
 
-            for ( const auto& node : nodes ) {
+            const auto& first_node = *sorted_nodes.begin();
+            for ( std::size_t i = 1; i < sorted_nodes.size(); i++ ) {
+                const auto& node = sorted_nodes[i];
                 if ( node->location() == first_node->location() )
                     continue;
```
bbannier, line 207: We seem to assume below that all nodes have different locations, so with the iteration starting at `1` here this should be technically dead code.

```suggestion
```

hilti/toolchain/src/compiler/validator.cc
```diff
@@ -233,14 +239,18 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
     void operator()(Function* n) final {
         checkNodeAttributes(n->nodeTag(), n->attributes(), "function");
 
+        auto is_hook = n->ftype()->flavor() == type::function::Flavor::Hook;
         if ( auto attrs = n->attributes() ) {
```
bbannier, line 243: Nit.

```suggestion
        if ( auto attrs = n->attributes() ) {
            auto is_hook = n->ftype()->flavor() == type::function::Flavor::Hook;
 
```

hilti/toolchain/src/compiler/validator.cc
```diff
@@ -196,27 +196,33 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
...
+                    // Try all previous nodes and see if this is a valid overload for each
+                    for ( std::size_t j = 0; j < i; j++ ) {
+                        if ( auto previous_decl = sorted_nodes[j]->tryAs<declaration::Function>() ) {
+                            auto current_fn_ty = current_decl->function()->ftype();
+                            auto previous_fn_ty = previous_decl->function()->ftype();
+
+                            if ( auto rc = isValidOverload(current_fn_ty, previous_fn_ty); ! rc )
+                                error(fmt("'%s' is not a valid overload: %s; previous definition in %s", id, rc.error(),
+                                          previous_decl->location()),
+                                      node);
```
bbannier, line 221: Nit.

```suggestion
                            if ( auto valid = isValidOverload(current_fn_ty, previous_fn_ty); ! valid )
                                error(fmt("'%s' is not a valid overload: %s; previous definition in %s", id, valid.error(),
                                          previous_decl->location()),
                                      node);
```

hilti/toolchain/src/compiler/validator.cc
```diff
@@ -196,27 +196,33 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
             if ( nodes.size() <= 1 )
                 continue;
 
-            const auto& first_node = *std::min_element(nodes.begin(), nodes.end(), [](const auto& a, const auto& b) {
-                return a->location() < b->location();
-            });
+            auto sorted_nodes = std::vector<Declaration*>(nodes.begin(), nodes.end());
+            std::sort(sorted_nodes.begin(), sorted_nodes.end(),
+                      [](const auto& a, const auto& b) { return a->location() < b->location(); });
```
bbannier, line 201: Nit.

```suggestion
                      [](const auto* a, const auto* b) { return a->location() < b->location(); });
```

hilti/toolchain/src/compiler/validator.cc
```diff
@@ -196,27 +196,33 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
...
 
-            for ( const auto& node : nodes ) {
+            const auto& first_node = *sorted_nodes.begin();
```
bbannier, line 203: Nit.

```suggestion
            const auto* first_node = *sorted_nodes.begin();
```

bbannier (dismissed):

hilti/toolchain/src/ast/types/function.cc
evantypanski: I'll commit that as-suggested, but this makes me wonder if more thought should go into hooks. I think we could probably find errors there.

(ugh)

PR #1994, opened 2025-03-12
Simple parser (micro-)benchmarking

evantypanski:
Draft PR as I work on it, but thought it might be useful/interesting.

Initial run with the two cases I ran:

```
--------------------------------------------------------------------------------------
Benchmark                                            Time             CPU   Iterations
--------------------------------------------------------------------------------------
benchmark_parser/Benchmark::WithUnit/100         12910 ns        12849 ns        55722
benchmark_parser/Benchmark::WithUnit/1000       120478 ns       120480 ns         5742
benchmark_parser/Benchmark::WithUnit/10000     1140806 ns      1140793 ns          588
benchmark_parser/Benchmark::WithUnit/100000   11261903 ns     11261919 ns           62
benchmark_parser/Benchmark::Regex/100             5099 ns         5102 ns       136973
benchmark_parser/Benchmark::Regex/1000           39491 ns        39493 ns        17754
benchmark_parser/Benchmark::Regex/10000         379056 ns       378989 ns         1840
benchmark_parser/Benchmark::Regex/100000       3784368 ns      3784473 ns          184
```

I still want to get some details from previous releases (if possible) and maybe get a way to modify the C++ code generated from Spicy (maybe just from adding the files to SOURCES or something)

this just piggybacks on the `--enable-benchmark` configure flag. It generates a binary `spicy-rt-parsing-benchmark`

### DIFF
--- a/spicy/runtime/CMakeLists.txt
+++ b/spicy/runtime/CMakeLists.txt
@@ -99,3 +99,7 @@ target_link_libraries(spicy-rt-tests
                       PRIVATE $<IF:$<CONFIG:Debug>,spicy-rt-debug-objects,spicy-rt-objects>)
 target_link_libraries(spicy-rt-tests PRIVATE $<IF:$<CONFIG:Debug>,spicy-rt-debug,spicy-rt> doctest)
 add_test(NAME spicy-rt-tests COMMAND ${PROJECT_BINARY_DIR}/bin/spicy-rt-tests)
+
+if (${USE_BENCHMARK})
+    add_subdirectory(tests/benchmarks/)
+endif ()
--- a/spicy/runtime/tests/benchmarks/CMakeLists.txt
+++ b/spicy/runtime/tests/benchmarks/CMakeLists.txt
@@ -1,0 +1,23 @@
+# Copyright (c) 2020-now by the Zeek Project. See LICENSE for details.
+
+set(BENCH_SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/parsers.spicy")
+set(BENCH_MODULE Benchmark)
+
+list(TRANSFORM BENCH_MODULE PREPEND Benchmark_ OUTPUT_VARIABLE _generated_sources)
+list(TRANSFORM _generated_sources APPEND ".cc" OUTPUT_VARIABLE _generated_sources)
+list(APPEND _generated_sources "Benchmark___linker__.cc")
+
+add_custom_command(
+    OUTPUT ${_generated_sources}
+    COMMAND ${CMAKE_BINARY_DIR}/bin/spicyc -x ${CMAKE_CURRENT_BINARY_DIR}/Benchmark
+            "${BENCH_SOURCES}"
+    DEPENDS spicyc ${BENCH_SOURCES}
+    COMMENT "Generating C++ code for Benchmark")
+
+add_executable(spicy-rt-parsing-benchmark parsing.cc ${_generated_sources})
+target_compile_options(spicy-rt-parsing-benchmark PRIVATE "-Wall")
+target_link_libraries(spicy-rt-parsing-benchmark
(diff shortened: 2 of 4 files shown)


evantypanski:
I went back and ran these benchmarks until release 1.7 (1.6 had some options changes that made the `spicyc` command fail and I didn't look at it any more than that). Here are the results from the largest input:

| Release | `Benchmark::WithUnit` | `Benchmark::Regex`|
|---------|-------------------------|---------------------|
| main | 11261903 ns | 3784368 ns |
| 1.12 | 13688066 ns | 4265321 ns |
| 1.11 | 11623078 ns | 3776724 ns |
| 1.10 | 12080949 ns | 3939922 ns |
| 1.9 | 15838365 ns | 3774306 ns |
| 1.8 | 15894474 ns | 3764055 ns |
| 1.7 | 18471877 ns | 3867579 ns |

Note that these numbers can be quite different per-run. But, there's a noticeable speedup for units/vectors in that time it seems

rsmmr:
> 1.12	13688066 ns	4265321 ns

> Note that these numbers can be quite different per-run. 

Do you have a sense what's going on with 1.12? Could that just be a measurement glitch, or did we actually make it worse?

evantypanski:
> Do you have a sense what's going on with 1.12?

I'm pretty sure that's just the relative flakiness of the measurements. I just ran the benchmark 20 times and 4 of them were around the 1.12 values in the table, so I'd guess those were just unlucky. Fortunately, none get to the 1.7 release numbers for units, though - so the trend stands

I just ran them to see a general trend though, which I think it shows pretty well. Not sure why they're so variable with nothing else going on on my laptop ¯\\\_(ツ)_/¯

evantypanski:
So this seems ready to review, I initially wanted to do a bit more but I couldn't find elegant ways to do it. Namely I wanted to add in a way to modify a parser after generating its C++. But, I think that's best done manually. This will just be a simple microbenchmark that we can add to as we find useful benchmarks - this is what I have.

I have a couple small optimizations that probably won't change much practically. It will go in a different PR.

To me, the next steps are to optimize for these cases and use that improvement to track progress. The SSL analyzer will respond to improvements in unit performance. I don't have Regex heavy analyzers, I'm not even sure if improvements there would make a big difference in any current analyzers.

So 1) cut down generated unit overhead when building the parser as much as we can, then 2) hopefully couple the CFG based analyses in order to see significant improvements in the SSL analyzer. Then we'll have some nice tangible results (though, there are already tangible results looking at past releases!)

spicy/runtime/CMakeLists.txt
```diff
@@ -99,3 +99,7 @@ target_link_libraries(spicy-rt-tests
                       PRIVATE $<IF:$<CONFIG:Debug>,spicy-rt-debug-objects,spicy-rt-objects>)
 target_link_libraries(spicy-rt-tests PRIVATE $<IF:$<CONFIG:Debug>,spicy-rt-debug,spicy-rt> doctest)
 add_test(NAME spicy-rt-tests COMMAND ${PROJECT_BINARY_DIR}/bin/spicy-rt-tests)
+
+if (${USE_BENCHMARK})
+    add_subdirectory(src/benchmarks/)
```
rsmmr, line 104: Let's move this into `tests/benchmarks`, just for clarity.

spicy/runtime/src/benchmarks/parsing.cc
```diff
@@ -0,0 +1,73 @@
...
+
+    const spicy::rt::Parser* parser = nullptr;
+    for ( auto* p : spicy::rt::parsers() ) {
+        if ( p->name == parser_name ) {
+            parser = p;
+            break;
+        }
+    }
+
+    assert(parser);
```
rsmmr, line 42: I'd do a hard error/abort here that triggers in release builds as well in case something goes wrong somewhere, just so that it's clear what happened.

rsmmr (approved):

PR #1997, opened 2025-03-25
Fix inactive stack size check during module initialization.

### LINKED ISSUE #1967: Stack overflow for function call in module scope not rejected
The following code aborts with a stack overflow error from the OS.

```ruby
module hello;

function foo(n: uint64) {
    foo(n);
}

foo(134);
```

For some parsing code we inject additional code into function bodies to measure the stack size, and perform a controlled abort if the stack grows too large. We should do the same for this code.

rsmmr:
The stack check wasn't active when running on the main fiber.

Closes #1967.


hilti/runtime/src/fiber.cc
```diff
@@ -173,6 +173,33 @@ detail::Fiber::Fiber(Type type) : _type(type), _fiber(std::make_unique<::Fiber>(
     switch ( type ) {
         case Type::Main: {
             ::fiber_init_toplevel(_fiber.get());
+
+#if __x86_64__ || __arm__ || __arm64__ || __aarch64__ || __i386__
```
bbannier, line 177: Suggest to move this just around the use of `__builtin_frame_address` since the rest should be generic.

hilti/runtime/src/fiber.cc
```diff
@@ -173,6 +173,33 @@ detail::Fiber::Fiber(Type type) : _type(type), _fiber(std::make_unique<::Fiber>(
...
+            // We assume/require that the main stack is at least at large as
+            // our shared fiber stack. With that, we can compute a stack
+            // starting address that's safely inside the actual stack. That
+            // address can then be used for stack size checking in checkStack()
+            // to catch recursions going too deep while we're running inside
+            // the main fiber.
+            const auto min_size = configuration::get().fiber_shared_stack_size;
+
+            struct rlimit limit;
```
bbannier, line 186: Nit.
```suggestion
           rlimit limit;
```

hilti/runtime/src/fiber.cc
```diff
@@ -173,6 +173,33 @@ detail::Fiber::Fiber(Type type) : _type(type), _fiber(std::make_unique<::Fiber>(
...
+            if ( ::getrlimit(RLIMIT_STACK, &limit) < 0 )
+                throw RuntimeError("could not get current stack size");
+
+            if ( limit.rlim_cur < min_size )
+                throw RuntimeError(fmt("process stack size too small, need at least %zu KB", min_size / 1024));
+
+            // There's a bit of fuzziness here as the current frame won't start
+            // exactly at the beginning of the stack---but should be good
+            // enough.
+            _fiber->stack = reinterpret_cast<char*>(__builtin_frame_address(0)) - min_size;
```
bbannier, line 196: Took me a second to remember that stacks grown down so we need `-` 🤦.

hilti/runtime/src/fiber.cc
```diff
@@ -173,6 +173,33 @@ detail::Fiber::Fiber(Type type) : _type(type), _fiber(std::make_unique<::Fiber>(
...
+            _fiber->stack_size = min_size;
+
+#else
+#error "unsupported architecture in hilti::rt::detail::Fiber::Fiber()"
+#endif
+
             // ASAN stack size will be set dynamically later.
```
bbannier, line 203: Either this is stale, or above fiber setup should happen only if we do not `HILTI_HAVE_ASAN`.

bbannier (approved):
Nice, looks like we already almost did this.

hilti/runtime/src/fiber.cc
rsmmr: I'm not sure honestly, will just leave as is.

PR #1999, opened 2025-04-01
Let `configure` work with CMake 4.

rsmmr:
(Not checking for a `cmake4` binary; unlikely to have that but not a
valid `cmake`.)

### DIFF
--- a/configure
+++ b/configure
@@ -37,14 +37,15 @@ command="$0 $*"
 cmake_exe="<no cmake>"
 for i in cmake cmake3; do
     if which $i >/dev/null; then
-        $i --version | grep -q "cmake.*version 3" && cmake_exe=$(which $i)
+        version="$($i --version 2>&1 | grep "cmake.*version" | awk -F '[ .]' '{print $3}')"
+        test -n "${version}" && test "${version}" -ge 3 && cmake_exe=$(which $i)
         break
     fi
 done
 
 which "${cmake_exe}" > /dev/null 2>&1 || {
     echo "\
-This package requires CMake, please install it first, then you may
+This package requires CMake >= 3, please install it first, then you may
 use this configure script to access CMake equivalent functionality.\
 " >&2;
     exit 1;


configure
```diff
@@ -37,14 +37,15 @@ command="$0 $*"
...
     fi
 done
 
 which "${cmake_exe}" > /dev/null 2>&1 || {
     echo "\
-This package requires CMake, please install it first, then you may
+This package requires CMake >= 3, please install it first, then you may
 use this configure script to access CMake equivalent functionality.\
 " >&2;
     exit 1;
```
bbannier, line 51: Nit: I am unsure all this is worth it since we treat `cmake` and `cmake3` interchangeably anyway. Parsing the version string mainly prevents us from somebody e.g., linking a completely unrelated binary to say `/usr/bin/cmake`, something which should be rare and break a lot of other things before. Also, since we use `which` this likely breaks if e.g., `cmake` is an alias.

WDYT about just requiring `cmake` in path and using that? While at it we could allow aliases, e.g.,

```sh
cmake_exe=""
if command -v cmake >/dev/null; then
    cmake_exe=cmake
fi

command -v "${cmake_exe}" > /dev/null 2>&1 || {
    echo "\
This package requires CMake >= 3, please install it first, then you may
use this configure script to access CMake equivalent functionality.\
" >&2;
    exit 1;
}
```

bbannier (dismissed):

configure
bbannier: Scratch part of that: `command -v` is bash-only, so we cannot use it like that in `./configure` which only requires `sh`; we probably need to continue using `which`.

The other thing I seem to need is a tweak to `hilti/toolchain/CMakeLists.txt` since cmake-4 seems to leave `CMAKE_OSX_SYSROOT` empty on my system:

```patch
diff --git a/hilti/toolchain/CMakeLists.txt b/hilti/toolchain/CMakeLists.txt
index 00282c69b..81efff4e5 100644
--- a/hilti/toolchain/CMakeLists.txt
+++ b/hilti/toolchain/CMakeLists.txt
@@ -156,8 +156,10 @@ hilti_link_object_libraries_in_tree(hilti PUBLIC)
 
 # Additional flags
 if (APPLE)
-    set(addl_cxx_flags "-isysroot ${CMAKE_OSX_SYSROOT}")
-    set(addl_ld_flags "-isysroot ${CMAKE_OSX_SYSROOT}")
+    if (NOT CMAKE_OSX_SYSROOT STREQUAL "")
+        set(addl_cxx_flags "-isysroot ${CMAKE_OSX_SYSROOT}")
+        set(addl_ld_flags "-isysroot ${CMAKE_OSX_SYSROOT}")
+    endif ()
 endif ()
 
 # HILTI library directories
```

bbannier:
Backported to release/1.11.

PR #2002, opened 2025-04-01
Extend unit `switch` based on look-ahead to support blocks of items.

### LINKED ISSUE #1971: Unit switch with lookahead routing to block not understood by Spicy parser
I would like to use a unit switch with lookahead and add a block with multiple parsers to a case. This works for unit switch over expressions, but not for lookahead.

```ruby
type X = unit {
    switch {
        -> { # ERROR: syntax error, unexpected '{', expecting ':'
            : b"a";
            : b"a";
        }
        -> : b"b";
    };
};
```

To model something like this I would need to move the block to its own unit.

I am marking this _Feature Request_, but I am pretty sure it is just an unintended limitation in our parser.

rsmmr:
Closes #1971.


spicy/toolchain/include/ast/types/unit-items/switch.h
```diff
@@ -45,23 +45,18 @@ class Case : public Node {
     }
 
     /** Factory function for a default case. */
```
evantypanski, line 47: note this comment seems stale

evantypanski (approved):
Awesome

PR #2003, opened 2025-04-01
Fix minor typos in comments and docs

evantypanski:
Also, I have a question:

Is it "an HILTI" or "a HILTI"? I picked "a HILTI"

also these are the lines highlighted by the highlight change:

Screenshot 2025-04-01 at 4 33 52 PM


bbannier (approved):
> Is it "an HILTI" or "a HILTI"? I picked "a HILTI"

Since the _h_ is pronounced I agree that it is probably _a HILTI_.

PR #2004, opened 2025-04-02
Fine-tune recent CMake 4 tweak.

rsmmr:


CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 
 cmake_minimum_required(VERSION 3.15.0)
 
-if (APPLE AND CMAKE_VERSION VERSION_GREATER_EQUAL 4.0.0)
+if (APPLE AND CMAKE_VERSION VERSION_GREATER_EQUAL 4.0.0 AND NOT CMAKE_OSX_SYSROOT)
```
bbannier, line 5: Nit: technically we do not need to check `CMAKE_VERSION` anymore.

```suggestion
if (APPLE AND NOT CMAKE_OSX_SYSROOT)
```

bbannier (approved):
FTR, the related Zeek PR would be https://github.com/zeek/zeek/pull/4326.

LGTM assuming you were able to reproduce the issue locally (I wasn't), and this patch fixes it.

CMakeLists.txt
rsmmr: It's just to be extra sure, not reason to change anything on CMake < 4.

rsmmr:
> LGTM assuming you were able to reproduce the issue locally 

I was.

bbannier:
Backported to release/1.11.

PR #2006, opened 2025-04-02
Fix a path in the CMake config for the recent benchmark.

rsmmr:
The manual path wouldn't work inside Zeek.


bbannier:
Suggest to use https://github.com/zeek/spicy/pull/2005 instead of the approach here.

rsmmr:
> Suggest to use #2005 instead of the approach here.

I think this one is actually the better one, it's documented behavior for CMake to replace names of executable targets accordingly.

bbannier (approved):

PR #2008, opened 2025-04-02
Proposal: Allow optional semicolon after unit hooks.

evantypanski:
This is a proposal, but it's easy to do so I'm proposing through a PR

It has become a frustration for me, even as I get better writing Spicy parsers, that semicolons can't come after hooks. The main reason is I may have some unit:

```
public type Data = unit {
    x: uint8;
};
```

Then I want to print that unit or add something for debugging, so I naturally just insert the hook before the semicolon (since the hook is tied to the field it makes sense the semicolon terminates the field to me):

```
public type Data = unit {
    x: uint8 { print $$; };
};
```

But that errors! This happens basically every time I try to do this but it's really not necessary.

I originally had this for unit hooks not attached to fields but I dropped it because there's no reason for that one, and it's uglier.

### DIFF
--- a/spicy/toolchain/src/compiler/parser/parser.yy
+++ b/spicy/toolchain/src/compiler/parser/parser.yy
@@ -875,6 +875,8 @@ opt_skip_     :
                 SKIP                             { $$ = true; }
               | /* empty */                      { $$ = false; }
 
+opt_semi      : ';' | /* empty */
+
 opt_unit_field_args
               : '(' opt_exprs ')'                { $$ = std::move($2); }
               | /* empty */                      { $$ = hilti::Expressions(); }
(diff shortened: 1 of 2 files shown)


awelzel:
> But that errors! This happens basically every time I try to do this but it's really not necessary.

Thanks for bringing this up. I've hit this a lot, too, but never dared to open a ticket :-)

rsmmr:
> I would want our documentation to unilaterally choose no-semicolon. Then semicolons get added only if they get that same intuition as me, but it has no real consequence. Maybe a formatter can go in and remove those trailing semicolons. I probably would manually remove them if I notice before committing.

This all sounds like an argument *against* the proposal: it seems odd to introduce something that we believe should not be used (even be actively removed!)

I'm with Benjamin here: not a fan of changing the permitted syntax,

PR #2010, opened 2025-04-03
And `ends_with` from `bytes` and `string`

bbannier:


evantypanski (approved):
Looks good and works for my minor want, thanks!

PR #2011, opened 2025-04-04
Add `reserve` calls to vectors with known size

evantypanski:
This applies to vectors like

```
module Test;

public type Data = unit {
    : uint8[] &size=1000;
};
```

So you'll get a reserve call before parsing:

```
__dst.reserve(::hilti::rt::integer::safe<std::uint64_t>{1000U});
```

this improves the largest vector with `&size` benchmark by about 10-20% locally, from 11997705ns to 10186651ns. The effect is smaller the smaller the vector is, but I don't see a slowdown for those cases

Seems like this likely won't affect many real parsers, but it's a "why not" kind of optimization


awelzel:
How is the vector populated during parsing? Just `push_back()` / `emplace_back()` ?

If so, another nice side-effect is that it avoids the double-size-of-vector allocation strategy used by `libstdc++` (not sure what `libc++` does). If `uint[8] &size=20` is filled with `push_back()`, the vector's capacity will end-up 32, and for a `&size=513` it'd be 1024. So this also saves a bit of memory :-)

evantypanski:
> How is the vector populated during parsing? Just push_back() / emplace_back() ?

Yep it's just `emplace_back`:

```
    __dst.reserve(::hilti::rt::integer::safe<std::uint64_t>{1000U});
    while ( ! (::spicy::rt::detail::atEod(__data, __cur, (__lhs_1=(::hilti::rt::StrongReference<::spicy::rt::filter::detail::Filters>())))) ) {
            // ...
            ::hilti::rt::tuple::assign(std::tie(__elem, __cur), ::hilti::rt::integer::unpack<uint8_t>(__cur, ::hilti::rt::ByteOrder{::hilti::rt::ByteOrder::Network}).valueOrThrow());
            // ...

        if ( ! (__stop) ) {
            __dst.emplace_back(__elem);
        }   
```

I wonder if that should be `push_back` instead?

But yeah, saving the memory is nice. Though I did notice that it doesn't check to make sure the vector is actually used before reserving, so fields that aren't stored may potentially reserve memory that is never used. Oops

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -277,6 +277,13 @@ struct ProductionVisitor : public production::Visitor {
                         pushState(std::move(pstate));
                         pb->initializeUnit(p.location());
                     }
+
+                    if ( p.meta().field() && ! p.meta().field()->isTransient() && addl_param ) {
+                        if ( auto a = p.meta().field()->attributes()->find(hilti::attribute::Kind::Size);
```
rsmmr, line 282: Do you really want `&size` here? That's the number of bytes that the vector is being parsed from. For `uint8` elements, that happens to be the right number, but for like `uint16` it wouldn't be. 

It seems that the optimization should instead be applied to cases like `: uint8[1000]` where we have the actual vector size in terms of number of elements?

evantypanski: Oops, thanks for pointing that out, I considered it then wrongly dismissed it when implementing

I'll see what I can do for this, anything with a known size should be possible, but for units it may get tricky. I'd rather not add in too much logic for this too, though, since it seems like it'd mostly be a relatively small difference for normal sized vectors and it'd be more maintenance.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -277,6 +277,13 @@ struct ProductionVisitor : public production::Visitor {
...
+                             a && addl_param->type()->type()->isA<hilti::type::Vector>() )
+                            builder()->addExpression(builder()->memberCall(builder()->id(addl_param->id()), "reserve",
+                                                                           {*a->valueAsExpression()}));
```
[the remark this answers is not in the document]
evantypanski: converted this to a draft in order to address this and #2016 here

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -513,8 +513,19 @@ struct ProductionVisitor : public production::Visitor {
...
-        if ( ! unit && p.meta().field() )
+        if ( ! unit && p.meta().field() ) {
+            auto field = p.meta().field();
+            auto repeat = field->repeatCount();
+            // Vectors with stored elements should reserve the amount
+            // they will parse, if known.
+            if ( ! field->isTransient() && repeat && field->parseType()->type()->isA<hilti::type::Vector>() &&
+                 ! repeat->type()->type()->isA<hilti::type::Null>() ) {
+                builder()->addExpression(
+                    builder()->assign(destination(), builder()->default_(p.meta().field()->parseType()->type())));
```
evantypanski, line 524: This default assignment is just since the optional field normally sets during `valueOrInit` but that's only called if it's LHS. Is there another way to do this or is this fine?

spicy/toolchain/src/compiler/codegen/parser-builder.cc
evantypanski: I just ditched the `&size` part, it's too much to try to guess with skip fields and whatnot in order to correlate to the number of units, for example. Though you probably already came to that conclusion

spicy/toolchain/src/compiler/codegen/parser-builder.cc
bbannier: This is partially due this happening in the outer `stage2` function, not the inner `stage1` which already receives a proper value.

Doing this in `stage2` has the other issue that a stateful repeat expression would now be evaluated twice, once here and again in `stage1`; this can cause incorrect behavior similar to #2016. I would suggest to move his `reserve` call into `stage1`, and use a temporary there if we would use it multiple times, e.g.,

```patch
diff --git a/spicy/toolchain/src/compiler/codegen/parser-builder.cc b/spicy/toolchain/src/compiler/codegen/parser-builder.cc
index f21a158c8..aca01b4b5 100644
--- a/spicy/toolchain/src/compiler/codegen/parser-builder.cc
+++ b/spicy/toolchain/src/compiler/codegen/parser-builder.cc
@@ -513,19 +513,8 @@ struct ProductionVisitor : public production::Visitor {
         Expressions args = {state().data,   state().begin,      state().cur,  state().trim,
                             state().lahead, state().lahead_end, state().error};
 
-        if ( ! unit && p.meta().field() ) {
-            auto field = p.meta().field();
-            auto repeat = field->repeatCount();
-            // Vectors with stored elements should reserve the amount
-            // they will parse, if known.
-            if ( ! field->isTransient() && repeat && field->parseType()->type()->isA<hilti::type::Vector>() &&
-                 ! repeat->type()->type()->isA<hilti::type::Null>() ) {
-                builder()->addExpression(
-                    builder()->assign(destination(), builder()->default_(p.meta().field()->parseType()->type())));
-                builder()->addExpression(builder()->memberCall(destination(), "reserve", {repeat}));
-            }
+        if ( ! unit && p.meta().field() )
             args.push_back(destination());
-        }
 
         auto call = builder()->memberCall(state().self, id, args);
         builder()->addAssign(builder()->tuple({state().cur, state().lahead, state().lahead_end, state().error}), call);
@@ -1581,10 +1570,23 @@ struct ProductionVisitor : public production::Visitor {
     void operator()(const production::Epsilon* /* p */) final {}
 
     void operator()(const production::Counter* p) final {
+        auto* count = p->expression();
+
+        // Vectors with stored elements should reserve the amount they will parse, if known.
+        if ( auto* field = p->meta().field();
+             field && ! field->isTransient() && field->parseType()->type()->isA<hilti::type::Vector>() ) {
+            if ( auto* repeat = field->repeatCount(); repeat && ! repeat->type()->type()->isA<hilti::type::Null>() ) {
+                // Create a temporary so stateful size expressions are only evaluated once.
+                count = builder()->addTmp("repeat", repeat);
+
+                builder()->addExpression(builder()->memberCall(destination(), "reserve", {count}));
+            }
+        }
+
         auto body = builder()->addWhile(builder()->local("__i",
                                                          builder()->qualifiedType(builder()->typeUnsignedInteger(64),
                                                                                   hilti::Constness::Mutable),
-                                                         p->expression()),
+                                                         count),
                                         builder()->id("__i"));
 
         pushBuilder(body);
```

evantypanski: Oh, that's fair enough, I actually had it elsewhere but forgot to move it back when I decided not to add the reserve with `&size` instead only with the count. You're right this works better. It would've been necessary (or nicer?) to do before with `&size` in order to avoid passing the `&size` expression around within the parser.

It looks a lot better where you're suggesting without the `&size`, sorry I completely forgot that was my reason and that it's not necessary anymore

bbannier: Ah nice, the update diff looks great (just using`p->expression()` is much nicer than my hasty suggestion of `field->repeatCount()`).

Only a nit, but if you are okay with it I would like to add back the empty lines before "blocks" preceded by comments so this is less compressed.

bbannier (approved):

timwoj:
It looks like this change is triggering an oss-fuzz bug: https://oss-fuzz.com/testcase-detail/6213181104259072

bbannier:
@timwoj I created an issue in https://github.com/zeek/spicy-png to make the parser more robust.

As for the Spicy-side change, I believe this exposes an existing issue in this parser, but is not an issue in Spicy (so IMO no release blocker). Any parser which relied on lists being expanded dynamically to their final size like we did previously would have also failed along the way.[^stop]

[^stop]: Not 100% sure what happens for parsers also using `stop`, I'll look into that tomorrow.

evantypanski:
> As for the Spicy-side change, I believe this exposes an existing issue in this parser, but is not an issue in Spicy

My opinion is that it's a bit of both, a parser should limit its array bounds if it knows that the object is huge and accepting it would probably OOM.

But, an attacker wouldn't necessarily need to provide that data anymore - they just say they're sending tons of data but don't, so they get the same result with less effort.

Seems like to me it's reasonable to limit the amount we'd reserve to some reasonably high number, just in case. It's just a minor optimization after all, it's not critical to reserve the exact amount immediately

PR #2014, opened 2025-04-09
Disallow illegal characters in hook IDs and a couple other places.

### LINKED ISSUE #2013: Hook names can have non-ID characters to start with strange consequences
I was messing around with hooks and noticed that the names are allowed to be quite strange. This results in weird behavior when compiling.

Example:

```
module Test;

public type Data = unit {
    name: uint8;

    on %  0 1 2 name { }
};
```

This compiles. I used `spicy-driver` to demonstrate that this actually ends up printing the non-blank characters before `name`:

```
$ spicy-driver test.spicy
%012
```

Which is... just kind of weird?

I can't say this is impactful whatsoever, but I'd kind of prefer this error instead (though the [default catch-all rule](https://github.com/zeek/spicy/blob/d132c06d20245787d67d46acd1a52929f077ff3b/spicy/toolchain/src/compiler/parser/scanner.ll#L258) seems a little vague)

Culprit seems to be the [`HOOK_ID`](https://github.com/zeek/spicy/blob/d132c06d20245787d67d46acd1a52929f077ff3b/spicy/toolchain/src/compiler/parser/scanner.ll#L28) mode in the scanner to some extent.

rsmmr:
The lexer wasn't handling all cases.

Adding a test only for the hook IDs, doesn't seem worth the trouble
for the rest.

Closes #2013.


evantypanski (approved):
Great

PR #2019, opened 2025-04-11
Split attributes between HILTI and Spicy.

### LINKED ISSUE #1955: Model Spicy and HILTI attributes separately
Current HILTI `attribute::Kind` holds both Spicy as well as HILTI attributes (e.g., `parse-at` cannot appear in HILTI code). We should attempt to model them separately; this might not be trivial since in C++ enums cannot be inherited.

> I realized we have another issue with the new `Kind` enum: Spicy-side values now reside inside HILTI headers, which isn't great (things like `MaxSize` and `ParseAt` is nothing that HILTI should know about). Not immediately sure how to fix that, it's really like we'd need two enum types.

_Originally posted by @rsmmr in https://github.com/zeek/spicy/issues/1936#issuecomment-2535260921_

rsmmr:
A previous change had moved knowledge of Spicy attributes into the
HILTI layer. This splits that out again to separate concerns. The main
challenge is that we were using C++ enums, which could not be extended
by Spicy to add its own attributes. We now replace those enums values
with a custom helper class so that we can define constants at
different places.

Other than that new helper class, most of the changes are pretty
mechanical, with the main objective being that Spicy only refers to
its own attributes for anything that can appear in Spicy source code,
and use HILTI attributes only when interacting with the HILTI layer.

Closes #1955.


hilti/toolchain/src/compiler/validator.cc
```diff
@@ -139,10 +139,14 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
     // Ensures that the node represented by tag is allowed to have all of the
     // provided attributes. This does not use any context, if more information
     // is needed, then do the check elsewhere.
-    void checkNodeAttributes(node::Tag tag, AttributeSet* attributes, const std::string_view& where) {
+    void checkNodeAttributes(Node* n, node::Tag tag, AttributeSet* attributes, const std::string_view& where) {
```
evantypanski, line 142: Would it be reasonable to now use `n->nodeTag()` and remove the `tag` from this function?

evantypanski (approved):
Great, I like this separation

Just one minor nit

PR #2022, opened 2025-04-14
Fix `&size` expressions evaluating multiple times.

### LINKED ISSUE #2016: `&size` expressions execute multiple times resulting in issues with side effects
I was considering [Robin's comment](https://github.com/zeek/spicy/pull/2011#discussion_r2034981782) for vectors with reserve and an `&size` expression with side effects. It turns out that the `&size` expression gets evaluated multiple times I think. Example:

```
module Test;

public type Data = unit {
    len: uint8;
    len_bytes: bytes &size=self.len++;
    len_plus_one_bytes: bytes &size=self.len;

    on %done { print self; }
};
```

Parsing gets one too many byte in `len_plus_one_bytes` (and skips a byte!):

```
$ printf '\x01123456789' | spicy-driver increment.spicy
[$len=3, $len_bytes=b"1", $len_plus_one_bytes=b"345"]
```

The generated code looks like:

```
        // Begin parsing production: Variable: len_bytes -> bytes;
        ::hilti::rt::optional::valueOrInit((*__self).len_bytes) = ::spicy::rt::detail::extractBytes(__data, __cur, ::hilti::rt::integer::safe<uint64_t>(::hilti::rt::optional::value((*__self).len)++), ::hilti::rt::Bool(false), "increment.spicy:5:5-5:38"sv, (__lhs_1=(::hilti::rt::StrongReference<::spicy::rt::filter::detail::Filters>())));
        __cur = __cur.advance(::hilti::rt::integer::safe<uint64_t>(::hilti::rt::optional::value((*__self).len)++));
        if ( __trim ) {
            (*__data).trim(__cur.begin());
        }

        // End parsing production: Variable: len_bytes -> bytes;
```

There are two increment calls, one in `valueOrInit` and one in `__cur.advance`, so parsing `len_plus_one_bytes` uses the doubly-incremented `self.len`

evantypanski:
Fixes #2016

Prevents `&size` accesses getting evaluated multiple times

This is a little more tricky than other attributes that I think already do this (`&max-size`, `&requires`) since `&size` is thrown around a bit more. Unfortunately that means it's a little messy and returns the expression and passes it in in annoying ways.

I tried a couple other solutions like storing the expression in the attribute or state. I don't think those were going anywhere, but open to other suggestions.

I'll rebase #2011 on this, but this change makes that PR a bit more annoying so I'll wait until this is reviewed in case it changes drastically


rsmmr:
> I was about to suggest that we could avoid creating temps when the expression is constant/side-effect-free, but I'm now thinking that's better left to a future optimization pass doing constant-propagation 

Well, maybe. We could check for constants now already, too; shouldn't be hard. We could then create a helper function, say `getAttributeValue(attr)` that does `valueAsExpression()` and checks if that's an `expression:Ctor` and constant. If yes, it just returns the expression itself; if not, it creates at temp, assigns the expression, and returns the temp. That function we could then use for any attribute taking an expression to address the issue a bit more generically.

evantypanski:
> I'm not quite sure if you're saying you looked at other attributes, and they don't suffer from this issue?

Yes, nothing else seems to evaluate it multiple times and/or create a temporary. I tried to break a few attributes in the same way and couldn't.

I did add a helper to create a temporary and used it when attributes' expressions are retrieved. The second commit does it locally in the `ProductionVisitor`, the third commit brings it into the Builder to use more generally, not sure which would be preferred so I kept them separate. My preference is probably within the Builder

Putting it in the Builder means that it can be used in the few extra places, but some places continue to use `valueAsExpression` -- if `dd` is necessary and created in a block separate from the `valueAsExpression` call. I didn't think it was worth the effort to change that logic if they're obviously used only once.

Also as a random plus from this, I do kind of like temporaries here even if they're only used once in generated code - makes it more obvious where these random expressions comes from. But that's secondary

hilti/toolchain/src/ast/builder/builder.cc
```diff
@@ -62,6 +62,18 @@ Expression* Builder::addTmp(const std::string& prefix, QualifiedType* t, Express
     return id(tmp);
 }
 
+Expression* Builder::addAttributeTmp(Attribute* attr, const std::string& prefix) {
```
rsmmr, line 65: I'd say this is Spicy-specific functionality, so would not put it into the HILTI builder. Keeping it in the visitor would be one option if that's sufficient, otherwise the Spicy parser builder is be another common place for such helper functionality.

hilti/toolchain/include/ast/builder/builder.h
```diff
@@ -509,6 +509,9 @@ class Builder : public builder::NodeFactory {
         block()->_add(context(), local(std::move(id), t, std::move(args), std::move(m)));
     }
 
+    // Adds a temporary to store an attribute's expression, if necessary.
+    Expression* addAttributeTmp(Attribute* attr, const std::string& prefix);
```
rsmmr, line 513: Mostly personal taste  but I find the "add...Tmp" potentially misleading given that the function might end up *not* adding a tmp (for constants). `evaluateAttributeExpression` maybe? (Fine to leave as is if you prefer.)

rsmmr (dismissed):
👍

hilti/toolchain/include/ast/builder/builder.h
evantypanski: Sure. `maybeAddAttributeTmp` wasn't catchy :)

rsmmr (approved):

PR #2035, opened 2025-04-23
CI: No more Ubuntu 20.04

awelzel:
EOL in April 2025, remove it now.


bbannier (approved):
Thanks for thinking of Spicy!

PR #2036, opened 2025-04-23
Limit `reserve` calls on vectors to minimize OOM.

evantypanski:
Fixing OOM https://github.com/zeek/spicy/pull/2011#issuecomment-2825144687

Parsers should handle their own memory allocations, especially since they should *expect* to allocate that memory. Regardless, the `reserve` call is a small optimization, so I think it would be nice to limit that. The limit is now `262144U` which seems plenty to me to see benefits from reserving the memory while minimizing chances of exploding.

@bbannier may disagree though, I don't mind either way. Feel free to close or adjust too.

This has the nice side effect of making the error messages nicer again, after the `reserve` change it leaked some implementation details that the user probably doesn't care about.


spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -1575,9 +1575,14 @@ struct ProductionVisitor : public production::Visitor {
         // Fields with stored elements should reserve the amount they will parse
         if ( auto* field = p->meta().field(); field && ! field->isTransient() ) {
             // Create a temporary so stateful size expressions are only evaluated once.
-            repeat = builder()->addTmp("repeat", repeat);
+            auto uint64 = builder()->qualifiedType(builder()->typeUnsignedInteger(64), hilti::Constness::Const);
+            repeat = builder()->addTmp("repeat", uint64, repeat);
 
-            builder()->addExpression(builder()->memberCall(destination(), "reserve", {repeat}));
+            // Limit the number of reserved elements just in case
+            auto max_reserve = builder()->integer(1U << 18);
```
bbannier, line 1582: The general idea of this optimization is that we amortize some of the initial resizes when we start with capacity zero and repeatedly double it whenever elements beyond it are added; once we have some capacity these resizes become rarer.

With that I think it would make sense to pick a much, much smaller value for `max_reserve`, e.g., `512` or less instead of `262144`. This also helps when the vector parsing is stopped via `stop` as we do not allocated "too much" then. Additionally this should also be more conservative if the vector contains some bigger type.

```suggestion
            // Cap the number of reserved elements to avoid allocating
            // excessive much memory. The vector parsing might be stopped with
            // `stop` in which case we would not store the declared repeat
            // count anyway.
            auto max_reserve = builder()->integer(512);
            auto num_elements = builder()->min(max_reserve, repeat);
```

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -1575,9 +1575,14 @@ struct ProductionVisitor : public production::Visitor {
         // Fields with stored elements should reserve the amount they will parse
         if ( auto* field = p->meta().field(); field && ! field->isTransient() ) {
             // Create a temporary so stateful size expressions are only evaluated once.
-            repeat = builder()->addTmp("repeat", repeat);
+            auto uint64 = builder()->qualifiedType(builder()->typeUnsignedInteger(64), hilti::Constness::Const);
+            repeat = builder()->addTmp("repeat", uint64, repeat);
```
bbannier, line 1579: Nice!

Nit: `uint64` is a pointer, so we could use `auto*`. We are absolutely inconsistent with that though.

bbannier (approved):
I think this makes a lot of sense, and I also like the cleaned up error message a lot.

I left a comment on making this even softer, but I think we really want this.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
rsmmr: As a side note, agree with using `auto*` as a general style guideline. The inconsistency is, too a large degree, because of that massive, earlier AST refactor which did a lot of search-and-replace on types. (If I were better with clang's AST tooling, I'd actually try do some automated conversion for all  `Node`-derived `auto` pointers.)

bbannier: I opened https://github.com/zeek/spicy/pull/2037 to automatically lint and fix this.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
evantypanski: Sure, seems mostly fine, since we should probably be conservative

It's no longer as big of a speedup (in fact no noticeable speedup for vectors with elements of single bytes), but maybe it would be for large elements. Regardless probably worth it to avoid OOM.

I don't see much utility in saying you'll parse `X` elements then planning on stopping early in some cases, though. Seems like there are better ways in spicy of doing that which won't reserve.

PR #2037, opened 2025-04-24
Use `auto*` instead of `auto`.

bbannier:
This patch uses `auto*` instead of `auto` when creating pointer values. Instances were identified and fixed automatically with `clang-tidy`'s `readability-qualified-auto` check which we also enable.

The only manual changes in this patch are in the `config.cc.in` files.


evantypanski (approved):
I only get one change when running clang tidy so this seems good - thanks!

The one change is unrelated to `auto` and may be (?) useless:

```
diff --git i/spicy/runtime/include/parser.h w/spicy/runtime/include/parser.h
index 55699e5f..4eb687c2 100644
--- i/spicy/runtime/include/parser.h
+++ w/spicy/runtime/include/parser.h
@@ -395,7 +395,7 @@ namespace detail {
  */
 template<typename UnitRef>
 inline void registerParser(::spicy::rt::Parser& p, // NOLINT(google-runtime-references)
-                           std::string linker_scope, UnitRef /* not used, just for template instantiation */,
+                           std::string linker_scope, const UnitRef& /* not used, just for template instantiation */,
                            const hilti::rt::TypeInfo* /* utype */) {
     // Note: This may may be called before spicy::rt::init(), and during
     // hilti::rt::init(). Cannot rely on any library functionality being

```

PR #2038, opened 2025-04-25
Prevent escape of non-HILTI exception in lower-level driver functions.

bbannier:
While we already did some handling of non-HILTI exceptions in Spicy's high-level drivers[^spicy-driver] (even though this would allow escape of exceptions not deriving from `std::exception`), we did not do the same for the lower level processing functionality in the actual driver. This is the interface used by e.g., Zeek, and already an `std::exception` thrown from a parser would lead to the exception propagating into Zeek and ultimately causing a `std::terminate` there, see e.g., [this oss-fuzz
report](https://oss-fuzz.com/testcase-detail/6088469254569984). In that particular case a call to `std::vector::reserve` exceeding the available memory caused raising of a `std::length_error`.

This patch extends the exception handling in the driver so all exceptions are handled by now transforming any non-HILTI exceptions into HILTI exceptions like expected by callers.

[^spicy-driver]: https://github.com/zeek/spicy/blob/32251850bd8808d18c5dd9306dc0d56c7ebd5d9d/spicy/toolchain/bin/spicy-driver.cc#L331-L334

### DIFF
--- a/spicy/runtime/src/driver.cc
+++ b/spicy/runtime/src/driver.cc
@@ -367,6 +367,15 @@ driver::ParsingState::State driver::ParsingState::_process(size_t size, const ch
         DRIVER_DEBUG(e.what());
         _done = true;
         throw;
+    } catch ( const std::exception& e ) {
+        DRIVER_DEBUG(e.what());
+        _done = true;
+        throw hilti::rt::Exception(e.what());
+    } catch ( ... ) {
+        auto* what = "non-standard exception thrown";
+        DRIVER_DEBUG(what);
+        _done = true;
+        throw hilti::rt::Exception(what);
     }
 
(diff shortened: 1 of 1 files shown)


bbannier:
This is definitely a candidate for backporting and zeek-7.2.0, but we might not make the cut for rc1; worst case we land in zeek-7.2.1.

rsmmr (approved):
The change makes sense. 

That said, the specific case mentioned leaves me wondering if we should catch that `std::length_error` right when it happens and turn it into a HILTI runtime exception there. That's a larger issue, though, as I don't think we're are consistent in how we're handling `std` exceptions during runtime.

bbannier:
Also backported to `release/1.11` and `release/1.12`.

PR #2040, opened 2025-04-25
Add example `FileCheck` tests

evantypanski:
I chatted with @bbannier about LLVM [`lit`](https://llvm.org/docs/CommandGuide/lit.html) and [`FileCheck`](https://llvm.org/docs/CommandGuide/FileCheck.html) and think it may be useful to integrate at least `FileCheck` into the tests. It'll be especially helpful for optimizations, I think.

I included one test for the vector reserve functionality (which is tough to test otherwise) and one that duplicates an existing [optimization test](https://github.com/zeek/spicy/blob/main/tests/hilti/optimization/peephole-default-void.hlt).

Curious if there are opinions on these tests? We can add more of them and start using them for optimizer work immediately, I think.

There's a bunch of stuff you can do, for example for the optimization test I could've also tested both the optimized and non-optimized versions by changing `--check-prefix`. It's really useful for generated code, I find.

I'm also curious if CI has `FileCheck`


rsmmr (commented):
Mind elaborating what this gains us compared to using pure btes? I don't really know FileCheck, but it seems not too far away from using `btest-diff` and generally `@TEST-EXEC`. 

Generally speaking, I'm reluctant to add another testing framework, and in particular using two different ones inside a single test (like `peephole-default-void-filecheck.hlt` does). Not saying we definitely cannot, but would like to understand the value better.

evantypanski:
> Mind elaborating what this gains us compared to using pure btest?

Sure, I love pitching `FileCheck` :) (this PR was just a draft to have something to point to before a full pitch but I'll go with it)

1) It lets us test something that's otherwise borderline untestable. The `reserve` call is an example here, where I don't really see a clean way to test a Spicy parser for that *single* call in such a simple way. That one could be tested with `sed` and `btest-diff`, but that gets tricky when you're looking at even a few more lines that you want to test.
2) It's more resilient to small changes, since you can assign variables based on regex then use them later. I did that in the `reserve` test with the `REPEAT` variable - so a small change to the variable's name won't cause a failure. You have extra options like `CHECK-DAG` that can be useful for generated code that may appear in an arbitrary order.
3) Everything is in the same file. I think it's easier to reason about a test when you can see exactly what it expects within the test. I really don't like having to find a baseline file in order to figure out what a test is testing, which happens pretty frequently, actually.

So overall I find it more resilient to changes and easier to follow for some tests. I don't think it's a drop-in replacement for `btest-diff`, there's utility to that too, I just like having more fine-grained testing for certain cases. One of the reasons I brought this up now is because tests for optimizations will be a lot nicer with `FileCheck`, since you can ensure some patterns are there and appear in a certain order. With `btest-diff` it gets harder to maintain that because you have to intuit the original intent of the test, with more complex optimizations that'll only become more of a problem.

The downside is that it's an extra program to include in CI and dev environments and it's something else to learn (since it has some unique syntax elements). The `peephole-default-void-filecheck.hlt` might be better with just `btest-diff`, it was just a simple test to pitch `FileCheck`'s utility.

Also for what it's worth, `FileCheck` isn't itself a runner, it's just a program that takes two inputs and compares some expected values, so I could use it within `btest` pretty easily. Using `lit` (which is basically `btest`) would be a whole other thing that I have no intent of getting into

rsmmr:
> Sure, I love pitching FileCheck :) (

Thanks, that's helpful!

Ok, after reading up a bit more about `FileCheck` as well, I'm on board. It does seem useful, and something hard to do otherwise. I'm not saying that I find the markup super readable/intuitive, but this may just be one of those use cases for which essentially any solution would end up with its quirks.

The only thing I'd suggest is keeping those `@TEST-REQUIRES: which FileCheck` in all the relevant tests, so that it doesn't become a hard dependency for running tests.

bbannier:
> I'm also curious if CI has FileCheck

It does for the jobs named `clang*` since we install `llvm-dev`. The binary is currently available as `FileCheck-19` or from `/usr/lib/llvm-19/bin/FileCheck`.

evantypanski:
> The binary is currently available as FileCheck-19 or from /usr/lib/llvm-19/bin/FileCheck.

Would retrieving this in a test require a separate helper script to find the `FileCheck-XX` or `FileCheck` binary? Has that been done elsewhere in tests or would that be new?

From a minimal search I don't see anything quite the same but I may be missing something, just wondering if that's a new problem.

bbannier:
> Would retrieving this in a test require a separate helper script to find the `FileCheck-XX` or `FileCheck` binary? Has that been done elsewhere in tests or would that be new?

I wouldn't use a script since it would depend too much on the exact env. Instead I'd just make the binary available under the expected name, e.g.,

```patch
diff --git a/ci/Dockerfile b/ci/Dockerfile
index d6df1a9e3..f03c7974b 100644
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -44,6 +44,7 @@ RUN apt-get update \
  && curl https://apt.llvm.org/llvm-snapshot.gpg.key -o /etc/apt/trusted.gpg.d/llvm.asc \
  && apt-get update \
  && apt-get install -y --no-install-recommends llvm-19-dev clang-19 libclang-19-dev clang-tidy-19 libclang-rt-19-dev \
+ && update-alternatives --install /usr/bin/FileCheck FileCheck $(which FileCheck-19) 100 \
  # Install a recent CMake for `SKIP_LINTING` support.
  && test -f /usr/share/doc/kitware-archive-keyring/copyright || \
     curl https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | tee /usr/share/keyrings/kitware-archive-keyring.gpg >/dev/null \
```

evantypanski:
Okay, I marked this as ready. I cut down a few of the tests I had before because I don't think they're necessarily better as `FileCheck` tests, especially if they won't always be run. Instead I just grabbed the `reserve` vectors test. Then a Hilti test that I think it better with `FileCheck` so you see it all in one file (before it was pretty hard for me to see what it was testing...)

So all this does is add two tests, but I'd like to start adding tests as `FileCheck` tests :)

I'd also be comfortable just doing the `reserve` test and the CI change if that seems better than repurposing a test

tests/hilti/output/optimization/const.hlt
```diff
@@ -0,0 +1,113 @@
+# @TEST-DOC: Tests optimizations performing constant folding.
+#
+# @TEST-REQUIRES: which FileCheck
+# @TEST-EXEC: hiltic %INPUT -pg | FileCheck %INPUT -check-prefix=NOOPT
+# @TEST-EXEC: hiltic %INPUT -p | FileCheck %INPUT -check-prefix=OPT
+# @TEST-EXEC: hiltic %INPUT -p -D optimizer
+# @TEST-EXEC: cat .stderr | sort -k 3 > log
+# @TEST-EXEC: btest-diff log
```
rsmmr, line 8: For many of the optimizer tests, the debug baselines are very sensitive to changes, and need updating all the time. So maybe when switching to FileCheck, we could also stop diffing the debug output?

rsmmr (approved):
> So all this does is add two tests, but I'd like to start adding tests as FileCheck tests :)

Fine with me. Accepting FileCheck into our tool set, means it should be "equal": feel free to use it whenever it makes sense.

PR #2041, opened 2025-04-28
Upgrade to llvm-20 in CI

bbannier:


evantypanski (commented):
The `DO NOT MERGE` commit should probably not be in there, right?

Also super nit, but it doesn't matter: 11daa7c66655a1fa5857519030cf9656d641a964 commit message has a typo, 'dedicatred' -> 'dedicated'

PR #2046, opened 2025-05-06
Backport bumps to freebsd platforms in CI

bbannier:
Unfortunately freebsd images go away in cirrus, so we have to bump them to be able to continue running CI.


bbannier:
Part of https://github.com/zeek/spicy/pull/2045 now.

PR #2048, opened 2025-05-07
Fix `hiltic -p` printing invalid functions.

evantypanski:
I noticed doing optimizations that methods would get an extra `method` in the function declaration, which made the resulting code from `hiltic -p` not compile. This should fix that.

I mostly only glanced through the test changes, but they look better to me. It makes the cases I noticed compile, where they wouldn't before.


hilti/toolchain/src/compiler/printer.cc
```diff
@@ -35,15 +35,20 @@ namespace {
...
         if ( ftype.isWildcard() ) {
             _out << "<function>";
             return;
         }
 
-        if ( ftype.flavor() != type::function::Flavor::Standard )
+        if ( ftype.flavor() == type::function::Flavor::Standard )
+            _out << "function ";
+        else
             _out << to_string(ftype.flavor()) << ' ';
```
rsmmr, line 47: If we renamed `Flavor::Standard` to `Flavor::Function` (along with a corresponding string mapping) we could save this special-case here.

hilti/toolchain/src/compiler/printer.cc
```diff
@@ -35,15 +35,20 @@ namespace {
 struct Printer : visitor::PreOrder {
     Printer(printer::Stream& out) : _out(out) {} // NOLINT
 
-    void printFunctionType(const type::Function& ftype, const ID& id) {
+    void printFunctionType(const type::Function& ftype, const ID& id, function::CallingConvention calling_conv) {
```
rsmmr, line 38: I'm wondering now if the calling convention should really be part of the function type, rather than the function itself.

rsmmr (approved):
Looks good. Left a couple of minor ideas, but not sure it's worth the trouble. Generally, there are various other cases where HILTI code doesn't print out correctly, fixing that would be a major effort at this point unfortunately.

hilti/toolchain/src/compiler/printer.cc
evantypanski: I made #2052 for this - it turned out a bit more effort than I wanted to spend on it right now, but I left the WIP patch in that issue if we pick it up in the future

evantypanski:
> Generally, there are various other cases where HILTI code doesn't print out correctly, fixing that would be a major effort at this point unfortunately.

Probably not worth the effort, this case was just useful for me since I'm doing some interprocedural optimization work and it's useful to use the `-p` output to see if it will compile again after optimizations

Technically we could have a whole suite of tests to see if the hilti output is idempotent... but that is even less worth it :)

PR #2049, opened 2025-05-07
Fix compilation error for GCC15

i2z1:
Include cstdint header to support uint8_t/uint64_t in GCC 15+

### DIFF
--- a/hilti/toolchain/include/ast/node-tag.h
+++ b/hilti/toolchain/include/ast/node-tag.h
@@ -3,6 +3,7 @@
 #pragma once
 
 #include <array>
+#include <cstdint>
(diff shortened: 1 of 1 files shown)


bbannier (approved):
Thanks for the patch @i2z1, pretty happy to see that this is the only thing needed to build with gcc-15. We currently do not have this compiler in CI, but I have confirmed locally that this is all that is needed, and that the test suite passes.

bbannier:

i2z1:
Unfortunately, zeek does not provide Arch Linux packages as build target, so this is the only way to get up to date release - try to build it with Arch's repo mainline GCC15.

PR #2050, opened 2025-05-07
Prefer stdout over stderr for `--help` messages.

evantypanski:
Related: https://github.com/zeek/zeek/pull/4153

This makes help messages greppable. It also makes incorrect arguments not print the whole help message, instead they prompt the user to use `--help`

I'm happy to go through and nitpick which stream each should go through, but generally I'd prefer big pieces of information to go through stdout so it's slightly easier to `grep`. 🤷 

it also makes the errors nicer, for example with `spicyc`:

```
etyp:/Users/etyp $ spicyc --aoeu
Try 'spicyc --help' for usage
[error] option '--aoeu' not supported

etyp:/Users/etyp $ spicyc -o
Try 'spicyc --help' for usage
[error] option '-o' requires an argument
```


hilti/toolchain/src/compiler/driver.cc
```diff
@@ -73,7 +73,7 @@ Driver::Driver(std::string name, const hilti::rt::filesystem::path& argv0) : _na
 Driver::~Driver() {
     if ( _driver_options.report_times )
         try {
-            util::timing::summary(std::cerr);
+            util::timing::summary(std::cout);
```
rsmmr, line 76: This one I'd prefer to stay on stderr, the reason being that it's augmenting any normal output to stdout, so that the two don't get mixed together.

hilti/toolchain/src/compiler/driver.cc
```diff
@@ -431,14 +431,19 @@ Result<Nothing> Driver::parseOptions(int argc, char** argv) {
 
             case 'h': usage(); return Nothing();
 
-            case '?': usage(); return error("unknown option");
+            case '?':
+                std::cerr << "Try '" << argv[0] << " --help' for usage\n";
+                if ( optopt )
+                    fatalError(fmt("option '%s' requires an argument", argv[optind - 1]));
+                else
+                    fatalError(fmt("option '%s' not supported", argv[optind - 1]));
```
rsmmr, line 439: How about making this just one line of output:

```suggestion
            case '?':
                if ( optopt )
                    fatalError(fmt("option '%s' requires an argument; try --help for usage", argv[optind - 1]));
                else
                    fatalError(fmt("option '%s' not supported; try --help for usage", argv[optind - 1]));
```

hilti/toolchain/src/compiler/driver.cc
```diff
@@ -431,14 +431,19 @@ Result<Nothing> Driver::parseOptions(int argc, char** argv) {
...
 
             default:
                 if ( hookProcessCommandLineOption(c, optarg) )
                     break;
 
-                usage();
-                return error(fmt("option %c not implemented", c));
+                std::cerr << "Try '" << argv[0] << " --help' for usage\n";
+                fatalError(fmt("option '%s' not implemented", argv[optind - 1]));
```
rsmmr, line 446: Like above.

spicy/toolchain/bin/spicy-driver.cc
```diff
@@ -248,8 +232,13 @@ void SpicyDriver::parseOptions(int argc, char** argv) {
...
             case 'h': usage(); exit(0);
-            case '?': usage(); exit(1); // getopt reports error
-            default: usage(); fatalError(fmt("option %c not supported", c));
+            case '?': [[fallthrough]];
+            default:
+                std::cerr << "Try '" << argv[0] << " --help' for usage\n";
+                if ( optopt )
+                    fatalError(fmt("option '%s' requires an argument", argv[optind - 1]));
+                else
+                    fatalError(fmt("option '%s' not supported", argv[optind - 1]));
```
rsmmr, line 241: Like above.

spicy/toolchain/bin/spicy-driver.cc
```diff
@@ -320,7 +309,7 @@ int main(int argc, char** argv) try {
     driver.finishRuntime();
 
     if ( driver.driverOptions().report_times )
-        hilti::util::timing::summary(std::cerr);
+        hilti::util::timing::summary(std::cout);
```
rsmmr, line 312: Would prefer stderr here again,.

rsmmr (approved):

PR #2054, opened 2025-05-09
Drop unnecessary Bison include from CMake function.

bbannier:
We do not seem to need any includes from Bison, but where previously adding some include path in a pretty ad hoc way. This patch performs minimal cleanup of this. There is probably the opportunity to clean up how we add dependencies with Flex and Bison (e.g., using by refactoring how we propagate this and using functionality like `add_flex_bison_dependency` at the lowest level), but I punt on that here.

### DIFF
--- a/cmake/Util.cmake
+++ b/cmake/Util.cmake
@@ -2,11 +2,12 @@
 
 ### A collection of small helpers for the HILTI/Spicy build system.
 
-# Add the likely Bison include directory to Bison sources to avoid version mismatches.
-function (bison_source src output_dir)
-    string(REGEX REPLACE "/bin/bison" "/include" bison_include "${BISON_EXECUTABLE}")
-    set_source_files_properties(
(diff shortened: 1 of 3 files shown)


cmake/Util.cmake
```diff
@@ -2,11 +2,12 @@
 
 ### A collection of small helpers for the HILTI/Spicy build system.
 
-# Add the likely Bison include directory to Bison sources to avoid version mismatches.
-function (bison_source src output_dir)
-    string(REGEX REPLACE "/bin/bison" "/include" bison_include "${BISON_EXECUTABLE}")
-    set_source_files_properties(
-        ${src} PROPERTIES INCLUDE_DIRECTORIES "${bison_include};${FLEX_INCLUDE_DIR};${output_dir}")
+# Add the likely Flex and Bison include directories to source files.
```
awelzel, line 5: Given there's no bison include dir to find, maybe make this a bit more vagure? Your call.

```suggestion
# Add required include directories for a  Flex / Bison source file.
```

awelzel (approved):

PR #2056, opened 2025-05-13
Cleanups for working with generated hlt code

bbannier:
This PR contains a number of cleanups for working with generated parser code. I used for reducing generated code.


hilti/toolchain/src/compiler/printer.cc
```diff
@@ -429,8 +429,20 @@ struct Printer : visitor::PreOrder {
...
+        // to use in C++ codegen. These expressions have no actual equivalent
+        // in the syntax.
+        //
+        // To still somehow capture them in HILTI output render them by
+        // declaring a variable with no constructor arguments; this is distinct
+        // from HILTI default initialization which uses assignment syntax. This
+        // is not 100% equivalent, but allows rendering valid code.
+        if ( auto* init = n->init() ) {
+            if ( init->isA<expression::Void>() )
+                _out << "()";
```
rsmmr, line 442: An alternative could be rendering without the `()` and adding a comment, like:

```
local tuple<view<stream>, int<64>, const iterator<stream>, optional<hilti::RecoverableFailure>> __result; # no-init
```

But either way is fine.

rsmmr (approved):

hilti/toolchain/src/compiler/printer.cc
bbannier: I would have preferred that, but it looks like we no syntax for inline comments. We cannot directly inject it here since the `;` is rendered elsewhere, but we could probably do some nasty tricks like printing it explicitly here before the comment; this would lead to code like

```
T  __result; # no-init;
```

PR #2058, opened 2025-05-14
Fix `bytes` iterator dereference operation.

### LINKED ISSUE #2057: Incorrect coercion of bytes iterator leads to C++ compiler error
A user reported a failure like this on Slack:

```ruby
module foo;

global xs: vector<tuple<data: uint8>>;

for (i in b"") {
    xs.push_back((i, ));
}
```

When compiling this we run into a C++ compiler error since the value bound to `i` is an unwrapped `uint8`, not a wrapped integer we expect elsewhere.

This is an issue similar to #2047 which I hoped we didn't need to change to wrapping since it was unclear to me what the performance implications in iteration-heavy code might be, but I was wrong.

We should probably convince ourself with an iteration-heavy benchmark that wrapping adds negligible overhead and simply wrap these; if overhead shows up we should extend the tuple machinery to also handle `uint8` values.

A fix should be backported to `release/1.11` and `release/1.13`.

rsmmr:
It now returns a safe integer.

We likely have further instances of this issue, and may eventually
need to change the internal storage for `bytes` and `stream` to safe
integers. Leaving that for later to limit impact here.

Closes #2057.


bbannier (dismissed):
Thanks, this looks good.

I also looked into this issue this morning. On my perf concerns from the ticket, the current implementation was already doing many other conversions to safe integers internally and also pulled a lock on the bytes control block, so it already was pretty expensive; making this additional conversion makes only a small difference of a couple percent.

rsmmr:
> already doing many other conversions to safe integers internally

Yeah, that was my thinking already as well. Good to verify.

PR #2062, opened 2025-05-15
Fix cyclic usage of units types inside other types.

### LINKED ISSUE #2061: Recursive `tuple` fail when compiling C++
The [Redis analyzer](https://github.com/evantypanski/spicy-redis) currently fails to compile. I believe this is a regression, since I had compiled it before (at least since putting in the [Redis analyzer PR in Zeek](https://github.com/zeek/zeek/pull/4288))

I narrowed it down a bit, this seems to reproduce the issue for me:

```
module Test;

type Data = unit {
    var x: tuple<Data>;
};
```

It seems related to the recursive `tuple<Data>` inside of a unit variable. Note that this is a bit more indirect in the analyzer. Also note that `vector<Data>` compiles.

It's actually hard for me to diagnose right now since I'm getting other compiler errors at older releases too, so maybe some compiler shenanigans are happening somewhere.

If it's useful: `spicy-config --cxx` outputs a compiler that is `Apple clang version 16.0.0 (clang-1600.0.26.6)`

rsmmr:
We had cyclic dependencies partially solved through the `&on-heap`
conversion, but weren't consistently applying it to types nested
inside others.

Closes #2061.

- **Add API method to ctors to see if they create a reference.**
- **Small tweak to builder API so that we can create index expression with operands other than integers.**
- **Fix cyclic usage of units types inside other types.**

### DIFF
--- a/hilti/toolchain/include/ast/builder/builder.h
+++ b/hilti/toolchain/include/ast/builder/builder.h
@@ -314,8 +314,8 @@ public:
         return expressionUnresolvedOperator(operator_::Kind::Call, {id(id_, m), tuple(v, m)}, m);
     }
 
-    auto index(Expression* value, unsigned int index, const Meta& m = Meta()) {
-        return expressionUnresolvedOperator(operator_::Kind::Index, {value, integer(index, m)}, m);
+    auto index(Expression* value, Expression* index, const Meta& m = Meta()) {
+        return expressionUnresolvedOperator(operator_::Kind::Index, {value, index}, m);
     }
 
     auto size(Expression* op, const Meta& m = Meta()) {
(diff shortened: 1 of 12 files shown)


rsmmr:
Cool. 

I a bit worried about merging this into the upcoming release since the change is quite subtle and I wouldn't be surprised if there edge cases where things might need more tweaking. That said, I confirmed that the analyzers built into Zeek are working.

bbannier:
> That said, I confirmed that the analyzers built into Zeek are working.

It also seems to work with analyzers pulled in by https://github.com/zeek/spicy-analyzers.

rsmmr:
Ok, a bit more cleanup and now ready for review. I'd definitely merge this into `main`, releases branches we can decide.

evantypanski (approved):
Ok I spent a while trying to break this and couldn't find anything. I also like that it gets rid of some random-looking numbers in the `parent` calls

PR #2067, opened 2025-05-19
Error for redefined locals from statement inits.

### LINKED ISSUE #2065: `while` and `if` initializers sometimes lead to C++ errors
The following Spicy code has an error in C++:

```
module Test;

while (local i: int32 = 1; i < 10) {
    local i: int32 = 2;
}
```

Error:

```
$ spicyc -j test.spicy
/private/var/folders/w6/h7tg594x0yng0_yzvcsn6mv00000gn/T/Test_859d06215870f69e-ab01021f8b6c055f.cc:22:45: error: redefinition of 'i'
   22 |         ::hilti::rt::integer::safe<int32_t> i = ::hilti::rt::integer::safe<std::int32_t>{2};
      |                                             ^
/private/var/folders/w6/h7tg594x0yng0_yzvcsn6mv00000gn/T/Test_859d06215870f69e-ab01021f8b6c055f.cc:20:47: note: previous definition is here
   20 |     for ( ::hilti::rt::integer::safe<int32_t> i = ::hilti::rt::integer::safe<std::int32_t>{1}; i < ::hilti::rt::integer::safe<std::int32_t>{10};  ) {
      |                                               ^
1 error generated.
```

Also happens for `if`

This could should be rejected with `redefinition of 'i' defined in`

evantypanski:
Closes #2065

This now properly errors in Hilti from code that redefines a local inside a statement with an initializer, like:

```
for (i in vec) {
    local int<64> i = 5;
}
```

This avoids a C++ error later.

It seems fine to add to two different scopes - the node needs it for the condition, then the body needs it for redefinitions. Otherwise it seems like it'll cause some hacky cases one way or another, but not sure if there are edge cases I'm missing.


hilti/toolchain/src/compiler/scope-builder.cc
```diff
@@ -120,23 +118,42 @@ struct Visitor : visitor::PostOrder {
...
+    }
 
     void operator()(statement::If* n) final {
-        if ( n->init() )
-            n->getOrCreateScope()->insert(n->init());
+        if ( auto* init = n->init() ) {
+            n->getOrCreateScope()->insert(init);
+
+            // Also add this to the true body to avoid redefinitions
+            n->true_()->getOrCreateScope()->insert(init);
```
bbannier, line 133: We need it for the `else` branch as well since the following is valid Spicy code (modelling semantics also in e.g., C):

```ruby
if (local a = 0) {
    # Empty.
} else {
    print a;
}
```

Let's also add a test for this.

hilti/toolchain/src/compiler/scope-builder.cc
```diff
@@ -120,23 +118,42 @@ struct Visitor : visitor::PostOrder {
...
     }
 
     void operator()(statement::While* n) final {
-        if ( auto* x = n->init() )
-            n->getOrCreateScope()->insert(x);
+        if ( auto* init = n->init() ) {
+            n->getOrCreateScope()->insert(init);
+
+            // Also add this to the body to avoid redefinitions
+            n->body()->getOrCreateScope()->insert(init);
```
bbannier, line 155: Ditto, but for HILTI code:

```ruby
module foo {

import hilti;

while ( local auto a = 0 ) {
    # Empty.
} else {
    hilti::print(a);
}

}
```

This should also have a test.

bbannier (changes requested):

hilti/toolchain/src/compiler/scope-builder.cc
evantypanski: Dang, I should've tested that, I thought about it but thought it'd be weird if the scope extended inside the `else` - my bad!

PR #2068, opened 2025-05-20
Only build benchmarks and tests when building test targets

bbannier:
This came up during https://github.com/Homebrew/homebrew-core/pull/224043. I also changed the setup so unit tests are not build by default.


rsmmr (commented):
> They are not interesting for Spicy users, but add considerable build
time.

I'd prefer keeping to build these by default. Having them built implicitly when running `ctest.all` seems surprising, including because *that* is now slow the first time it's running. Plus, during development, I do sometimes catch stuff I broke because some test isn't compiling (and I never run "ninja check`).

bbannier:
> I'd prefer keeping to build these by default.

In that case I'd like to keep the current logic of not adding the to `ALL`, but I'll add a user-facing flag defaulting to `ON` which which forces building them.

rsmmr:
I'm not sure what the problem is with generally always building them?

bbannier:
> I'm not sure what the problem is with generally always building them?

They add build time to anybody building Spicy since they add ~100 additional compilation jobs. Most upstream projects do not care about it, e.g., we probably do not care in zeek/zeek. The reason this originally came up was that for Homebrew the zeek-7.2 formula failed to build since google-benchmark did not build there; I believe they absolutely do not care about our benchmarks or tests.

3rdparty/CMakeLists.txt
```diff
@@ -18,7 +18,10 @@ set(CMAKE_CXX_FLAGS "${flags}")
 option(DOCTEST_NO_INSTALL "Skip the installation process" ON)
 add_subdirectory(doctest)
 
-option(JUSTRX_HAVE_BENCHMARK "Benchmark is provided" ON)
+set(DISABLE_TESTS ON)
+# If we'd enable tests above we would need to pick up the benchmarking
+# library here to avoid declaring the target twice.
+set(JUSTRX_HAVE_BENCHMARK ON)
```
bbannier, line 24: If you are wondering why this changed from a new option to setting the already existing variable, I made this change since it was slightly unusual to shadow an option from justrx (though it still worked on my CMake version); it is also different for how we treat any other 3rdparty library's flags. I fixed building of benchmarks in https://github.com/rsmmr/justrx/pull/15 so it should now be possible to build and run them there.

CMakeLists.txt
```diff
@@ -275,6 +275,14 @@ add_custom_target(
 add_custom_target(tests COMMENT "Building test targets")
 add_dependencies(tests hilti-tests spicy-tests)
 
+# By default Test targets are not built, but we want them to be build as part
+# of the default dev experience. Projects using Spicy can still disable them
+# when bundling Spicy by setting the the option.
+option(SPICY_ENABLE_TESTS "Build Spicy unit tests and benchmarks" ON)
```
rsmmr, line 281: This comment feels misleading: by default, they *are* on, no? (I assume the comment refers to `EXCLUDE_FROM_ALL`, but still, suggest to rephrase.)

rsmmr (approved):
This works for me, thanks. Indeed, this would have been my next suggestion as well: introduce a setting for building the tests, but keep it on by default for a pure Spicy build.

I'll look at the justrx PR in a minute.

CMakeLists.txt
bbannier: This was unintentional, I simply confused myself. This should be fixed now.

rsmmr (dismissed):

PR #2069, opened 2025-05-22
Coverity fixes

evantypanski:
Round of fixes for some issues marked as being fixed but were probably missed in the awesome UI. I did not dig to find out if they're actually fine, I'd rather just shut them up :)

There are a couple of "uncaught" exceptions in `spicy-driver.cc` that may need to be manually resolved, I think they were fixed by 6fc414d? Not looking super hard, though - maybe I missed something. CIDs 1586737 & 1586307

Also fixing 1586205 caused a `stack-use-after-scope` address sanitizer issue [in CI](https://cirrus-ci.com/task/6232057568821248?logs=build#L462) so I dumped it

### DIFF
--- a/spicy/toolchain/src/compiler/codegen/parsers/types.cc
+++ b/spicy/toolchain/src/compiler/codegen/parsers/types.cc
@@ -109,7 +109,7 @@ struct Visitor : public visitor::PreOrder {
     Expression* result = nullptr;
 
     auto pb() { return tp->pb; }
-    auto state() { return pb()->state(); }
+    const auto& state() { return pb()->state(); }
     auto builder() { return pb()->builder(); }
     auto context() { return pb()->context(); }
     auto pushBuilder(std::shared_ptr<Builder> b) { return pb()->pushBuilder(std::move(b)); }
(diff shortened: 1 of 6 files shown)


spicy/toolchain/src/compiler/codegen/parsers/types.cc
```diff
@@ -108,7 +108,7 @@ struct Visitor : public visitor::PreOrder {
     TypeParser* tp;
     Expression* result = nullptr;
 
-    auto pb() { return tp->pb; }
+    auto& pb() { return tp->pb; }
```
bbannier, line 111: Nit:

```suggestion
    auto* pb() { return tp->pb; }
```

bbannier (approved):
Thanks!

Could you wait for the lint job to finish? It is currently running [here](https://cirrus-ci.com/task/6509412799479808).

spicy/toolchain/src/compiler/codegen/parsers/types.cc
evantypanski: oops that was the wrong thing 🤦 it was meant to go on state, which is a const reference

bbannier (approved):

PR #2073, opened 2025-05-27
Prevent throwing naked exception when yielding from aborted fiber.

bbannier:
It is not clear to me why we did not throw a `RuntimeError` in the first place, but in any case we did not provide a proper stringification which would lead to a very cryptic error for this already obscure error scenario.

With this patch we instead throw a Runtime error with some context.


rsmmr (approved):
I'm still not sure yet this is the actual problem but the change looks fine either way, so good to merge.

PR #2074, opened 2025-05-28
Fix fiber abortion.

rsmmr:
When aborting a fiber, we need to activate it once more, to then leave
it for good by raising an `AbortException`. Problem was that that
exception ended up being caught by user code because it was derived
from `std::exception`. This change removes the base class so that the
exception is guaranteed to go back to the managing fiber code, where
we just ignore it.

### DIFF
--- a/hilti/runtime/src/fiber.cc
+++ b/hilti/runtime/src/fiber.cc
@@ -257,8 +257,11 @@ detail::Fiber::Fiber(Type type) : _type(type), _fiber(std::make_unique<::Fiber>(
     };
 }
 
-HILTI_EXCEPTION(AbortException, RuntimeError);
-AbortException::~AbortException() = default;
+// Exception raised by a fiber resuming operation in case it has been aborted
+// in the meantime. This must not be derived from `std::exception` to guarantee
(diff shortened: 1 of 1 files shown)


bbannier (approved):
Thanks for the quick fix, this is pretty subtle.

Throwing this exception is pretty weird, but we need to be able to skip over user code following calls to `yield`, so changing this to use e.g., a return value from `Fiber::yield` would at the least require rewriting generated code if it is at all possible. At least now we have some documentation for `AbortException`.

Could you cherry-pick this into `release/1.13`? With that we can mark zeek/zeek#4509 as needing a backport for zeek-7.2.x.

rsmmr:
> Fiber::yield would at the least require rewriting generated code if it is at all possible.

Yeah, that would be major change across the code, and not for the better I'd say.

> At least now we have some documentation for AbortException.

:-) 

> Could you cherry-pick this into release/1.13? 

Will do.

bbannier:
Backported to 1.13 as 3a7e11b1106319dcac947e263f8329ad97a30720.

PR #2075, opened 2025-05-29
Cleanup HILTI resolver's `checkForHooks`.

evantypanski:
This started just cleaning up the function with cases that didn't seem possible, but I realized that this function just always returns `false`. It never gets called with multiple matches on a hook function, so in practice this function just adds a bit of unnecessary complexity. Maybe I'm missing something?

This is nitpicky though, so if we'd rather not remove this that's fine by me, it just seemed a bit misleading when I was trying to understand how hooks are resolved (and, indeed, did mislead me a bit).


rsmmr (approved):
It's a good change. Seems the need for that function went away with some earlier changes to the operator resolving.

PR #2076, opened 2025-06-03
Add codspeed integration

bbannier:


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,56 @@
+# Copyright (c) 2020-2023 by the Zeek Project. See LICENSE for details.
```
awelzel, line 1:
```suggestion
# Copyright (c) 2025- by the Zeek Project. See LICENSE for details.
```

?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,56 @@
...
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  benchmarks:
+    name: Run benchmarks
+    runs-on: ubuntu-22.04
```
awelzel: Any reason not to go to 24.04?

awelzel (approved):
:tada:

.github/workflows/benchmark.yml
bbannier: This is the copyright line we use consistently in Spicy. Since there is some automation around updating it I'd like to keep this consistent (I am also not really familiar with it). I believe we made a move away from date ranges anyway, so I'd suggest cleaning that up separately.

.github/workflows/benchmark.yml
bbannier: Nothing beyond keeping this consistent with [our default CI image](https://github.com/zeek/spicy/blob/992444228151b193379e93403f8d486a6bf37878/ci/Dockerfile). Since I'd like to bump this to ubuntu-24.04 in https://github.com/zeek/spicy/pull/2077 to be able to check with more recent versions of GCC I'll bump here as well.

PR #2077, opened 2025-06-03
Speed up checking of iterator compatibility

### LINKED ISSUE #1663: bytes for loop performance
In the context zeek/zeek#3580 it was observed that a `for` loop over a bytes object to xor decode the content takes ~30% of the total runtime in that setting. This ticket is meant as a generic "speed up bytes iteration".

The test is iterating 500k times over a 134 bytes instance in `%init`. The SPICY_THREADED=1 version creates a short lived thread in spicy-driver's main to avoid glibc using an optimized non-atomic shared_ptrs (#1600) This reflects better the performance within an application like Zeek and makes performance significantly worse. Python is added as baseline which executes significantly faster.

No sure how relevant this is, for looping over large bytes might not be a common use case.

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `echo "x" \| ../build/bin/spicy-driver loop.hlto` | 4.922 ± 0.022 | 4.903 | 4.945 | 1.76 ± 0.01 |
| `echo "x" \| SPICY_THREADED=1 ../build/bin/spicy-driver loop.hlto` | 7.147 ± 0.015 | 7.132 | 7.161 | 2.56 ± 0.01 |
| `python3 loop.py` | 2.791 ± 0.007 | 2.783 | 2.797 | 1.00 |

The non-dwarf flamegraph (suspect std::shared_ptr above the flat bars like in zeek/zeek#3580).

flame

```
#loop.spicy
module Loop;

global xs: bytes = b"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";

public type X = unit {

 on %init {
  local i = 0;
  local sum  = 0;
  while ( i < 500000 ) {
    for ( c in xs )
      sum += c;
    i++;
  }
  print "sum", sum;
 }
};
```

```
# loop.py
xs = b"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"


def main():
    i = 0
    s = 0
    while i < 500000:
        for c in xs:
            s += c
        i += 1

    print("sum", s)


if __name__ == "__main__":
    main()
```

bbannier:
We were previously using a control block which held a weak_ptr to the
protected data. This was pretty inefficient for a number of reasons:

- access to the controlled data always required a `weak_ptr::lock` which
  created a temporary shared_ptr copy and immediately destroyed it after
  access
- to check whether the control block was expired we used `lock` instead
  of `expired` which introduced the same overhead
- to check compatibility of iterators we compared shared_ptrs to the
  control data which again required full locks instead of using
  `owner_before`

This patch introduces a new control block data structure and uses it
across all classes which previously held ad hoc implementations
(`Bytes`, `Map`, `Set`, `Vector`). The main improvement is that we now
separate tracking liveliness and the data, and use a better
implementation for control block equality checks. With the new
implementation I see throughput improvements across the board for
anything needing to iterate, e.g., I see bytes iteration being up to 30x
faster in trivial setups; the code from the original issue is now 10x
faster.

Closes #1663.


hilti/runtime/include/util.h
```diff
@@ -649,4 +650,95 @@ constexpr std::size_t hashCombine(std::size_t hash1, std::size_t hash2, Hashes..
...
+
+    ControlBlockRef& operator=(const ControlBlockRef&) = default;
+    ControlBlockRef& operator=(ControlBlockRef&&) = default;
+
+    bool isValid() const { return _data && ! _control.expired(); }
+
+    // Get reference to the controlled object. Throws if the object has expired or the data is invalid..
+    const Data& get() const {
+        if ( ! _data )
+            throw InvalidIterator("iterator is invalid");
```
bbannier: It is slightly weird how we use iterator-specific exceptions in a class which claims to be pretty general, but this captures our current use. Instead of doing that we could also make this less general (e.g., move to some `iterator.h`), or allow users to pass the exception type as a template parameter (then we'd probably change at least this error message).

bbannier: I added a commit doing the latter.

hilti/runtime/include/util.h
```diff
@@ -649,4 +650,97 @@ constexpr std::size_t hashCombine(std::size_t hash1, std::size_t hash2, Hashes..
         return result;
 }
 
+template<typename Data, typename Error>
+class ControlBlockRef;
+
+// Helper class for loosely tracking liveliness of some memory. Code using this
+// would instantiate an instance of this pointing to some data. It can then
+// pass out references to that control block by calling `asRef`. One can get
+// values from these references as long as the original control block is live.
```
rsmmr:
```suggestion
// values from these references as long as the original control block is live (which is checked).
```

hilti/runtime/include/util.h
```diff
@@ -649,4 +650,97 @@ constexpr std::size_t hashCombine(std::size_t hash1, std::size_t hash2, Hashes..
         return result;
 }
 
+template<typename Data, typename Error>
+class ControlBlockRef;
+
+// Helper class for loosely tracking liveliness of some memory. Code using this
```
rsmmr: Please document the new classes, incl. at least public members, using Doxygen-style.

hilti/runtime/include/util.h
```diff
@@ -649,4 +650,97 @@ constexpr std::size_t hashCombine(std::size_t hash1, std::size_t hash2, Hashes..
...
+    ControlBlock& operator=(const ControlBlock&) = default;
+    ControlBlock& operator=(ControlBlock&&) = default;
+
+    friend bool operator==(const ControlBlock& a, const ControlBlock& b) {
+        return std::tie(a._control, a._data) == std::tie(b._control, b._data);
+    }
+
+    friend bool operator!=(const ControlBlock& a, const ControlBlock& b) { return ! (a == b); }
+
+    /*implicit*/ operator Ref() const {
```
rsmmr:
```suggestion
    /* implicit */ operator Ref() const {
```

hilti/runtime/include/util.h
```diff
@@ -649,4 +650,97 @@ constexpr std::size_t hashCombine(std::size_t hash1, std::size_t hash2, Hashes..
...
+        if ( ! _control )
+            _control = std::make_shared<bool>();
+
+        return {_control, _data};
+    }
+
+    // Invalidate all references to this control block.
+    void reset() { _control.reset(); }
```
rsmmr: Should this clear `_data` as well, just to avoid accidental continued use? (Or is it explicitly supported to create a new control pointer for the same data?)

hilti/runtime/include/util.h
```diff
@@ -649,4 +650,97 @@ constexpr std::size_t hashCombine(std::size_t hash1, std::size_t hash2, Hashes..
...
 
+template<typename Data, typename Error>
+class ControlBlockRef;
+
+// Helper class for loosely tracking liveliness of some memory. Code using this
+// would instantiate an instance of this pointing to some data. It can then
+// pass out references to that control block by calling `asRef`. One can get
+// values from these references as long as the original control block is live.
+template<typename Data, typename Error>
+class ControlBlock {
```
rsmmr: How about moving the new classes into their own `rt::util::control` namespace (`control::Block`, `control::Reference`)? That's closer to how we structure things elsewhere, and a bit more clear too I'd say.

hilti/runtime/include/util.h
```diff
@@ -649,4 +650,97 @@ constexpr std::size_t hashCombine(std::size_t hash1, std::size_t hash2, Hashes..
...
+    using Ref = ControlBlockRef<Data, Error>;
+
+    ControlBlock() = default;
+    ControlBlock(Data* data) : _data(data) {}
+
+    ControlBlock(const ControlBlock&) = default;
+    ControlBlock(ControlBlock&&) = default;
+
+    ControlBlock& operator=(const ControlBlock&) = default;
+    ControlBlock& operator=(ControlBlock&&) = default;
```
rsmmr: I'm wondering if we should `delete` these. Everything still compiles if I do, and conceptually it seems we don't really *want* to create multiple copies of the same block, do we?

hilti/runtime/include/util.h
```diff
@@ -649,4 +650,97 @@ constexpr std::size_t hashCombine(std::size_t hash1, std::size_t hash2, Hashes..
...
+
+    ControlBlockRef& operator=(const ControlBlockRef&) = default;
+    ControlBlockRef& operator=(ControlBlockRef&&) = default;
+
+    bool isValid() const { return _data && ! _control.expired(); }
+
+    // Get reference to the controlled object. Throws if the object has expired or the data is invalid..
+    const Data& get() const {
+        if ( ! _data )
+            throw Error("bound object is invalid");
```
rsmmr:
```suggestion
            throw Error("underlying object is invalid");
```

As this is potentially user-facing (more instances below).

rsmmr (commented):
I'm a bit torn here. It's great to see the speed-up of course, and factoring out the control block logic makes a lot of sense too. However, there's a drawback of not relying on `lock()` anymore to get access to the data: the code is no longer thread-safe. 

That's one of these things that we currently don't need: the way we use Spicy doesn't require thread-safety, and we have some other places that aren't thread-safe either. But iterators are a pretty fundamental piece to the runtime, and if we ever wanted to become thread-safe, we'd have to revert some of this (and it would probably take a while to even just notice the issue).

Let me ask: the PR does multiple optimizations. Do you know how much performance we'd lose if we switched back to access to the controlled data through `shared_ptr`+lock, but left the other improvements (`expired`/`owner_before`) in?

hilti/runtime/include/util.h
bbannier: Hmm, I think we need the copy and move ctrs and assignment operators so we can use these in classes for which we want to support these operations. The implementations for copies should not copy the control block though, otherwise we'd get spooky action at a distance where a iterators into a container copy would get invalidated if the original object gets updated; I'll update these to use a separate `_control`.

hilti/runtime/include/util.h
bbannier: Setting `_data` here does nothing since all access to it is via a reference instance which store their own value. I believe for an expired reference it should be impossible to get access to the value behind that `_data` since all access to data is guarded by checks for the reference's `_control`.

bbannier:
> I'm a bit torn here. It's great to see the speed-up of course, and factoring out the control block logic makes a lot of sense too. However, there's a drawback of not relying on `lock()` anymore to get access to the data: the code is no longer thread-safe.
> 
> [..]
> 
> Do you know how much performance we'd lose if we switched back to access to the controlled data through `shared_ptr`+lock, but left the other improvements (`expired`/`owner_before`) in?

At least for the `iterate_bytes benchmark` not going through the `shared_ptr` costs about 2x more, but the more significant improvement is in the `owner_before`-style check for compatibility.

That said, I believe thread-safety is a red hering since even before access to the data behind the iterator was not threadsafe (e.g., `vector` data could be mutated concurrently while we accessed it through a still valid iterator). The only thread-safety feature of `shared_ptr` is that its refcount is atomic, and a threadsafe runtime library we would need a lot of reworking in how we store and access data. If user code was using iterators to access const containers concurrently before this should still work now.

I'd still opt for accessing the data via the controlled raw ptr since 2x is a significant performance impact for loop-heavy code. With the new data structure we will have a single place to work on should we ever try to make the runtime library threadsafe.

CMakeLists.txt
```diff
@@ -109,7 +109,14 @@ if (USE_SANITIZERS)
 endif ()
 
 if (USE_WERROR)
-    set(werror_flags "-Werror")
+    set(werror_flags "-Werror -Wno-error=array-bounds")
+
+    # With versions >=13.0 GCC gained `-Warray-bounds` which reports false
+    # positives, see e.g., https://gcc.gnu.org/bugzilla/show_bug.cgi?id=111273.
+    if (CMAKE_COMPILER_IS_GNUCXX AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 13.0)
+        set(werror_flags "-Wno-error=array-bounds")
```
timwoj: Doesn't this overwrite the previous setting of `werror_flags`, removing `-Werror`?

hilti/runtime/include/util.h
```diff
@@ -649,4 +650,122 @@ constexpr std::size_t hashCombine(std::size_t hash1, std::size_t hash2, Hashes..
...
+
+/// Helper class for loosely tracking liveliness of some memory.
+///
+/// Code using this would instantiate an instance of this pointing to some data.
+/// It can then pass out references to that control block by calling `Ref`.
+/// One can get values from these references as long as the original control
+/// block is live (this is checked).
+///
+/// \tparam Data the type of data this Block controls
+/// \tparam Error the type of exception a Reference should throw if it became stale
```
rsmmr: Sorry, I don't want to be a pain, but that's not the style we use elsewhere. Maybe my "Doxygen-style" was misleading: what I meant is use `/** ... */` and `@xxx`; and no ` @brief` needed (which I see further below), first sentence is always the summary. That's more like "Javadoc-style" I guess? Essentially let's just try to stay consistent with our comment/doc style.

bbannier: No problem at all, this just shows that I don't have enough practice writing Doxygen comments 🙈.

re: `\brief`, at least with my Doxygen version this would correctly emit the summary in the class overview (the method view always shows everything so it doesn't matter there), but I changed this to be consistent with our current practice for now.
Screenshot 2025-06-06 at 11 31 27 AM

Screenshot 2025-06-06 at 11 33 06 AM

rsmmr: > \brief, at least with my Doxygen version this would correctly emit the summary in the class overview (

Right, Doxygen is flexible in what it parses, I indeed meant this for consistency with our practice elsewhere.

PR #2079, opened 2025-06-04
Fix C++ compilation issue with weak->strong refs.

### LINKED ISSUE #2063: Forwarded unit variable reference causes C++ compilation issue
I found this when messing around with Redis analyzer code for a workaround with #2061. This isn't actually coming up in real code, just something I noticed.

The given Spicy code has a C++ compilation error:

```
module Test;

public type Data = unit(depth: uint8&) {
    inner: Inner(depth);
};

type Inner = unit(depth: uint8&) {
    x: uint64;
};
```

Error:

```
$ spicyc -j test.spicy
/private/var/folders/w6/h7tg594x0yng0_yzvcsn6mv00000gn/T/Test_6d63c02a2073970a-893bb4472c469795.cc:128:64: error: ambiguous conversion for functional-style cast from 'ValueReference<SafeInt<unsigned char, SafeIntException>>' to '__hlt::Test::Inner'
  128 |         ::hilti::rt::optional::valueOrInit((*__self).inner) = (__hlt::Test::Inner(__self->__p_depth.derefAsValue()));
      |                                                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/private/var/folders/w6/h7tg594x0yng0_yzvcsn6mv00000gn/T/Test_6d63c02a2073970a-893bb4472c469795.cc:70:12: note: candidate constructor
   70 |     Inner::Inner(::hilti::rt::StrongReference<::hilti::rt::integer::safe<uint8_t>> __p_depth) : __p_depth(std::move(__p_depth)) {
      |            ^
/private/var/folders/w6/h7tg594x0yng0_yzvcsn6mv00000gn/T/Test_6d63c02a2073970a-893bb4472c469795.cc:73:12: note: candidate constructor
   73 |     Inner::Inner(std::optional<std::optional<::hilti::rt::integer::safe<uint64_t>>> x) : Inner() {
      |            ^
1 error generated.
```

~I'm kind of confused about this case, changing the type of `Inner::x` makes it not fail, same with removing the `&convert`. So this is as minimal as I'm getting right now :)~

Seems like there's a conflict between constructing the version with a unit variable and the ctor with `x`. It doesn't quite trigger when I'd expect it to, though.

evantypanski:
Fixes #2063

strong refs can't get coerced from weak refs directly

Also did a little cleanup since I don't like single-letter variable names :)


tests/hilti/types/reference/strong-from-weak-ref.hlt
```diff
@@ -0,0 +1,21 @@
+# @TEST-EXEC: ${HILTIC} -j %INPUT >output
+# @TEST-EXEC: btest-diff output
```
rsmmr, line 4: Let's add a `@TEST-DOC`.

tests/spicy/types/unit/params-conflict.spicy
```diff
@@ -0,0 +1,13 @@
+# @TEST-DOC: Ensures the unit parameter constructor doesn't conflict with the field
```
rsmmr: Suggest to add the ticket number where this is coming from as well. (I usually add `; regression test for #xx`, but doesn't really matter how.)

rsmmr (approved):

PR #2080, opened 2025-06-05
Catch if methods are implemented multiple times.

### LINKED ISSUE #2044: Hilti allows multiple method definitions, leading to C++ error
I can't find a way to reproduce this via Spicy, so this probably isn't a big deal. The following (hilti) code leads to a C++ compiler error:

```
module Test {

import hilti;

type T = struct {
    method int<64> test() &static { return 1; }
};

# Conflicting definition!
method int<64> T::test() &static {
    return 2;
}

global T t;

t.test();

}
```

Error:

```
$ hiltic -j test.hlt
/private/var/folders/w6/h7tg594x0yng0_yzvcsn6mv00000gn/T/Test_99c07d899d482988-e82c8fc9d785225f.cc:50:22: error: redefinition of 'test'
   50 | auto __hlt::Test::T::test() -> ::hilti::rt::integer::safe<int64_t> {
      |                      ^
/private/var/folders/w6/h7tg594x0yng0_yzvcsn6mv00000gn/T/Test_99c07d899d482988-e82c8fc9d785225f.cc:45:29: note: previous definition is here
   45 | inline auto __hlt::Test::T::test() -> ::hilti::rt::integer::safe<int64_t> {
      |                             ^
1 error generated.
[error] hiltic: JIT compilation failed
```

Note that this is a proper hilti error for functions, so overload checking exists, just doesn't get used for methods I think. This also happens if you just have multiple definitions of a function, you don't need the inline one.

rsmmr:
Closes #2044.


hilti/toolchain/src/compiler/validator.cc
```diff
@@ -136,6 +136,8 @@ struct VisitorPre : visitor::PreOrder, public validator::VisitorMixIn {
 struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
     using hilti::validator::VisitorMixIn::VisitorMixIn;
 
+    std::unordered_set<uint32_t> _method_declarations; // tracks methods already seen
```
evantypanski: Maybe add a comment that ~this~ `uint32_t` is based on the `ast::DeclarationIndex` if that won't get hashed? This just feels a little unintuitive to me until I looked at the code that used it. That could also just be my lack of experience with that in the Spicy code, though :)

evantypanski (approved):

hilti/toolchain/src/compiler/validator.cc
rsmmr: Even better: I've made the `DeclarationIndex` hashable, so that we can use it as the index directly.

PR #2082, opened 2025-06-05
Enforce that block-local declarations must be variables.

### LINKED ISSUE #1988: Compiler segfault for block-local declared struct
The following incorrect code fails with a segfault at compile time:

```ruby
module Test;

for (x in b"123") {
    type X = struct {
        a: uint8;
    };
}
```

Looking at the backtrace, the error comes from [this access](https://github.com/zeek/spicy/blob/c9ed1a1254c166a73d46f5390192bc12c92343db/hilti/toolchain/src/compiler/scope-builder.cc#L121) to `Declaration::declaration()` which returns a nullptr.

This has been an issue since at least v1.4.0.

rsmmr:
Logically we already had this restriction in place, but so far it
would lead to either a crash or an internal error.

Closes #1988.

### DIFF
--- a/hilti/toolchain/src/compiler/validator.cc
+++ b/hilti/toolchain/src/compiler/validator.cc
@@ -555,6 +555,11 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
         }
     }
 
+    void operator()(statement::Declaration* n) final {
+        if ( ! n->declaration()->isA<declaration::LocalVariable>() )
+            error("only variables can be declared inside local scopes", n);
+    }
+
     void operator()(statement::Return* n) final {
         auto* func = n->parent<Function>();
 


hilti/toolchain/src/compiler/validator.cc
```diff
@@ -555,6 +555,11 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
         }
     }
 
+    void operator()(statement::Declaration* n) final {
+        if ( ! n->declaration()->isA<declaration::LocalVariable>() )
+            error("only variables can be declared inside local scopes", n);
```
evantypanski: I have a couple nitpicks here

1) Both `const` and `type` are allowed by the parser, then later rejected. I like that because they get nicer error messages (though maybe it could apply to more decls). But - the `const` check is in [Spicy's validator](https://github.com/zeek/spicy/blob/207e1cb51e5cdb970f19b802f56a43d2f26e6aba/spicy/toolchain/src/compiler/validator.cc#L320-L324) and has a different message. Maybe those can be combined into this one?
2) Since globals are variables too, it would be nice if this said locals:

```suggestion
            error("only local variables can be declared inside local scopes", n);
```

rsmmr: > But - the const check is in [Spicy's validator](https://github.com/zeek/spicy/blob/207e1cb51e5cdb970f19b802f56a43d2f26e6aba/spicy/toolchain/src/compiler/validator.cc#L320-L324) and has a different message. Maybe those can be combined into this one?

Indeed, that const check can just go, it's covered by the new check. Done.

> error("only local variables can be declared inside local scopes", n);

That sounds slightly misleading to me. At least the way I read it, it would put the emphasis on "*local* variables", as a contrast to "*global* variables". So leaving this, but slightly tweaking it to add a "(not XXX)", with `XXX` being the declaration what the validator found.

Also adding a test that I forgot to commit.

PR #2083, opened 2025-06-05
Remove unused function parameters

evantypanski:
This is a draft - it's pretty (...very) rough, just thought I'd put it in as a draft since all tests pass (well I ruined one when cleaning up but I added a comment about it)

There are still open questions about how to track declarations (which, I tried a few things and they didn't work great, so reverted to IDs - that's probably bad though). I also have a bunch of random TODOs in there since the quality is pretty rough. But, here's a start :)

The main thing TODO, though, is to actually place remove function arguments before the call itself, just in case one of them has a side effect.


bbannier:
This seems to treat the "public module interface" differently than other places where we rewrite code. The general assumption we make elsewhere is that interface (well, at least its methods and hooks) of `public` units might be used in code we do not see (in particular: custom C++ code), so we do not rewrite it; for non-`public` units OTOH we assume that we always see all code. What this patch does is different in that methods of public interface need to be explicitly declared with `&keep-signature`.

If this pass used `public` like others it probably would need to be structured slightly differently.

Re: hooks, these usually do not pass a lot of data, so probably okay to ignore for now, else we'd need to analyze all bodies (unless hook is in `public` unit and could be implemented in C++ we do not see).

evantypanski:
> The general assumption we make elsewhere is that interface (well, at least its methods and hooks) of public units might be used in code we do not see (in particular: custom C++ code)

Yeah I have a check to see if the *field* is public. I can add one for functions. Then checking for public structs would just be checking its parent like I do for `cxxname`, though that could probably just be changed to just be a check for public. Is there more there? All I see is one extra condition and one changed condition.

I mean, I'd prefer to still be able to optimize the like `__parse_stage1`-type functions with this, but if those are guaranteed to stick around exactly then that's a big chunk of potential speedup gone. I figured the `parseX` functions were the only entrypoints

Meanwhile I pushed a fixup to not remove parameters if the parent type is public, which cuts down on the opportunities to optimize a decent amount

> What this patch does is different in that methods of public interface need to be explicitly declared with &keep-signature.

I think it's still necessary in some cases, even for non-public units, because the [parse method signatures are baked into the runtime](https://github.com/zeek/spicy/blob/main/spicy/runtime/include/typedefs.h#L21-L35). Even with checks for public units, removing that `&keep-signature` makes some tests fail to compile. Seems like non-public filters do this, for example.

> hooks, these usually do not pass a lot of data, so probably okay to ignore for now, else we'd need to analyze all bodies

Analyzing hooks is exactly the same as everything else though, I think. As long as they're not public I don't see an issue with removing stuff from them, is that wrong? This pass currently works for hooks with multiple definitions. It avoids overloads with multiple definitions, though.

hilti/toolchain/include/ast/operator-registry.h
```diff
@@ -112,7 +129,8 @@ class Registry {
     std::map<ID, std::vector<const Operator*>>
         _operators_by_builtin_function; // initialized operators by builtin call operators; empty ID collect all without
                                         // a static name
-    std::map<ID, std::vector<const Operator*>> _operators_by_method; // initialized operators by method
+    std::map<ID, std::vector<const Operator*>> _operators_by_method;           // initialized operators by method
+    std::map<uintptr_t, std::vector<expression::ResolvedOperator*>> _resolved; // operators to all resolved instances
```
rsmmr: I haven't really reviewed the PR but I noticed this piece: Architecturally, this isn't a good place for this state. The operator registry represents global state that should remain independent of any particular AST. However, this new map gets updated regularly as AST nodes are being created and destroyed. (Note how the existing maps are not storing AST nodes: `Operator` isn't derived from `Node`).

Better to stick to our typical pattern of collecting information inside a visitor iterating over the AST. Two ideas where to put that: 

1. If this is information that only the optimizer needs, I think it would be best to put it right into the optimizer code: create a new visitor there that collects all the information about the AST that the optimizer needs during a preprocessing pass. 

2. Alternatively, if that doesn't work, the `ASTContext` maintains state about the AST that can be accessed from anywhere. But even if putting it there, still use a visitor to compute the data, not have nodes dynamically register themselves, as that makes it hard to follow and easy to break.

hilti/toolchain/include/ast/operator.h
```diff
@@ -338,6 +338,9 @@ class Operator {
      */
     auto typename_() const { return _typename(); }
 
+    /** Returns a globally unique numeric identifier for the operator. */
+    uintptr_t identity() const { return reinterpret_cast<uintptr_t>(this); }
```
rsmmr: One more question: why do we need this opaque value? Looks like it's just used as the index into `_resolved`. Could we just store `Operator*` in there directly?

hilti/toolchain/include/ast/operator-registry.h
evantypanski: Sure, I was iffy about putting the uses in the operator registry, but it seemed decently relevant because of the name. And the operator concept confused me a bit

I'd prefer putting it in the `ASTContext` since I want to use uses in more places, and I don't want to calculate them each time or shove them all in one pass. Another example would be when changing the function return type, or when removing dead code or something it may be useful to go through uses to check for side effects.

Overall it's the same principle with tracking uses though so it shouldn't be a big change

hilti/toolchain/include/ast/operator.h
evantypanski: Yeah I had `Operator*` but I thought this looked more similar to nodes. I guess uses tracking probably won't expand to anything but operators, so this shouldn't be necessary.

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1553,298 @@ struct MemberVisitor : OptimizerVisitor {
...
+                    case expression::keyword::Kind::Captures: keyword_id = "__captures"; break;
+                    case expression::keyword::Kind::Scope: return; // Scope is different
+                    default: util::cannotBeReached();
+                }
+                removeUsed(ftype, function_id, keyword_id);
+            }
+            case Stage::PRUNE_USES: return;
+            case Stage::PRUNE_DECLS: return;
+        }
+    }
```
bbannier: This all seems unneeded, do you recall why you added this? There should be nothing magical about IDs like `__self` which would just be any ID, so your existing machinery seems to handle this.

tests/hilti/output/optimization/remove-params.hlt
```diff
@@ -0,0 +1,99 @@
...
+}
+
+# The only caller of this has side effects
+# CHECK: function void called_with_side_effects(uint<64> x)
+function void called_with_side_effects(uint<64> x) { }
+
+# One of multiple callers of this has a side effect in params
+# CHECK: function void called_with_one_side_effect(uint<64> x, uint<64> y)
+function void called_with_one_side_effect(uint<64> x, uint<64> y) { }
+
```
bbannier, line 110: This seems to work well, but I had to convince myself with a couple more tests.

Unless I am overlooking something here, the tests here seem centered around side effects. Could you add one or more for use as well, e.g., something like

```suggestion
# CHECK: function void last_used(uint<64> a, uint<64> c) {
function void last_used(uint<64> a, uint<64> b, uint<64> c) {
    hilti::print(a);
    hilti::print(c);
}

# CHECK: last_used(1, 3);
last_used(1, 2, 3);

```

tests/hilti/output/optimization/remove-method-params.hlt
```diff
@@ -0,0 +1,48 @@
...
+    method void method_inline(uint<64> x) {}
+
+    # CHECK: method void method_stubbed();
+    # CHECK-NOT: method void method_stubbed(uint<64> x);
+    method void method_stubbed(uint<64> x);
+
+    # CHECK: hook void hook_one() {}
+    # CHECK-NOT: hook void hook_one(uint<64> x)
+    hook void hook_one(uint<64> x) {}
+};
```
bbannier: The inline hook definition fails during C++ compilation and I assume we do not generate code like that, a working version would be

```suggestion
    # CHECK: hook void hook_one();
    # CHECK-NOT: hook void hook_one(uint<64> x);
    hook void hook_one(uint<64> x);
};

# CHECK: hook void S::hook_one()
# CHECK-NOT: hook void S::hook_one(uint<64> x)
hook void S::hook_one(uint<64> x) {}

```

You could check that the file still compiles after rewrites by adding (I guess the same applies to `remove-params.hlt`)

```sh
# @TEST-EXEC: hiltic %INPUT -dj
```

tests/hilti/output/optimization/remove-method-params.hlt
```diff
@@ -0,0 +1,48 @@
...
+
+# CHECK: s.method_inline()
+s.method_inline(1);
+
+# CHECK: s.method_stubbed()
+s.method_stubbed(1);
+
+# CHECK: s.hook_one()
+s.hook_one(1);
+
```
bbannier, line 52: This branch now works as I'd expect for `public` units, could you add a test for that, e.g.,

```suggestion

public type P = struct {
    method void method_inline(uint<64> y) {}
    method void method_stubbed(uint<64> y);
    hook void hook_one(uint<64> y);
};

# FIXME(bbannier): unclear why this doesn't match, probably related to order.
# _CHECK: method void P::method_stubbed(uint<64> x) {
method void P::method_stubbed(uint<64> y) {}


# FIXME(bbannier): unclear why this doesn't match, probably related to order.
# _CHECK: hook void P::hook_one(uint<64> y) {
hook void P::hook_one(uint<64> y) {}

global P p;

# CHECK: p.method_inline(1);
p.method_inline(1);

# CHECK: p.method_stubbed(1);
p.method_stubbed(1);

# CHECK: p.hook_one(1);
p.hook_one(1);

```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1553,298 @@ struct MemberVisitor : OptimizerVisitor {
...
+            return;
+
+        auto params = ftype->parameters();
+
+        // Ensure they're sorted
+        std::sort(unused.unused_params.begin(), unused.unused_params.end(), std::greater<>());
+        for ( std::size_t index : unused.unused_params ) {
+            assert(index < params.size());
+            params.erase(params.begin() + static_cast<std::ptrdiff_t>(index));
+        }
```
bbannier: I am still not exactly sure I understand how this removal loop works and how it does not break removals in the middle, but it looks like it doesn't.

Could you expand the comment and use a version which makes sure to only remove the specified element (and not from from it til `end`)?, e.g.,

```suggestion
        // Ensure they're sorted in decending order so we remove from the back.
        std::sort(unused.unused_params.begin(), unused.unused_params.end(), std::greater<>());
        for ( std::size_t index : unused.unused_params ) {
            assert(index < params.size());
            auto pos = params.begin() + static_cast<std::ptrdiff_t>(index);
            params.erase(pos, std::next(pos));
        }
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1553,298 @@ struct MemberVisitor : OptimizerVisitor {
...
+    void removeUsed(type::Function* ftype, const ID& function_id, const ID& param_id) {
+        int i = 0;
+        auto& unused = fn_unused_params[function_id];
+        for ( auto param : unused.unused_params ) {
+            if ( ftype->parameters()[param]->id() == param_id ) {
+                unused.unused_params.erase(unused.unused_params.begin() + i);
+                return;
+            }
+            i++;
+        }
```
bbannier: Similarly.

```suggestion
        auto& unused = fn_unused_params.at(function_id);

        for ( auto [i, param] : util::enumerate(unused.unused_params) ) {
            assert(ftype->parameters().size() >= param);
            if ( ftype->parameters()[param]->id() == param_id ) {
                auto pos = unused.unused_params.begin() + static_cast<std::ptrdiff_t>(i);
                unused.unused_params.erase(pos, std::next(pos));
                return;
            }
        }
```

I also replaced the use of `operator[]` here which would insert missing entries with `at` which throws if the key is not present since we should always have the entry.

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1553,298 @@ struct MemberVisitor : OptimizerVisitor {
...
+        auto* ntuple = builder()->expressionCtor(builder()->ctorTuple(params));
+        if ( is_method )
+            replaceNode(fn_use->op2(), ntuple, "removing unused arguments from method call");
+        else
+            replaceNode(fn_use->op1(), ntuple, "removing unused arguments from call");
+    }
+
+    void pruneFromUses(const ID& function_id, const Operator* op) {
+        auto unused = fn_unused_params[function_id];
```
bbannier:
```suggestion
        auto unused = fn_unused_params.at(function_id);
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1553,298 @@ struct MemberVisitor : OptimizerVisitor {
...
+            if ( ! use )
+                continue;
+            removeArgs(use, unused.unused_params);
+        }
+
+        unused.removed_uses = true;
+    }
+
+    void pruneFromDecl(const ID& function_id, type::Function* ftype) {
+        auto unused = fn_unused_params[function_id];
```
bbannier:
```suggestion
        auto unused = fn_unused_params.at(function_id);
```

Actually not even sure we need the copy, but probably less convoluted with it.

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1553,298 @@ struct MemberVisitor : OptimizerVisitor {
...
+
+    void operator()(expression::Name* n) final {
+        auto opt_enclosing_fn = enclosingFunction(n);
+        if ( ! opt_enclosing_fn )
+            return;
+        auto [ftype, function_id] = *opt_enclosing_fn;
+
+        switch ( _stage ) {
+            case Stage::COLLECT: {
+                auto& unused = fn_unused_params[function_id];
```
bbannier:
```suggestion
                auto& unused = fn_unused_params.at(function_id);
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1553,298 @@ struct MemberVisitor : OptimizerVisitor {
...
+    bool prune_decls(Node* node) override {
+        _stage = Stage::PRUNE_DECLS;
+
+        clearModified();
+        visitor::visit(*this, node);
+
+        return isModified();
+    }
+
+    void removeArgs(hilti::Node* call, const std::vector<std::size_t>& positions) {
```
bbannier: You always pass this a `ResolvedOperator*` and also do an unconditional cast below, so this can be simplified by avoiding going through `Node*`,

```suggestion
    void removeArgs(expression::ResolvedOperator* call, const std::vector<std::size_t>& positions) {
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1553,298 @@ struct MemberVisitor : OptimizerVisitor {
...
+        if ( ! call->isA<operator_::function::Call>() && ! call->isA<operator_::struct_::MemberCall>() )
+            return;
```
bbannier: A lot of the existing code in this file would silently skip unexpected data, but in this case we directly control the input in this file and could be a little more drastic to catch errors, e.g.,

```suggestion
            logger().fatalError(util::fmt("expected Call or MemberCall node, but got %s", call->typename_()));
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1553,298 @@ struct MemberVisitor : OptimizerVisitor {
...
+                pruneFromDecl(function_id, ftype);
+                break;
+            }
+        }
+    }
+
+    std::optional<std::tuple<type::Function*, ID>> enclosingFunction(Node* n) {
+        for ( auto* current = n->parent(); current; current = current->parent() ) {
+            if ( auto* fn_decl = current->tryAs<declaration::Function>() ) {
+                return std::make_tuple(fn_decl->function()->ftype(), fn_decl->functionID(context()));
```
bbannier: Nit.

```suggestion
                return std::tuple(fn_decl->function()->ftype(), fn_decl->functionID(context()));
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1553,298 @@ struct MemberVisitor : OptimizerVisitor {
...
+                break;
+            }
+            else if ( auto* field = current->tryAs<declaration::Field>(); field && field->inlineFunction() ) {
+                return std::make_tuple(field->inlineFunction()->ftype(), field->fullyQualifiedID());
```
bbannier: Nit.

```suggestion
                return std::tuple(field->inlineFunction()->ftype(), field->fullyQualifiedID());
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1553,298 @@ struct MemberVisitor : OptimizerVisitor {
...
+            }
+        }
+
+        return std::nullopt;
```
bbannier: Nit.

```suggestion
        return {};
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1553,298 @@ struct MemberVisitor : OptimizerVisitor {
...
+                continue;
+
+            auto* tup = ctor->ctor()->tryAs<ctor::Tuple>();
+            if ( ! tup )
+                continue;
+
+            for ( std::size_t i = 0; i < tup->value().size(); i++ ) {
+                if ( tup->value()[i]->isA<operator_::function::Call>() )
+                    return true;
+            }
```
bbannier: Let's us a range-based `for` loop,

```suggestion
            for ( auto* arg : tup->value() ) {
                if ( arg->isA<operator_::function::Call>() )
                    return true;
            }
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1553,298 @@ struct MemberVisitor : OptimizerVisitor {
...
+        if ( ! call->isA<operator_::function::Call>() && ! call->isA<operator_::struct_::MemberCall>() )
+            return;
+        if ( positions.empty() )
+            return;
+
+        auto* fn_use = call->as<expression::ResolvedOperator>();
+        bool is_method = call->isA<operator_::struct_::MemberCall>();
+
+        // Get the params as a tuple
+        auto* ctor = is_method ? fn_use->op2()->tryAs<expression::Ctor>() : fn_use->op1()->tryAs<expression::Ctor>();
```
bbannier: There is probably a nice follow-up in here to drop implicit `self` as well, i.e., turn methods into free-standing functions.

hilti/toolchain/include/ast/attribute.h
```diff
@@ -88,6 +88,7 @@ const Kind Debug("&debug");
 const Kind Default("&default");
 const Kind HavePrototype("&have_prototype");
 const Kind Internal("&internal");
+const Kind KeepSignature("&keep-signature");
```
bbannier: It is unfortunate that we need a new attribute for this, but it looks like none of the existing ones exactly fit (`&always-emit`?). WDYT about renaming this so it is more about intend and less about handling?, e.g.,

```suggestion
const Kind PublicSignature("&public-signature");
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1588,8 +1876,13 @@ void detail::optimizer::optimize(Builder* builder, ASTRoot* root) {
...
               return std::make_unique<MemberVisitor>(builder, hilti::logging::debug::Optimizer);
           }},
-         {"types", [](Builder* builder) -> std::unique_ptr<OptimizerVisitor> {
+         {"types",
+          [](Builder* builder) -> std::unique_ptr<OptimizerVisitor> {
               return std::make_unique<TypeVisitor>(builder, hilti::logging::debug::Optimizer);
+          }},
+
+         {"remove_unused_params", [](Builder* builder) -> std::unique_ptr<OptimizerVisitor> {
+              return std::make_unique<FunctionParamVisitor>(builder, hilti::logging::debug::Optimizer);
```
bbannier: If you want you could append a `,` after the last element so diffs here get smaller in the future.

Like I mentioned to you offline, the peephole optimizer which runs after all "normal" passes removes statements of the form

```cpp
self.__error = __error;
__error = self.__error;
```

I.e., after your new pass we have another pass which removes uses of `__error` which is passed as an arg, so potentially a missed opportunity to rewrite signatures.

In its current form #1697 does not remove _any_ writes into `self` since it isn't smart enough about exceptions; we'd probably need a new pass which walks the CFG to detect that these two statements appear in a pair with nothing which could throw in between.

hilti/toolchain/src/compiler/optimizer.cc
evantypanski: Huh, the tests seem to pass without it. I was having troubles with `__captures` in particular in the test suite. I guess if it seems to work now it's not needed, not sure what changed though!

hilti/toolchain/include/ast/operator-registry.h
rsmmr: > I'd prefer putting it in the ASTContext since I want to use uses in more places, and I don't want to calculate them each time or shove them all in one pass. Another example would be when changing the function return type, or when removing dead code or something it may be useful to go through uses to check for side effects.

Question is: do you need it outside of the optimizer? If no, then I would suggest to keep the state there as well. You can still compute it only once, just wrap it into some state object, and pass that into the various optimization passes.  That might even generalize to other optimization state as well.

Generally, as we build out the optimization infrastructure, I think it would be good to make that as self-contained as possible. Logically, I think it is pretty much self contained to begin with (AST in, optimized AST out), and the less we introduce dependencies with other parts of the code, the easier both sides will be to maintain. 

That said, if you imagine to use this outside of the optimizer as well, then ASTContext is a good place.

evantypanski: > Question is: do you need it outside of the optimizer?

Not at the moment, you're right it could very well be contained and hoisted out if it's needed, I'll just do that. Sorry for the confusion there

tests/hilti/output/optimization/remove-method-params.hlt
evantypanski: 👍 

And yeah the order is finicky with these, that's why I use `@TEST-START-NEXT` pretty aggressively

hilti/toolchain/src/compiler/optimizer.cc
evantypanski: I agree the comment was very vague, I'll fix that

the `erase` change doesn't seem necessary, it'll only remove one element if `last` isn' provided I think, is that wrong?

bbannier: > the erase change doesn't seem necessary, ...

Right, I probably thought of `string::erase` with index.

hilti/toolchain/src/compiler/optimizer.cc
evantypanski: yeah uses used to be nodes and I never updated those to be more precise, I also very smartly casted a `ResolvedOperator` to a `ResolvedOperator` in `usesContainSideEffects` :) (I'll fix that too)

hilti/toolchain/include/ast/attribute.h
evantypanski: > `&always-emit`?

Yeah using that breaks `import-from-with-name-clash`, there are also other side-effects like more `parseX` functions getting emitted (which is presumably the reason why that test fails). I figured using a new attribute would express the intent better, and it'll just be used whenever we need a stable signature, but not necessarily always have the definition

`public-signature` seems like a better name, though

hilti/toolchain/src/compiler/optimizer.cc
evantypanski: > after your new pass we have another pass which removes uses of __error which is passed as an arg, so potentially a missed opportunity to rewrite signatures.

That's almost certainly necessary for what I want to do, since my hope is that *some* high-use units will have parse methods that just return `__cur` so that we can remove the tuple return type. Since `__error` is also returned it's kind of a big deal that some cases get missed. Will have to visit that, there are technically 2 steps until that matters though (changing function return types and constant propagation with `__result`)

hilti/toolchain/src/compiler/optimizer.cc
evantypanski: I don't mind the implicit `self` too much, but it is a bit annoying that it changes which operand holds the function call arguments

Logically it's just a call with an extra `self` - it could potentially keep as a `MemberCall`, but align more closely with `Call`. That would be nice. Naively just subclassing `Call` but adding `op2` as the `self` argument seems pretty reasonable to me. I guess at that point may just say the presence of `op2` means it's a member call, then ditch `MemberCall` since it's just a `Call` at that point. But I haven't thought hard about the implications there

bbannier: I actually was thinking of something else: if a method makes no use of `self` it would be nice to capture that, but realistically that is probably something which is extremely rare in code we generate. Really nothing to do here.

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -81,9 +94,13 @@ class OptimizerVisitor : public visitor::MutatingPreOrder {
     enum class Stage { COLLECT, PRUNE_USES, PRUNE_DECLS };
     Stage _stage = Stage::COLLECT;
     declaration::Module* _current_module = nullptr;
+    const OperatorUses& op_uses;
```
bbannier: I like that this is `const` so users don't accidentally add data with `operator[]`, but let's not make this a const ref since that disables assignment and move operators (we currently don't use them, but still).

Let's instead make this a private pointer member, and provide access to the data. Currently users _should_ check for presence with `count` and can only then get the data with `at`; our accessor could encapsulate that completely, e.g., 

```cpp
    const OperatorUses::mapped_type* uses(const Operator* x) const {
        if ( _op_uses->count(x) == 0 )
            return nullptr;

        return &_op_uses->at(x);
    }

private:
    const OperatorUses* _op_uses;
```

hilti/toolchain/src/ast/declarations/function.cc
```diff
@@ -16,3 +16,10 @@ node::Properties declaration::Function::properties() const {
 
     return Declaration::properties() + std::move(p);
 }
+
+ID declaration::Function::functionID(ASTContext* ctx) {
+    if ( auto* prototype = ctx->lookup(linkedPrototypeIndex()) )
+        return prototype->fullyQualifiedID();
+
+    return fullyQualifiedID();
+}
```
bbannier, line 25: This almost feels like a workaround for `ID` not dealing well with IDs of imported functions where we seem to prepend the importing module's name. If we cannot have it in `ID` itself it probably belongs into `Declaration`.

evantypanski: Since it relies on `linkedPrototypeIndex()` and a lookup, I don't think that's possible. It seems to me to be a unique problem to functions, that you can declare in one place, implement in another, and they're both declarations.

FWIW changing this to declaration and using `declarationIndex()` breaks tests

bbannier: Thanks for checking!

hilti/toolchain/include/ast/declarations/function.h
```diff
@@ -51,6 +48,12 @@ class Function : public Declaration {
         _linked_prototype_index = index;
     }
 
+    /**
+     * Returns the ID for the function declaration, regardless of the module which
+     * it was implemented in.
```
rsmmr:
```suggestion
     * it was implemented in. Only valid once the AST has been resolved.
```

hilti/toolchain/include/ast/attribute.h
```diff
@@ -94,6 +94,7 @@ const Kind Nosub("&nosub");
 const Kind OnHeap("&on-heap");
 const Kind Optional("&optional");
 const Kind Priority("&priority");
+const Kind PublicSignature("&public-signature");
```
rsmmr: Wondering if we could just call this `&public`, aligning it with the `public` linkage that has similar semantics. (unless you guys discussed that already?)

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -73,6 +73,19 @@ static auto idFeatureFromConstant(const ID& featureConstant) -> std::optional<st
     return {{type_id, feature}};
 };
 
+using OperatorUses = std::map<const Operator*, std::vector<expression::ResolvedOperator*>>;
+
+// Collects uses of resolved operators
+struct CollectUsesPass : public hilti::visitor::PreOrder {
```
rsmmr, line 80: Ack, I think this structure works well.

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1579,272 @@ struct MemberVisitor : OptimizerVisitor {
...
+        clearModified();
+        visitor::visit(*this, node);
+
+        return isModified();
+    }
+
+    void removeArgs(expression::ResolvedOperator* call, const std::vector<std::size_t>& positions) {
+        if ( ! call->isA<operator_::function::Call>() && ! call->isA<operator_::struct_::MemberCall>() )
+            logger().fatalError(util::fmt("expected Call or MemberCall node, but got %s", call->typename_()));
+        if ( positions.empty() )
```
rsmmr, line 1629:
```suggestion

        if ( positions.empty() )
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1579,272 @@ struct MemberVisitor : OptimizerVisitor {
...
+        auto* ctor = is_method ? call->op2()->as<expression::Ctor>() : call->op1()->as<expression::Ctor>();
+        auto* tup = ctor->ctor()->as<ctor::Tuple>();
+
+        // Make new parameters
+        Expressions params;
+        for ( std::size_t i = 0; i < tup->value().size(); i++ ) {
+            if ( std::find(positions.begin(), positions.end(), i) == positions.end() )
+                params.push_back(tup->value()[i]);
+        }
+        auto* ntuple = builder()->expressionCtor(builder()->ctorTuple(params));
```
rsmmr, line 1645:
```suggestion

        auto* ntuple = builder()->expressionCtor(builder()->ctorTuple(params));
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1579,272 @@ struct MemberVisitor : OptimizerVisitor {
...
+
+        auto params = ftype->parameters();
+
+        // Ensure they're sorted in descending order so we remove from the back.
+        std::sort(unused.unused_params.begin(), unused.unused_params.end(), std::greater<>());
+        for ( std::size_t index : unused.unused_params ) {
+            assert(index < params.size());
+            params.erase(params.begin() + static_cast<std::ptrdiff_t>(index));
+        }
+        recordChange(ftype, "removing unused function parameters");
```
rsmmr, line 1685:
```suggestion

        recordChange(ftype, "removing unused function parameters");
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1579,272 @@ struct MemberVisitor : OptimizerVisitor {
...
+        ftype->setParameters(builder()->context(), params);
+    }
+
+    /** Determines if the uses of this operator contain any side effects. */
```
rsmmr: Suggest to expand the comment a bit to say what we count as side effect. Reading the code, it's arguments that contain function calls?

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1579,272 @@ struct MemberVisitor : OptimizerVisitor {
...
+                return;
+            }
+        }
+    }
+
+    void operator()(expression::Name* n) final {
+        auto opt_enclosing_fn = enclosingFunction(n);
+        if ( ! opt_enclosing_fn )
+            return;
+        auto [ftype, function_id] = *opt_enclosing_fn;
```
rsmmr, line 1845:
```suggestion

        auto [ftype, function_id] = *opt_enclosing_fn;
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1579,272 @@ struct MemberVisitor : OptimizerVisitor {
...
+    }
+
+    /** Removes the param_id as used within the function. */
+    void removeUsed(type::Function* ftype, const ID& function_id, const ID& param_id) {
+        auto& unused = fn_unused_params.at(function_id);
+
+        for ( auto [i, param] : util::enumerate(unused.unused_params) ) {
+            assert(ftype->parameters().size() >= param);
+            if ( ftype->parameters()[param]->id() == param_id ) {
+                auto pos = unused.unused_params.begin() + static_cast<std::ptrdiff_t>(i);
```
rsmmr: Nit: Wouldn't it be easier to just use an iterator for the outer loop (instead of the ` : enumerate(...)` )?

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1557,6 +1579,272 @@ struct MemberVisitor : OptimizerVisitor {
...
+        if ( ! ftype || ! n->parent()->isA<type::Struct>() )
+            return;
+
+        const auto& function_id = n->fullyQualifiedID();
+
+        switch ( _stage ) {
+            case Stage::COLLECT: {
+                if ( fn_unused_params.count(function_id) > 0 )
+                    return;
+                // Create the unused params
```
rsmmr, line 1776:
```suggestion

                // Create the unused params
```

rsmmr (approved):
Looks good. I didn't look too closely at when exactly we're removing the parameters, but didn't spot anything obvious at least. You have tried this with a few existing analyzers I suppose?

hilti/toolchain/include/ast/attribute.h
evantypanski: `&public` makes sense to me. I think I initially had it as `&keep-signature`, so @bbannier suggested `&public-signature`. Once it's `&public-signature` it makes sense to just call it `&public` though - but took a couple iterations to get there

hilti/toolchain/src/compiler/optimizer.cc
evantypanski: Yeah, "operator containing side effects" right now are just function calls arguments that contain function calls, aligning with a current simplification with the CFG optimizations I think. It'll make sense in the future to have a better analysis for that

I'll update the comment though

PR #2084, opened 2025-06-06
Make `-Wstringop-overflow` not fatal with gcc.

bbannier:


evantypanski (approved):
Seems reasonable, seems like that warning is pretty flaky for people

PR #2085, opened 2025-06-06
When C++ compilation fails, ask user for help.

### LINKED ISSUE #2066: Add issue report to failed C++ compilation
When C++ code from users fails to compile, users get confused. I've seen it a couple of times on the Spicy Slack channel, where users don't expect C++ errors and try to debug it as if it's something wrong on their end, when really it's an issue with Spicy. There may also be cases where C++ compilation fails and they do not submit a bug report, but instead just change the code, so we miss compiler issues from users.

I'd suggest submitting a Spicy bug report when JIT compilation fails (or similar cases). This could be just a link to Spicy's issues, OR we could create some temporary file that has some specifics for them to submit. That would be really nice since it makes their job easy, but maybe it'd contain source code that they don't want to share.

One obvious downside is that you can use custom C++, which may then fail to compile. Honestly I don't think it's that big of an issue because 1) that's not very common and 2) if a user is doing that, they'll know to be cautious. We could also put some extra qualifier ("if this is not in custom C++ code...")

rsmmr:
This records the C++ compiler in a dedicated output file
"hilti-jit-error.log", and asks the user to file a ticket.

[Added] If the environment variable `HILTI_JIT_SHOW_CXX_OUTPUT` is set, the output
will also go to stderr. We set this variable inside CI.

Before, the process management code was trying to capture the output
in memory, but that didn't actually work, it always ended up on stderr
anyways. Now we tell reproc to just write all compiler output into a
temporary file, where we can then copy it from if needed.

Closes #2066.


evantypanski:
Yeah this is exactly what I had in mind! Couple points

1) I'd prefer it point to the new issue page (https://github.com/zeek/spicy/issues/new) rather than the issues list
2) I tried in a plugin just to see what most people would see, but the log file doesn't appear, would suggest checking that and making sure it's clear first. Maybe also include the absolute path to the file?

But overall this looks a lot nicer. There's a small part of me that would want an option to put the compiler output to stderr, but it's a small part.

Also I really like that it includes version info! Depending how many compilers are ever going to be supported that sometimes gets annoying - for example, `cl` doesn't have a `--version` flag :)

But this seems really nice and convenient to me

rsmmr:
> I tried in a plugin just to see what most people would see, but the log file doesn't appear, would suggest checking that and making sure it's clear first.

I'm not sure what you mean here. How can I reproduce what you're seeing (or not)?

rsmmr:
> Think of CI or ephemeral build environments

That's a good point. I was originally actually thinking about adding an command line option to show the output, but ended up deciding against, because (1) option proliferation and (2) I'm guessing that usually one would at first not be using the option until running into a problem--but at that point one can just as well "cat hilti-jit-error.log" instead of restarting with the option.

That said, seems there are two use cases where it'd be helpful to output the diagnosticis directly:

1. CI/automated builds
2. Iteration during Spicy development when C++ code might be broken for a while when working on the C++ codegen.

I'm still think an option isn't the right thing here. We could do an environment variable, but adding one just for this feels a bit obscure. Is there an existing one that we could attach to?

Another idea that would at least address at subset of cases, would be  outputting `hilti-jit-error.log` to the console when Spicy is built in debug mode. But not sure I like changing behavior between debug and release builds.


Other ideas? (Keep in mind that this counts as "internal Spicy errors", it's not something that Spicy users should regularly run into; and in the cases they do, they'll probably do so locally first, before CI even enters the picture).

rsmmr:
Thanks for the input. Thinking I'll go with an environment variable letting the output go to stderr, but skipping further customization like going into a different file or selecting stdout.

hilti/toolchain/src/compiler/jit.cc
```diff
@@ -24,6 +24,8 @@
 #include <hilti/compiler/detail/cxx/unit.h>
 #include <hilti/compiler/jit.h>
 
+#include "filesystem/include/ghc/filesystem.hpp"
```
evantypanski: What is this used for? I see `rt::filesystem` but not `ghc`

hilti/toolchain/src/compiler/jit.cc
```diff
@@ -194,6 +182,45 @@ JIT::~JIT() {
...
+That file contains diagnostics from the C++ compiler. Please also
+include your Spicy code that triggered this error, ideally reduced to
+a small snippet still exhibiting the issue if you can.
+
+===
+
+)",
+                               rt::filesystem::absolute(UserDiagnosticsFile));
+
+        if ( auto* show_output = getenv("HILTI_JIT_SHOW_CXX_OUTPUT"); show_output && *show_output ) {
```
evantypanski, line 207: This is a matter of taste, but since this makes no distinction between the values, I'd prefer this toggle just check for the presence of the env var rather than having to set it to some random value. This way `HILTI_JIT_SHOW_CXX_OUTPUT= spicyc -j test.spicy` works. Happy either way though

```suggestion
        if ( hilti::rt::getenv("HILTI_JIT_SHOW_CXX_OUTPUT").has_value() ) {
```

hilti/toolchain/src/compiler/jit.cc
```diff
@@ -605,6 +632,42 @@ Result<Nothing> JIT::JobRunner::_waitForJobs() {
...
+    // Copy the output of the process over into the diagnostics file
+    std::ifstream in(job.output);
+    if ( ! in ) {
+        logger().warning(
+            util::fmt("could not open process output file %s for reading: %s", job.output, ::strerror(errno)));
+        return;
+    }
+
+    if ( add_header )
+        out << "HILTI version " << hilti::configuration().version_string_long << "\n\n";
```
evantypanski, line 660: Super-nit: Since this is using `add_header` again, I'd prefer it to be in the same block as the other `add_header` above, just to keep things together

tests/Scripts/canonify-hiltic
```diff
@@ -11,7 +11,7 @@ cat \
   | sed '/\] loaded module/                {s#/\([^/ ]\{1,\}/\)\{1,\}\([^/ ]\{1,\}\)#<...>/\2#g;}' \
   | sed '/[^ ]\{1,\}\.[^ ]\{1,\}:[0-9]\{1,\}:/      {s#/\([^/ ]\{1,\}/\)\{1,\}\([^/ ]\{1,\}\)#<...>/\2#g;}' \
   | sed '/[^ ]\{1,\}\.[^ ]\{1,\}:[0-9]\{1,\}-[0-9]\{1,\}:/      {s#/\([^/ ]\{1,\}/\)\{1,\}\([^/ ]\{1,\}\)#<...>/\2#g;}' \
-  | sed 's#[("]\(/[^/\]\{1,\}\)\{1,\}\([^/\]\{1,\}:[0-9-]\{1,\}\)[")]#(<...>/\1\2)#g' \
+  | sed 's#[("]n\(/[^/\]\{1,\}\)\{1,\}\([^/\]\{1,\}:[0-9-]\{1,\}\)[")]#(<...>/\1\2)#g' \
```
evantypanski: what's this `n`?

evantypanski (approved):
I have a few nits but nothing much - hopefully people report more bugs after this gets in a release!

hilti/toolchain/src/compiler/jit.cc
rsmmr: Ah, that must have been an unintended fix-it, removing.

hilti/toolchain/src/compiler/jit.cc
rsmmr: Actually I don't like it working without any value at all, and we're using the same pattern elsewhere already, so prefer to leave as is.

hilti/toolchain/src/compiler/jit.cc
rsmmr: The thing is that we need to open `out` in between, in append mode, that's why it's split.

tests/Scripts/canonify-hiltic
rsmmr: oops, another accident.

PR #2088, opened 2025-06-10
Fix scope resolution of local variables.

### LINKED ISSUE #2086: Incorrect resolution order for globals and local scopes leads to C++ error
The following Spicy code causes a C++ error:

```
module Test;

global x = 0;

{
    x = x + 1;
    local x = 5;
}
```

Error:

```
$ spicyc -j ../test.spicy
In file included from /private/var/folders/w6/h7tg594x0yng0_yzvcsn6mv00000gn/T/Test_89b935601bd807c3-8e5e080e761d75b5.cc:1:
In file included from /Users/etyp/src/spicy/hilti/runtime/include/hilti/rt/libhilti.h:11:
In file included from /Users/etyp/src/spicy/hilti/runtime/include/hilti/rt/configuration.h:11:
In file included from /Users/etyp/src/spicy/hilti/runtime/include/hilti/rt/util.h:23:
In file included from /Users/etyp/src/spicy/hilti/runtime/include/hilti/rt/types/time.h:11:
In file included from /Users/etyp/src/spicy/hilti/runtime/include/hilti/rt/types/interval.h:12:
In file included from /Users/etyp/src/spicy/hilti/runtime/include/hilti/rt/safe-int.h:6:
/Users/etyp/src/spicy/hilti/runtime/include/hilti/rt/3rdparty/SafeInt/SafeInt.hpp:601:23: error: static assertion failed due to requirement 'safeint_internal::numeric_type<std::optional<SafeInt<unsigned long long, hilti::rt::integer::detail::SafeIntException>>>::isInt': Integer type required
  601 |         static_assert(safeint_internal::numeric_type< T >::isInt, "Integer type required");
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/etyp/src/spicy/hilti/runtime/include/hilti/rt/3rdparty/SafeInt/SafeInt.hpp:634:96: note: in instantiation of template class 'safeint_internal::int_traits<std::optional<SafeInt<unsigned long long, hilti::rt::integer::detail::SafeIntException>>>' requested here
  634 |             isBothLT32Bit = (safeint_internal::int_traits< T >::isLT32Bit && safeint_internal::int_traits< U >::isLT32Bit),
      |                                                                                                ^
/Users/etyp/src/spicy/hilti/runtime/include/hilti/rt/3rdparty/SafeInt/SafeInt.hpp:652:55: note: in instantiation of template class 'safeint_internal::type_compare<unsigned long long, std::optional<SafeInt<unsigned long long, hilti::rt::integer::detail::SafeIntException>>>' requested here
  652 |         IntZone_UintLT32_UintLT32 = safeint_internal::type_compare< T,U >::isBothUnsigned && safeint_internal::type_compare< T,U >::isBothLT32Bit,
      |                                                       ^
/Users/etyp/src/spicy/hilti/runtime/include/hilti/rt/3rdparty/SafeInt/SafeInt.hpp:3457:19: note: in instantiation of template class 'IntRegion<unsigned long long, std::optional<SafeInt<unsigned long long, hilti::rt::integer::detail::SafeIntException>>>' requested here
 3457 |         method = (IntRegion< T,U >::IntZone_UintLT32_UintLT32  ? AdditionState_CastIntCheckMax :
      |                   ^
/Users/etyp/src/spicy/hilti/runtime/include/hilti/rt/3rdparty/SafeInt/SafeInt.hpp:6773:27: note: in instantiation of template class 'AdditionMethod<unsigned long long, std::optional<SafeInt<unsigned long long, hilti::rt::integer::detail::SafeIntException>>>' requested here
 6773 |     AdditionHelper< T, U, AdditionMethod< T, U >::method >::template AdditionThrow< E >( (T)rhs, lhs, ret );
      |                           ^
/private/var/folders/w6/h7tg594x0yng0_yzvcsn6mv00000gn/T/Test_89b935601bd807c3-8e5e080e761d75b5.cc:29:15: note: in instantiation of function template specialization 'operator+<unsigned long long, std::optional<SafeInt<unsigned long long, hilti::rt::integer::detail::SafeIntException>>, hilti::rt::integer::detail::SafeIntException>' requested here
   29 |         x = x + ::hilti::rt::integer::safe<std::uint64_t>{1U};
      |               ^
1 error generated.
```

Note that I mention resolution because the `local x = 5` after shouldn't make a difference, but it does.

rsmmr:
If usage of a local comes before its declaration, we now no longer
resolve that usage to this local. It'll either be resolved to an
upper layer ID (if there is one of the same name), or rejected if it's
otherwise unknown.

Closes #2086.


hilti/toolchain/include/ast/node.h
```diff
@@ -464,6 +464,30 @@ class Node {
...
+
+        if ( ! recurse )
+            return false;
+
+        for ( const auto* c : _children ) {
+            if ( c && c->hasChild(n, true) )
+                return true;
+        }
+
+        return false;
```
bbannier: Nit: I'd prefer passing `recurse` instead of introducing the dependency on the early return just above. We could also write more clearly as

```suggestion
        return std::any_of(_children.begin(), _children.end(),
                           [&](const auto* c) { return c && c->hasChild(n, recurse); });
```

hilti/toolchain/src/ast/scope-lookup.cc
```diff
@@ -60,6 +63,28 @@ std::pair<bool, Result<std::pair<Declaration*, ID>>> hilti::scope::detail::looku
             }
         }
 
+        if ( d->isA<declaration::LocalVariable>() ) {
```
bbannier, line 66: Like we discussed in the context of the node contains check in #1697, this could be nicer if statements always had consistent locations, but they do not. In that case we could have directly compared source locations of both the original node and the declaration via the decl scope instead of iterating block statements 🤷.

bbannier (dismissed):

hilti/toolchain/src/ast/scope-lookup.cc
rsmmr: Yeah, but even if generally we'd have better location information, I'd be reluctant to rely on that for such comparisons. In the end it's the AST structure that's the ground truth.

hilti/toolchain/include/ast/node.h
rsmmr: > I'd prefer passing recurse I

+1

>  write more clearly as

Not convinced that's "more clearly". :)

PR #2089, opened 2025-06-10
When printing anonymous bitfields inside a struct, lift up the fields.

### LINKED ISSUE #1660: String representation of anonymous bitfield in units
```
module Test;

import spicy;

public type Header = unit {

  : bitfield(16) {
    fin: 0;
    rsv: 1..3;
    opcode: 4..7;
    remaining: 8..15;
  } &bit-order=spicy::BitOrder::MSB0;

  on %done {
    print "fin", self.fin, "opcode", self.opcode;

    print self;
  }
};

$ printf "\x82\xff" | spicy-driver  ./test.spicy 
fin, 1, opcode, 2
[$<anon>=(1, 0, 2, 255)]
```

As discussed in Slack, it's a bit surprising that the "lifted" fields are not shown in the string representation of the unit and instead the `$<anon>` part shows up.

rsmmr:
This now prints, e.g.,  `[$fin=1, $rsv=0, $opcode=2, $remaining=255]`
instead of `[$<anon>=(1, 0, 2, 255)]`.

In addition, we also prettify non-anonymous bitfields. They now print
as, e.g., `[$y=(a: 4, b: 8)]` instead of `[$y=(4, 8)]`.

The implementation is a bit of hack (because we normally don't convey
type information through `cxx::*` elements) but can't of think
anything nicer.

Tested through updates to existing tests.

Closes #1660.


rsmmr:
@bbannier @evantypanski still need review here

bbannier:
It seems this needed to be so hacky since in generated code bitfields are almost completely gone. The approach here works for the use case of printing a `bitfield` if it is part of a `struct` (from a `unit`), but still does nothing for anything else, e.g.,

```ruby
module Test;

type BF = bitfield(16) {
    fin: 0;
    rsv: 1..3;
    opcode: 4..7;
    remaining: 8..15;
};
global x: BF;

public type Header = unit {
    h: BF;

    on %done {
        print self;

        # Via global.
        x = self.h;
        print x;

        # Via local.
        local y = self.h;
        print y;
    }
};
```

```console
$ spicy-driver -d foo.spicy -f /dev/zero
[$h=(fin: 0, rsv: 0, opcode: 0, remaining: 0)]
(0, 0, 0, 0)
(0, 0, 0, 0)
```

We might be able to just call this unsupported.

Alternatively, a way to support that would be to store track the typeinfo for globals as well, and push the printing from structs down to locals (and also add a copy for globals), or instead expose bitfields so they can track their own typeinfo directly.

hilti/runtime/include/types/bitfield.h
```diff
@@ -53,53 +57,69 @@ ptrdiff_t elementOffset() {
...
+            out += ", ";
+
+        if ( is_anonymous )
+            out += fmt("$%s=%s", b.name, v.to_string());
+        else
+            out += fmt("%s: %s", b.name, v.to_string());
+    }
+
+    return out;
 }
```
bbannier, line 82: I find this function slightly weird:

- we pass a `TypeInfo*` even though we already have it in `x.ti`; at the very least we should `assert(x.ti == type_info)`.
- `is_anonymous` has nothing to do with the bitfield, but instead with how it is embeded into a unit. WDYT about just deciding on one of the renderings, and implementing this as a `to_string` instead?

hilti/runtime/include/types/bitfield.h
```diff
@@ -53,53 +57,69 @@ ptrdiff_t elementOffset() {
...
+    type_info::Value bitfield(&empty, type_info);
+    for ( const auto& [b, v] : type_info->bitfield->iterate(bitfield) ) {
+        if ( ! out.empty() )
+            out += ", ";
+
+        out += fmt("$%s=(not set)", b.name);
+    }
+
+    return out;
 }
```
bbannier, line 109: If we wouldn't have to deal with `is_anonymous` this would almost be covered by the existing `std::string to_string(const std::optional<T>& x, adl::tag /*unused*/)` but for the rendering of the internals of an unset bitfield. TBH I'd be fine with just not showing that and using the default stringification for `optional` (shows something like `(not set)`). Is there another reason we'd want to treat bitfield differently?

bbannier (approved):
I am not sure this became much simpler since it wants to add a lot of bells and whistles. I personally would be fine with this being less powerful, added some comments for that below.

hilti/runtime/include/types/bitfield.h
rsmmr: > * we pass a `TypeInfo*` even though we already have it in `x.ti`; at the very least we should `assert(x.ti == type_info)`.

It's all a bit subtle but there is a case where the Bitfield doesn't have its type info set, but we know the typeinfo at the time we call `render()` (it's a default initialized Bitfield). 

> * `is_anonymous` has nothing to do with the bitfield, but instead with how it is embedded into a unit. WDYT about just deciding on one of the renderings, and implementing this as a `to_string` instead?

I tried to stay with the output style we're currently generating.  That said, I could see changing how bitfields are printed to use the same style as structs/units, but I don't think it would buy us much in terms of saving code/complexity here (because we still need the `optional` case, see below). Plus, I actually like printing bitfields differently, so that one can visually distinguish from structs. (And indeed I just tried changing it to use the struct style, and do not really like it ...)

hilti/runtime/include/types/bitfield.h
rsmmr: > for the rendering of the internals of an unset bitfield. TBH I'd be fine with just not showing that and using the default stringification for optional

Don't think that would work well because, with the lifting, we want to show the individual bit ranges as if they were direct fields of the struct; that's why they all get their own `not set`.

In the end it's this lifting that causes most of the complexity here, but that's something quite useful, so I think it's worth the trouble.

PR #2090, opened 2025-06-12
Gracefully exit `spicy-driver` (&more) for profiling information

### LINKED ISSUE #1996: Missing profiling report on parse error
When passed the `-Z` flag tools like `spicy-driver` report execution counts on exit, e.g.,

```ruby
# foo.spicy
module foo;

public type X = unit {
    : b"ABC";
};
```

```console
$ echo 'ABC' | spicy-driver -Z foo.spicy
#
# Profiling results
#
#name                                                   count       time      avg-%    total-%          volume
hilti/func/foo::X::__parse_foo__X_stage2                    1       4292       5.91       5.91               -
hilti/func/foo::X::__parse_stage1                           1       5000       6.89       6.89               -
hilti/func/foo::X::parse3                                   1       5708       7.86       7.86               -
hilti/func/foo::__register_foo_X                            1      11125      15.33      15.33               -
hilti/total                                                 1      72583     100.00     100.00               -
spicy/prepare/input/foo::X                                  1       2792       3.85       3.85               -
spicy/unit/foo::X                                           1       4667       6.43       6.43               3
spicy/unit/foo::X::_anon                                    1       3958       5.45       5.45               3
```

If we encounter a parse error this report is missing.

```console
$ echo '123' | spicy-driver -Z /tmp/bla.spicy
[error] terminating with uncaught exception of type spicy::rt::ParseError: expected bytes literal "ABC" but input starts with "123" (foo.spicy:4:7-4:12)
```

evantypanski:
Fixes #1996

The exceptions bubbled up to `main` then got caught in those exception handlers with `spicy-driver`, so something like `spicy-driver -Z` would not emit profiling information if there was a parse error. This makes runtime exceptions get caught sooner so that the calling code can use the same error mechanisms as already exist in order to gracefully exit.

I know that making this catch runtime errors is a relatively large shift, but it already returns a `Result` so it seems pretty reasonable to me to use that instead, then all of its callers handle that anyway.

This does make some annoying test changes, but I don't really see changing `spicy-dump`s error as a big deal. Can change that too to finish the runtime like `spicy-driver` etc.


spicy/runtime/src/driver.cc
```diff
@@ -187,13 +187,18 @@ Result<spicy::rt::ParsedUnit> Driver::processInput(const spicy::rt::Parser& pars
...
+                r = parser.parse3(unit, data, {}, {});
+            }
+            else {
+                DRIVER_DEBUG(fmt("resuming parsing input (eod=%s)", data->isFrozen()));
+                r->resume();
+            }
+        } catch ( const hilti::rt::RuntimeError& e ) {
+            return Error(fmt("terminating with uncaught exception of type %s: %s",
+                             hilti::rt::demangle(typeid(e).name()), e.what()));
         }
```
bbannier: This also makes sense to me, but I'd propose we tweak it a little.

- Our `Result` is not a sum type and can only capture a string, so it is practically impossible to tweak it for the exact context of the caller. Let's not say _terminated_, but e.g., _processing failed_. Also, the exception isn't really _uncaught_ anymore.
- Let's make sure this function does not throw at all instead of only handling parse errors; one way to do that would be to use a function `try` block. We should also convert all exceptions, not just `RuntimeError`. We need to update the docstring of `processInput`.

```cpp
Result<spicy::rt::ParsedUnit> Driver::processInput(const spicy::rt::Parser& parser, std::istream& in,
                                                   int increment) try {
    // Previous body ...

    return std::move(*unit);
} catch ( const std::exception& e ) {
    return Error(fmt("processing failed with exception of type %s: %s", hilti::rt::demangle(typeid(e).name()), e.what()));
} catch ( ... ) {
    return Error(fmt("parsing failed with non-standard exception %s",
                     hilti::rt::demangle(typeid(std::current_exception()).name())));
}
```

evantypanski: Yeah I was a little worried about changing the message too much, but I do like that better.

spicy/runtime/include/driver.h
```diff
@@ -236,8 +236,8 @@ class Driver {
      * @param increment if non-zero, will feed the data in small chunks at a
      * time; this is mainly for testing parsers; incremental parsing
      *
-     * @return error if the input couldn't be fed to the parser (excluding parse errors)
-     * @throws HILTI or Spocy runtime error if the parser into trouble
+     * @return error if the input couldn't be fed to the parser or an
+     * exception is thrown
```
bbannier: Nit:

```suggestion
     * @return error if the input couldn't be fed to the parser or parsing failed
```

bbannier (dismissed):
Oh wow, that error message was in a lot more baselines than I expected. Not ideal, but our error messages are also not a stable API 🤷.

PR #2096, opened 2025-06-25
Switch compilation to C++20.

bbannier:


hilti/runtime/include/fiber.h
```diff
@@ -310,8 +310,10 @@ class Resumable {
      *
      * @param f function to be executed
      */
-    template<typename Function, typename = std::enable_if_t<std::is_invocable_v<Function, resumable::Handle*>>>
-    Resumable(Function f) : _fiber(detail::Fiber::create()) {
+    template<typename Function>
+    Resumable(Function f)
+        requires(std::is_invocable_v<Function, resumable::Handle*>)
```
evantypanski, line 315: I'm so happy for constraints :)

hilti/toolchain/include/ast/node.h
```diff
@@ -243,9 +244,9 @@ class Node {
     /** Returns the node tag associated with the instance's class. */
     node::Tag nodeTag() const {
         // Get the last non-zero tag. The last element(s) may be unset
-        for ( auto it = _node_tags.rbegin(); it != _node_tags.rend(); it++ ) {
-            if ( *it != 0 )
-                return *it;
+        for ( unsigned short _node_tag : std::ranges::reverse_view(_node_tags) ) {
```
evantypanski: why not `auto` for `_node_tag`?

CMakeLists.txt
```diff
@@ -128,6 +128,10 @@ if (USE_WERROR)
         # `__builtin_memcpy` in e.g., its `bits/char_traits.h`.
         string(APPEND werror_flags " -Wno-error=restrict")
 
+        # When optimizing GCC often reports false positives around maybe
+        # uninitialized values, e.g., in its string implementation. These are
+        # extremely tedious to silence, so just disable the warning altogether.
+        string(APPEND werror_flags " -Wno-error=maybe-uninitialized")
```
evantypanski: Kind of a bummer for this, but I guess there are other analyses we run that would catch these

hilti/toolchain/include/ast/node.h
bbannier: Looks like the fix-it for`modernize-loop-convert` didn't preserve the `auto`, brought it back.

CMakeLists.txt
bbannier: Completely agree, but it seems that the integration of the GCC compiler and its `libstdc++` is pretty poor since they trigger diagnostics deep inside `libstdc++` headers for code which should be fine (e.g., a `string` copy). I cannot recall the last time these diagnostics produced anything of value for us, and hopefully we'll catch them in another build or in Coverity.

I discussed the use of `-Wno-error=..` with Robin last week, and we landed on still showing "handled" errors being potentially confusing since it might be unclear whether anything needs to be done (probably debatable since they'd also make an `--enable-werror` build pass, but still), so I switched all of these to `-Wno-..` so they do not show up at all.

bbannier:
> @bbannier I fixed the Fedora42/opensuse issues if you want to take a look - All CI passes!

Awesome, thanks! This looks good to me now and I will set it _Ready for review_.

spicy/toolchain/bin/spicy-batch-extract.cc
```diff
@@ -22,7 +22,7 @@ static void processPreBatchedInput(std::string needle, std::istream& in, std::os
     out << magic << '\n';
 
     std::set<std::string> needles = {std::move(needle)};
-    auto is_needle = [&](const std::string& n) { return needles.find(n) != needles.end(); };
+    auto is_needle = [&](const std::string& n) { return needles.contains(n); };
```
evantypanski, line 25: ¯\\\_(ツ)_/¯ seems like this lambda doesn't give any readability benefits anymore, maybe just change uses to use `contains`?

evantypanski (approved):
just one little extra comment from me

spicy/toolchain/bin/spicy-batch-extract.cc
bbannier: Sure, but that same argument could already have been made before where we could just have used `needles.count(n)` 🤷.

There are probably many things we could clean up, but I'd prefer to defer that to some other PRs.

rsmmr (approved):
Hooray for designated initializers!

PR #2097, opened 2025-06-25
Move calling convention from function to function type.

### LINKED ISSUE #2052: Function types should (?) hold the calling convention
From https://github.com/zeek/spicy/pull/2048#discussion_r2079041980

This is cleanup. Right now, the function type doesn't hold its calling convention. Instead declarations do. But, it seems to fit nicely there for a few reasons:

1) Both `declaration::Field` and `hilti::Function` need separate calling convention instances that could both go in the function type
2) Some builders get simplified from only having function types take the calling convention
3) The calling convention in the grammar goes into a declaration, even though lexically it's sandwiched between parts that go in a function type

This won't have much user impact, but could simplify code a little.

Here's a patch to get started, I stopped when I realized it's taking more time than I wanted. May be worth starting with: 

[fntype-cc.patch](https://github.com/user-attachments/files/20103714/fntype-cc.patch)

rsmmr:
Includes slight cleanup for printing out methods: we now prefix them
just with `method` instead of `method extern`.

This is based on the patch in #2052.

Closes #2052.


hilti/toolchain/include/ast/declarations/field.h
```diff
@@ -37,6 +34,13 @@ class Field : public Declaration {
             return child<QualifiedType>(0);
     }
 
+    // std::optional<type::function::CallingConvention> callingConvention() const {
```
evantypanski, line 37: was this leftover from the patch?

evantypanski (approved):

PR #2098, opened 2025-06-25
Allow registering a module twice.

### LINKED ISSUE #1085: Add a test which validates that running the same module multiple times does not lead to accidental state sharing
We currently assign a linker scope for each imported module which should ensure global state from different module instances is not shared. We should add a test to validate that this holds.

rsmmr:
So far, if one compiled the same HILTI module twice, each into its own
HLTO, then when loading the two HLTOs, the runtime system would skip
the second instance. However, that's not really what we want: a module
could intentionally be part of multiple HLTOs, in which case each
should get its own copy of that module state (i.e., its globals).

This change allows the same module to be registered multiple times,
with the HLTO linker scope distinguishing between the instances at
runtime, as usual. To make that work, we change how we compute the
linker scope: so far that was a hash of the source file, but we the
scopes to differ even if the sources are identical, so we're switching
to a random number.

Closes #1085.


hilti/toolchain/src/compiler/cxx/linker.cc
```diff
@@ -57,16 +59,11 @@ void cxx::Linker::finalize() {
...
-    // Create a scope string that's likely to be unique to this linker module.
-    std::size_t hash = 0;
-    for ( const auto& [id, path] : _modules ) {
-        std::ifstream ifs(path);
-        std::string content((std::istreambuf_iterator<char>(ifs)), (std::istreambuf_iterator<char>()));
-        hash = rt::hashCombine(hash, std::hash<std::string>()(content));
-    }
+    // Create a scope string that's unique to this linker module.
+    std::mt19937 mt19937((std::random_device()()));
+    auto scope = hilti::rt::fmt("%" PRIx64, std::uniform_int_distribution<>{}(mt19937));
```
bbannier: This is pretty blunt, and it effectively means that each compilation produces a distinct HLTO file. It would still behave surprisingly if one e.g., copies a HLTO file and loads both the original and the copy.

Do you recall why we decided to [not register the same module twice](https://github.com/zeek/spicy/blob/94d0700f3006b327f0b8305dc922e0df49e8b7ed/hilti/runtime/src/init.cc#L94-L99) in the first place? I wonder why we couldn't just remove that code instead of working around it with non-determinism.

Another deterministic alternative could be to inject the scope instead of reading it from the module. We could e.g., derive it from the path of the library which seems easy to explain.

rsmmr: > This is pretty blunt, and it effectively means that each compilation produces a distinct HLTO file. It would still behave surprisingly if one e.g., copies a HLTO file and loads both the original and the copy.

If it's actually the same *HLTO*, that will still be caught because the scope will be the same (there's a unit test for that). But yes, if one compiles it twice, and loads both HLTOs, that will now be accepted. However, I don't see how to otherwise distinguish legitimate uses cases from accidents.

> Do you recall why we decided to [not register the same module twice](https://github.com/zeek/spicy/blob/94d0700f3006b327f0b8305dc922e0df49e8b7ed/hilti/runtime/src/init.cc#L94-L99) in the first place? 

I believe that was the "avoid accidents" case: the assumption was that if I load the same module twice, that's a mistake. In some sense that's still the case now, except that we change the definition of "same module" from "same source code" to "same compilation product (i.e., HLTO)".

The reason that we used a hash originally was to produce stable output I believe. The random values aren't that nice (and blunt indeed!) but not sure what else to key on instead (see below re/ paths).

> I wonder why we couldn't just remove that code instead of working around it with non-determinism.
 
I see that as two different things: we could actually remove that code, the only result would be now accepting the copy-of-an-HLTO case. The different scopes, on the other hand, are to ensure that we keep the modules apart at runtime.

> Another deterministic alternative could be to inject the scope instead of reading it from the module. We could e.g., derive it from the path of the library which seems easy to explain.

I thought about that too but not sure that's really helping with much: the only difference I see is that now we would actually no longer detect the copy-of-HLTO case anymore (because the paths would differ). Or am I missing something?

bbannier: > > Do you recall why we decided to [not register the same module twice](https://github.com/zeek/spicy/blob/94d0700f3006b327f0b8305dc922e0df49e8b7ed/hilti/runtime/src/init.cc#L94-L99) in the first place?
> 
> I believe that was the "avoid accidents" case: the assumption was that if I load the same module twice, that's a mistake. In some sense that's still the case now, except that we change the definition of "same module" from "same source code" to "same compilation product (i.e., HLTO)".

I have the feeling that issues around users loading the exact same module multiple times are probably rare since the most likely place for this to appear is manual parsing with e.g., `spicy-driver` where the user manually passes all files. In Zeek this is abstracted away, and I suspect it might be the same for other tools embedding Spicy.

> > Another deterministic alternative could be to inject the scope instead of reading it from the module. We could e.g., derive it from the path of the library which seems easy to explain.
> 
> I thought about that too but not sure that's really helping with much: the only difference I see is that now we would actually no longer detect the copy-of-HLTO case anymore (because the paths would differ). Or am I missing something?

In addition to reproducible builds it seems conceptually more simple to me -- instead of relying on magic embedded data, we'd more directly translate what the user wanted (_"load the module here"_ where _module_ directly translates to a tangible file).

> The reason that we used a hash originally was to produce stable output I believe. The random values aren't that nice (and blunt indeed!) but not sure what else to key on instead.

Yes exactly. Likely not a real user concern, but embedding random data will completely break caching of builds, at least for the C++ linker file.

rsmmr: > random data will completely break caching of builds

ah, that's a good point, didn't think about that. let me see if I can make the paths work.

bbannier:
> See how this looks now. All changes in new commit (will squash later).

I like it!

tests/Baseline/hilti.ast.basic-module/output
```diff
@@ -6,12 +6,16 @@
 
 #include <hilti/rt/libhilti.h>
 
-extern const char* __hlt_hlto_scope;
+extern uint64_t __hlt_hlto_scope;
```
bbannier, line 9: Could you add `#include <cstdint>` to `libhilti.h`? We already seem to pull it in implicitly, but would be good to be explicit.

hilti/runtime/include/util.h
```diff
@@ -786,4 +786,25 @@ class Reference {
 
 } // namespace control
 
+/**
+ * Clone of `std::experimental::scope_exit`that calls an exit function on destruction.
```
bbannier:
```suggestion
 * Clone of `std::experimental::scope_exit` that calls an exit function on destruction.
```

Also in `base/util.h`.

hilti/runtime/include/util.h
```diff
@@ -786,4 +786,25 @@ class Reference {
...
+ */
+template<typename EF>
+struct scope_exit {
```
bbannier, line 793: Could you replace the literal copy in `base/util.h` with

```cpp
using rt::scope_exit;
```

hilti/runtime/src/library.cc
```diff
@@ -110,6 +117,17 @@ hilti::rt::Result<Nothing> hilti::rt::Library::remove() const {
...
+uint64_t hilti::rt::Library::currentScope() {
+    std::string scope;
+
+    if ( _current_path )
+        scope = _current_path->native();
+    else
+        scope = fmt("<scope-%>" PRIu64, ++_scope_counter);
+
+    return std::hash<std::string>()(scope);
+}
```
bbannier, line 129: We don't really need to go via the intermediary string

```suggestion
uint64_t hilti::rt::Library::currentScope() {
    if ( _current_path )
        return std::hash<std::string>()(_current_path->native());

    return ++_scope_counter;
}
```

rsmmr: I'll leave this as is, feels better to me to hash in both cases.

hilti/runtime/include/util.h
rsmmr: Ah, I had tried reusing it differently at first, which didn't work. This does.

hilti/runtime/src/library.cc
bbannier: All creating the extra string for `_scope_counter` does it add some "salt" to the hash (plus force a bunch more allocations which aren't that bad given how seldom this code runs), but if `std::hash` is any good it doesn't add anything to e.g., reduce the likelihood of collisions.

PR #2099, opened 2025-06-25
Fully implement iterator interface for `set::Iterator`.

bbannier:
We use our container iterators like stdlib iterators, e.g., we expect that they can work with `std::back_inserter`. This requires that our iterators implement the stdlib iterator interface, either by deriving from `std::iterator_category`, or by providing needed dependent types by hand. In fact, with C++20 and its support for concepts the standard library might more strictly enforce this requirement and non-complying iterators might stop compiling.

This patch implements the full iterator interface for `set::Iterator`.

### DIFF
--- a/hilti/runtime/include/types/set.h
+++ b/hilti/runtime/include/types/set.h
@@ -38,6 +38,12 @@ class Iterator {
     typename S::V::iterator _iterator;
 
 public:
+    using iterator_category = typename S::V::iterator::iterator_category;
+    using value_type = typename S::V::value_type;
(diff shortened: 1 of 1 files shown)


bbannier:
I pulled this from https://github.com/zeek/spicy/pull/2096. Zeek wants to potentially build with C++20 (see ﻿https://github.com/zeek/zeek/pull/4549) and since it builds Spicy-generated parsers as part of the Zeek build, these parsers also need to contain valid C++20.

PR #2103, opened 2025-06-26
Add missing deref operations for struct try-member/has-member operators.

### LINKED ISSUE #2100: Internal error for ternaries and `HasMember` check
I bisected and the culprit is 5b228f8b4be1c7464e03b9901d2171b1af971a79

The following Spicy code gets an internal error (not quite convinced it's minimal):

```
module Test;

function stringify(data: Data) {
    local elements = data?.elements ? data.elements : data.elements;
    for (ele in elements) {
        stringify(ele);
    }
}

type Data = unit {
    elements: Data[1];
};
```

Error:

```
$ spicyc -c test.spicy
internal error: unexpected type, want type::Struct but have type::ValueReference
[1]    52210 abort      spicyc -c test.spicy
```

Popped up when working on https://github.com/zeek/zeek/pull/4570

rsmmr:
The regular member access operators already did this.

Closes #2100.

### DIFF
--- a/hilti/toolchain/src/compiler/codegen/operators.cc
+++ b/hilti/toolchain/src/compiler/codegen/operators.cc
@@ -910,7 +910,11 @@ struct Visitor : hilti::visitor::PreOrder {
     void operator()(operator_::struct_::HasMember* n) final {
         const auto& id = n->op1()->as<expression::Member>()->id();
 
-        if ( auto* f = n->op0()->type()->type()->as<type::Struct>()->field(id); f->isOptional() )
+        auto* type = n->op0()->type()->type();
+        if ( type->isReferenceType() )
+            type = type->dereferencedType()->type();
+
+        if ( auto* f = type->as<type::Struct>()->field(id); f->isOptional() )
             result = fmt("%s.has_value()", memberAccess(n, id));
         else
             result = "true";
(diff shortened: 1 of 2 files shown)


hilti/toolchain/src/compiler/codegen/operators.cc
```diff
@@ -920,7 +924,11 @@ struct Visitor : hilti::visitor::PreOrder {
         const auto& id = n->op1()->as<expression::Member>()->id();
         assert(! lhs);
 
-        if ( auto* f = n->op0()->type()->type()->as<type::Struct>()->field(id); f->isOptional() ) {
+        auto* type = n->op0()->type()->type();
+        if ( type->isReferenceType() )
+            type = type->dereferencedType()->type();
+
+        if ( auto* f = type->as<type::Struct>()->field(id); f->isOptional() ) {
```
evantypanski, line 931: Random unrelated question: Is this `TryMember` useful? I see it's slightly different than just `<struct>.<field>` if the field is undefined without a default, but wondering if that manifests anywhere. I have just used `.` for field accesses in all cases, just potentially with a `?.` check first

evantypanski (approved):

hilti/toolchain/src/compiler/codegen/operators.cc
rsmmr: The main use case is on the Zeek side inside EVT files, documented here: https://github.com/zeek/zeek-docs/blob/5ff21a13afd35f4a4364b1ed3fd152c73411e8ce/devel/spicy/reference.rst?plain=1#L448. Example: https://github.com/zeek/zeek/blob/22958f7cdf1810ca4aa6a98c0e6a2cbe9eebca0e/testing/btest/spicy/tuple-arg.zeek#L48. While that example is simple, it would also work if that `self.?software` was nested deeply inside some larger other expression (the exception would abort evaluation of the expression, flagging to Zeek to leave the field unset). 

I don't know if anybody actually uses this (or is even aware of it ...) but personally I think it's pretty neat. :-)

PR #2104, opened 2025-07-01
Random cleanup

evantypanski:
I got enough little things randomly accumulating that figured I'd get it off my plate

The `auto op` change was caught by local warnings, but I thought that would've been fixed by #2037

And a weird one here was the copyright change - I thought that was fixed by e8200c9 but it skipped that file

tl;dr: the robots are failing us


hilti/toolchain/src/compiler/resolver.cc
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2020-2023 by the Zeek Project. See LICENSE for
+// Copyright (c) 2020-now by the Zeek Project. See LICENSE for details.
```
bbannier, line 1: There's a couple more of these. Could you update `tests/Scripts/license-header.py` and fix the remaining cases?

```diff
diff --git a/tests/Scripts/license-header.py b/tests/Scripts/license-header.py
index cb88d28dc..c5777f90e 100755
--- a/tests/Scripts/license-header.py
+++ b/tests/Scripts/license-header.py
@@ -6,8 +6,8 @@ import sys
 for f in sys.argv[1:]:
     with open(f) as input:
         if not any(
-            re.match(r".*Copyright.*by\ the\ Zeek\ Project", x)
+            re.match(r".*Copyright\ \(c\)\ 2020-now\ by\ the\ Zeek\ Project", x)
             for x in input.readlines()
         ):
-            print(f"{f} does not seem to contain a license header")
+            print(f"{f} does not seem to contain a valid license header")
             sys.exit(1)
```

hilti/toolchain/include/ast/operators/function.h
bbannier, line 1: Yeah, I am not sure why clang-tidy doesn't catch these since it even has a corresponding `function.cc` which is part of the build; there are many more like this, e.g., `hilti/toolchain/include/ast/operators/real.h`. Fine to fix here now, but also okay if you tack these onto other PRs.

hilti/toolchain/include/ast/operator-registry.h
```diff
@@ -120,7 +120,7 @@ class Registry {
  * operator available under that name.
  */
 inline auto get(std::string_view name) {
-    if ( auto op = Registry::singleton().byName(name) )
+    if ( const auto* op = Registry::singleton().byName(name) )
```
bbannier, line 123: This seems to be a similar issue with diagnostics in headers. I get a diagnostic from ~clang-tidy~ clangd in my editor here, but nothing when running `clang-tidy` on this header or the corresponding `operator-registry.cc`, I wonder if we miss some configuration knob.

bbannier (dismissed):

hilti/toolchain/include/ast/operators/function.h
evantypanski: I've been fixing them when I notice in a file I'm editing, but sometimes I just look at a file and change them but they don't live anywhere for a bit, hence why they're here. I didn't systematically go through and get all of them

hilti/toolchain/src/compiler/resolver.cc
evantypanski: Ah good catch, I just grepped for the beginning of the old copyright string and only found that one. Weirdly precommit found an issue with `.github/workflows/email-failure.yml` but not `.github/workflows/email-success.yml`. hmph

hilti/toolchain/include/ast/operator-registry.h
evantypanski: seems like this case is because of the dash in the file name? I'll double check first

bbannier: Wow, I think you are right 🤦

evantypanski: I'll fix it here and run clang tidy on all files as a sanity check :)

evantypanski: Should be fixed, I also ran `readability-qualified-auto` on the extra files

EDIT: Welp, I'll run more checks on them :)

evantypanski:
CI lint failure looked like a FP? [maybe related](https://github.com/llvm/llvm-project/issues/110502)

```
FAILED: hilti/toolchain/CMakeFiles/hilti-objects.dir/src/ast/builder/builder.cc.o 
/usr/bin/cmake -E __run_co_compile --launcher=/usr/bin/ccache --tidy="/usr/bin/clang-tidy-20;--extra-arg-before=--driver-mode=g++" --source=/tmp/cirrus-ci-build/hilti/toolchain/src/ast/builder/builder.cc -- /usr/bin/clang++-20  -I/tmp/cirrus-ci-build/3rdparty/reproc/reproc++/include -I/tmp/cirrus-ci-build/build/include -I/tmp/cirrus-ci-build/hilti/toolchain/include -I/tmp/cirrus-ci-build/3rdparty -I/tmp/cirrus-ci-build/hilti/runtime/include -fsanitize=address -fno-omit-frame-pointer -fno-optimize-sibling-calls -O1 -shared-libasan -Werror -g -std=c++17 -fPIC -Wall -O0 -MD -MT hilti/toolchain/CMakeFiles/hilti-objects.dir/src/ast/builder/builder.cc.o -MF hilti/toolchain/CMakeFiles/hilti-objects.dir/src/ast/builder/builder.cc.o.d -o hilti/toolchain/CMakeFiles/hilti-objects.dir/src/ast/builder/builder.cc.o -c /tmp/cirrus-ci-build/hilti/toolchain/src/ast/builder/builder.cc
/tmp/cirrus-ci-build/hilti/toolchain/include/hilti/compiler/detail/cxx/elements.h:109:94: error: use c++14 style type templates [modernize-type-traits,-warnings-as-errors]
  109 |     template<typename... T, typename enable = std::enable_if_t<(... && std::is_convertible_v<T, std::string_view>)>>
      |                                                                                              ^
      |                                                                                              _t
```

I'd like a better way to get those ctors accessible, but forwarding makes the build fail and more stuff required. Not looking to do a big cleanup here either, though it may just be necessary if this isn't good. I'd rather make an issue for further work, this was just a random cleanup :)

hilti/toolchain/include/compiler/detail/cxx/elements.h
```diff
@@ -106,7 +106,8 @@ class ID : public detail::IDBase<ID, normalizeID> {
     ID(std::string_view s, AlreadyNormalized n) : Base(s, n) {}
 
     /** Concatenates multiple strings into a single ID, separating them with `::`. */
-    template<typename... T, typename enable = std::enable_if_t<(... && std::is_convertible_v<T, std::string_view>)>>
+    template<typename... T,
+             typename enable = std::enable_if_t<(... && std::is_convertible_v<T, std::string_view>)>> // NOLINT
```
bbannier: Nit: Let's only silence that specific lint, and maybe insert a link to the issue. Indention gets a little funky since we don't want `NOLINT` to be part of the docstring, but that'd be fine with me.

```suggestion
    template<typename... T,
             typename enable = std::enable_if_t<
                 (... && std::is_convertible_v<
                             T, // NOLINT(modernize-type-traits), see https://github.com/llvm/llvm-project/issues/110502
                             std::string_view>)>>
```

hilti/toolchain/include/ast/id.h
```diff
@@ -31,7 +31,8 @@ class ID : public detail::IDBase<ID> {
     ID(std::string_view s, AlreadyNormalized n) : Base(s, n) {}
 
     /** Concatenates multiple strings into a single ID, separating them with `::`. */
-    template<typename... T, typename enable = std::enable_if_t<(... && std::is_convertible_v<T, std::string_view>)>>
+    template<typename... T,
+             typename enable = std::enable_if_t<(... && std::is_convertible_v<T, std::string_view>)>> // NOLINT
```
bbannier: I reran the `lint` job locally and it looks like this `NOLINT` is not needed.

```suggestion
    template<typename... T, typename enable = std::enable_if_t<(... && std::is_convertible_v<T, std::string_view>)>>
```

PR #2105, opened 2025-07-03
Allow registering multiple modules in the same scope.

bbannier:
In 2ac6fa1 we switched to dynamically
injecting scopes for libraries. This assumed that
`Library::currentScope` was called exactly once per collection of
inputs, but this is not the case if there are multiple modules in such a
collection which each need to register. `Library::currentScope` was
called once per module, but if one does not compile to a HLTO file it
uses an implementation which returns a new value for each invocation.
This was a problem in e.g., Zeek where we do not use HLTO files, but
instead generate C++ code for all files of a collection

This patch changes generated code so it only override the scope variable
embedded in generated code if it was not already set up. We need an
extra twist for the case of separately compiled linker files as tested
by `tests/hilti/hiltic/cc/separate-units.hlt`; for that we add a new
method to `Library` which allows to detect whether `currentScope` would
return a stable or generated value.

### DIFF
--- a/hilti/runtime/include/linker.h
+++ b/hilti/runtime/include/linker.h
@@ -3,4 +3,5 @@
 #pragma once
 
 #define HILTI_EXPORT __attribute__((visibility("default")))
+#define HILTI_HIDDEN __attribute__((visibility("hidden")))
 #define HILTI_WEAK __attribute__((weak))
--- a/hilti/toolchain/src/compiler/cxx/linker.cc
+++ b/hilti/toolchain/src/compiler/cxx/linker.cc
@@ -62,7 +62,7 @@ void cxx::Linker::finalize() {
     // `0` encoding unset. We will inject the actual scope at runtime when the
     // library is loaded.
     const auto& cxx_namespace = _codegen->context()->options().cxx_namespace_intern;
-    unit->add(fmt("HILTI_EXPORT HILTI_WEAK uint64_t %s_hlto_scope = 0;", cxx_namespace));
+    unit->add(fmt("HILTI_HIDDEN uint64_t %s_hlto_scope = 0;", cxx_namespace));
 
     for ( const auto& j : _joins ) {
         for ( const auto& c : j.second ) {
--- a/tests/Baseline/hilti.hiltic.print.empty/output
+++ b/tests/Baseline/hilti.hiltic.print.empty/output
@@ -32,5 +32,5 @@ extern void __hlt::Foo::__register_module() {
 
 const char HILTI_EXPORT HILTI_WEAK * __hlt_hlto_library_version = <...>
 const char HILTI_EXPORT HILTI_WEAK * __hlt_hlto_bind_to_version = <...>
-HILTI_EXPORT HILTI_WEAK uint64_t __hlt_hlto_scope = 0;
+HILTI_HIDDEN uint64_t __hlt_hlto_scope = 0;
 
--- a/tests/Baseline/hilti.hiltic.print.globals/output
+++ b/tests/Baseline/hilti.hiltic.print.globals/output
@@ -55,5 +55,5 @@ extern void __hlt::Foo::__register_module() {
 
 const char HILTI_EXPORT HILTI_WEAK * __hlt_hlto_library_version = <...>
 const char HILTI_EXPORT HILTI_WEAK * __hlt_hlto_bind_to_version = <...>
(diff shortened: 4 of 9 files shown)


hilti/runtime/src/library.cc
```diff
@@ -118,14 +118,10 @@ hilti::rt::Result<Nothing> hilti::rt::Library::remove() const {
...
-    std::string scope;
-
     if ( _current_path )
-        scope = _current_path->native();
-    else
-        scope = fmt("<scope-%>" PRIu64, ++_scope_counter);
+        return std::hash<std::string>{}(_current_path->native());
 
-    return std::hash<std::string>()(scope);
+    return ++_scope_counter;
```
rsmmr: Let's skip this commit, I really dislike returning a hash in one case and a meaningful value is another.

hilti/toolchain/src/compiler/cxx/unit.cc
```diff
@@ -74,7 +74,11 @@ void Unit::_addModuleInitFunction() {
         add(extern_scope);
 
         cxx::Block register_;
-        register_.addStatement(fmt("%s = ::hilti::rt::Library::currentScope()", scope));
+
+        // Set the scope if it wasn't already set up.
+        register_.addIf(fmt("::hilti::rt::Library::isLoading() || %s == 0", scope),
```
rsmmr: This is getting pretty complex/tricky for the generated code. How about a slightly different twist where we pass a  pointer to `__hlto_scope` into the runtime library and then centralize this logic there, like:
```
void hilti::rt::Library::setScope(uint64_t* scope) {
    if ( ::hilti::rt::Library::isLoading() || *scope == 0 )
        *scope = ::hilti::rt::Library::currentScope();
}
```

And we then probably don't even need `isLoading()` and `currentScope()` anymore, and can just inline those two inside `setScope()`, along with some generous comments explaining what's going on.

And then in unit.cc we simply generate:
```
    register_.addStatement(fmt("::hilti::rt::Library::setScope(&%s)", scope));
```

hilti/toolchain/src/compiler/resolver.cc
```diff
@@ -1105,7 +1105,7 @@ struct VisitorPass2 : visitor::MutatingPostOrder {
 
     void operator()(expression::Keyword* n) final {
         if ( n->kind() == expression::keyword::Kind::Scope && ! n->type()->isResolved() ) {
-            auto* ntype = builder()->qualifiedType(builder()->typeString(), Constness::Const);
+            auto* ntype = builder()->qualifiedType(builder()->typeUnsignedInteger(64), Constness::Const);
```
rsmmr, line 1108: I'm surprised none of our tests failed on this one.

spicy/lib/spicy_rt.hlt
```diff
@@ -24,8 +24,8 @@ public type Sink = struct {
     method void connect_filter(strong_ref<*> unit &requires-type-feature="is_filter");
     method void connect_mime_type(string mime_type);
     method void connect_mime_type(bytes mime_type);
-    method void connect_mime_type(string mime_type, string scope) &internal;
-    method void connect_mime_type(bytes mime_type, string scope) &internal;
+    method void connect_mime_type(string mime_type, uint<64> scope) &internal;
+    method void connect_mime_type(bytes mime_type, uint<64> scope) &internal;
```
rsmmr, line 28: Likewise, not sure why our tests didn't catch this.

hilti/toolchain/src/compiler/cxx/unit.cc
bbannier: Great idea. Let me change that to accept a mutable reference so we don't have to worry about null value.

hilti/runtime/src/library.cc
bbannier: Fair enough.

hilti/toolchain/src/compiler/cxx/unit.cc
rsmmr: Pointer is preferred for mutable arguments, per our coding conventions (not documented, but still :) ).

bbannier: Ah, I could have sworn it was the other way around outside of `Node` functions. Changing to a pointer.

rsmmr (dismissed):

PR #2106, opened 2025-07-03
[DO NOT MERGE] Experiment changing benchmark return type

evantypanski:
This is an experiment with directly changing the return type of `Inner` in the benchmarks. There were a couple goals here:

1) See what challenges come up when changing function returns
2) See what benchmark changes we can expect when that lands

Overall it seems 1) is: a lot more random issues than I thought. Then 2) is: locally it was significant, but not huge. The more units get tossed around to parse random things, the more it will matter. We'll see what CI benchmarks say about it.

This doesn't do any other optimizations (in fact: removes others) so it's also not as effective as we should be able to eventually get. I'm just experimenting


evantypanski:
+15% ish on the benchmarks after removing other optimizations, and it's only part of the optimizations. ¯\\\_(ツ)_/¯

good enough for the experiment

PR #2110, opened 2025-07-08
Cleanup&change operator precedence for `in` and `!in`

### LINKED ISSUE #1905: Unintuitive precedence for unary `*` and `in` operators
Given this program:

```
module Test;

# This works fine
print *begin(b"hi") == 'h';

# This has a `spicyc` compiler error:
# [error] [..] cannot coerce expression 'begin(b"hi")' of type 'iterator<bytes>' to type 'uint<8>'
print *begin(b"hi") in set<uint8>('h', 'i');

# This works again
print (*begin(b"hi")) in set<uint8>('h', 'i');
```

I get an error on that second `print` - but I don't see a good case for `*` to not go first.

It might also get the right precedence but not get validated right, but I haven't looked very deeply yet.

Is this expected? It feels like a bug to me. Should also check other similar cases

evantypanski:
Closes #1905 

This is a decent amount of cleanup, plus a docs change and a small precedence change. Technically it is a breaking change - `in` and `!in` now have different precedence. But, I tried a bunch of analyzers (eg Zeek analyzers, spicy-analyzers) and they all compile. I think it's so unintuitive for `in` and `!in` to bind so tightly that I don't mind that.

There are misc. other tiny changes described in e0df94183df4f6ff0ec989a90910b7d52bdce610 as well

### DIFF
--- a/doc/programming/language/precedence.rst
+++ b/doc/programming/language/precedence.rst
@@ -1,0 +1,49 @@
+
+.. _precedence:
+
+===================
+Operator Precedence
+===================
+
+In Spicy, operator precedence is generally pretty similar to other
+languages, with some Spicy-specific operators added. The following
+table summarizes Spicy's operator precedence, from highest to lowest:
+
+.. list-table:: Operator Precedence
+   :widths: 20 80
+   :header-rows: 1
+
+   * - Associativity
+     - Operators
+   * - Left
+     - ``.``, ``[``, ``?.``, ``.?``
+   * - Right
+     - ``!``, ``*`` (unary), ``~``, ``-`` (unary), ``|...|``, ``--``, ``++``
+   * - Right
+     - ``**``
+   * - Left
+     - ``%``, ``*``, ``/``
+   * - Left
+     - ``+``, ``-``
+   * - Left
+     - ``<<``, ``>>``
+   * - Left
+     - ``&``
+   * - Left
+     - ``^``
+   * - Left
+     - ``|``
+   * - Left
+     - ``<``, ``>``, ``>=``, ``<=``
+   * - Left
+     - ``==``, ``!=``
+   * - Left
+     - ``in``, ``!in``
+   * - Left
+     - ``&&``
+   * - Left
+     - ``||``
+   * - Left
+     - ``?``, ``:``
+   * - Left
+     - ``=``, ``-=``, ``+=``, ``*=``, ``/=``
--- a/spicy/toolchain/src/compiler/parser/parser.yy
+++ b/spicy/toolchain/src/compiler/parser/parser.yy
@@ -320,6 +320,8 @@ static std::vector<hilti::DocString> _docs;
 %type <uint64_t> const_uint
 %type <bool>     opt_skip opt_skip_
 
+// NOTE: Operator precedence is documented in doc/programming/language/precedence.rst
+// Update that file when any changes are made here.
 %left '=' MINUSASSIGN PLUSASSIGN TIMESASSIGN DIVIDEASSIGN
 %left '?' ':'
 %left OR
--- a/doc/programming/language/index.rst
+++ b/doc/programming/language/index.rst
@@ -30,6 +30,7 @@ Spicy pretty quickly.
     variables
     types
     statements
+    precedence
     packing
     conditional
     appendix


spicy/toolchain/src/compiler/parser/parser.yy
```diff
@@ -1062,25 +1053,23 @@ expr_e        : CAST type_param_begin qtype type_param_end '(' expr ')'   { $$ =
...
               | DOLLARDOLLAR                     { $$ = builder->expressionName(std::move("__dd"), __loc__);}
               | DOLLAR_NUMBER                    { // $N -> $@[N] (with $@ being available internally only, not exposed to users)
                                                    auto captures = builder->expressionKeyword(hilti::expression::keyword::Kind::Captures, builder->qualifiedType(builder->typeName("~hilti::Captures"), hilti::Constness::Mutable), __loc__);
                                                    auto index = builder->expressionCtor(builder->ctorUnsignedInteger($1, 64, __loc__), __loc__);
                                                    $$ = builder->expressionUnresolvedOperator(hilti::operator_::Kind::Index, {std::move(captures), std::move(index)}, __loc__);
                                                  }
+    ;
+
+call_expr     : scoped_id '(' opt_exprs ')'      { $$ = builder->expressionUnresolvedOperator(hilti::operator_::Kind::Call, {builder->expressionName(std::move($1)), builder->expressionCtor(builder->ctorTuple(std::move($3), __loc__))}, __loc__); }
```
rsmmr: Is it necessary to limit the callee to an ID? I don't think we're using anything else there right now, but in principle we could.

doc/programming/language/precedence.rst
```diff
@@ -0,0 +1,49 @@
+
+.. _precedence:
+
+===================
+Operator precedence
```
rsmmr:
```suggestion
Operator Precedence
```

doc/programming/language/precedence.rst
```diff
@@ -0,0 +1,49 @@
...
+===================
+
+While Spicy is like C, its operators have their own precedence. This
+matters for non-C operators like ``in``. The following table
+summarizes Spicy's operator precedence, from highest to lowest:
```
rsmmr: I think I'd just remove the reference to "C" here, like:

```suggestion
In Spicy, operator precedence is generally pretty similar to other languages, with some Spicy-specific operators added. The following table
summarizes Spicy's operator precedence, from highest to lowest:
```

rsmmr (commented):
Much better, thanks for cleaning this up. 

One request: please apply the same transformation to the HILTI parser, so that we keep them in sync; shouldn't be too difficult as the expressions are very similar. 

(Btw, I'm still looking for a way to avoid the duplication of Bison code between the Bison parsers for HILTI and Spicy; would be nice if we could reuse parts where the grammar matches, instead of copy/pasting.)

spicy/toolchain/src/compiler/parser/parser.yy
evantypanski: Not really, it just increased the number of shift/reduce conflicts since now it would have to choose amongst all expressions rather than a subset.

My preference would be to keep it `scoped_id`, then if something else becomes callable, yank those out of `expr` into a separate `callable` production. Then it can be limited and the GLR parser won't be working with a bunch of parallel states

but maybe I'm just not used to the GLR parser lifestyle :) I pushed a fixup commit for that if you want to see what it's like, but I prefer the `scoped_id` in order to limit ambiguities

rsmmr: If we ever need it, it's most likely not just a specific syntactic construct to add, but actually full expressions: I'm picturing overloading expressions of certain types with a call operator, which our infrastructure would support just fine. With that in mind, I'd prefer to keep `expr` for the callee even with the additional conflicts just so that we know we could go that route still---unless there's something we know is not going to work in the new version (we had plenty conflicts already, but it doesn't seem to hurt anywhere so far). 

Same applies to the new HILTI grammar then.

rsmmr (dismissed):

evantypanski:
Oops, I forgot to baseline the test so I added a baseline, sorry! I'll merge tomorrow regardless

PR #2115, opened 2025-07-15
Clean up temporary JIT output files.

rsmmr:
Closes https://github.com/zeek/zeek/issues/4658.

### DIFF
--- a/hilti/toolchain/include/compiler/jit.h
+++ b/hilti/toolchain/include/compiler/jit.h
@@ -191,9 +191,11 @@ private:
     std::vector<hilti::rt::filesystem::path> _objects;
 
     struct Job {
+        ~Job();
+
         std::string cmdline;
         std::unique_ptr<reproc::process> process;
-        hilti::rt::filesystem::path output; // path to file capturing process output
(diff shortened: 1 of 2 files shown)


hilti/toolchain/include/compiler/jit.h
```diff
@@ -191,9 +191,11 @@ class JIT {
     std::vector<hilti::rt::filesystem::path> _objects;
 
     struct Job {
+        ~Job();
+
         std::string cmdline;
         std::unique_ptr<reproc::process> process;
-        hilti::rt::filesystem::path output; // path to file capturing process output
+        hilti::rt::filesystem::path output; // path to file capturing process output; file will be deleted by destructor
```
bbannier, line 198: This works for me. Another approach would be to lift `TemporaryFile` from tests into e.g., utils and use it here so this works automatically, but I am fine either way.

bbannier (approved):

hilti/toolchain/include/compiler/jit.h
rsmmr: I'll leave as is.

PR #2118, opened 2025-07-16
Backports for `release/1.11`

bbannier:


tests/Baseline/spicy.types.sink.reassembler.wrong-init-seq/output
```diff
@@ -1,2 +1,2 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
-[error] terminating with uncaught exception of type spicy::rt::SinkError: sink cannot update initial sequence number after activity has already been seen (<...>/wrong-init-seq.spicy:20:19-38:1)
+[error] terminating with uncaught exception of type spicy::rt::SinkError: sink cannot update initial sequence number after activity has already been seen (<...>/wrong-init-seq.spicy:14:9-14:50)
```
bbannier, line 2: The is the only really review-worthy change in this PR. It is not clear to me why it is needed only on this branch but not elsewhere, but the change seems to make sense to me.

rsmmr: Same here, for both statements. :-)

rsmmr (approved):

PR #2119, opened 2025-07-16
Fix missing inline functions in enum prototypes.

rsmmr:
Our prototype generation could miss function bodies for `inline`
functions.


bbannier (approved):

bbannier:
Also backported to `release/1.13`

PR #2126, opened 2025-07-25
Fix dataflow extraction of reads and writes in assignments.

### LINKED ISSUE #2123: Build out handling of assignments in CFG code
Here's one of those places that assignment expressions will bite. There are TONS of complicated interactions that can happen here (I'm also thinking assignments on the right-hand-side of a shortcircuited operator like `||`). But, here's a simple case that gets broken:

```
module Foo {

import hilti;

function void f() {
    local uint<64> x;
    x = 1;
    local bool b = ((x = 5) == 5);
    hilti::print(x);
}

f();

}
```

Since the `local bool b` is removed, the `x = 5` is not output, so with optimizations this prints `1`, without it prints `5`.

I haven't dove further into this, but you may also hit problems if you just keep all declarations that have assignments in them if you blindly assume that the assignment shadows previous ones because of short circuiting. For example, have to make sure that both possible values of `x` are live after this:

```
x = 5;
local bool b = (y == 5) || ((x = 6) == 6);
```

Or else you might wrongly assume that `x == 6` after that point.

_Originally posted by @evantypanski in https://github.com/zeek/spicy/pull/1697#discussion_r2126797032_

bbannier:
We previously would not correctly recognize value assignments if they
were contained in other assignments, e.g., constructs of the form

    x = (y = 1);

This patch fixes the extraction for such cases.

Closes #2123.

### DIFF
--- a/hilti/toolchain/src/compiler/cfg.cc
+++ b/hilti/toolchain/src/compiler/cfg.cc
@@ -453,9 +453,11 @@ std::string CFG::dot() const {
             auto in_out = [&]() -> std::string {
                 auto to_str = [&](const auto& xs) {
                     std::vector<std::string> ys;
-                    for ( const auto& [_, stmts] : xs ) {
+                    for ( const auto& [decl, stmts] : xs ) {
+                        std::vector<std::string> xs;
                         for ( const auto& stmt : stmts )
-                            ys.push_back(escape(stmt->print()));
+                            xs.push_back(escape(stmt->print()));
+                        ys.push_back(util::fmt("%s: %s", decl->id(), util::join(xs, ", ")));
                     }
 
                     std::ranges::sort(ys);
--- a/hilti/toolchain/src/compiler/cfg.cc
+++ b/hilti/toolchain/src/compiler/cfg.cc
@@ -8,6 +8,7 @@
 #include <iterator>
 #include <map>
 #include <optional>
+#include <ranges>
 #include <string>
 #include <unordered_map>
 #include <unordered_set>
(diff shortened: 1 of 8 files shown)


hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -589,25 +587,34 @@ struct DataflowVisitor : visitor::PreOrder {
...
+                node = node->parent();
+
+                // We either climb up to the original statement, or stop when
+                // we have found a use. A particular name can only appear on a
+                // single side of an assignment (if the same identifier appears
+                // elsewhere it would be a different name instance), so
+                // stopping once we have the first use extracts the desired
+                // information, e.g., `x = (x = 1)` only ever writes to `x`,
+                // but does not read it, even though `x` also appears on the
+                // RHS.
```
evantypanski: Random point: `x` would technically be a use in the generated C++, this matters for implicit type conversions. IDK how I'd trigger it for us, so I won't harp on that too much

(oops meant to start a review)

tests/hilti/cfg/assign.hlt
```diff
@@ -6,7 +6,22 @@ module foo {
 function uint<64> fn() {
     local auto x = 1;
     x = x + 1;
-    return x;
```
evantypanski, line 9: was this intentionally removed?

evantypanski (commented):
this does still seem to be removing declarations that have needed assignment expressions though:

```
module Test {

import hilti;

function void f() {
    local x = "bad";
    local x2 = x = "good";
    hilti::print(x);
}

f();

}
```

```
$ HILTI_OPTIMIZER_ENABLE_CFG=1 hiltic -j test.hlt
bad
$ hiltic -j test.hlt
good
```

hilti/toolchain/src/compiler/cfg.cc
bbannier: Good point, I added a read here now which also leads to simpler code.

tests/hilti/cfg/assign.hlt
bbannier: Unintentional, but thankfully I didn't hide any bugs.

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -844,8 +837,11 @@ void CFG::_populateDataflow() {
                 if ( auto it = transfer.in.find(decl); it != transfer.in.end() ) {
                     const auto& [_, prev] = *it;
 
+                    changed |= ! transfer.kill.contains(decl);
+                    auto& kill = transfer.kill[decl];
+
                     for ( const auto& p : prev ) {
-                        auto [_, inserted] = transfer.kill.insert(p);
+                        auto [_, inserted] = kill.insert(p);
```
evantypanski, line 844: Related to short-circuiting, this shouldn't always kill the previous declarations, since it might be context-dependent on whether it gets updated or not in assignment expressions:

```
function void f(int<64> i) {
    local x = "bad";
    local x2 = (i == 10) || ((x = "good") == "good");
    hilti::print(x);
}
```

Depending on `i`, `x` may end up as `good` or `bad`. It looks like `x = "bad"` is added to the kill set after `x2`, which would be wrong. Happy to punt that for later, but just pointing it out

Image

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -586,46 +584,41 @@ struct DataflowVisitor : visitor::PreOrder {
...
+
+                        in_assignment = true;
+                    }
+                }
+                node = node->parent();
+            } while ( node && node != root.value() );
         }
 
+        if ( in_assignment )
+            ; // Nothing, handled above.
```
evantypanski, line 620: what's the point of `in_assignment`? ~seems like a leftover~

oops nevermind there's an `else if` after that I completely skipped :)

evantypanski (approved):
Seems reasonable to leave the short circuiting issue for later if needed, ~just one other minor comment~

hilti/toolchain/src/compiler/cfg.cc
bbannier: Opened https://github.com/zeek/spicy/issues/2129 to track this further, I wonder how long we can keep the approach of storing statements in the CFG.

hilti/toolchain/src/compiler/cfg.cc
bbannier: Yeah, this is what I mentioned in the [comment above](https://github.com/zeek/spicy/pull/2126#issuecomment-3132890796). Overall this approach is pretty fragile.

PR #2130, opened 2025-07-29
Avoid removing `__captures` from non-public units.

evantypanski:
This caused a regression/C++ compiler error in spicy-nats.

this is the same solution as before https://github.com/zeek/spicy/pull/2083#discussion_r2174989210 - but I made it specifically only apply for `__captures`. Apparently the other tests pass because their units are public, whereas when I hit the issue I hadn't made public units stable yet.

Not sure why it only applies to `__captures` but it seems like the case.

### DIFF
--- a/tests/spicy/optimization/remove-captures-param.spicy
+++ b/tests/spicy/optimization/remove-captures-param.spicy
@@ -1,0 +1,18 @@
+# @TEST-REQUIRES: which FileCheck
+# @TEST-EXEC: spicyc -p %INPUT | FileCheck %INPUT
+# @TEST-EXEC: spicyc -dj %INPUT
+
+# @TEST-DOC: Ensures captures are not removed from non-public units
+
+module Test;
+
+type Data = unit {
+    # CHECK: hook void __on_x(vector<bytes> __captures);
+    x: /(abc)/ {
+        print $1;
+    }
+
+    # Ensure it is removed when captures aren't used.
+    # CHECK: hook void __on_y();
+    y: /(def)/ {}
+};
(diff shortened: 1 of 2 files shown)


tests/spicy/optimization/remove-captures-param.spicy
```diff
@@ -0,0 +1,11 @@
+# @TEST-EXEC: spicyc -j %INPUT
+
+# @TEST-DOC: Ensures captures are not removed from non-public units
+
+module Test;
+
+type Data = unit {
```
bbannier, line 9: Could you add a check that we do remove the `__captures` parameter if it is unused? Maybe simplest to add some `FileCheck` assertions, e.g.,

```ruby
# @TEST-REQUIRES: which FileCheck
# @TEST-EXEC: spicyc -p %INPUT -o output.hlt
# @TEST-EXEC: FileCheck %INPUT < output.hlt
# @TEST-EXEC: spicyc -dj %INPUT

# @TEST-DOC: Ensures captures are not removed from non-public units

module Test;

type Data = unit {
    # CHECK: hook void __on_x(vector<bytes> __captures);
    x: /(abc)/ {
        print $1;
    }
    # CHECK: hook void __on_y();
    y: /(def)/ {}
};
```

I gave the fields names to be independent of how we name hooks for anon fields, but matching on `__on__anon` and `__on__anon_2` would work as well. I also changed the regexps to actually capture something to refer to, but since we do not run the parsers this currently makes no difference.

bbannier (dismissed):

tests/spicy/optimization/remove-captures-param.spicy
evantypanski: I made #2131 for the out-of-range captures issue, I forgot the parens but I think we should be able to statically determine that

PR #2133, opened 2025-08-01
Fix issues reported by coverity

bbannier:


evantypanski (approved):
those moves sure are ugly, though

PR #2134, opened 2025-08-07
Add bit ranges to type information for bitfields.

rsmmr:
- **Fix unit test for bitfields.**
- **Add bit ranges to type information for bitfields.**

### DIFF
--- a/hilti/runtime/include/type-info.h
+++ b/hilti/runtime/include/type-info.h
@@ -409,13 +409,18 @@ public:
      * Constructor.
      *
      * @param name ID of the field
+     * @param lower lower bit of the field
+     * @param upper upper bit of the field
      * @param type type of the field
      * @param offset offset of the field inside the bitfield's storage tuple.
      */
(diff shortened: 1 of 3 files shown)


hilti/runtime/include/type-info.h
```diff
@@ -409,13 +409,18 @@ class Bits {
      * Constructor.
      *
      * @param name ID of the field
+     * @param lower lower bit of the field
+     * @param upper upper bit of the field
      * @param type type of the field
      * @param offset offset of the field inside the bitfield's storage tuple.
      */
-    Bits(const char* name, const TypeInfo* type, std::ptrdiff_t offset) : name(name), type(type), offset(offset) {}
+    Bits(const char* name, unsigned int lower, unsigned upper, const TypeInfo* type, std::ptrdiff_t offset)
```
evantypanski:
```suggestion
    Bits(const char* name, unsigned int lower, unsigned int upper, const TypeInfo* type, std::ptrdiff_t offset)
```

may not even be worth it, but `lower` and `upper` use two different styles to get the `unsigned int`

evantypanski (dismissed):

PR #2135, opened 2025-08-07
Add `&cxx-any-as-ptr` attribute.

rsmmr:
This is a very niche attribute extending the host application API:
Given an external C++ function that's made available to HILTI/Spicy
through `&cxxname`, if that function has a parameter of type `any`,
HILTI normally passes the corresponding argument as a *reference* to the
argument's original type. This works well if the C++ implementation
uses a template parameter to receive the
argument, which is how we use `any` inside the HILTI/Spicy runtime
libraries at various points. It however doesn't work well if one doesn't
want to (or cannot) use templates, for example because the function is
part of a library without a corresponding C++ template header.

For that case, this PR adds a new attribute `&cxx-any-as-ptr` that can be
combined (only) with `&cxxname`: it changes the passing convention for
`any` parameters to pass a *pointer* instead of a reference. The
pointer is still typed according to the value that the caller passes, but
now it can matched against functions taking a `void *`. Along
with the type information that HILTI, provides  to the C++ function as a
subsequent argument to any `any` parameter (which we already did; no change), the receiving
function can now, for example, create a runtime
`type_info::Value` instance for the `any` argument and work with that
generically. 

If HILTI is asked to generate a prototype for an
external function, it now also generates that with `void *` if the new
attribute is present.

For the record, I considered changing the calling convention to always use pointers, but
by-reference has some advantages for the cases where it works, for
example by being able to pass RHS values.

If this is all hard to comprehend, see
`tests/hilti/codegen/cxx-any-as-ptr.hlt` for an example. :)

The patch also includes a small fix for generating the C++ prototype
for functions with `any` parameters: the prototype missed including the type information 
as an additional argument coming with `any` parameters, even though we already called such 
functions that way.


spicy/toolchain/include/ast/attribute.h
```diff
@@ -20,6 +20,7 @@ const Kind Chunked("&chunked");
 const Kind Convert("&convert");
 const Kind Count("&count");
 const Kind Cxxname("&cxxname");
+const Kind CxxAnyAsPtr("&cxx-any-as-ptr");
```
evantypanski, line 23: is this currently usable from Spicy?

hilti/toolchain/src/compiler/codegen/codegen.cc
```diff
@@ -703,7 +714,13 @@ std::vector<cxx::Expression> CodeGen::compileCallArguments(const node::Range<Exp
     unsigned int i = 0;
     for ( const auto& p : params ) {
         Expression* arg = (i < args.size() ? args[i] : p->default_());
-        x.emplace_back(compile(arg, p->kind() == parameter::Kind::InOut));
+
+        if ( p->type()->type()->isA<type::Any>() && function &&
+             function->attributes()->has(hilti::attribute::kind::Cxxname) &&
```
evantypanski: Since `has` just uses `find` it may be worth removing it.. the next line uses `find` for the same thing and it's just as valid

rsmmr:
> Would it make sense for this to apply to parameters instead of the whole function? Logically it seems to fit nicer with the parameter

Good point. I had originally envisioned this as a migrating to some kind of v2 of function call conventions, until I realized that depending on use case both behaviors can be useful. With that, moving to a parameter attribute makes sense, and means a smaller diff too.

spicy/toolchain/include/ast/attribute.h
rsmmr: Yes, except that in the new version of putting the attribute on the parameter, Spicy needs a tweak because the yacc parser doesn't allow that yet. Also adding a test for the Spicy side.

hilti/toolchain/src/compiler/codegen/codegen.cc
rsmmr: +1, I'll kick out the `has()` method.

hilti/toolchain/src/compiler/validator.cc
```diff
@@ -260,6 +260,16 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
             if ( ! n->body() && ! is_hook && ! attrs->has(hilti::attribute::kind::Cxxname) )
                 error(fmt("function '%s' must have a body or be declared with &cxxname", n->id()), n);
         }
+
+        for ( const auto& p : n->ftype()->parameters() ) {
+            if ( p->attributes()->find(hilti::attribute::kind::CxxAnyAsPtr) ) {
+                if ( ! n->attributes()->find(hilti::attribute::kind::Cxxname) )
```
evantypanski, line 266: just noting that above it checks `if ( auto* attrs = n->attributes() )` but I don't think there's a way for `n->attributes() == nullptr` but this uses `attributes` outside of that check

evantypanski (approved):
👍

hilti/toolchain/src/compiler/validator.cc
rsmmr: I believe that check further up is legacy: since the Big-AST-Rewrite™, attribute sets are always initialized to an empty set.

PR #2136, opened 2025-08-07
Include all fields into a struct's type information, even if not emitted.

rsmmr:
So far a host application couldn't learn about any fields in a HILTI
`struct` that, although defined by the HILTI type, were skipped from
the C++ struct generated by the codegen (e.g., because it's anonymous
and hence won't receive a value anyways). Now we include these fields
into the struct's type information as well, so that they can still be
discovered during introspection. The field's type information gains a
new flag for querying whether it has been emitted. However, when
iterating a struct's field *values* (in constrast to its field
*types*) we continue to skip these fields since they don't have a
value after all. This also means that existing host applications
shouldn't run into any trouble when iterating the values, nothing
changes there.

### DIFF
--- a/hilti/runtime/src/tests/type-info.cc
+++ b/hilti/runtime/src/tests/type-info.cc
@@ -54,15 +54,19 @@ namespace {
 const hilti::rt::TypeInfo __ti_Test_X =
     {"Test::X", "Test::X", nullptr,
      new hilti::rt::type_info::Struct(std::vector<hilti::rt::type_info::struct_::Field>(
-         {hilti::rt::type_info::struct_::Field{"i", &hilti::rt::type_info::int32, offsetof(Test::X, i), false, false},
-          hilti::rt::type_info::struct_::Field{"s", &hilti::rt::type_info::string, offsetof(Test::X, s), false, false},
-          hilti::rt::type_info::struct_::Field{"y", &type_info::__ti_Test_Y, offsetof(Test::X, y), false, false}}))};
+         {hilti::rt::type_info::struct_::Field{"i", &hilti::rt::type_info::int32, offsetof(Test::X, i), false, false,
+                                               true},
+          hilti::rt::type_info::struct_::Field{"s", &hilti::rt::type_info::string, offsetof(Test::X, s), false, false,
+                                               true},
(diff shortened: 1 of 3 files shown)


hilti/toolchain/src/compiler/codegen/types.cc
```diff
@@ -910,24 +910,29 @@ struct VisitorTypeInfoDynamic : hilti::visitor::PreOrder {
...
-                fmt("::hilti::rt::type_info::struct_::Field{ \"%s\", %s, static_cast<std::ptrdiff_t>(offsetof(%s, "
-                    "%s)), %s, %s%s }",
-                    cxx::ID(f->id()), cg->typeInfo(f->type()), cxx_type_id, cxx::ID(f->id()), f->isInternal(),
-                    f->isAnonymous(), accessor));
+            std::string offset;
+
+            if ( ! f->isNoEmit() )
+                offset = fmt("static_cast<std::ptrdiff_t>(offsetof(%s, %s))", cxx_type_id, cxx::ID(f->id()));
+            else
+                offset = "std::size_t{0}";
```
evantypanski: Should this be casted to `std::ptrdiff_t`?

Also maybe it should be an invalid value from `offsetof` like `-1`? That way it won't clash with a real field

hilti/runtime/src/tests/type-info.cc
```diff
@@ -174,6 +178,34 @@ TEST_CASE("anonymous fields") {
...
+    CHECK_EQ(s->fields().size(), 1U);
+    CHECK(! s->fields()[0].get().isEmitted());
+
+    // We shouldn't see this field when iterating.
+    int count = 0;
+    for ( const auto& _ : type_info::value::auxType<type_info::Struct>(v)->iterate(v) ) {
+        (void)_;
+        count++;
+    }
+    CHECK_EQ(count, 0);
```
evantypanski: this could just use `ranges` if you like this more?:


```suggestion
    int count = std::ranges::distance(type_info::value::auxType<type_info::Struct>(v)->iterate(v));
    CHECK_EQ(count, 0);
```

rsmmr:
I'm going to pull this PR. It has a problem with retrieving offsets for structure fields (because that depends on exact field order); plus that way of aligning offsets with fields has always been fragile, and to some degree only worked by chance. I'll file a new PR with a different approach. Sorry for the noise.

evantypanski (approved):

Issue #2031, opened 2025-04-17
Propagate constants through function returns


evantypanski:
Given some hilti code:

```
module Test {
function int<32> one() {
    return 1;
}

function void calls_one() {
    local i = one();
}

calls_one();
}
```

The `local i = one();` check can be replaced with `one(); local i = 1;`

This itself isn't very useful, but there are a couple other considerations that would substantially speed up parsers:

1) This should "reduce" tuples, so if the return is `tuple<int<64>, int<32>>` and the second `int<32>` is always `1`, then it would make the function instead return `int32` (only the first).
2) This should also accept "passthroughs" from function parameter -> return with no uses as no use, so they can be removed.

The general steps are:

1) For each return in a function, determine if it's the same as all others (or a given tuple element is the same)
2) Ensure that, for variables, there are no other uses of that variable
3) Remove the return value from each return and modify the return type
4) Remove any assignments at each call site
5) Replace each assignment at return them with direct assignments afterwards

These may end up pretty tricky, but seem doable. However, this applies largely to generated functions, so it could also just check for known patterns in generated code.

PR #2138, opened 2025-08-11
Add constant propagation pass.

### LINKED ISSUE #2137: Control flow constant propagation
This is hoisted out from a point in #1690 so that I can assign myself :) - at this point it seems like I'll have a PR next week.

The point is:

> Migrate optimizations implemented in ConstantFoldingVisitor to control-flow based approach.
>
> Currently ConstantFoldingVisitor implements optimizations for simplifying logic around constants. If we migrate this optimization to a control-flow based approach we can make us of it for locals with known values as well.

This is a prereq for #2031 so I picked it up

I'm pretty sure the best way for this is to implement a separate dataflow analysis from the gen/kill sets in the CFG already. With that I can make my own transfer function to catch extra scenarios, like passing references to functions. Then anything that has two possible values becomes not-a-constant. So I'm making my own worklist algorithm.

Another part of this could be generalizing the approach I took with a worklist dataflow algorithm so all you have to do is provide a meet function, transfer function, init, and direction - then we can do dataflow analyses in the "traditional" way. IDK

Either way, this one turned out to be relatively straightforward and do what I want for #2031 - but a but more to go. I don't expect speedups from this yet - but it should enable a couple of other things.

evantypanski:
Closes #2137

This will be a draft for now but it does exactly what I want for #2031 - couple bits I think still need cleanup

This also implements my own dataflow - mainly to add the "not a constant" and have its own logic. Maybe it could use the CFG stuff in the future.

It also uses a fair few visitors, I may have gone overboard ;)

There's also removing the constant folding visitor and implementing its logic here, may do that in another PR (ie differentiate "folding" from "propagation")


evantypanski:
This causes test failures and an infinite loop in the Redis analyzer in Zeek, I'll look into that.

Also, this gives about a 3% speedup on an all-SSL pcap in Zeek with the Spicy SSL analyzer. Given that about half the runtime is in Spicy, a 6% speedup here would equate to that, so the 6% improvement in vectors is actually a real improvement in real code. I honestly didn't expect that here, we may have some good opportunities to speed up parsers with just constant propagation and folding with the CFG in the future.

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -977,6 +977,271 @@ struct ConstantFoldingVisitor : OptimizerVisitor {
...
+
+        struct TransferVisitor : hilti::visitor::PreOrder {
+            ConstantMap& constants;
+            NameNACer name_nac;
+
+            TransferVisitor(ConstantMap& constants) : constants(constants), name_nac(constants) {}
+
+            // Tries to evaluate an expression to a constant value given a map of known constants.
+            Expression* evaluate(Expression* expr) {
+                if ( expr->isA<expression::Ctor>() )
```
rsmmr: Quick fly-by comment: not all ctors are constant. Sets for example are not. You probably want `expr->isConstant()` here, probably in conjunction with the `Ctor` check.

tests/Baseline/spicy.optimization.default-parser-functions/opt.hlt
```diff
@@ -61,9 +61,8 @@ method tuple<const view<stream>, int<64>, const iterator<stream>, optional<hilti
     {
         hilti::debugIndent("spicy");
         local iterator<stream> __begin = begin(__cur);
-        local strong_ref<stream> filtered = Null;
 
-        if ( ! filtered ) 
+        if ( ! Null )
```
bbannier, line 65: I tried to understand why we emit this `Null` since as written in this HILTI source code it is invalid. It still compiles since in the AST we keep the coercions (from `Null` to the `strong_ref`, and from a `strong_ref` to a `bool`).
  Not folding branches over a constant `Null` like we do for `True` and `False` seems like it leaves potentially a lot of perf on the table since we in some cases do not remove writes to e.g., `__result` which we'd otherwise inline as well (and we already see a pretty big effect from the inlining done here).
 
This particular construct looks like this in the AST:

```
|| [debug/ast-final]             - statement::If [parent @s:12ef] (bla2.hlt:4:4-4:21) [@s:12ee]
|| [debug/ast-final]               - <empty>
|| [debug/ast-final]               - expression::LogicalNot [parent @s:12ee] (bla2.hlt:4:8-4:16)  (const)  (resolved) [@e:12ec]
|| [debug/ast-final]                 - expression::Coerced [parent @e:12ec] (bla2.hlt:4:9-4:16)  (const)  (resolved) [@e:1731]
|| [debug/ast-final]                   - expression::Ctor [parent @e:1731] (bla2.hlt:3:40-3:43)  (const)  (resolved) [@e:1a41]
|| [debug/ast-final]                     - ctor::Coerced [parent @e:1a41] (bla2.hlt:3:40-3:43) [@c:1a42]
|| [debug/ast-final]                       - ctor::Null [parent @c:1a42] (bla2.hlt:3:40-3:43) [@c:1a43]
|| [debug/ast-final]                         - QualifiedType <const="true" extern="false" side="rhs"> [parent @c:1a43] (bla2.hlt:3:40-3:43) [@q:1a44]
|| [debug/ast-final]                           - type::Null <declaration="-" type="-" unified="null" wildcard="false"> [parent @q:1a44] (bla2.hlt:3:40-3:43) (resolved) [@t:1a45]
|| [debug/ast-final]                       - ctor::StrongReference [parent @c:1a42] (bla2.hlt:3:40-3:43) [@c:1a46]
|| [debug/ast-final]                         - QualifiedType <const="true" extern="false" side="rhs"> [parent @c:1a46] [@q:1a47]
|| [debug/ast-final]                           - type::StrongReference <declaration="-" type="-" unified="strong_ref(stream)" wildcard="false"> [parent @q:1a47] (resolved) [@t:1a48]
|| [debug/ast-final]                             - QualifiedType <const="false" extern="false" side="rhs"> [parent @t:1a48] (bla2.hlt:3:21-3:26) [@q:1a49]
|| [debug/ast-final]                               - type::Stream <declaration="-" type="-" unified="stream" wildcard="false"> [parent @q:1a49] (bla2.hlt:3:21-3:26) (resolved) [@t:1a4a]
|| [debug/ast-final]                                 - QualifiedType <const="false" extern="false" side="rhs"> [parent @t:1a4a] (bla2.hlt:3:21-3:26) [@q:1a4b]
|| [debug/ast-final]                                   - type::stream::View <declaration="-" type="-" unified="view::stream" wildcard="false"> [parent @q:1a4b] (bla2.hlt:3:21-3:26) (resolved) [@t:1a4c]
|| [debug/ast-final]                                     - QualifiedType <const="false" extern="false" side="rhs"> [parent @t:1a4c] (bla2.hlt:3:21-3:26) [@q:1a4d]
|| [debug/ast-final]                                       - type::stream::Iterator <declaration="-" type="-" unified="iterator(stream)" wildcard="false"> [parent @q:1a4d] (bla2.hlt:3:21-3:26) (resolved) [@t:1a4e]
|| [debug/ast-final]                                         - QualifiedType <const="false" extern="false" side="rhs"> [parent @t:1a4e] (bla2.hlt:3:21-3:26) [@q:1a4f]
|| [debug/ast-final]                                           - type::UnsignedInteger <declaration="-" type="-" unified="uint8" width="8" wildcard="false"> [parent @q:1a4f] (bla2.hlt:3:21-3:26) (resolved) [@t:1a50]
|| [debug/ast-final]                   - QualifiedType <const="true" extern="false" side="rhs"> [parent @e:1731] (bla2.hlt:4:8-4:16) [@q:1733]
|| [debug/ast-final]                     - type::Bool <declaration="-" type="-" unified="bool" wildcard="false"> [parent @q:1733] (bla2.hlt:4:8-4:16) (resolved) [@t:1734]
|| [debug/ast-final]                 - QualifiedType <const="true" extern="false" side="rhs"> [parent @e:12ec] (bla2.hlt:4:8-4:16) [@q:12eb]
|| [debug/ast-final]                   - type::Bool <declaration="-" type="-" unified="bool" wildcard="false"> [parent @q:12eb] (bla2.hlt:4:8-4:16) (resolved) [@t:12ea]
|| [debug/ast-final]               - statement::Block [parent @s:12ee] (bla2.hlt:4:19-4:21) [@s:12ed]
|| [debug/ast-final]               - <empty>
```

What we seem to miss is constant folding in over coercions to `bool`. Just special-casing for values of `strong_ref` from `Null` might enable this use case. Let's either include this in this PR, or create a follow-up issue.

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -977,6 +977,271 @@ struct ConstantFoldingVisitor : OptimizerVisitor {
...
+            ConstantMap new_out = result.in[n];
+            transfer(n, new_out);
+
+            // If it changed, add successors to worklist
+            ConstantMap old_out = result.out[n];
+            if ( old_out != new_out ) {
+                result.out[n] = new_out;
+                for ( auto succ_id : result.cfg.graph().neighborsDownstream(n->identity()) ) {
+                    const auto* succ_node = result.cfg.graph().getNode(succ_id);
+                    if ( std::ranges::find(worklist, *succ_node) == worklist.end() )
```
bbannier, line 1139: WDYT about changing `worklist` to e.g., an `unordered_set` to reduce the impact of these repeated scans? With that the `push` would become an `insert`, and the `pop` could be a `auto n = worklist.extract(worklist.begin()).value();`.

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -977,6 +977,271 @@ struct ConstantFoldingVisitor : OptimizerVisitor {
...
+                if ( auto* name = assign->target()->tryAs<expression::Name>() ) {
+                    if ( auto* decl = name->resolvedDeclaration() ) {
+                        auto* const_val = evaluate(assign->source());
+                        constants[decl] = {.expr = const_val, .not_a_constant = const_val == nullptr};
+                    }
+                }
+            }
+
+            void operator()(declaration::LocalVariable* decl) override {
+                constants.erase(decl);
```
bbannier: Not needed.


```suggestion
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -977,6 +977,271 @@ struct ConstantFoldingVisitor : OptimizerVisitor {
...
+                if ( decl->init() ) {
+                    auto* const_val = evaluate(decl->init());
```
bbannier: Nit.


```suggestion
                if ( auto* init = decl->init() ) {
                    auto* const_val = evaluate(init);
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -977,6 +977,271 @@ struct ConstantFoldingVisitor : OptimizerVisitor {
...
+            TransferVisitor(ConstantMap& constants) : constants(constants), name_nac(constants) {}
+
+            // Tries to evaluate an expression to a constant value given a map of known constants.
+            Expression* evaluate(Expression* expr) {
+                if ( expr->isConstant() && expr->isA<expression::Ctor>() )
+                    return expr;
+
+                if ( auto* name = expr->tryAs<expression::Name>() ) {
+                    if ( auto* decl = name->resolvedDeclaration(); decl && constants.contains(decl) ) {
+                        auto val = constants.at(decl);
```
bbannier:
```suggestion
                        const auto& val = constants.at(decl);
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -977,6 +977,271 @@ struct ConstantFoldingVisitor : OptimizerVisitor {
...
 };
 
+struct ConstantPropagationVisitor : OptimizerVisitor {
+    using OptimizerVisitor::OptimizerVisitor;
+    using OptimizerVisitor::operator();
+
+    struct ConstantValue {
+        Expression* expr;
+        // NAC
+        bool not_a_constant = false;
```
bbannier:
```suggestion
        Expression* expr = nullptr;
        // NAC
        bool not_a_constant = false;
```

tests/Baseline/spicy.optimization.default-parser-functions/opt.hlt
evantypanski: I was considering making a follow-up issue to completely remove the current `ConstantFoldingVisitor`, then putting the "folding" logic into the new `ConstantPropagationVisitor`. So that would probably include this now-useful boost to its folding abilities.

I think the folding visitor also conflates constant folding with some dead code removal. I noticed that parts of this are a bit iffy when it comes to blocks, there are points where we generate block-scopes, then assign `result` within it, so constant-propagation would cause a compilation error, like this:

```
    <whatever> result;
    {
        auto used_by_function = 0;
        result = my_fn(used_by_function);
    }
    return result;
```

so `my_fn` is used within a scope, but it can't be constant propagated since that's outside of a scope. I think that's the case for the `! Null` checks here - we actually can't propagate those since it'd bring it outside of the necessary scope. I actually don't think this PR as-is will either, since it marks the function calls as NAC - we would need to tell if it's pure to propagate calls.

My suggestion would be:

1) Move folding logic into `ConstantPropagationVisitor` including `! Null` checks into a boolean (special case seems reasonable to me here)
2) Take the logic for removing unreachable `else` blocks and move that into a `ScopeHoistingVisitor` which will remove scopes if needed, including unreachable ones (which would be exactly `if (False)` or similar.

tl;dr: I'd prefer this go to a new followup issue that would fold the `! Null` checks in `ConstantPropagationVisitor`. Then add another visitor to do some scope hoisting in order to potentially get the benefits you mention.

hilti/toolchain/src/compiler/optimizer.cc
evantypanski: I'll look into this a little more, ideally they are sorted - it'll be significantly better to run in roughly topological order (with modifications for loops) so preserving order for the worklist would be best. But I'm not sure if it's overshadowed by this lookup. Not to say it currently matters much though

bbannier: Ah I see, this is why you wanted `CFG::nodes` to return a sorted list. Like you say, also unsure whether this matters currently as the CFGs we compute are not crazy big, so fine to punt.

evantypanski: I'll probably be able to do this in a followup. I added in some debug statements for the current state vs. a reverse topological order, the reverse topological order seems to take about 3/4 the iterations in the worklist algorithm so that seems like the way forward.

As a followup though, so the scope of this PR doesn't explode :)

tests/spicy/output/vector/reserve-known-elements.spicy
```diff
@@ -1,7 +1,7 @@
 # @TEST-DOC: Checks that a reserve call gets generated for vectors of known size
 #
 # @TEST-REQUIRES: which FileCheck
-# @TEST-EXEC: spicyc -p %INPUT | FileCheck %INPUT
+# @TEST-EXEC: spicyc -g -p %INPUT | FileCheck %INPUT
```
bbannier: `-g` disables basically every cleanup, let's maybe avoid doing that. We should be fine if we match on `# CHECK: {{[a-z_]+}}.reserve(128 <= 100 ? 128 : 100)` instead which seems localized enough.

tests/hilti/output/optimization/cfg-constprop.hlt
```diff
@@ -0,0 +1,87 @@
...
+    if (b) {
+        x = 42;
+        # Indirection through y since function calls don't get propagation
+        # CHECK: local uint<64> y = 42;
+        local uint<64> y = x;
+        hilti::print(y);
+        return;
+    }
+
+    # CHECK: local uint<64> y = 5;
```
bbannier, line 52: Very cool seeing control flow in action.

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -174,6 +174,9 @@ GraphNode CFG::_addBlock(GraphNode predecessor, const Nodes& stmts, const Node*
         else if ( auto* try_catch = c->tryAs<statement::Try>() )
             predecessor = _addTryCatch(predecessor, *try_catch);
 
+        else if ( auto* switch_ = c->tryAs<statement::Switch>() )
+            predecessor = _addSwitch(predecessor, *switch_);
+
```
bbannier, line 176: Nit: these aren't really ordered, at least not alphabetically. Let's either do that, or move this up to after e.g., `If` where we have more branching stuff.

bbannier (dismissed):

tests/spicy/output/vector/reserve-known-elements.spicy
evantypanski: in the future we'll probably also fold ternaries like that, but for now your suggestion seems fine too. I guess if/when we fold ternaries making that into `reserve(100)` isn't bad either.

tests/hilti/output/optimization/cfg-constprop.hlt
evantypanski: honestly it's like magic to me still

PR #2142, opened 2025-08-12
Rework how we store the field offsets inside a struct.

rsmmr:
Rework how we store the field offsets inside a struct.

So far Spicy was tracking unit field offsets by storing an additional
vector in each compiled struct that was recording begin/end position
of the fields. That was fragile because it depended on the order of
the structs' fields matching the order of unit fields exactly,
including which ones were emitted at all. That happened to work, but
was cumbersome to maintain that way. And with all the upcoming
optimization work, depending on field specifics in the final struct,
certainly isn't not a good ideas either.

This work changes the vector to a map indexed by field ID, making
maintenance and access of the information much easier and well
defined. While somewhat slower, offset tracking is already an opt-in
feature (`spicyc -Q`) that assumes performance isn't critical.

This also brings one new feature: In addition to the fields' offsets,
we now also record beginning and end offsets of the whole struct
itself. We store that in a map entry hardcoded to `self` (because
that's a reserved ID, it cannot conflict with a field name).

Tested through the existing (and updated) spicy-dump code/tests.

- **Prettify usage message.**
- **Rework how we store the field offsets inside a struct.**


spicy/toolchain/bin/spicy-dump/printer-json.cc
```diff
@@ -89,6 +89,16 @@ nlohmann::json JSONPrinter::convert(const hilti::rt::type_info::Value& v) {
...
         case TypeInfo::Struct: {
+            static auto make_json_offsets_object = [](const auto& tuple) {
+                auto o = json::object();
+                o["start"] = tuple::get<0>(tuple).Ref();
+
+                if ( const auto& end = tuple::get<1>(tuple) )
+                    o["end"] = end->Ref();
+
+                return o;
+            };
```
bbannier, line 100: Could you inline this again? The code is pretty small, and having a templated helper would cause instantiations for lots of different types.

spicy/toolchain/bin/spicy-dump/printer-json.cc
```diff
@@ -109,31 +119,31 @@ nlohmann::json JSONPrinter::convert(const hilti::rt::type_info::Value& v) {
...
-            auto offsets = json::object();
-            const auto& offsets_ = spicy::rt::get_offsets_for_unit(*struct_, v);
-            if ( _options.include_offsets && offsets_ ) {
-                auto fields = struct_->fields();
+            const auto* field_offsets = spicy::rt::get_offsets_for_unit(*struct_, v);
 
-                for ( const auto&& [index, offset] : enumerate(*offsets_) ) {
-                    auto o = json::object();
+            if ( _options.include_offsets && field_offsets ) {
+                auto json_map = json::object();
```
bbannier: Nit.

```suggestion
            if ( const auto* field_offsets = spicy::rt::get_offsets_for_unit(*struct_, v);
                 field_offsets && _options.include_offsets ) {
                auto json_map = json::object();
```

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -795,17 +795,9 @@ struct ProductionVisitor : public production::Visitor {
...
-            assert(field->index());
-            auto* index = builder()->addTmp("index", builder()->integer(*field->index()));
-            builder()->addMemberCall(__offsets, "resize", {builder()->sum(index, builder()->integer(1))});
-
-            builder()->addAssign(builder()->index(__offsets, builder()->integer(*field->index())),
-                                 builder()->tuple(
-                                     {cur_offset,
-                                      builder()->optional(builder()->qualifiedType(builder()->typeUnsignedInteger(64),
-                                                                                   hilti::Constness::Const))}));
+            // _offsets[field] = (self.offset(), null);
```
bbannier, line 798:
```suggestion
```

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -921,12 +913,14 @@ struct ProductionVisitor : public production::Visitor {
 
         if ( pb->options().getAuxOption<bool>("spicy.track_offsets", false) ) {
             assert(field->index());
+
+            // _offsets[field] = (offsets[field].begin(), self.offset());
```
bbannier, line 917:
```suggestion
```

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -2634,6 +2628,14 @@ void ParserBuilder::initializeUnit(const Location& l) {
 
     saveParsePosition();
 
+    if ( options().getAuxOption<bool>("spicy.track_offsets", false) ) {
+        // _offsets["self"] = (self.offset(), null);
```
bbannier, line 2632:
```suggestion
```

spicy/toolchain/src/compiler/codegen/parser-builder.cc
```diff
@@ -2644,6 +2646,16 @@ void ParserBuilder::finalizeUnit(bool success, const Location& l) {
 
     saveParsePosition();
 
+    if ( options().getAuxOption<bool>("spicy.track_offsets", false) ) {
+        // _offsets["self"] = (offsets["self"].begin(), self.offset());
```
bbannier, line 2650:
```suggestion
```

spicy/toolchain/bin/spicy-dump/printer-json.cc
rsmmr: I don't think this turns into many instances: the `auto` parameter always resolves to the same type, I just saved me the effort off typing it out because it's pretty complex. There's nothing else templated here either as far as I can see.

spicy/toolchain/src/compiler/codegen/parser-builder.cc
rsmmr: All these comments give the Spicy logic that the subsequent code generates, meant to make it easier to follow the logic.

PR #2151, opened 2025-08-19
Sort the CFG for constant propagation

evantypanski:
This is a followup from [this comment](https://github.com/zeek/spicy/pull/2138#discussion_r2282952585). Instead of using a set, it adds a sort to the CFG so we can traverse it the most effective way. For constant propagation, that's reverse postorder. Note that it doesn't store the result because we currently treat CFGs as pretty ephemeral, happy to change that if it seems useful

I took the mean number of iterations until convergence in `populate_dataflow` for three scenarios: unsorted, postorder, and reverse-postorder when compiling the Redis analyzer. Here are the results:

```
Mean for out-postorder: 105.69
Mean for out-reverse-postorder: 27.41
Mean for out-unsorted: 97.59
```

So reverse postorder wins pretty easily, since fewer iterations == better

I also touched up a couple parts. The change for equality with not-a-constant in `operator==` helped fix a possible infinite loop :)

### DIFF
--- a/hilti/toolchain/src/compiler/cfg.cc
+++ b/hilti/toolchain/src/compiler/cfg.cc
@@ -5,6 +5,7 @@
 #include <algorithm>
 #include <cassert>
 #include <cstdint>
+#include <functional>
 #include <iterator>
 #include <map>
 #include <optional>
@@ -68,6 +69,31 @@
 #include <hilti/hilti/ast/types/vector.h>
 
 namespace hilti::detail::cfg {
+
+std::deque<GraphNode> CFG::postorder() const {
+    std::deque<GraphNode> sorted;
+
+    std::unordered_set<NodeId> visited;
+
+    std::function<void(NodeId)> dfs_visit = [&](NodeId node_id) {
+        if ( visited.contains(node_id) )
+            return;
+
+        visited.insert(node_id);
+
+        for ( const auto& neighbor_id : g.neighborsDownstream(node_id) )
+            dfs_visit(neighbor_id);
+
+        const auto* node = g.getNode(node_id);
+        assert(node);
+        sorted.push_back(*node);
+    };
+
+    // This will sort all reachable nodes.
+    dfs_visit(_begin->identity());
+
+    return sorted;
+}
 
 // Helper function to check whether some `inner` node is a child of an `outer` node.
 static bool _contains(const Node& outer, const Node& inner) {
--- a/hilti/toolchain/src/compiler/cfg.cc
+++ b/hilti/toolchain/src/compiler/cfg.cc
@@ -606,7 +606,7 @@ struct DataflowVisitor : visitor::PreOrder {
                     transfer.gen[decl] = root;
                     break;
                 };
-            };
+            }
         }
 
         // Since we do not know whether the called function is pure always keep it.
--- a/hilti/toolchain/src/compiler/optimizer.cc
+++ b/hilti/toolchain/src/compiler/optimizer.cc
@@ -987,6 +987,10 @@ struct ConstantPropagationVisitor : OptimizerVisitor {
         bool not_a_constant = false;
 
         bool operator==(const ConstantValue& other) const {
+            // If both are NAC, what's in expr doesn't matter
+            if ( not_a_constant && other.not_a_constant )
+                return true;
+
             return expr == other.expr && not_a_constant == other.not_a_constant;
         }
     };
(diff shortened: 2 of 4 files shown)


hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -69,6 +70,38 @@
...
+        recursion_stack.erase(node_id);
+        visited.insert(node_id);
+
+        const auto* node = g.getNode(node_id);
+        assert(node);
+        sorted.push_back(*node);
+    };
+
+    // This will sort all reachable nodes.
+    dfs_visit(_begin->identity());
```
bbannier, line 93: This kind of visitation will miss nodes which are not reachable from the beginning of the block which is probably not acceptable in general, but might be okay for some use cases, e.g.,

```ruby
function bar() {
	return;
	print 1;
}
```

evantypanski: Yeah, that's intentional. I can add that into the doc comment? But I'm hesitant to change it, I don't think it's useful for most passes to consider nodes that aren't reachable from the beginning - unless you're specifically targeting those nodes. Also, there is no clear order for those, I guess shoving them at the end would be ok but in every case I'm considering that would just add useless work. Especially if unreachable code just gets removed later anyway

bbannier: Ah I see. I was looking for ways how we could fold `postorder` into `nodes`, but with that context having a separate function makes a lot of sense. Let's capture that only reachable nodes are included in the doc string.

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -69,6 +70,38 @@
...
+    std::unordered_set<NodeId> visited;
+    // For cycle detection
+    std::unordered_set<NodeId> recursion_stack;
+
+    std::function<void(NodeId)> dfs_visit = [&](NodeId node_id) {
+        if ( recursion_stack.contains(node_id) || visited.contains(node_id) )
+            // We either already added this or will add this.
+            return;
+
+        recursion_stack.insert(node_id);
```
bbannier: Isn't this the same as inserting into `visited` _before_ we climb down?

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1097,21 +1102,24 @@ struct ConstantPropagationVisitor : OptimizerVisitor {
         visitor::visit(tv, n.value());
     }
 
-    void populate_dataflow(AnalysisResult& result, const ConstantMap& init) {
-        std::vector<detail::cfg::GraphNode> worklist;
+    void populate_dataflow(AnalysisResult& result, const ConstantMap& init, const ID& function_name) {
+        auto postorder = result.cfg.postorder();
+        auto worklist = std::deque(postorder.begin(), postorder.end());
```
bbannier: Wondering whether we shouldn't have `postorder` return a `deque` in the first place. It wouldn't cost us much, and a `deque` also already has other nice properties as well (e.g., cheaper append).

hilti/toolchain/src/compiler/cfg.cc
evantypanski: seems like it, that was mainly there since I experimented with a few other algorithms too, good catch

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -69,6 +70,32 @@
...
 namespace hilti::detail::cfg {
 
+std::deque<GraphNode> CFG::postorder() const {
+    std::deque<GraphNode> sorted;
+
+    std::unordered_set<NodeId> visited;
+
+    std::function<void(NodeId)> dfs_visit = [&](NodeId node_id) {
+        if ( visited.contains(node_id) )
+            // We either already added this or will add this.
```
bbannier: Outdated. Since the code is simpler now the comment doesn't add much, so I'd probably drop it altogether. If we keep it lets update it.

```suggestion
            // We already added this.
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1097,21 +1101,24 @@ struct ConstantPropagationVisitor : OptimizerVisitor {
         visitor::visit(tv, n.value());
     }
 
-    void populate_dataflow(AnalysisResult& result, const ConstantMap& init) {
-        std::vector<detail::cfg::GraphNode> worklist;
+    void populate_dataflow(AnalysisResult& result, const ConstantMap& init, const ID& function_name) {
+        auto worklist = result.cfg.postorder();
+        assert(worklist.size() >= 1);
```
bbannier, line 1107: Wondering whether a comment here would help (e.g., "we always expect at least begin/end nodes"), but maybe just exiting before the first `pop_front` instead of the `assert` would be sufficiently self-explanatory.

bbannier (dismissed):
🚀

PR #2153, opened 2025-08-21
Fold `! Null` to true in constants

### LINKED ISSUE #2150: Extend constant folding to include `! Null`
This was part of #2138 with regards to `Null` getting propagated into `if (!Null)` in generated code:

I tried to understand why we emit this `Null` since as written in this HILTI source code it is invalid. It still compiles since in the AST we keep the coercions (from `Null` to the `strong_ref`, and from a `strong_ref` to a `bool`).
  Not folding branches over a constant `Null` like we do for `True` and `False` seems like it leaves potentially a lot of perf on the table since we in some cases do not remove writes to e.g., `__result` which we'd otherwise inline as well (and we already see a pretty big effect from the inlining done here).
 
This particular construct looks like this in the AST:

```
|| [debug/ast-final]             - statement::If [parent @s:12ef] (bla2.hlt:4:4-4:21) [@s:12ee]
|| [debug/ast-final]               - <empty>
|| [debug/ast-final]               - expression::LogicalNot [parent @s:12ee] (bla2.hlt:4:8-4:16)  (const)  (resolved) [@e:12ec]
|| [debug/ast-final]                 - expression::Coerced [parent @e:12ec] (bla2.hlt:4:9-4:16)  (const)  (resolved) [@e:1731]
|| [debug/ast-final]                   - expression::Ctor [parent @e:1731] (bla2.hlt:3:40-3:43)  (const)  (resolved) [@e:1a41]
|| [debug/ast-final]                     - ctor::Coerced [parent @e:1a41] (bla2.hlt:3:40-3:43) [@c:1a42]
|| [debug/ast-final]                       - ctor::Null [parent @c:1a42] (bla2.hlt:3:40-3:43) [@c:1a43]
|| [debug/ast-final]                         - QualifiedType <const="true" extern="false" side="rhs"> [parent @c:1a43] (bla2.hlt:3:40-3:43) [@q:1a44]
|| [debug/ast-final]                           - type::Null <declaration="-" type="-" unified="null" wildcard="false"> [parent @q:1a44] (bla2.hlt:3:40-3:43) (resolved) [@t:1a45]
|| [debug/ast-final]                       - ctor::StrongReference [parent @c:1a42] (bla2.hlt:3:40-3:43) [@c:1a46]
|| [debug/ast-final]                         - QualifiedType <const="true" extern="false" side="rhs"> [parent @c:1a46] [@q:1a47]
|| [debug/ast-final]                           - type::StrongReference <declaration="-" type="-" unified="strong_ref(stream)" wildcard="false"> [parent @q:1a47] (resolved) [@t:1a48]
|| [debug/ast-final]                             - QualifiedType <const="false" extern="false" side="rhs"> [parent @t:1a48] (bla2.hlt:3:21-3:26) [@q:1a49]
|| [debug/ast-final]                               - type::Stream <declaration="-" type="-" unified="stream" wildcard="false"> [parent @q:1a49] (bla2.hlt:3:21-3:26) (resolved) [@t:1a4a]
|| [debug/ast-final]                                 - QualifiedType <const="false" extern="false" side="rhs"> [parent @t:1a4a] (bla2.hlt:3:21-3:26) [@q:1a4b]
|| [debug/ast-final]                                   - type::stream::View <declaration="-" type="-" unified="view::stream" wildcard="false"> [parent @q:1a4b] (bla2.hlt:3:21-3:26) (resolved) [@t:1a4c]
|| [debug/ast-final]                                     - QualifiedType <const="false" extern="false" side="rhs"> [parent @t:1a4c] (bla2.hlt:3:21-3:26) [@q:1a4d]
|| [debug/ast-final]                                       - type::stream::Iterator <declaration="-" type="-" unified="iterator(stream)" wildcard="false"> [parent @q:1a4d] (bla2.hlt:3:21-3:26) (resolved) [@t:1a4e]
|| [debug/ast-final]                                         - QualifiedType <const="false" extern="false" side="rhs"> [parent @t:1a4e] (bla2.hlt:3:21-3:26) [@q:1a4f]
|| [debug/ast-final]                                           - type::UnsignedInteger <declaration="-" type="-" unified="uint8" width="8" wildcard="false"> [parent @q:1a4f] (bla2.hlt:3:21-3:26) (resolved) [@t:1a50]
|| [debug/ast-final]                   - QualifiedType <const="true" extern="false" side="rhs"> [parent @e:1731] (bla2.hlt:4:8-4:16) [@q:1733]
|| [debug/ast-final]                     - type::Bool <declaration="-" type="-" unified="bool" wildcard="false"> [parent @q:1733] (bla2.hlt:4:8-4:16) (resolved) [@t:1734]
|| [debug/ast-final]                 - QualifiedType <const="true" extern="false" side="rhs"> [parent @e:12ec] (bla2.hlt:4:8-4:16) [@q:12eb]
|| [debug/ast-final]                   - type::Bool <declaration="-" type="-" unified="bool" wildcard="false"> [parent @q:12eb] (bla2.hlt:4:8-4:16) (resolved) [@t:12ea]
|| [debug/ast-final]               - statement::Block [parent @s:12ee] (bla2.hlt:4:19-4:21) [@s:12ed]
|| [debug/ast-final]               - <empty>
```

What we seem to miss is constant folding in over coercions to `bool`. Just special-casing for values of `strong_ref` from `Null` might enable this use case. Let's either include this in this PR, or create a follow-up issue.

_Originally posted by @bbannier in https://github.com/zeek/spicy/pull/2138#discussion_r2282947421_
            

----------------------

See the original reply for more context, but my proposal for this would be to extend the constant propagation visitor to include "folding" logic (like `1+1` turns into `2`, or `! Null` terns into `True`). Then take parts of the current folding visitor and make it into a "scope hoisting" pass to reduce `if (True)` into just its inner block. This would enable more optimizations. But, there are lots of configurations to achieve the same thing.

evantypanski:
Closes #2150 

I need to think more about bringing the constant folding visitor into the propagation visitor - since the folding visitor also has the propagation of feature constants, putting it in the propagation visitor without the CFG enabled makes code fail to compile at C++. I am considering making optimizing visitors take whether or not CFG is enabled in order to determine whether they should run certain parts, but that's getting more complex so just letting this go for now.

The function calls still won't get constant-propagated for a couple reasons, we may be able to special case assigning a variable then immediately returning it (after flattening blocks a little more potentially).

The speedups here would be from combining with other optimizations (propagating `__result`, or extra places where #2031 applies)


rsmmr:
> just fold a coercion to a bool from `Null`-like things to `False`.

Hmm ... Folding the coercion sounds right, but I don't think the optimizer is the right place for that, as we have a separate constant folder that works on the AST already. I quickly tried to get this in there, and it seems to work with two small changes: one to the ctor coercer, and one to dead code optimizer. Here's the patch:

```
diff --git a/hilti/toolchain/src/compiler/coercer.cc b/hilti/toolchain/src/compiler/coercer.cc
index b6e5637fd..f53ad4064 100644
--- a/hilti/toolchain/src/compiler/coercer.cc
+++ b/hilti/toolchain/src/compiler/coercer.cc
@@ -84,6 +84,11 @@ struct VisitorCtor : visitor::PreOrder {
             result = builder->ctorWeakReference(t->dereferencedType());
             return;
         }
+
+        if ( dst->type()->isA<type::Bool>() ) {
+            result = builder->ctorBool(false);
+            return;
+        }
     }
 
     void operator()(ctor::List* n) final {
diff --git a/hilti/toolchain/src/compiler/optimizer.cc b/hilti/toolchain/src/compiler/optimizer.cc
index cb3f9a1e7..60cb3cbc1 100644
--- a/hilti/toolchain/src/compiler/optimizer.cc
+++ b/hilti/toolchain/src/compiler/optimizer.cc
@@ -836,9 +836,15 @@ struct ConstantFoldingVisitor : OptimizerVisitor {
     }
 
     std::optional<bool> tryAsBoolLiteral(Expression* x) {
-        if ( auto* expression = x->tryAs<expression::Ctor>() )
-            if ( auto* bool_ = expression->ctor()->tryAs<ctor::Bool>() )
+        if ( auto* expression = x->tryAs<expression::Ctor>() ) {
+            auto* ctor = expression->ctor();
+
+            if ( auto* x = ctor->tryAs<ctor::Coerced>() )
+                ctor = x->coercedCtor();
+
+            if ( auto* bool_ = ctor->tryAs<ctor::Bool>() )
                 return {bool_->value()};
+        }
 
         return {};
     }
```

I tried this with the following HILTI test and the two branches now get folded/optimized as expected:

```
module Test {

    if ( Null )
        assert False;

    if ( ! Null )
        assert True;

}
```

evantypanski:
> I think the final pre-optimizer AST should just end up with those booleans, coerced to from Null

Is that possible (in its current form)? Since the optimizer creates the `Null` values from constant propagation, they won't go back to constant folding to allow the optimizations I'm looking to get (namely eventually getting rid of the `__result` temporary there). Coercing the `Null` to `False` in the optimizer catches the cases that get constant-propagated too, doing it in the folder seems to only catch the naive cases that we don't generate

I'll draft this and look at doing a bit more with this PR, I was trying to split up some work into this smaller PR but it doesn't seem like that'll fully work out

evantypanski:
Okay, I think I got this in a better state - see my last comment for a bit about the approach.

It ends up being a little complex, but the benefit here is that the optimizer and coercion play more nicely (in my opinion)

hilti/toolchain/src/ast/ast-context.cc
```diff
@@ -783,12 +783,19 @@ Result<Nothing> ASTContext::_transform(Builder* builder, const Plugin& plugin) {
...
+            break;
 
-    // Optimization may have left some computed node state unset, such as a
-    // canonical IDs. Do a final resolver run to get that in shape.
-    if ( auto rc = _resolve(builder, plugin::registry().hiltiPlugin()); ! rc )
-        return rc;
+        first = false;
+
+        // Optimization may have left some computed node state unset, such as a
+        // canonical IDs. Do another resolver run to get that in shape.
```
rsmmr: The comment sounds a bit narrow now that we're relying on more work by the coercer?

tests/Baseline/hilti.ast.basic-module/debug.log
```diff
@@ -22,6 +22,7 @@
 [debug/compiler]     [HILTI] building scopes
 [debug/compiler]     [HILTI] resolving AST
 [debug/compiler] [HILTI] validating (post)
+[debug/compiler] [HILTI] validating (post)
```
rsmmr: Not totally critical but I think it would be worth making this log as (say) `[HILTI] validating (post optimizer)". Maybe turn the boolean to `_validate()` into a enum, or give it an additional string argument for passing the text logged inside the parentheses?

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -1221,8 +1227,13 @@ struct ConstantPropagationVisitor : OptimizerVisitor {
                 auto const_val = const_it->second;
 
                 if ( ! const_val.not_a_constant ) {
-                    recordChange(n, util::fmt("propagating constant value in %s", n->id()));
-                    replaceNode(n, node::deepcopy(context(), const_val.expr));
+                    Node* to_replace = n;
+                    // Replace the coercion, too, so that the coercer reruns.
+                    if ( auto* coerced = n->parent()->tryAs<expression::Coerced>() )
+                        to_replace = coerced;
```
rsmmr, line 1208: Ah, this is more elegant than I expected! :-)

hilti/toolchain/src/ast/ast-context.cc
```diff
@@ -781,6 +785,10 @@ Result<Nothing> ASTContext::_transform(Builder* builder, const Plugin& plugin) {
 }
 
 Result<Nothing> ASTContext::_optimize(Builder* builder) {
+    if ( logger().isEnabled(logging::debug::CfgInitial) ) {
+        hilti::detail::cfg::dump(logging::debug::CfgInitial, _root);
+    }
+
```
rsmmr: Matter of taste but I'd move the `isEnabled()` check into `dump()`. If you leave it here, which is fine, remove the  braces for coding style.

hilti/toolchain/src/ast/ast-context.cc
```diff
@@ -797,6 +805,10 @@ Result<Nothing> ASTContext::_optimize(Builder* builder) {
             return rc;
     }
 
+    if ( logger().isEnabled(logging::debug::CfgFinal) ) {
+        hilti::detail::cfg::dump(logging::debug::CfgFinal, _root);
+    }
+
```
rsmmr: Same.

hilti/toolchain/src/ast/ast-context.cc
evantypanski: I considered both and left it this way all else being roughly equal because I was influenced by [this post](https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html) a while ago. Not all of the points stand, but since I found both equal, I went with the one that followed that more closely

(the rule of thumb being "push ifs up")

tests/Baseline/hilti.ast.basic-module/debug.log
evantypanski: wow, I feel silly - there are already 2 post runs for validating, one being after the `optimize` call. Not sure how I missed it, I think I was just tunnel visioned on the `_optimize` function. I'll just drop the commit and forget about it...

rsmmr: Alright, I didn't realize either that nothing is happening between those two. :)

PR #2155, opened 2025-08-26
Fix CFG representation of `switch` statements

bbannier:


bbannier:
Closing this since this is part of #2159 already.

PR #2159, opened 2025-08-29
Fix aliasing and access computation

bbannier:


hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -804,6 +804,17 @@ struct DataflowVisitor : visitor::PreOrder {
...
+        // Switch statements are reflected in the CFG as local variables and
+        // different branches.
+        //
+        // TODO(bbannier): We currently model different switch cases as
+        // separate branches, but removing a case would remove the whole switch
+        // statement. Prevent that by explicitly requesting the the variable
+        // (which means also its switch statement) to be kept if we have any
+        // cases.
+        if ( auto* switch_ = x->parent()->tryAs<statement::Switch>(); switch_ && ! switch_->cases().empty() )
+            transfer.keep = true;
```
evantypanski, line 892: This is a little confusing to me; could we not remove the `switch` and just replace it with a scope if it's just the `default`? I know the extra variable but maybe I'm missing how that might change this case

though: we should generally be pretty careful with removing any control flow logic, so I don't see it as bad to keep it. We're already approximating it since we don't have basic blocks and terminators, so something might get lost - better not risk it much

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -800,6 +879,17 @@ struct DataflowVisitor : visitor::PreOrder {
         // the finalizer has no side effects.
         if ( auto* s = x->type()->type()->tryAs<type::Struct>(); s && s->field("~finally") )
             transfer.keep = true;
+
+        // Switch statements are reflected in the CFG as local variables and
+        // different branches.
+        //
+        // TODO(bbannier): We currently model different switch cases as
+        // separate branches, but removing a case would remove the whole switch
+        // statement. Prevent that by explicitly requesting the the variable
```
evantypanski: nit:


```suggestion
        // statement. Prevent that by explicitly requesting the variable
```

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -687,21 +694,22 @@ struct DataflowVisitor : visitor::PreOrder {
...
-                        in_assignment = true;
-                    }
                 }
                 node = node->parent();
             } while ( node && node != root.value() );
         }
 
-        if ( in_assignment )
-            ; // Nothing, handled above.
+        if ( stmt->isA<expression::Assign>() ) {
```
evantypanski: could we perhaps rename `stmt`? It's a little weird that there is a check to see if a variable named `stmt` is an expression. This was the case before, but this condition jumps out as weird to me now

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -166,6 +169,56 @@ GraphNode CFG::_addGlobals(GraphNode predecessor, const Node& root) {
...
+        return predecessor;
+
+    // Add parameters.
+    for ( auto* param : fn->ftype()->parameters() ) {
+        if ( ! param )
+            continue;
+
+        auto d = _getOrAddNode(param);
+        _addEdge(_begin, d);
+        _begin = d;
```
evantypanski: should this be changing `predecessor` instead to not assume `_begin`?

If not, should we completely remove the return value and `predecessor` parameter, since it just seems like it gets returned as-is always?

I think this causes `start` to branch in the CFG when there are params:

Image

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -711,25 +764,24 @@ struct DataflowVisitor : visitor::PreOrder {
             // Nothing, handled above.
         }
 
-        else if ( auto* d = stmt->tryAs<statement::Declaration>() ) {
+        else if ( stmt->isA<statement::Declaration>() )
```
evantypanski: More of a question: Why are decl statements handled differently than decls? Put another way: why is this `else if` case necessary, rather than being the same as the `stmt->tryAs<Declaration>()` case?

hilti/toolchain/src/compiler/cfg.cc
bbannier: I initially tried that, but it seemed tricky and also slightly messy since we'd need to introduce both a variable to the scope as well as keep a block scope for the default case. This might be possible, but I just didn't want to delay this further by digging more.

That said, the more I work with it the more I am convinced that `keep` is just a marker for tech debt which we might need to tackle at some point down the line anyway.

hilti/toolchain/src/compiler/cfg.cc
bbannier: I renamed it in an additional patch.

hilti/toolchain/src/compiler/cfg.cc
bbannier: `hilti::Declaration` is just a `hilti::Node` while `hilti::statement::Declaration` is a `hilti::Statement`. The way to thinkn about this is `hilti::Declaration` is for generic declarations, e.g., identifiers in a module while `hilit::statement::Declaration` is for statements introducing declarations like e.g., locals.

hilti/toolchain/src/compiler/cfg.cc
bbannier: Ups, definitely, the function already returns `predecessor`. I fixed the whole function to use `predecessor` in place of `_begin` previously.

PR #2161, opened 2025-09-04
Enable control-flow based optimizations by default.

bbannier:
We initially landed optimizations dependent on control and data flow behind a feature flag since the implementation had known issues. These have now been fixed, so we enable these optimizations by default. The feature flag stays in place and can be used to disable them now.


hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -2453,9 +2453,10 @@ bool detail::optimizer::optimize(Builder* builder, ASTRoot* root, bool first) {
           2}},
     };
 
-    // TODO(bbannier): Control-flow based optimizations are not ready for
-    // prime-time yet and behind a feature guard.
-    bool has_cfg = rt::getenv("HILTI_OPTIMIZER_ENABLE_CFG") == "1";
+    // Control-flow based optimizations are behind
+    // a feature guard which defaults to ON.
+    auto flag = rt::getenv("HILTI_OPTIMIZER_ENABLE_CFG");
+    bool has_cfg = ! flag || flag == "1";
```
rsmmr:
```suggestion
    bool has_cfg = (! flag || flag == "1");
```

rsmmr (approved):

PR #2172, opened 2025-09-18
Fix misnamed variable [skip CI].

evantypanski:
From #2171

I accidentally didn't swap a condition when I changed the default to be including dataflow so the variable says to omit... oops. At least it doesn't change the logic

### DIFF
--- a/hilti/toolchain/src/compiler/cfg.cc
+++ b/hilti/toolchain/src/compiler/cfg.cc
@@ -545,7 +545,7 @@ std::string CFG::dot(bool omit_dataflow) const {
         node_ids.insert({n->identity(), id});
 
         std::optional<std::string> xlabel;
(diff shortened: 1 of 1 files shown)


hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -1062,7 +1062,7 @@ void CFG::_populateDataflow() {
 static std::string dataflowDot(const hilti::Statement& stmt) {
     auto cfg = hilti::detail::cfg::CFG(&stmt);
     auto omit_dataflow = rt::getenv("HILTI_OPTIMIZER_OMIT_CFG_DATAFLOW") == "1";
-    return cfg.dot(! omit_dataflow);
+    return cfg.dot(omit_dataflow);
```
bbannier, line 1065: This looks like a strong signal that we should add a very basic test.

bbannier (approved):

PR #2173, opened 2025-09-22
Add function return propagation

### LINKED ISSUE #2031: Propagate constants through function returns

evantypanski:
Closes #2031

This isn't quite ready, but it's roughly feature-complete. I had to restrict this a *lot* in a lot of weird ways, I want to try to clean some of that.

But, first I'm going to do some of the other changes that are necessary to make this actually change generated parsers, I won't sink time into it before seeing that. So this will stay a draft until I see the value from that.

Those steps are:

1) Unnecessary scope blocks need to be removed (because `{...}` blocks are all through the code and prevent propagation)
2) Need something which will propagate returns of the form `__result = func_call(...); return __result;` - right now function calls won't propagate.
3) Need to register "passthroughs" so this optimization will apply if there is a function call return in an intermediary function

Those are a decent amount of work themselves so it might be a while; I will probably stack PRs. I don't really want to merge this extra complexity if we aren't getting any speedup yet

### DIFF
--- a/tests/hilti/output/optimization/propagate-returns.hlt
+++ b/tests/hilti/output/optimization/propagate-returns.hlt
@@ -53,6 +53,10 @@ type S = struct {
     method tuple<uint<64>, uint<64>, uint<64>, uint<64>> passthrough1(uint<64> p1, uint<64> p2, uint<64> p3, uint<64> p4);
     method tuple<uint<64>, uint<64>, uint<64>, uint<64>> passthrough2(uint<64> p1, uint<64> p2, uint<64> p3, uint<64> p4);
 
+    # This is basically an end-to-end test for parse functions
+    method tuple<uint<64>, uint<64>, uint<64>, uint<64>> parse_stage2(uint<64> not_returned, uint<64> p1, uint<64> p2, uint<64> p3, uint<64> p4);
+    method tuple<uint<64>, uint<64>, uint<64>, uint<64>> parse_stage1(uint<64> p1, uint<64> p2, uint<64> p3, uint<64> p4);
+
 };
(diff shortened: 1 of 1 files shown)


evantypanski:
I kind of hacked together the support for methods, "passthroughs," and a peephole analyzer change that should work on parsers. What's holding it back now is that there is a block scope - which I believe @bbannier has some ideas for tearing down block scopes.

My guess is that this has some bugs that will become apparent when it actually applies to parsers, but theoretically it should provide a nice little speedup when scopes are flattened a bit. Until then I'll keep it in this rather horrific state of hackiness just so we see a speedup before introducing all of this complexity.

PR #2177, opened 2025-10-01
Fix C++ code generation for struct parameters passed as references.

bbannier:
Previously we could generate C++ moves for constructor arguments passed as references. This is semantically incorrect, but seems to have had no ill effects on code we generated. Since it could still confuse readers and static analysis tools fix this to only generate moves if the argument was passed as `copy`, but write out copies for all other argument passing. This more clearly reflects what was already happening.

### DIFF
--- a/hilti/toolchain/include/compiler/detail/cxx/elements.h
+++ b/hilti/toolchain/include/compiler/detail/cxx/elements.h
@@ -290,6 +290,8 @@ struct Argument : public DeclarationBase {
     }
 
     bool operator!=(const Argument& other) const { return ! operator==(other); }
+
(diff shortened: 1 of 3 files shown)


evantypanski (approved):
one day we could probably solidify when `move` is generated a bit more. Feels ad-hoc as is

bbannier:
> Feels ad-hoc as is

Absolutely agree, to me the string check feels extra dirty. The issue is that elements used during C++ codegen are very thin wrapper, e.g., `Argument` already gets constructed from `string` values which it just accumulates. I considered also passing a bool to capture whether an arg was passed by ref, but I think that would introduce brittleness.

PR #2179, opened 2025-10-08
Fix a couple issues with explicit null casts

### LINKED ISSUE #2154: Coercion from `Null` gets unhandled internal error
This Spicy code gives a big internal error when running with `spicyc -j test.spicy`:

```
module Test;

cast<bool>(Null);
```

```
$ spicyc -j test.spicy
[internal-error] codegen: type type::Null unhandled for coercion

--- Aborting
# 0   libhilti.dylib                      0x0000000103f0be5c _ZN5hilti2rt9BacktraceC1Ev + 140
...
# 23  dyld                                0x000000019c86f154 start + 2476
[1]    58113 abort      spicyc -j test.spicy
```

evantypanski:
Seems like only null casts had the internal error issue with unhandled coercions here, not sure if there's a better way for this, but this is the best I found

Fixes #2154


hilti/toolchain/src/compiler/codegen/coercions.cc
```diff
@@ -307,8 +316,11 @@ cxx::Expression CodeGen::coerce(const cxx::Expression& e, QualifiedType* src, Qu
         // If only difference is constness, nothing to do.
         return e;
 
-    if ( auto* t = dst->type()->tryAs<type::Optional>(); t && ! src->type()->isA<type::Optional>() )
-        return fmt("%s(%s)", compile(dst, codegen::TypeUsage::Storage), e);
+    if ( auto* t = dst->type()->tryAs<type::Optional>(); t && ! src->type()->isA<type::Optional>() ) {
+        // Null should not include the expression
```
bbannier: `rt::Null` has a user-defined conversion to `std::optional`, so I was hoping avoiding the special-casing would work, but it does not.

```
struct Null {
    template<typename T>
    operator std::optional<T>() {
        return std::nullopt;
    }
};

void fn() {
    auto x1 = static_cast<std::optional<int>>(Null{});
    auto x2 = static_cast<std::optional<SafeInt<int>>>(Null{});
}
```

Here `x2` is rejected since `SafeInt` explicitly requires an integer type which `Null` is not. I think this is due to user-defined conversions only running after standard conversions.

All this to say, maybe capture this in a comment here, e.g.,




```suggestion
        // Special case `Null`. Since standard conversions run before the
        // user-defined conversion operation from `Null` to `std::optional` we
        // might end up using a standard conversion which might be invalid.
```

tests/spicy/types/null/coercion.spicy
```diff
@@ -0,0 +1,12 @@
...
+# @TEST-DOC: Ensures explicit null casts compile; regression test for #2154
+
+module Foo;
+
+type X = struct {};
+
+print cast<bool>(Null);
+print cast<X&>(Null);
+print cast<optional<int32>>(Null);
+print cast<optional<X>>(Null);
```
bbannier, line 13: Could you either baseline the result, or alternatively use `assert` to check that this gets us what we expect?

hilti/toolchain/src/compiler/codegen/coercions.cc
evantypanski: when templates are involved sometimes C++ resolution just becomes black magic

PR #2183, opened 2025-10-21
Reimplement two runtime types

rsmmr:
This re-implements the runtime representation for tuples and
optionals, which so far still using `std::tuple` and `std::optional`
directly. While behind the scenes, we keep doing so, the API now
becomes custom `rt::Tuple` and `rt::Optional` classes. They allows us
to tailor API & implementation to HILTI's needs, and can facilitate
future extensions and optimizations. This is similar to what we do for
most other runtime types already. 

The new implementations aren't exactly concise, but most of that is pretty much boiler-plate to provide similar API to the as the standard types, which C++ doesn't always make easy. (Note that the new APIs don't accept the standard types in place of the custom ones.)

- **No longer use `std::*` for working with HILTI tuples.**
- **New tuple implementation.**
- **No longer use `std::*` for working with HILTI optionals.**
- **New `Optional` implementation.**


hilti/runtime/include/result.h
```diff
@@ -168,7 +168,7 @@ class Result {
     explicit operator bool() const { return hasValue(); }
 
     /** Converts the result to an optional that's set if it represents a successful return value. */
-    operator std::optional<T>() const { return hasValue() ? std::make_optional(value()) : std::nullopt; }
+    // operator Optional<T>() const { return hasValue() ? hilti::rt::optional::make(value()) : std::nullopt; } // TODO
```
bbannier: This seems in general useful and I am surprised it was unused. Let's either keep it and add a small tests, or drop it.

hilti/runtime/include/type-info.h
```diff
@@ -415,17 +415,19 @@ class Bits {
...
     Bits(const char* name, unsigned int lower, unsigned int upper, const TypeInfo* type, std::ptrdiff_t offset)
-        : name(name), lower(lower), upper(upper), type(type), offset(offset) {}
+        : name(name), lower(lower), upper(upper), type(type), _offset(offset) {}
 
     const std::string name;   /**< ID of the field, with an empty string indicating no name */
     const unsigned int lower; /**< lower bit of the field */
     const unsigned int upper; /**< upper bit of the field */
     const TypeInfo* type;     /**< type of the field */
 
+    auto offset() const { return _offset; } // TODO: Remove
```
bbannier: Looks like this could go now.

hilti/runtime/src/tests/integer.cc
```diff
@@ -195,26 +195,29 @@ TEST_CASE("unpack") {
     using Result32 = Result<Tuple<integer::safe<uint32_t>, Bytes>>;
     using Result64 = Result<Tuple<integer::safe<uint64_t>, Bytes>>;
 
+    // TODO: Can we make this nicer, ideally without the lambdas?
+    static auto make_result16 = [](integer::safe<uint16_t> i, Bytes b) -> Result16 { return tuple::make(i, b); };
+    static auto make_result32 = [](integer::safe<uint32_t> i, Bytes b) -> Result32 { return tuple::make(i, b); };
+    static auto make_result64 = [](integer::safe<uint64_t> i, Bytes b) -> Result64 { return tuple::make(i, b); };
+
```
bbannier, line 203: The noisy bit is the conversion from an `int` to the `integer::safe`, but with deductions it would look like this without a helper:

```cpp
    CHECK_EQ(integer::unpack<uint16_t>("\x01\x00"_b, ByteOrder::Little),
             Result(tuple::make(integer::safe<uint16_t>(1), ""_b)));
```

Introducing an extra helper makes this only slightly shorter:

```cpp
template<typename T>
auto make_result(T i, Bytes b) {
    return Result{tuple::make(integer::safe<T>(i), b)};
}

...

    CHECK_EQ(integer::unpack<uint16_t>("\x01\x00"_b, ByteOrder::Little), make_result<uint16_t>(1, ""_b));
```

hilti/runtime/src/tests/integer.cc
```diff
@@ -195,26 +195,29 @@ TEST_CASE("unpack") {
     using Result32 = Result<Tuple<integer::safe<uint32_t>, Bytes>>;
     using Result64 = Result<Tuple<integer::safe<uint64_t>, Bytes>>;
```
bbannier, line 197: Both of these are actually unused.

hilti/runtime/include/types/bytes.h
```diff
@@ -487,7 +487,7 @@ class Bytes : protected std::string {
      */
     Tuple<Bytes, Bytes> split1() const {
         auto p = hilti::rt::split1(str());
-        return {p.first, p.second};
+        return tuple::make<Bytes, Bytes>(p.first, p.second);
```
bbannier: Ups, we should have moved here in the first place. I'd also opt for dropping the explicit types for maintainability.

```suggestion
        return tuple::make(std::move(p.first), std::move(p.second));
```

hilti/runtime/include/types/bytes.h
```diff
@@ -507,7 +507,7 @@ class Bytes : protected std::string {
      */
     Tuple<Bytes, Bytes> split1(const Bytes& sep) const {
         auto p = hilti::rt::split1(str(), sep);
-        return {p.first, p.second};
+        return tuple::make<Bytes, Bytes>(p.first, p.second);
```
bbannier:
```suggestion
        return tuple::make(std::move(p.first), std::move(p.second));
```

hilti/runtime/src/types/string.cc
```diff
@@ -127,12 +127,12 @@ Vector<std::string> string::split(std::string_view s, std::string_view sep) {
 
 Tuple<std::string, std::string> string::split1(const std::string& s) {
     auto pair = hilti::rt::split1(s);
-    return {pair.first, pair.second};
+    return tuple::make(pair.first, pair.second);
```
bbannier:
```suggestion
    return tuple::make(std::move(pair.first), std::move(pair.second));
```

hilti/runtime/src/types/string.cc
```diff
@@ -127,12 +127,12 @@ Vector<std::string> string::split(std::string_view s, std::string_view sep) {
...
 }
 
 Tuple<std::string, std::string> string::split1(const std::string& s, const std::string& sep) {
     auto pair = hilti::rt::split1(s, sep);
-    return {pair.first, pair.second};
+    return tuple::make(pair.first, pair.second);
```
bbannier:
```suggestion
    return tuple::make(std::move(pair.first), std::move(pair.second));
```

hilti/runtime/include/types/optional.h
```diff
@@ -2,93 +2,383 @@
...
+    constexpr Optional(std::in_place_t, auto&&... args)
+        : std::optional<T>(std::in_place, std::forward<decltype(args)>(args)...) {}
+
+    /**
+     * Constructor from a value of different type U that's convertible to T.
+     */
+    template<typename U>
+    constexpr Optional(U&& v)
+        requires(std::is_constructible_v<T, U> && ! std::is_same_v<std::decay_t<U>, Optional> &&
+                 ! std::is_same_v<std::decay_t<U>, T>)
```
bbannier: Can we use the same `requires` clause from below? It seems to be the one we use in most places here.

hilti/runtime/include/types/optional.h
```diff
@@ -2,93 +2,383 @@
...
+}
+
+/**
+ * Constructs an optional initialized to a value constructed from given
+ * arguments. This is similar to ``std::make_optional``.
+ */
+template<class T, class... Args>
+constexpr Optional<T> make(Args&&... args) {
+    return Optional<T>(std::in_place, std::forward<Args>(args)...);
 }
```
bbannier, line 351: In contrast to the `Tuple` functions these don't add anything over constructors, and deduction seems to work the same, e.g., this works:

```cpp
auto x = Optional(1); // Optional<int>
```

Can we omit these?

hilti/runtime/include/types/null.h
```diff
@@ -4,21 +4,23 @@
...
  * Represents HILTI's "null" type.
  */
 struct Null {
     template<typename T>
-    operator std::optional<T>() {
-        return std::nullopt;
+    operator hilti::rt::Optional<T>() {
+        return {};
     }
 };
```
bbannier: `Optional` can already construct from a `Null` and this is not needed anymore.

```suggestion
struct Null {};
```

hilti/runtime/include/types/reference.h
```diff
@@ -238,43 +238,70 @@ class ValueReference {
      * @throws NullReference if the instance does not currently refer to a valid value
```
bbannier: This shouldn't throw `NullReference` anymore.

Could we add a test for assignments from empty references?

hilti/runtime/include/types/reference.h
```diff
@@ -238,43 +238,70 @@ class ValueReference {
...
+        // assignment operator.
+        try {
+            if ( auto* ptr = _get() )
+                *ptr = *other._get();
+            else
+                _ptr = std::make_shared<T>(*other._get());
+
+            return *this;
+        } catch ( ... ) {
+            cannot_be_reached();
```
bbannier, line 270: This seems not ideal, do you recall why we don't just rethrow?

hilti/runtime/include/types/reference.h
```diff
@@ -238,43 +238,70 @@ class ValueReference {
...
-                cannot_be_reached();
             }
-        }
+            else
+                _ptr = std::make_shared<T>(*other._get());
 
-        return *this;
+            return *this;
+        } catch ( ... ) {
+            cannot_be_reached();
```
bbannier, line 301: Same here.

hilti/runtime/include/types/tuple.h
```diff
@@ -2,107 +2,285 @@
...
+    constexpr bool hasValue(size_t idx) const { return idx < MaxElements ? (_mask & (1ULL << idx)) != 0 : false; }
+
+    friend auto operator<=>(const TupleBase& t1, const TupleBase& t2) = default;
+
+protected:
+    /** Maximum number of tuple elements supported. */
+    static constexpr size_t MaxElements = 64;
+
+    /** Constructor marking all elements as initially unset. */
+    TupleBase() = default;
```
bbannier:
```suggestion
    TupleBase() noexcept = default;
```

Probably in most other places below as well.

hilti/runtime/include/types/tuple.h
```diff
@@ -2,107 +2,285 @@
...
+     *
+     * @tparam Idx index of the element to return
+     */
+    template<std::size_t Idx>
+    static ptrdiff_t elementOffset() {
+        Tuple t;
+        // This is pretty certainly not well-defined, but seems to work for us ...
+        // NOLINTNEXTLINE(clang-analyzer-security.PointerSub)
+        return (reinterpret_cast<const char*>(&std::get<Idx>(t)) - reinterpret_cast<const char*>(&t));
+    }
```
bbannier, line 205: We don't need to spend time at runtime here.


```suggestion
    static ptrdiff_t elementOffset() {
        static Tuple t;
        // This is pretty certainly not well-defined, but seems to work for us ...
        // NOLINTNEXTLINE(clang-analyzer-security.PointerSub)
        static auto offset = reinterpret_cast<const char*>(&std::get<Idx>(t)) - reinterpret_cast<const char*>(&t);

        return offset;
    }
```

hilti/runtime/include/types/tuple.h
```diff
@@ -2,107 +2,285 @@
...
+ * For efficiency, we currently limit the maximum number of tuple elements to
+ * 64. This could be changed if deemed necessary.
  */
+class TupleBase {
+public:
+    /**
+     * Returns true if the element at index `idx` is set. If `idx` is
+     * beyond the number of valid tuple elements the result is undefined.
+     */
+    constexpr bool hasValue(size_t idx) const { return idx < MaxElements ? (_mask & (1ULL << idx)) != 0 : false; }
```
bbannier:
```suggestion
    constexpr bool hasValue(size_t idx) const noexcept {
        return idx < MaxElements ? (_mask & (1ULL << idx)) != 0 : false;
    }
```

hilti/runtime/include/types/tuple.h
```diff
@@ -174,40 +324,19 @@ ptrdiff_t elementOffset() {
...
- * unset if the expression threw `AttributeNotSet` (all other exception are
+ * unset if the expression threw `AttributeNotSet` (all other exceptions are
  * passed through)
  */
 template<typename Func>
 auto wrap_expression(Func&& f) {
     using element_t = std::invoke_result_t<Func>;
     try {
-        return std::optional<element_t>(f());
+        return Optional<element_t>(f());
```
bbannier, line 336: Nit: There are many places here where we spell out template arguments even though we don't need to.

```suggestion
        return Optional(f());
```

hilti/runtime/src/types/regexp.cc
```diff
@@ -112,10 +112,10 @@ Tuple<integer::safe<int32_t>, stream::View> regexp::MatchState::advance(const st
 
     if ( rc >= 0 ) {
         _pimpl->_done = true;
-        return std::make_tuple<integer::safe<int32_t>, stream::View>(rc, std::move(ndata));
+        return tuple::make<integer::safe<int32_t>, stream::View>(rc, std::move(ndata));
```
bbannier: This could be

```suggestion
    return tuple::make(integer::safe{rc}, std::move(ndata));
```

hilti/toolchain/src/compiler/codegen/codegen.cc
```diff
@@ -61,7 +61,7 @@ struct GlobalsVisitor : hilti::visitor::PostOrder {
         // the values, in particular wrt destruction when the runtime shuts
         // down.
         for ( auto g : globals ) {
-            g.type = fmt("std::optional<%s>", g.type);
+            g.type = fmt("hilti::rt::Optional<%s>", g.type);
```
bbannier: We usually use a full path.


```suggestion
            g.type = fmt("::hilti::rt::Optional<%s>", g.type);
```

More instance below.

hilti/toolchain/src/compiler/codegen/coercions.cc
```diff
@@ -97,12 +97,13 @@ struct Visitor : public hilti::visitor::PreOrder {
         if ( dst->type()->isA<type::Optional>() ) {
             // Create tmp to avoid evaluation "expr" twice.
             auto tmp = cg->addTmp("opt", cg->compile(src, codegen::TypeUsage::Storage));
-            result = {fmt("(%s = (%s), %s.has_value() ? std::make_optional(*%s) : std::nullopt)", tmp, expr, tmp, tmp),
+            result = {fmt("(%s = (%s), %s.hasValue() ? hilti::rt::optional::make(*%s) : hilti::rt::Null())", tmp, expr,
```
bbannier, line 100:
```suggestion
                fmt("[&](const auto& __t) { return ::hilti::rt::tuple::make(%s); }(%s)", util::join(exprs, ", "), expr);
```

Also below.

hilti/toolchain/src/compiler/codegen/ctors.cc
```diff
@@ -67,9 +67,9 @@ struct Visitor : hilti::visitor::PreOrder {
             types.emplace_back(itype);
 
             if ( auto* x = n->bits(b->id()) )
-                values.emplace_back(fmt("std::make_optional(%s)", cg->compile(x->expression())));
+                values.emplace_back(fmt("hilti::rt::optional::make(%s)", cg->compile(x->expression())));
```
bbannier:
```suggestion
                values.emplace_back(fmt("::hilti::rt::optional::make(%s)", cg->compile(x->expression())));
```

Also below.

hilti/toolchain/src/compiler/codegen/types.cc
```diff
@@ -207,7 +207,7 @@ struct VisitorDeclaration : hilti::visitor::PreOrder {
                     auto t = cg->compile(f->type(), codegen::TypeUsage::Storage);
 
                     if ( f->isOptional() )
-                        t = fmt("std::optional<%s>", t);
+                        t = fmt("hilti::rt::Optional<%s>", t);
```
bbannier:
```suggestion
                        t = fmt("::hilti::rt::Optional<%s>", t);
```

Also below.

hilti/toolchain/src/compiler/codegen/types.cc
```diff
@@ -804,18 +811,26 @@ struct VisitorTypeInfoDynamic : hilti::visitor::PreOrder {
     std::optional<cxx::Expression> result;
 
     void operator()(type::Bitfield* n) final {
+        // TODO: Add tupleType() to type::Bietfield()?
```
bbannier: Seems fine to me as is.

hilti/toolchain/src/compiler/cxx/elements.cc
```diff
@@ -460,12 +460,13 @@ std::string cxx::type::Struct::str() const {
...
-                                                               [&](const auto& x) {
-                                                                   auto& l = std::get<declaration::Local>(x);
-                                                                   return fmt("std::optional<%s> %s", l.type, l.id);
-                                                               }),
-                                               ", ");
+            auto locals_ctor_args =
+                util::join(util::transform(locals_user,
+                                           [&](const auto& x) {
+                                               auto& l = std::get<declaration::Local>(x);
+                                               return fmt("hilti::rt::Optional<%s> %s", l.type, l.id);
```
bbannier:
```suggestion
                                               return fmt("::hilti::rt::Optional<%s> %s", l.type, l.id);
```

Also below.

hilti/toolchain/tests/util.cc
```diff
@@ -61,7 +61,7 @@ TEST_CASE("cacheDirectory") {
         SUBCASE("env override") { ::setenv("SPICY_CACHE", "OVERRIDE", 1); }
         SUBCASE("no env override") { ::unsetenv("SPICY_CACHE"); }
 
-        CHECK_EQ(hilti::util::cacheDirectory(configuration), hilti::rt::filesystem::path("BUILD") / "cache" / "spicy");
+        CHECK_EQ(*hilti::util::cacheDirectory(configuration), hilti::rt::filesystem::path("BUILD") / "cache" / "spicy");
```
bbannier: I'd prefer to keep this is as is. If we want to deref the value lets prefix this with a `REQUIRE` check that it is set.

Also below.

hilti/runtime/include/util.h
```diff
@@ -114,7 +114,7 @@ struct ResourceUsage {
 ResourceUsage resource_usage();
 
 /** Returns the value of an environment variable, if set. */
-extern std::optional<std::string> getenv(const std::string& name);
+extern hilti::rt::Optional<std::string> getenv(const std::string& name);
```
bbannier: Nit: We are already inside `hilti::rt` so this could be just


```suggestion
extern Optional<std::string> getenv(const std::string& name);
```

There are many instances of this in this PR.

hilti/runtime/src/type-info.cc
```diff
@@ -51,3 +51,18 @@ const TypeInfo type_info::uint32{std::nullopt, "uint32", TO_STRING(hilti::rt::in
 const TypeInfo type_info::uint64{std::nullopt, "uint64", TO_STRING(hilti::rt::integer::safe<uint64_t>),
                                  new type_info::UnsignedInteger<uint64_t>()};
 const TypeInfo type_info::void_{std::nullopt, "void", TO_STRING_CONST("void"), new type_info::Void()};
+
+
+std::vector<std::pair<const type_info::bitfield::Bits&, type_info::Value>> type_info::Bitfield::iterate(
+    const type_info::Value& v) const {
+    auto elements = _tuple_ti->tuple->iterate(v);
+
+    std::vector<std::pair<const bitfield::Bits&, Value>> values;
```
bbannier, line 60: Nit.


```suggestion
    std::vector<std::pair<const bitfield::Bits&, Value>> values;
    values.reserve(std::min(bits().size(), elements.size()));    
```

hilti/runtime/src/util.cc
```diff
@@ -68,9 +68,9 @@ hilti::rt::ResourceUsage hilti::rt::resource_usage() {
     return stats;
 }
 
-std::optional<std::string> hilti::rt::getenv(const std::string& name) {
+hilti::rt::Optional<std::string> hilti::rt::getenv(const std::string& name) {
     if ( auto* x = ::getenv(name.c_str()) )
-        return {x};
+        return std::string(x);
```
bbannier: This was already inconsistent before, let's fix it while we touch it.

```suggestion
        return {std::string(x)};
```

Also fine should be

```suggestion
        return {{x}};
```

hilti/runtime/include/types/optional.h
```diff
@@ -2,93 +2,383 @@
 
 #pragma once
 
+#include <compare>
```
bbannier: My clang-tidy thinks this is unused.

spicy/runtime/include/util.h
bbannier, line 1: This now has unused includes `<optional>` and `<tuple>`. There are many instances of this elsewhere, typically with `<optional>`.

hilti/runtime/include/types/bitfield.h
```diff
@@ -72,23 +76,29 @@ inline std::string render(const Bitfield<Ts...>& x, const hilti::rt::TypeInfo* t
         if ( ! out.empty() )
             out += ", ";
 
+        std::string s;
+        if ( v )
+            s = v.to_string();
+        else
+            s = "(not set)";
```
bbannier: Does not really matter here, but allocating twice isn't ideal.

```suggestion
        auto s = v ? v.to_string() : "(not set)";
```

hilti/runtime/include/types/bitfield.h
```diff
@@ -108,8 +118,8 @@ inline std::string render(const std::optional<Bitfield<Ts...>>& x, const hilti::
     return out;
 }
 
-} // namespace detail
-} // namespace bitfield
+} // namespace bitfield::detail
+
```
bbannier, line 117:
```suggestion
```

bbannier (dismissed):
This looks good to me, and seems just a cleanup; I looked for places where we now use more expensive `Tuple` instead of `tuple` without need, but didn't notice anything.

I left a bunch of comments, but most are smallish cleanups. I hope this isn't too discouraging.

hilti/runtime/src/tests/integer.cc
rsmmr: Just a test, leaving as it is.

hilti/runtime/src/tests/integer.cc
rsmmr: They are used just below.

hilti/runtime/include/types/tuple.h
rsmmr: But then we end up with all these statics instantiated for every tuple type. `elementOffset()` is only used at startup for creating type info, I think that's fine; plus, maybe the compiler optimizes out the needed work anyways.

hilti/runtime/include/types/reference.h
rsmmr: It's to support `noexcpt` on the assignment operators. I don't quite recall *why* we need that, but it was required somewhere (this is preexisting).

hilti/toolchain/src/compiler/codegen/coercions.cc
rsmmr: I don't like having to create lambda functions here, the original (and pre-existing) code seems fine to me.

hilti/runtime/include/types/optional.h
rsmmr: This does add the `std::decay_t` over the constructor, which is I believe the main reason for having `std::make_optional` as well, so will leave as is to match that.

hilti/runtime/include/types/tuple.h
rsmmr: I had seen trouble with return type deduction if the template args weren't given here, so leaving as is. (And not sure which other places you're referring to.)

hilti/runtime/include/types/optional.h
rsmmr: We need the `decay_t` here, but we can unify them all: the ones below should probably also use that `decay_t`, too. Beyond that, the one distinction we need is the additional check for `is_same<Optional>` in a couple of cases. I've cleaned them all up.

PR #2185, opened 2025-10-21
Fix FPs in `FunctionParamVisitor` when parameters are shadowed.

### LINKED ISSUE #2184: `FunctionParamVisitor` fails to remove some redundant parameters
If a function body contains additional scopes with declarations which shadow function parameters, `FunctionParamVisitor` will not remove them, e.g.,

```ruby
module foo;

function fn(i: uint64, j: uint64): uint64 {
    {
        local j = 10;
        print j;
    }

    return i * 10;
}

print fn(1, 2);
```

```console
$ spicyc -p foo.spicy
module foo {

import hilti;
import spicy_rt;

function uint<64> fn(uint<64> i, uint<64> j) {
    {
        local uint<64> j = 10;
        hilti::print(j, True);
    }

    return i * 10;
}

hilti::print(fn(1, 2), True);

}
```

```console
$ spicyc --version
spicyc v1.15.0-dev.52 (7635e113)
```

bbannier:
The tracking of parameter use in `FunctionParamVisitor` relied on IDs to correlate seen `expression::Name`s and parameter declarations. This broke down when parameters where shadowed, e.g., we often end up shadowing the `__cur` parameter.

With this patch we additionally check that the resolved `Declaration` for a `Name` matches the parameter to take this into account.

Closes #2184.


hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -2155,16 +2156,45 @@ struct FunctionParamVisitor : OptimizerVisitor {
...
+                    break;
+                }
+                case expression::keyword::Kind::Self:
+                case expression::keyword::Kind::DollarDollar:
+                case expression::keyword::Kind::Scope:
+                    util::detail::internalError(util::fmt("unexpected keyword '%s'", name->print()));
+            }
+        }
+        else
+            util::detail::internalError(util::fmt("unexpected expression '%s'", name));
```
evantypanski, line 2183:
```suggestion
            util::detail::internalError(util::fmt("unexpected expression '%s'", name->print()));
```

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -2155,16 +2156,45 @@ struct FunctionParamVisitor : OptimizerVisitor {
...
 
+        std::string_view id = {};
+        const Declaration* resolved_declaration = nullptr;
+
+        if ( auto* x = name->tryAs<expression::Name>() ) {
+            id = x->id().str();
+            assert(x->resolvedDeclaration());
+            resolved_declaration = x->resolvedDeclaration();
+        }
+        else if ( auto* x = name->tryAs<expression::Keyword>() ) {
```
evantypanski, line 2170: could `const` this technically if you want, I wanted to `const` the `name` param but we can't :/


```suggestion
        else if ( const auto* x = name->tryAs<expression::Keyword>() ) {
```

PR #2194, opened 2025-11-05
Reimplement `rt::currentExecutable`

bbannier:


hilti/toolchain/src/base/util.cc
```diff
@@ -199,21 +173,25 @@ hilti::Result<hilti::rt::filesystem::path> util::findInPaths(const hilti::rt::fi
...
-#endif
-
-        rt::internalError(msg);
+hilti::rt::filesystem::path util::currentExecutable(const hilti::rt::filesystem::path& argv0) {
+    // We only use the error code for calls to `exists` to prevent any
+    // exceptions. If that function returns `true` the file exists and one can
+    // ignore the error code.
+    std::error_code ec;
+
+    // If `argv0` exists the executable was invoked directly.
```
rsmmr: I don't think this is true:  if argv0 is `exe` and there's `exe` inside the current path and also in `PATH`, this would return the former, but we're probably running the latter. 

I don't fully recall but I believe such ambiguities was the reason for using pathfind in the first place. Not quite sure what the right way forward is without it.

hilti/toolchain/src/base/util.cc
```diff
@@ -199,21 +173,25 @@ hilti::Result<hilti::rt::filesystem::path> util::findInPaths(const hilti::rt::fi
...
+    if ( rt::filesystem::exists(argv0, ec) )
+        return argv0;
```
rsmmr:
```suggestion
        return normalizePath(argv0);
```

hilti/toolchain/src/base/util.cc
```diff
@@ -199,21 +173,25 @@ hilti::Result<hilti::rt::filesystem::path> util::findInPaths(const hilti::rt::fi
...
+    // Otherwise it was found in `PATH` so look for it there.
+    if ( auto x = rt::getenv("PATH") ) {
+        for ( const auto& dir : split(*x, ":") ) {
+            if ( auto p = rt::filesystem::path(dir) / argv0; rt::filesystem::exists(p, ec) )
+                return p;
+        }
     }
 
-    return normalizePath(exe);
+    cannotBeReached();
```
rsmmr: I don't think `cannotBeReached` is the right course of action: it's not a Spicy-side logic error if `PATH` doesn't exist.

hilti/toolchain/src/base/util.cc
```diff
@@ -199,21 +173,25 @@ hilti::Result<hilti::rt::filesystem::path> util::findInPaths(const hilti::rt::fi
...
+
+    // If `argv0` exists the executable was invoked directly.
+    if ( rt::filesystem::exists(argv0, ec) )
+        return argv0;
+
+    // Otherwise it was found in `PATH` so look for it there.
+    if ( auto x = rt::getenv("PATH") ) {
+        for ( const auto& dir : split(*x, ":") ) {
+            if ( auto p = rt::filesystem::path(dir) / argv0; rt::filesystem::exists(p, ec) )
+                return p;
```
rsmmr:
```suggestion
        return normalizePath(p);
```

rsmmr: Also, custom `PATH` processing can be tricky. I'd at least remove whitespace around the components and make sure `p` is executable (same for argv0 above).

spicy/toolchain/bin/spicy-driver.cc
```diff
@@ -52,7 +52,7 @@ static void hookDeclineInput(const std::string& reason) { declined = true; }
 
 class SpicyDriver : public spicy::Driver, public spicy::rt::Driver {
 public:
-    explicit SpicyDriver() : spicy::Driver("spicy-driver", hilti::util::currentExecutable()) {
+    explicit SpicyDriver(std::string_view argv) : spicy::Driver("spicy-driver", hilti::util::currentExecutable(argv)) {
```
rsmmr:
```suggestion
    explicit SpicyDriver(std::string_view argv0) : spicy::Driver("spicy-driver", hilti::util::currentExecutable(argv0)) {
```

hilti/toolchain/src/base/util.cc
bbannier: I reimplemented this by examining`/proc/self/exe` or macos specific API.

hilti/toolchain/src/base/util.cc
```diff
@@ -200,20 +180,46 @@ hilti::Result<hilti::rt::filesystem::path> util::findInPaths(const hilti::rt::fi
...
-    const auto exe = PathFind::FindExecutable();
+    std::string error = "could not determine path of current executable";
+
+    if ( ! rt::filesystem::exists("/proc") || rt::filesystem::is_empty("/proc") )
+        error += ": /proc needs to be mounted";
 
-    if ( exe.empty() ) {
-        auto msg = std::string("could not determine path of current executable");
+    // On Linux and FreeBSD we use the proc file system.
+#if defined(__linux__) || defined(__FreeBSD__)
```
evantypanski: Could the Linux/FreeBSD get split into a different function from Apple? It's pretty confusing to follow

or, if you disagree, could the `#endif` get comments with the `if` that they close? eg

`#endif // defined(__linux__) || defined(__FreeBSD__)`

hilti/toolchain/src/base/util.cc
```diff
@@ -200,20 +180,46 @@ hilti::Result<hilti::rt::filesystem::path> util::findInPaths(const hilti::rt::fi
...
+    _NSGetExecutablePath(nullptr, &path_len);
+
+    // Get the actual value.
+    std::string path(path_len, '\0');
+    if ( _NSGetExecutablePath(path.data(), &path_len) != 0 )
+        // We already determined above that the path fits into the buffer, this
+        // should never fail.
+        rt::internalError(error);
+
+    return {std::move(path)};
```
evantypanski: According to [docs](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/dyld.3.html):

> Note that _NSGetExecutablePath() will return "a path" to
     the executable not a "real path" to the executable.

I think we should follow the symlink on Mac, if it is a symlink, to match the behavior of `/proc/self/exe`. I just tested on Mac/Linux and found they differ if I print the path:

Mac:

```
$ ln -s bin/spicyc test.out
$ ./test.out
/Users/etyp/src/spicy/build/test.out
```

Linux:

```
$ ln -s bin/spicyc test.out
$ ./test.out
"/home/etyp/src/spicy/build/bin/spicyc"
```

(not sure why Linux includes quotes but not Mac either)

hilti/toolchain/src/base/util.cc
bbannier: I completely removed the nesting so we now have some copied code which should makes things more readable.

hilti/toolchain/src/base/util.cc
bbannier: I added code to handle that, though it is probably a very niche use case since anywhere we care about this is to detect whether we are running from the build directory, and users who symlink their installed binaries would be have been fine (arguably more by accident).

Issue #1992, opened 2025-03-11
Function declarations with missing return are not rejected


bbannier:
We currently do not validate that non-`void` functions always return a value. Users might see a warning from their C++ compiler, and code might trap at runtime.

```ruby
module foo;

function f(): string {}

print f();
```

```console
$ spicyc -dj
/private/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/foo_449edd4c96aa0b67-e2b8330710e71844.cc:26:81: warning: non-void function does not return a value [-Wreturn-type]
   26 | static auto __hlt::foo::f() -> std::string { ::hilti::rt::detail::checkStack(); }
      |                                                                                 ^
1 warning generated.
[2]    54825 trace trap  spicyc foo.spicy -j
```

Instead of trapping the code might also return a bogus value like it does on my machine for a function returning an `uint64`.

Depending on the exact function this could hide until runtime, e.g.,

```ruby
function g(n: uint8): string {
    if (n == 1)
        return "123";
}

print g(0);
```

Detecting all of these scenarios requires control flow analysis for which we will add infrastructure with #1690.

PR #2201, opened 2025-11-18
Refactor and improve optimizer

rsmmr:
This is a large change refactoring, streamlining, and speeding up of the
optimizer and its individual passes. I'm seeing optimization speed-ups
in the order of 2-4 with large analyzers.

The main change is in the PR's final commit, which puts the
`optmizer/*` directories in place.  It wasn't really feasible to split
that out into smaller commits unfortunately, so that's a big chunk,
including both new code (mostly in `optimizer.{h,cc}` and
`pass.{h,cc}`) and re-factoring of existing code (mostly in
`passes/*.cc`). The initial commits in the PR, before that large one,
prepare for it by extending some APIs and moving some stuff around.

Conceptually, the core change with this PR is that we now have a
well-defined contract between optimizer and optimization passes:
before running a pass, the optimizer ensures that the AST is in
a consistent, fully-resolved state; and when a pass finishes, the pass
tells the optimizer what needs to be done to get the new AST back into
that state. See `optimizer::PassInfo` for the details.

We also substantially simplify the optimizer loop: it's no longer
nested across multiple phases & levels, and just runs all passes in
order until the AST stabilizes. Optionally, a individual pass maybe
itself repeated until stabilization, too, but that's the only nesting.

The passes themselves gain a new structure/API that clearly separates
between collection and mutation. We provide base visitor classes for
both kinds to derive from. The new structure no longer joins the
collection/mutation into a single visitor, which was hard to follow.

This new API enables the optimizer to track what parts of the AST were
changed by a pass, so that it can then re-resolve only the
corresponding subparts of the AST after each pass. We  track changes
at the function-level so that we just need to recompute state for any
functions that have changed. The exception is global code/types, where
we re-resolve the whole module.

~~The CFG computation moves from `detail::cfg` into the optimizer
namespace (`detail::optimizer::cfg`), reflecting that it's primarily
an optimizer concern.~~ The CFG ~~infrastructure~~ state is centralized so all
passes can share the same, cached CFGs when possible, rather than each
computing their own. We use the function-tracking mentioned above to
decide which CFGs need re-computation.

The previous type/member/function passes got merged into single
`dead-code-static` pass, with joint collection and mutation visitors.

The constant-folder pass now re-uses the AST resolver's corresponding
functionality. No more two implementations of constant folding.

In debug builds, we do extensive checks that after a pass has
finished, and the AST state updated as prescribed, the AST is indeed
fully resolved again. We do this by recomputing all AST state from
scratch and comparing it against the current one. If there's a
difference, that's hard internal error and the compiler will abort.
This check is expensive and disabled in release builds. It can also be
disabled even in debug builds by setting a new CMake variable
`HILTI_SKIP_EXPENSIVE_DEBUG_CHECKS`. We may want to do this for Zeek
debug builds, as the check slows them down quite a bit.

- **Consistently type function bodies as `statement::Block`.**
- **Extend resolver APIs.**
- **Speed up the graph data structure.**
- **Split out optimizer passes into their own compilation units.**
- **Remove `HILTI_OPTIMIZER_PASSES` and `HILTI_OPTIMIZER_ENABLE_CFG`.**
- **Move CFG tests under optimization topic.**
- **Rename cfg optimizer pass to `dead-code-cfg`.**
- **Refactor optimizer.**


hilti/toolchain/include/compiler/detail/optimizer/optimizer.h
```diff
@@ -0,0 +1,301 @@
...
+     * modified the AST, false otherwise.
+     */
+    using Callback = bool (*)(Optimizer* opt);
+
+    std::string name;      /**< descriptive name of the pass */
+    size_t order;          /**< order in which the pass should run (lower values run first) */
+    bool one_time = false; /**< if true, the pass runs only once in the first round */
+    bool iterate = false;  /**< if true, the pass is re-run until it makes no further modifications */
+    bitmask<PostProcessors> post_processors = PostProcessors::All; /**< post-processing required after modifications */
+    Callback run; /**< the callback executing the pass's main logic, per the contract described above */
```
evantypanski, line 233: I really like that this is just a function, so if you don't want to make a whole visitor, you don't have to.

One extra thing I've found with most optimization engines is that they let you just run your pass on a per-function basis, since most passes are entirely local. I could envision a special kind of `FunctionPass` whose run function gets called for each function, and includes that function, then it doesn't have to care about maintaining the state around functions.

Maybe the `run` callback isn't the right place, but the functionality would be nice.

hilti/toolchain/src/compiler/optimizer/optimizer.cc
```diff
@@ -0,0 +1,435 @@
...
+        HILTI_DEBUG(logging::debug::Optimizer, util::fmt("pass: %s with order %d (round %zu, pass iteration %zu)",
+                                                         pinfo.name, pinfo.order, round, iteration));
+        logging::DebugPushIndent _(logging::debug::Optimizer);
+
+        bool modified_by_pass = false;
+
+        {
+            util::timing::Collector _1(util::fmt("hilti/compiler/optimizer/pass/%s", pinfo.name));
+            auto _2 = _state.trackPass(&pinfo);
+            modified_by_pass = (*pinfo.run)(this);
```
evantypanski, line 321: Maybe in the future we could have an opportunity for purely local passes (ie just on function bodies) which only rerun the parts that changed

hilti/toolchain/src/compiler/optimizer/optimizer.cc
```diff
@@ -0,0 +1,435 @@
...
+            // corresponding CMake option as well.
+            _state.checkAST(pinfo.name);
+#endif
+#endif
+        }
+
+        if ( ! modified_by_pass || ! pinfo.iterate )
+            break;
+
+        if ( ++iteration >= 50 )
```
evantypanski: could this magic number be configured easily? If not it's fine, I could just see some analyses (constant propagation) getting a bit close in certain code

evantypanski (commented):
This is great, it's probably closer to what I'd expect when writing passes. I have some high-level thoughts, I didn't dive into the exact code logic in all cases. These are mostly things I'd wish for and potential other avenues, not saying everything here is 100% perfect and possible :)

First some high-level comments on the approach in the commit message:

> The passes themselves gain a new structure/API that clearly separates
between collection and mutation

I understand that in its current state, a pass being coupled with its collection makes sense. But, really these are *two* passes. There is an "analysis" pass that would not mutate, then a "transform" pass that does.

Some passes need a very particular analysis pass that only helps that one transformation. But, others still may have *multiple* which rely on an analysis pass. For example, I use the operator uses in #2173 and when removing unused function parameters. The CFG is used multiple times (which I consider an analysis pass). I think it's also going to be increasingly common for multiple passes to rely on a single analysis pass.

Theoretically these can be combined into the same pass (it might make sense for the function-signature passes), but then it would seem messy.

What would you think about making passes which mutate able to depend on any number of passes that do *not* mutate, then those passes can provide a mechanism to build-back from a changed state? This would change the `CFG` post-processor to just be an analysis pass which gets rebuilt based on that dependence-rebuilding API. Same for the `CollectorCallers` `Collector` pass.

> The CFG computation moves from detail::cfg into the optimizer
namespace (detail::optimizer::cfg), reflecting that it's primarily
an optimizer concern

Will this always be true? For example, supporting #1992 would mean looking at all incoming nodes to the `end` node and ensure that they are `return`. That's an analysis and has nothing to do with optimization.

I would actually like to couple the CFG less with the optimizer, not more.

> The constant-folder pass now re-uses the AST resolver's corresponding
functionality

This is great, it was in a weird state before. I think that constant folding *is* an optimization, though. It kind of blurs the lines between what should be in the optimizer and what should be elsewhere. This actually kind of goes into the previous point: treating these passes, even transformation passes, as solely "optimizer" passes seems like it puts us in a bit of a box.

I'm wondering if we could lift the constant folder into its own *pass* then just run that pass where it makes sense? This might be a bit of a slippery slope (is resolving actually just a transformation pass?) but it feels odd having folding in its own category off somewhere else to me

I understand, though, that folding is a special case and if anything the change in this is a nice decoupling. Just wanted to float that thought.

hilti/toolchain/include/compiler/detail/optimizer/optimizer.h
rsmmr: Yeah, I can see this being useful once we get more passes of that kind. Could be a separate callback, or maybe we'd support it at the visitor level, by having a corresponding base visitor class (right now, we don't have a base `Pass` class; could introduce that as well, but the `Callback` seemed sufficient for what we need at the moment).

Either way, I'm a bit reluctant once more to design an API without having more need for it, so would prefer to leave this for later as well.

hilti/toolchain/src/compiler/optimizer/optimizer.cc
rsmmr: That might go with that `FunctionPass` as well (need to keep in mind that a module has a body, too; anything that works on "code" needs to work on blocks in the end, not just on functions).

hilti/toolchain/src/compiler/optimizer/optimizer.cc
rsmmr: Do you mean "configured" at runtime or at compile-time? For the latter, sure, I could make it a central constant somewhere (we use the "50" at a number of similar places right now). I wouldn't want to make it a runtime option as users should never need to deal with this;  it's just protecting against bugs in the compiler.

evantypanski: Ah, yeah it's a bit more general of a problem.

I was referring to runtime, since it's an internal compiler error it is meant to be next-to-impossible to hit with normal use, even contrived examples. If users can hit an ICE then it should be configurable in my opinion, so hitting the limit on a user's file doesn't completely stop what they're doing.

Another option is provide a flag to not abort in that case, ie it just exits the `runPass` function here and stops running that pass. Then it'll hopefully come back and finish up. By default that would be off so users don't accidentally hit infinite loops (but we could keep the other iteration limits that might be harder to hit).

For the record I generated a Spicy file like:

```
module Test;

function f() {
    local a0 = 0;
    local a1 = a0 + 1;
    local a2 = a1 + 1;
    # ...
    local a100 = a99 + 1;
    print a100;
}

f();
```

which would now ICE. It's pretty darn contrived, but I can see hitting double-digit iterations in complex spicy functions, so 50 seems like it's low for something like constant propagation. If it's not even configurable then we're just leaving users potentially without a good remedy when it's not necessary

so I guess another option would be up it to 100, 500, or 1000 and I'd be happier too, at least for this particular counter

hilti/toolchain/include/compiler/detail/optimizer/optimizer.h
evantypanski: This one seems good as a followup. We already have a couple passes that would make use of it, but it wouldn't really buy us anything without my other comment (only rerunning local passes on functions that had their bodies changed).

hilti/toolchain/src/compiler/optimizer/optimizer.cc
rsmmr: I really don't like exposing this as a user-side option/flag (it's nothing a user should be concerned about) but let me factor it out into a compile-time constant and set it to 100. The reason for not setting it much higher is that when running into the issue, one would likely want to dump the intermediary ASTs to disk for inspection; and that yields as many files as this number specifies.

An alternative to your flag could be just to silently skip any remaining optimizations if the number is exceeded, but I don't really like that either. Either way, we can revisit in case this ever comes up.

evantypanski:
> Are you up for a full review?

Sure, hopefully I can this afternoon or tomorrow, since it's a big change. Sorry if before it was a bit of a high-level wishlist

hilti/toolchain/src/compiler/optimizer/optimizer.cc
```diff
@@ -0,0 +1,434 @@
...
+    _normalizeModificationState();
+
+    unsigned int round = 1;
+
+    while ( true ) {
+        // The following mimics ASTContext::_resolve(), but skips steps that
+        // the pass doesn't require. It also runs only on changed parts of the
+        // AST; it needs to run only HILTI's versions, no
+        // need to consider other compiler plugins.
+        HILTI_DEBUG(logging::debug::Optimizer, util::fmt("re-resolving AST, round %d", round));
```
evantypanski: "re-resolving" feels a bit too narrow here; perhaps "re-processing"?

Also just a question: why no need to consider compiler plugins here?

hilti/toolchain/src/compiler/optimizer/optimizer.cc
```diff
@@ -0,0 +1,434 @@
...
+
+        if ( ! cfg )
+            continue; // block no longer part of AST
+
+        auto actual = cfg->dot(false);
+        auto expected = CFG(block).dot(false);
+
+        if ( actual != expected ) {
+            std::cerr << "==== ACTUAL   ===\n\n" << actual << "\n\n";
+            std::cerr << "==== EXPECTED ===\n\n" << expected << "\n\n";
```
evantypanski: nit: LHS has 4 equals and RHS has 3! (I only spotted because my editor has a special glyph for 3 equals, not 4)


```suggestion
            std::cerr << "=== ACTUAL   ===\n\n" << actual << "\n\n";
            std::cerr << "=== EXPECTED ===\n\n" << expected << "\n\n";
```

hilti/toolchain/include/compiler/detail/optimizer/optimizer.h
```diff
@@ -0,0 +1,309 @@
...
+     */
+    static std::optional<std::pair<ID, std::string>> idFeatureFromConstant(const ID& feature_constant);
+
+    /**
+     * Extracts the innermost type, removing any wrapping in reference or
+     * iterator types recursively.
+     *
+     * @param type the type to unwrap
+     */
+    static QualifiedType* innermostType(QualifiedType* type);
```
evantypanski: could this just be a method on the type instead?

hilti/toolchain/include/compiler/detail/optimizer/optimizer.h
```diff
@@ -0,0 +1,309 @@
...
+
+    std::string name;      /**< descriptive name of the pass */
+    size_t order;          /**< order in which the pass should run (lower values run first) */
+    bool one_time = false; /**< if true, the pass runs only once in the first round */
+    bool iterate = false;  /**< if true, the pass is re-run until it makes no further modifications */
+    bitmask<Guarantees> guarantees = Guarantees::None; /**< AST guarantees the pass provides after processing */
+    Callback run; /**< the callback executing the pass's main logic, per the contract described above */
+
+    bool operator<(const PassInfo& other) const {
+        return order != other.order ? order < other.order : name < other.name;
```
evantypanski: maybe we should assert when adding that the names aren't equal, just in case? That way this remains deterministic

hilti/toolchain/include/compiler/detail/optimizer/optimizer.h
```diff
@@ -0,0 +1,309 @@
...
+    /**
+     * Callback executing the pass's main logic. This will typically run a
+     * `visitor::Mutator` over the AST to make modifications as desired; see
+     * the contract described above. The callback must return true if it
+     * modified the AST, false otherwise.
+     */
+    using Callback = bool (*)(Optimizer* opt);
+
+    std::string name;      /**< descriptive name of the pass */
+    size_t order;          /**< order in which the pass should run (lower values run first) */
```
evantypanski: Hm, ordering like this always worries me. In order to determine what order my pass should be in, I will probably open every other pass to look at their orders. Then most likely just throw my pass at the end anyway, like "highest number + 10"

It goes against the design a bit, but maybe an enum might work here? It means that it'll no longer be one place to register the pass, since you have to specify an order. But, the benefit is that you can see all other orders in one place, so I can make an informed decision about a pass's order immediately.

There's a world where passes can be strictly ordered based on the `order` so we don't have to fallback on the name, as well. That might be nice to avoid subtle behavior differences with ordering, though the hope would be that the behavior remains consistent no matter the order

hilti/toolchain/src/compiler/optimizer/optimizer.cc
```diff
@@ -0,0 +1,434 @@
...
+using namespace hilti::detail::optimizer;
+
+CFG* ASTState::cfg(statement::Block* block) {
+    assert(block);
+
+    if ( auto it = _cfgs.find(block); it != _cfgs.end() )
+        return it->second.get();
+    else
+        return _cfgs.emplace(block, std::make_unique<CFG>(block)).first->second.get();
+}
```
bbannier, line 38: Before this PR we only computed CFGs for either the global module block, or blocks which are function bodies. These CFGs were complete after also [adding globals and function parameters](https://github.com/zeek/spicy/blob/3a8e48d166f583a0e0cd432a751515d1cad1c1cc/hilti/toolchain/src/compiler/cfg.cc#L120-L121). If we compute a CFG for an arbitrary block it could be missing additional data from locals in a parent scope.

I'd suggest still accepting a `Block` here, but perform the tracking for the outermost block. For that we'd climb up to a function body or the global module block, and then use in places where we currently use `block`.

We could consider changing the CFG impl to allow computing full CFGs for nested blocks, but even if we did that it would still be hard to keep them consistent when their parent scopes change. With that we could probably simplify the tracking here and directly attach CFG change tracking to function change tracking. This would also allow more easily invalidate a CFG which currently isn't as streamlined.

hilti/toolchain/include/compiler/detail/optimizer/pass.h
```diff
@@ -0,0 +1,216 @@
...
+    ~Collector() override = default;
+
+    /** Returns the AST context being optimized. */
+    auto* context() const { return _optimizer->context(); }
+
+    /** Returns the optimizer instance passed into the constructor. */
+    auto* optimizer() const { return _optimizer; }
+
+    /** Returns the current optimizer state for use by the pass. */
+    auto* state() const { return _optimizer->state(); }
```
bbannier, line 43: Both `optimizer()` and `state()` give access to `builder()` which really shouldn't be called from a collector. I'd suggest 

- dropping `optimizer` (currently unused, but gives out a mutable `state()` and `builder()`), and
-  changing `state()` to return a `const auto*` (still allow inspecting state without mutation). To give access to`ASTState::cfg` it should become `const` and we'd need to declare `ASTState::_cfgs` `mutable`.

`context` could stay as is since it doesn't allow mutating anyway and making it return a `const` would require making all the context uses const-correct.

hilti/toolchain/src/compiler/optimizer/optimizer.cc
bbannier: Like Evan I am worried about somehow running into this limit with valid code. With an internal error we'd force users to rewrite their code. Would it be much harder to simply skip further steps if we hit the limit?

hilti/toolchain/src/compiler/optimizer/optimizer.cc
rsmmr: > I'd suggest still accepting a `Block` here, but perform the tracking for the outermost block. For that we'd climb up to a function body or the global module block, and then use in places where we currently use `block`.

What would we return (and cache) in that case? Returning the CFG for the outer block wouldn't be expected, would it?

How about instead documenting that only direct function/module bodies are supported and adding an assert to enforce that?

hilti/toolchain/include/compiler/detail/optimizer/pass.h
rsmmr: > Both `optimizer()` and `state()` give access to `builder()` which really shouldn't be called from a collector. I'd suggest

Good point.

> * dropping `optimizer` (currently unused, but gives out a mutable `state()` and `builder()`), and

Ok. Not totally great, because it seems natural to have access to the optimizer, also for symmetry with the Mutator, but seems fine for now. I'll drop it.

> * changing `state()` to return a `const auto*` (still allow inspecting state without mutation). To give access to`ASTState::cfg` it should become `const` and we'd need to declare `ASTState::_cfgs` `mutable`.

Suggest an easier change: we can remove `builder()` from `ASTState`. It's not needed because mutator passes can (and do) get to it through the optimizer already. And generally having a mutable state seems fine to me even from the collector (so that it can compute stuff without having to use `mutable`).

> `context` could stay as is since it doesn't allow mutating anyway

Technically, by having the context, one could create a custom builder. But seems fine to leave that loop-hole, it's really takes going out of the way to do that.

hilti/toolchain/src/compiler/optimizer/optimizer.cc
bbannier: If by `cfg(block)` we mean _dataflow facts about the `block`_ returning the function/module block CFG would be fine. The only place where that could become an issue is if the caller plans to use e.g., the CFG's `start` node directly for walking the CFG without going through a CFG node lookup. Since nobody currently does that I don't think that's a problem and returning the CFG for the whole block would still be fine.

hilti/toolchain/include/ast/node.h
```diff
@@ -643,6 +643,13 @@ class Node {
      */
     void replaceChild(ASTContext* ctx, Node* old, Node* new_);
 
+    /**
+     * Removes the node from its parent. The node will remain valid and can be
+     * re-inserted into the AST elsewhere later. Does nothing if the node has
+     * no parent. If the node has no parent, this is a no-op.
```
evantypanski: the two sentences say essentially the same thing


```suggestion
     * no parent.
```

hilti/toolchain/include/ast/visitor.h
```diff
@@ -211,42 +211,65 @@ class MutatingVisitorBase {
...
+
     /** Clears the flag recording that modifications have taken place. */
     auto clearModified() { _modified = false; }
 
     /**
-     * Replace a child node with a new node. This also logs a corresponding
-     * debug message to the stream passed to the constructor.
+     * Replace a child node with a new node.
+     *
+     * When overriding, the parent's implementation should be called.
```
evantypanski, line 231: It's not a big deal since I don't foresee this getting overridden much, but couldn't we guarantee the parent impl always gets called by making this non-virtual and having it call a virtual private method?

Maybe a couple more lines of code, but I expect that more than finding this restriction in a doc comment, even if it is clear

hilti/toolchain/include/compiler/detail/optimizer/optimizer.h
```diff
@@ -0,0 +1,306 @@
...
+ */
+class ASTState {
+public:
+    /** Returns the AST context being optimized. */
+    auto* context() { return _context; }
+
+    /**
+     * While a pass is running, returns information about it.
+     *
+     * @return the pass info, or null if no pass is currently running
```
evantypanski: but this has `assert(_pinfo)` and dereferences it

hilti/toolchain/include/compiler/detail/optimizer/optimizer.h
```diff
@@ -0,0 +1,306 @@
...
+    std::unordered_set<declaration::Module*> _modified_modules; // set of modified modules
+
+    std::unordered_map<statement::Block*, std::unique_ptr<CFG>> _cfgs; // cached CFGs
+};
+
+/**
+ * Enum specifying post-processing steps that optimization passes do *not*
+ * require after processing an AST. Generally, after each pass, the optimizer
+ * will re-run various resolver visitors to bring it back into a fully resolved
+ * and consistent state. For efficiency, passes can indicate through this enum
```
evantypanski:
```suggestion
 * and consistent state. For efficiency, passes can indicate through a bitmask
```

hilti/toolchain/include/compiler/detail/scope-builder.h
```diff
@@ -7,7 +7,20 @@
...
+ * Implements the corresponding functionality for the default HILTI compiler plugin.
+ *
+ * The result indicates whether any scopes were modified during the process.
+
+ * Note that this doesn't report if there are pre-existing IDs that weren't
+ * accessed at all during the current run. It will return "unchanged" in that
+ * case if no changes were made beyond such IDs.
+ *
+ * @param builder the builder to use
+ * @param root root node to process.
```
evantypanski:
```suggestion
 * @param node root node to process.
```

hilti/toolchain/include/compiler/detail/scope-builder.h
```diff
@@ -7,7 +7,20 @@
 
 namespace hilti::detail::scope_builder {
 
-/** Implements the corresponding functionality for the default HILTI compiler plugin. */
-void build(Builder* builder, ASTRoot* root);
+/**
+ * Implements the corresponding functionality for the default HILTI compiler plugin.
```
evantypanski, line 11: this was the comment here before, but I'm not really quite sure what it means? I don't know what "corresponding functionality" here is corresponding to. Maybe just remove this line?

hilti/toolchain/include/compiler/detail/resolver.h
```diff
@@ -8,6 +8,18 @@
...
+bool resolve(Builder* builder, Node* node);
+
+/**
+ * Implements just the type and expression coercion pass of the resolver. This
+ * can be run on any subpart of the AST. It assumes that expressions and types
+ * have already been resolved except for coercions. This makes as many passes
+ * over the AST as needed until all possible coercions have been applied.
+ *
+ * @param builder the builder to use
+ * @param root root node to process.
```
evantypanski:
```suggestion
 * @param node root node to process.
```

hilti/toolchain/src/ast/ast-context.cc
```diff
@@ -898,6 +897,28 @@ void ASTContext::_dumpState(const logging::DebugStream& stream) {
     logger().debugPopIndent(stream);
 }
 
+void ASTContext::_dumpState(std::ostream& out) const {
+    // This mostly duplicates the code from above but for ostreams. Not clear
+    // how to nicely cover both cases with just one version of the code.
```
evantypanski, line 902: IDK I guess you could bring it out into a separate function/method, then have a callback for the `<<` vs. `HILTI_DEBUG` logic

I don't think that's any cleaner, though!

hilti/toolchain/src/compiler/constant-folder.cc
```diff
@@ -358,3 +374,40 @@ Result<Ctor*> detail::constant_folder::fold(Builder* builder, Expression* expr)
...
+            }
+        }
+    }
+};
+
+bool detail::constant_folder::fold(Builder* builder, Node* node, bitmask<Style> style) {
+    bool modified = false;
+
+    while ( true ) {
+        // Pass 3 is in charge of coercion.
```
evantypanski: I think this comment is irrelevant/from the resolver


```suggestion
```

hilti/toolchain/src/compiler/optimizer/optimizer.cc
rsmmr: Ok, I'll make it use the outermost block and add a comment to the method about that.

hilti/toolchain/src/compiler/optimizer/optimizer.cc
rsmmr: Because everything has already been lowered to HILTI at the time the optimizer runs, no need to re-process any, e.g., Spicy-side elements. I'll add that to the comment.

hilti/toolchain/include/compiler/detail/scope-builder.h
rsmmr: It means it implements the scope-building callback for the plugin (`Plugin::ast_build_scopes`).. I'd clarify the comment but the same phrase is used across all the plugin callbacks, so I think I'll leave as is to not need to touch them all.

hilti/toolchain/src/ast/ast-context.cc
rsmmr: > I don't think that's any cleaner, though!

That's what I meant by "nicely". :-)

hilti/toolchain/include/ast/visitor.h
rsmmr: We could if we restricted *when* exactly it's called. I think it would then need to always be called either first before calling the private virtual method, or always afterwards. The way we currently override it, it would need to be called afterwards, but I'm not sure if that's generally right (well, I doubt we'll really ever override this much differently, but *conceptually* it seems cleaner to me to let the overriding method decide when to call the parent implementation).

hilti/toolchain/include/compiler/detail/optimizer/optimizer.h
rsmmr: Ack, clarified the comment.

hilti/toolchain/include/compiler/detail/optimizer/optimizer.h
rsmmr: Convinced. I've switch to an enum that implicitly defines order and name as well.  It's in separate commit, what do you think?

hilti/toolchain/include/compiler/detail/optimizer/optimizer.h
rsmmr: (No longer relevant with the new enum approach, see below).

PR #2203, opened 2025-11-21
Clean up pre-C++20 range-like utilities

bbannier:
Most of these are not needed anymore, and using the stdlib ranges should mean better performance and hopefully less bugs.

This came out of looking around for #2202, but turned out to be unrelated.

### DIFF
--- a/hilti/toolchain/include/base/util.h
+++ b/hilti/toolchain/include/base/util.h
@@ -116,32 +116,22 @@ extern std::pair<std::string, std::string> rsplit1(std::string s, const std::str
 Result<std::vector<std::string>> splitShellUnsafe(const std::string& s);
 
 /**
- * Returns a subrange of a vector, specified through zero-based indices. If
+ * Returns a subview of a range, specified through zero-based indices. If
  * indices are out of range, they are cut back to start/end of input.
  *
- * @param v vector to slice
+ * @param v range to slice
  * @param begin 1st index; if negative, counts from end Python-style
  * @param end one beyond last index; if negative, counts from end Python-style
  */
-template<typename T>
-std::vector<T> slice(const std::vector<T>& v, int begin, int end = -1) {
-    if ( begin < 0 )
-        begin = v.size() + begin;
-
-    if ( static_cast<size_t>(begin) > v.size() )
-        return {};
-
+template<std::ranges::sized_range R>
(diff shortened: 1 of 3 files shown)


hilti/toolchain/include/base/util.h
```diff
@@ -73,6 +73,13 @@ std::string fmt(const char* fmt, const Args&... args) {
 
 using hilti::rt::transform; // NOLINT(misc-unused-using-decls)
 
+/** Collect a view into a vector. */
+template<std::ranges::input_range R>
+auto toVector(R&& xs) {
+    // TODO(bbannier): replace callers with `std::ranges::to` in C++23 and remove this function.
```
evantypanski: Could we make this like `TODO(C++23)` then when we upgrade (in 1 or 5 years) we just grep for that?

not sure if there will be more cases.

hilti/runtime/include/util.h
```diff
@@ -258,6 +257,7 @@ bool endsWith(std::string_view s, std::string_view suffix);
 template<typename T, typename TIter = decltype(std::begin(std::declval<T>())),
          typename = decltype(std::end(std::declval<T>()))>
 constexpr auto enumerate(T&& iterable) {
+    // TODO(bbannier): replace callers with `std::views::enumerate` in C++23 and remove this function.
```
evantypanski: (also potential `TODO(C++23)` spot)

hilti/runtime/include/util.h
```diff
@@ -362,18 +364,21 @@ std::string escapeBytes(std::string_view s, bitmask<render_style::Bytes> style =
...
-std::string join(const T& l, const std::string& delim = "") {
+template<std::ranges::input_range T>
+std::string join(T&& l, std::string_view delim = "")
+    requires(std::is_constructible_v<std::string, std::ranges::range_value_t<T>>)
+{
     std::string result;
     bool first = true;
 
     for ( const auto& i : l ) {
         if ( not first )
```
evantypanski: while here could we change this to `! first`? It's one of those cases that compilers differ in behavior and I find it unclear. I know this wasn't touched so doesn't have to be here, just a request

evantypanski (approved):

hilti/runtime/include/util.h
bbannier: Sure, I did touch it up slightly after all.

PR #2208, opened 2025-11-24
Add pass which performs block flattening

bbannier:


hilti/toolchain/include/compiler/detail/optimizer/pass-id.h
```diff
@@ -19,6 +19,7 @@ enum PassID {
     DeadCodeCFG,
     ConstantPropagation,
     RemoveUnusedParameters,
+    FlattenBlocks,
```
rsmmr: Is this the best position in the order of passes? Asking because the commit message says  that other passes will benefit from having blocks flattened. Should we do it earlier, instead of waiting for the next iteration?

hilti/toolchain/src/compiler/optimizer/passes/flatten-blocks.cc
```diff
@@ -0,0 +1,158 @@
...
+    Function* function = nullptr;
+    for ( ; n; n = n->parent() ) {
+        if ( auto* f = n->tryAs<Function>() ) {
+            function = f;
+            break;
+        }
+    }
+
+    return function;
+}
```
rsmmr: Can use `n->parent<Function>()` instead of this helper.

hilti/toolchain/src/compiler/optimizer/passes/flatten-blocks.cc
```diff
@@ -0,0 +1,158 @@
...
+namespace {
+
+struct VariablesSort {
+    bool operator()(Node* a, Node* b) const { return a->identity() < b->identity(); }
+};
+
+// Enforce a consistent order of the declarations so their renaming below is deterministic.
+using Variables = std::map<declaration::LocalVariable*, std::unordered_set<expression::Name*>, VariablesSort>;
+
+struct Collector : optimizer::visitor::Collector {
```
rsmmr, line 23: Suggest to add a brief comment saying what this collector is collecting (I know other passes don't always do it either, but still, just helps reading).

hilti/toolchain/src/compiler/optimizer/passes/flatten-blocks.cc
```diff
@@ -0,0 +1,158 @@
...
+    void operator()(declaration::LocalVariable* local) override { (*variables)[local]; }
+
+    void operator()(expression::Name* name) override {
+        // Since this is a pre-order visitor we will always visit declarations of locals before their use.
+        if ( auto* decl = name->resolvedDeclaration()->tryAs<declaration::LocalVariable>();
+             decl && variables->contains(decl) )
+            (*variables)[decl].insert(name);
+    }
+
+    Variables* variables = nullptr;
```
rsmmr: Suggest to use the same pattern as other collectors: Put the collected state (i.e,, the instance of `Variables`) into the collector, then access from the mutator through `collector.variables`).

hilti/toolchain/src/compiler/optimizer/passes/flatten-blocks.cc
```diff
@@ -0,0 +1,158 @@
...
+
+    void operator()(statement::Block* block) override {
+        // Only work on blocks which are in other blocks. This excludes e.g.,
+        // blocks which are function bodies, or bodies of `if` or `while`.
+        auto* parent = block->parent()->tryAs<statement::Block>();
+        if ( ! parent )
+            return;
+
+        // We only work on blocks in functions, but not global blocks.
+        hilti::Function* function = containingFunction(block);
```
rsmmr:
```suggestion
        hilti::Function* function = block->parent<Function>();
```

hilti/toolchain/src/compiler/optimizer/passes/flatten-blocks.cc
```diff
@@ -0,0 +1,158 @@
...
+        if ( auto* f = n->tryAs<Function>() ) {
+            function = f;
+            break;
+        }
+    }
+
+    return function;
+}
+
+struct Mutator : optimizer::visitor::Mutator {
```
rsmmr, line 52: Likewise, a brief summary of what this mutator does wouldn't hurt.

hilti/toolchain/src/compiler/optimizer/passes/flatten-blocks.cc
```diff
@@ -0,0 +1,158 @@
...
+    Mutator(Optimizer* optimizer) : optimizer::visitor::Mutator(optimizer) {}
+
+    // Tracks names which we will introduce into the parent scope. We don't
+    // rely only on scope lookups since we only reresolve once all locals in a block
+    // have been renamed.
+    std::unordered_map<Function*, std::unordered_map<ID, Declaration*>> used_ids;
+
+    void operator()(statement::Block* block) override {
+        // Only work on blocks which are in other blocks. This excludes e.g.,
```
rsmmr:
```suggestion
        // Only work on blocks which are inside other blocks. This excludes e.g.,
```

hilti/toolchain/src/compiler/optimizer/passes/flatten-blocks.cc
```diff
@@ -0,0 +1,158 @@
...
+            decl->setID(id);
+
+            for ( auto* name : uses ) {
+                recordChange(name, util::fmt(R"(renaming reference "%s" -> "%s")", name->id(), id));
+                name->setID(id);
+            }
+        }
+
+        // Fold contents of block into parent.
+        auto contents = parent->children();
```
rsmmr: Use `parent->statements()`. Use of `children()` should generally be avoided when working with specific AST node types as that assumes knowledge about the internal node layout. I'd then also rename `contents` to `statements` for readability.

hilti/toolchain/src/compiler/optimizer/passes/flatten-blocks.cc
```diff
@@ -0,0 +1,158 @@
...
+        recordChange(parent, "inlining child block");
+        parent->clearChildren();
```
rsmmr: Similar: don't just manipulate "all children". Instead add a method `removeStatements()` to `statement::Block`.

hilti/toolchain/src/compiler/optimizer/passes/flatten-blocks.cc
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        // Copy the original contents in order, but inline the block.
+        for ( auto* c : contents ) {
+            if ( c == block ) {
+                recordChange(block, "inlining block");
+
+                auto children = block->children();
+                block->clearChildren();
+                parent->addChildren(context(), children);
```
rsmmr: Loop over child statements and use `parent->add(stmt)`, or add a new helper  method `add(Statements stmts)` to `Block`.

hilti/toolchain/src/compiler/optimizer/passes/flatten-blocks.cc
```diff
@@ -0,0 +1,158 @@
...
+            }
+
+            else
+                parent->addChild(context(), c);
```
rsmmr:
```suggestion
                parent->add(context(), c);
```

hilti/toolchain/src/compiler/optimizer/passes/flatten-blocks.cc
```diff
@@ -0,0 +1,158 @@
...
+                        // Since loops and conditionals can have a block we check whether the block actually
+                        // contains the declaration so we do not lifecycle variables declared in e.g., the loop
+                        // control block.
+                        if ( auto* local = a->tryAs<declaration::LocalVariable>();
+                             local && cfg::contains(*block, *local) ) {
+                            recordChange(local,
+                                         "resetting block-local variable at the end of block since block will "
+                                         "be removed");
+                            block->addChild(context(), builder()->assign(builder()->id(local->id()),
+                                                                         builder()->default_(local->type()->type())));
```
rsmmr, line 138: The commit message says "by now pointing
the alias to other invalid data". Might be matter of interpretation but technically `default_()` does produce something valid. I *believe* semantically `default()` does give us what we want, so maybe just don't refer to it as invalid.

hilti/toolchain/src/compiler/optimizer/passes/flatten-blocks.cc
```diff
@@ -0,0 +1,158 @@
...
+        recordChange(parent, "inlining child block");
+        parent->clearChildren();
+
+        // Variables declared in the block would have previously gone
+        // out of scope. Overwrite them to force any aliases to also
+        // see an update.
+        auto* cfg = state()->cfg(function->body());
+        if ( ! contents.empty() ) {
+            if ( auto* last = *contents.rbegin() ) {
+                const auto& successors = cfg->graph().neighborsDownstream(last->identity());
```
rsmmr:
```suggestion
        if ( ! contents.empty() ) {
            if ( auto* last = *contents.rbegin() ) {
                auto* cfg = state()->cfg(function->body());
                const auto& successors = cfg->graph().neighborsDownstream(last->identity());
```

rsmmr (commented):
Looks good, I like how it turned out now, mostly easy to follow (with the exception of the CFG tracking, but I think I got it :-). Main feedback is the manipulation of children instead of statements..

hilti/toolchain/include/compiler/detail/optimizer/pass-id.h
bbannier: If I move it before `DeadCodeCFG` that pass removes variables which are still used leading to a C++ compilation error. I had hoped that the new way we run passes would catch that, but it looks like it does not. I'll try to figure out where things go wrong.

bbannier: This seems to be due to `__captures` which we cannot really rename. I added a workaround, but if feels like we'd eventually need to properly model it (there's also an explicit workaround for it in `optimizer/passes/remove-unused-params.cc`)

https://github.com/zeek/spicy/blob/53c5d8b18bce414fff1f6d55275236086aa49390/hilti/toolchain/src/compiler/optimizer/passes/flatten-blocks.cc#L76-L82.

hilti/toolchain/include/ast/statements/block.h
```diff
@@ -13,6 +13,11 @@ class Block : public Statement {
 public:
     auto statements() const { return childrenOfType<Statement>(); }
 
+    void removeStatements() {
+        for ( auto* stmt : statements() )
+            removeChild(stmt);
```
rsmmr:
```suggestion
        clearChildren();
```

Here's it's actually ok to just remove all the children because we're inside the node class that knows its own layout.

PR #2214, opened 2025-11-25
Stop using reserved identifiers for generated IDs

bbannier:


rsmmr (commented):
I didn't know that `$` is even possible and dislike it quite a bit: when looking at the new output, my immediate reaction is "how does that even work?". 

How about using just a single underscore instead?

hilti/runtime/include/logging.h
```diff
@@ -38,7 +38,9 @@ void warning(std::string_view msg);
     }
 
 /** Shortcut to `hilti::rt::debug::setLocation`. */
-#define __location__(x) ::hilti::rt::debug::setLocation(x);
+// NOTE: This identifier should have a form compatible with one constructed
+// with `HILTI_INTERNAL_GLOBAL_ID`.
+#define hlt_internal_location(x) ::hilti::rt::debug::setLocation(x);
```
rsmmr: This is the one case where I don't quite like the new approach: this shows up prominently in `spicyc -c` output, and it now looks like a function call, but is a macro. I'm wondering if we should switch this to normal, namespaced & inlined function, so that the calls in the generated code look like this:

```
::hilti::rt::location("....")
```

At least it fits in with other runtime library functionality then.

hilti/runtime/src/tests/library-dummy.cc
```diff
@@ -10,7 +10,8 @@ extern "C" {
 int foo() { return RETURN_VALUE; }
 }
 
-const char HILTI_EXPORT HILTI_WEAK* __hlt_hlto_library_version = R"({
+// NOTE: This should be consistent with the identifier generated with `HILTI_INTERNAL_GLOBAL_ID`.
+const char HILTI_EXPORT HILTI_WEAK* hlt_internal_hlto_library_version = R"({
```
rsmmr: The macro should work here: 
```suggestion
const char HILTI_EXPORT HILTI_WEAK* HILTI_INTERNAL_GLOBAL(hlto_library_version) = R"({
```

(Needs to include `util.h` then.)

hilti/runtime/src/tests/type-info.cc
```diff
@@ -26,12 +26,12 @@ type Y = struct {
 */
 
 // Copied from output of hiltic.
-namespace __hlt::type_info {
+namespace hlt_internal::type_info {
```
rsmmr: Suggest to add another macros `HILTI_INTERNAL_NS` to build scoped identifiers.

hilti/toolchain/include/compiler/context.h
```diff
@@ -43,7 +43,8 @@ struct Options {
     std::vector<hilti::rt::filesystem::path> library_paths; /**< additional directories to search for imported files */
     std::string cxx_namespace_extern =
         "hlt"; /**< CXX namespace for generated C++ code accessible to the host application */
-    std::string cxx_namespace_intern = "__hlt"; /**< CXX namespace for generated internal C++ code */
+    // NOTE: This should be consistent with identifiers generated with `HILTI_INTERNAL_GLOBAL_ID`.
+    std::string cxx_namespace_intern = "hlt_internal"; /**< CXX namespace for generated internal C++ code */
```
rsmmr: I'd add a global constant to `util.h` containing `hlt_internal`.

hilti/toolchain/src/compiler/optimizer.cc
```diff
@@ -0,0 +1,2583 @@
...
+
+/**
+ * Visitor running on the final, optimized AST to perform additional peephole
+ * optimizations. Will run repeatedly until it performs no further changes.
+ */
+struct PeepholeOptimizer : visitor::MutatingPostOrder {
+    using visitor::MutatingPostOrder::MutatingPostOrder;
+
+    static bool isErrorId(std::string_view id) {
+        // Nominally the ID of the error field is `$error`, but we also allow `__error` for testing.
```
rsmmr: Comment is out of date (`$`).

spicy/toolchain/include/compiler/detail/codegen/parser-builder.h
```diff
@@ -571,10 +571,10 @@ class ParserBuilder {
     void syncAdvanceHook(std::shared_ptr<Builder> cond = {});
 
     /**
-     * Returns an expression referencing the current parse object's `__filters`
+     * Returns an expression referencing the current parse object's `$filters`
```
rsmmr:
```suggestion
     * Returns an expression referencing the current parse object's `HILTI_INTERNAL_GLOBAL(filters)`
```

spicy/toolchain/include/compiler/detail/codegen/parser-builder.h
```diff
@@ -571,10 +571,10 @@ class ParserBuilder {
...
      * member if that exists; otherwises return a `Null` expression. The result
      * of this method can be passed to runtime functions expecting a
-     * `__filters` argument.
+     * `$filters` argument.
```
rsmmr:
```suggestion
     * `HILTI_INTERNAL_GLOBAL(filters)` argument.
```

spicy/toolchain/src/compiler/parser/parser.yy
```diff
@@ -367,8 +366,17 @@ local_id      : IDENT                            { std::string name($1);
                                                    if ( ! driver->builder()->options().skip_validation ) {
                                                        if ( name.find('-') != std::string::npos)
                                                            hilti::logger().error(hilti::util::fmt("Invalid ID '%s': cannot contain '-'", name), __loc__.location());
+
                                                        if ( name.substr(0, 2) == "__" )
                                                            hilti::logger().error(hilti::util::fmt("Invalid ID '%s': cannot start with '__'", name), __loc__.location());
+
+                                                       const auto prefix_local = HILTI_INTERNAL_ID("");
```
rsmmr, line 374: Elsewhere I suggested adding a constant to util.h that stores the prefix. Then we can replace `HILTI_INTERNAL_ID("")` with that constant, here and elsewhere.

bbannier: I added an macro for the global internal namespace, but would like to keep computing `prefix_local` here since introducing another macro would cause even more duplication in `util.h`. I'll fix `prefix_global` to use the new macro though.

Similar edit in `hilti/toolchain/src/compiler/parser/parser.yy`.

hilti/runtime/src/tests/library-dummy.cc
```diff
@@ -10,7 +11,8 @@ extern "C" {
 int foo() { return RETURN_VALUE; }
 }
 
-const char HILTI_EXPORT HILTI_WEAK* __hlt_hlto_library_version = R"({
+// NOTE: This should be consistent with the identifier generated with `HILTI_INTERNAL_GLOBAL_ID`.
```
rsmmr: Stale comment.

hilti/toolchain/src/compiler/codegen/expressions.cc
```diff
@@ -133,9 +133,10 @@ struct Visitor : hilti::visitor::PreOrder {
         if ( decl->isA<declaration::GlobalVariable>() ) {
             if ( cg->options().cxx_enable_dynamic_globals ) {
                 if ( auto ns = fqid.namespace_(); ! ns.empty() )
-                    result = {fmt("%s->%s", cxx::ID(ns, "__globals()"), cxx::ID(fqid.local())), Side::LHS};
+                    result = {fmt("%s->%s", cxx::ID(ns, HILTI_INTERNAL_ID("globals()")), cxx::ID(fqid.local())),
+                              Side::LHS};
                 else
-                    result = {fmt("__globals()->%s", cxx::ID(fqid)), Side::LHS};
+                    result = {fmt(HILTI_INTERNAL_ID("globals()->%s"), cxx::ID(fqid)), Side::LHS};
```
rsmmr:
```suggestion
                    result = {fmt("%s->%s, HILTI_INTERNAL_ID("globals", cxx::ID(fqid)), Side::LHS};
```

tests/spicy/rt/release-vs-debug.cc
```diff
@@ -17,7 +17,7 @@
 #include <spicy/rt/libspicy.h>
 
 // Just dummy data.
-const char HILTI_EXPORT HILTI_WEAK * __hlt_hlto_library_version = R"({"created":1597144800.98031,"debug":false,"hilti_version":400,"magic":"v1","optimize":false})";
+const char HILTI_EXPORT HILTI_WEAK * hlt_internal_hlto_library_version = R"({"created":1597144800.98031,"debug":false,"hilti_version":400,"magic":"v1","optimize":false})";
```
rsmmr:
```suggestion
const char HILTI_EXPORT HILTI_WEAK * HILTI_INTERNAL_GLOBAL(hlto_library_version) = R"({"created":1597144800.98031,"debug":false,"hilti_version":400,"magic":"v1","optimize":false})";
```

tests/spicy/doc/my_http-host-parse2.cc
```diff
@@ -26,7 +26,7 @@ int main(int argc, char** argv) {
     stream->freeze();
 
     // Instantiate unit.
-    auto request = hilti::rt::reference::make_value<__hlt_my_http::MyHTTP::RequestLine>();
+    auto request = hilti::rt::reference::make_value<hlt_internal_my_http::MyHTTP::RequestLine>();
```
rsmmr:
```suggestion
    auto request = hilti::rt::reference::make_value<HILTI_INTERNAL_GLOBAL(my_http)::MyHTTP::RequestLine>();
```

rsmmr (approved):

PR #2218, opened 2025-12-03
Collection of cleanups

bbannier:
This PR contains a number of cleanups I made over in #2208 which can and should be applied independently.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -650,7 +650,7 @@ freebsd13_task:
 
 freebsd14_task:
   freebsd_instance:
-    image_family: freebsd-14-2
+    image_family: freebsd-14-3
     cpu: 8
     memory: 8GB
 


doc/faq.rst
```diff
@@ -34,6 +34,24 @@ to line 37, character 1 in the file ``foo.spicy``; ``foo.spicy:37:1-42:19`` is
...
+rejected and exception is thrown. At the same time, some the Spicy compiler
+performs some rewrites which remove unneeded code, i.e., code without "useful"
+side effects, in the case of above examples a statement ``1/0;`` could be
+removed while a statement like ``print 1/0;`` would not since it has a side
+effect.
+
+The automatic rewrites should never cause a parser which does not raise any
+exceptions to change behavior; if on the other hand a parser does raise
+exceptions Spicy takes slightly more liberties and allows the final program to
+raising fewer exceptions. Explicit ``throw`` statements are always respected.
```
rsmmr:
```suggestion
Spicy performs checks for correct usage, e.g., undefined integer operations
like ``1/0`` or ``MAX_INT + 1``, or dereferencing an invalid iterator. 
At the same time, however, the Spicy compiler may
perform some optimizations removing unneeded code, i.e., code without "useful"
side effects. In the case of above examples a statement ``1/0;`` could be
removed while a statement like ``print 1/0;`` would not since it has a side
effect (visible output).

These optimizations will never cause a parser which does not raise any
exceptions to change behavior. If on the other hand a parser does raise
exceptions, Spicy takes slightly more liberties and allows the final program to
raise fewer exceptions. Explicit ``throw`` statements are always respected.
```

PR #2219, opened 2025-12-04
 Fix tuple assignment when coercing individual elements.

### LINKED ISSUE #2175: Functions with tuple assignment returns run multiple times
The following code calls `ret_tup` twice in the tuple assignment case, but not for the single assignment:

```
module Test;

global glob: uint8 = 0;

function ret_tup() : tuple<uint8, uint8> {
    glob++;
    return (glob, glob);
}

function f() {
    local x = 0;
    local y = 0;
    print glob;
    (x, y) = ret_tup();
    print glob;
    local tup = ret_tup();
    print glob;
}

f();
```

```
$ spicyc -j test.spicy
0
2
3
```

since `ret_tup` only appears once in the tuple assignment, it should only get called once. Calling it multiple times in generated code is a bug

the generated hilti for the `(x, y) =...` call is:

```
(x, y) = (ret_tup()[0], ret_tup()[1]);
```

rsmmr:
This adds some infrastructure to support local temporaries inside
expressions. We then use that to avoid the double evaluation, but it
may also come in handy in more places in the future. 

Closes #2175.

- **Extend HILTI group expression to support a local temporary.**
- **Add helper to create a grouping with a temporary local.**
- **Fix tuple assignment when coercing individual elements.**


hilti/toolchain/src/compiler/codegen/expressions.cc
```diff
@@ -64,7 +64,40 @@ struct Visitor : hilti::visitor::PreOrder {
...
+                                        {}, std::move(cxx_init));
+
+            block.addTmp(cxx_local);
+
+            cg->popCxxBlock();
+
+            cxx::Formatter f;
+            f.ensure_braces_for_block = false;
+            f << block;
+            result = fmt("({%s %s;})", f.str(), cxx_expr);
```
evantypanski: Do we need to rely on non-standard statement expressions here? It feels a bit weird relying on a GCC/Clang extension, I would hope the code is more portable

eg why not something like a lambda that's immediately invoked:

```suggestion
            result = fmt("([&](){%s return %s;}())", f.str(), cxx_expr);
```

rsmmr:
> ```
>     local z = (x, y) = ret_tup();
> ```

Added a commit that fixes this.

hilti/toolchain/src/compiler/codegen/expressions.cc
rsmmr: Used this because we we're using it for `BuiltInFunction` already. Back then, I had considered using a lambda for it and don't recall why I decided against. But I think you're right, it's more suitable. I'll change. 

(And turns out `BuiltInFunction` isn't actually used anywhere, will remove that.)

hilti/toolchain/src/compiler/codegen/expressions.cc
```diff
@@ -32,39 +32,46 @@ struct Visitor : hilti::visitor::PreOrder {
...
 
     void operator()(expression::Ctor* n) final { result = cg->compile(n->ctor(), lhs); }
 
-    void operator()(expression::Grouping* n) final { result = fmt("(%s)", cg->compile(n->expression(), lhs)); }
+    void operator()(expression::Grouping* n) final {
+        auto cxx_expr = cg->compile(n->expression());
+
+        if ( auto* local = n->local() ) {
+            // We generate a (non-standard) statement expression here to create
+            // the local variable, similar to `expression::BuiltInFunction`.
```
evantypanski: this comment is stale now since it's no longer non-standard and builtin function is gone

rsmmr:
So after rebasing this to `main`, turns out there an issue with the new optimizer here: introducing the new local then requires full resolving (scope building, canonical IDs). However, the new code is part of the coercion logic, which the optimizer may call without triggering the other resolver pieces (using `ResolvedExceptCoercions`). Need to look into that.

Issue #2197, opened 2025-11-11
Consider allowing `[]` to coerce to default struct


evantypanski:
@0xxon posted on Slack about how `local test: TestValue = []` seems natural to initialize some `struct` with all default values, like:

```
type TestValue = struct {
    x: int64 &default=1;
};

function f() {
    local test: TestValue = [];
}
```

That's because the [docs](https://docs.zeek.org/projects/spicy/en/latest/programming/language/types.html#struct) show they can be initialized like `local my_struct: MyStruct = [$FIELD_1 = X_1, ..., $FIELD_N = X_N]`. So if you init 0 fields, it should be empty square brackets, but that is instead a [`list`](https://github.com/zeek/spicy/blob/ce2699e2ed72e0e3082c1637ec920e30421e8d81/spicy/toolchain/src/compiler/parser/parser.yy#L1150-L1153).

This actually makes some things impossible, like passing in an empty struct ctor to a function without an intermediate variable (I think, at least):

```
module Test;

public type Value = struct {
	x: int64 &default=1;
};

function f(val: Value): int64 {
	return val.x;
}

print f([$x=1]); # Works
print f([]); # error: call does not match any function: f(<const list<bool>>)
``` 

I don't think we can change that `[]` becomes an empty list. So I would propose that an empty list coerces to the default constructor of a struct

bbannier:
The target type for coercion might not be there in all cases, e.g., the following still requires a temporary:

```ruby
type X = struct { restrict: uint8; };
global s = "A default X is %s" % []; # Impossible to tell target type.
```

I think it would be nicer if we'd allow a RHS to explicitly specify a type, e.g., with a syntax `X([])` or `X()`.

Zeek has moved away from `[..]` syntax to more explicit `X($a=.., $b=..)` which also does not have that problem, so it might make sense to align Spicy syntax with that as well (that'd probably be `X()`).

PR #2220, opened 2025-12-05
Support coercion from empty list to struct.

rsmmr:
- **Support coercion from empty list to struct.**
- **Fix ambiguous C++ constructors for generated struct.**

Closes https://github.com/zeek/spicy/issues/2197.


hilti/runtime/include/types/struct.h
```diff
@@ -18,6 +18,18 @@ struct hasParameters {};
...
+ * Tag for struct constructors receiving type parameters as its arguments, to
+ * disambiguate them from other constructors.
+ */
+struct Parameters {};
+
+/**
+ * Tag for struct constructors receiving values for field initialization, to
+ * disambiguate them from other constructors.
+ */
+struct Inits {};
```
evantypanski, line 32: I'd personally prefer these two go in a `tag` namespace `...struct_::tag`

hilti/toolchain/include/compiler/detail/codegen/codegen.h
```diff
@@ -83,10 +83,14 @@ class CodeGen {
     cxx::Block compile(hilti::Statement* s, cxx::Block* b = nullptr);
     cxx::declaration::Function compile(Declaration* decl, type::Function* ft, declaration::Linkage linkage,
                                        AttributeSet* fattrs = {}, std::optional<cxx::ID> namespace_ = {});
+
+    enum class StructCtor { Inits, Parameters };
```
evantypanski: this name is a little confusing to me and I translated it to `CtorKind` to understand it better. Does that fit its intent a bit better?

hilti/toolchain/src/compiler/codegen/codegen.cc
```diff
@@ -727,10 +742,26 @@ std::vector<cxx::Expression> CodeGen::compileCallArguments(const node::Range<Exp
...
+            case StructCtor::Inits: x.emplace_back("::hilti::rt::struct_::Inits()"); break;
+            case StructCtor::Parameters: x.emplace_back("::hilti::rt::struct_::Parameters()"); break;
+            default: util::cannotBeReached();
+        }
+    }
+
+
+    x.reserve(x.size() + args.size());
+
     x.reserve(args.size());
```
evantypanski: this should be removed

hilti/toolchain/src/compiler/codegen/codegen.cc
```diff
@@ -727,10 +742,26 @@ std::vector<cxx::Expression> CodeGen::compileCallArguments(const node::Range<Exp
...
+
+    if ( struct_ctor ) {
+        switch ( *struct_ctor ) {
+            case StructCtor::Inits: x.emplace_back("::hilti::rt::struct_::Inits()"); break;
+            case StructCtor::Parameters: x.emplace_back("::hilti::rt::struct_::Parameters()"); break;
+            default: util::cannotBeReached();
+        }
+    }
+
+
```
evantypanski:
```suggestion
```

evantypanski (dismissed):
I only have nits, but yeah the args/params distinction will always need a tag so the extra logic will be necessary regardless

PR #2221, opened 2025-12-09
Make optimizer guarantees less granular.

rsmmr:
[
Below's the commit message. I measured performance impact and it's negligible, seems like this fine granular distinction wasn't really needed anyways. 

- With LDAP, current `main` is: 1.00 ± 0.02 times faster than this branch
- With SSL, current `main` is: 1.01 ± 0.01 times faster than this branch

]

This makes the guarantees that an optimizer pass can provide less
granular: it can either say the AST is fully resolved, or it's not. If
not, the full suite of
resolver/coercion/type-unification/scope-building will be run on any
changed parts of the AST. A pass can no longer just have some of those
parts run.

This is because it's too hard to understand when exactly what type of
guarantee if ok to give. For example, in an upcoming patch, the
coercer will start adding new temporaries. That then requires scope
building and a full resolver pass afterwards, meaning doing just
coercions will never be possible anymore.

More specifically, this commit includes:

- Remove `resolver::coerce()`.

- ~~Extend `resolver::resolve()` with an additional argument saying
  whether it should iterate until no further changes are being made
  (another issue we had: the optimizer only did a single round so far,
  which may not be sufficient).~~ When re-resolving an AST during optimization, loop over a sequence
  of scope-building, type unifications, and the standard resolver
  processing.

- Remove `optimizer::Guarantees::{ResolvedExceptCoercions,
  ScopesBuilt, TypesUnified}` and replace
  `optimizer::Guarantees::FullyResolved` with
  `optimizer::Guarantees::Resolved`. The latter now encompasses scope
  building and type-unification as well.

Inside the resolver code, I'm leavings the coercion visitor (V3) split
out, even though strictly speaking that isn't necessary anymore. But
seems fine to keep that semantic distinction internally.


evantypanski (commented):
I'm happy with the change to just one resolution flag, but this change makes me pause and second guess:

> Extend resolver::resolve() with an additional argument saying
whether it should iterate until no further changes are being made

Is this because we need to re-resolve with earlier steps after introducing a new node in #2219? I noticed test failures in debug, but it's actually fine in release mode

I worry here because introducing new nodes like that may break monotonicity if we're not careful. That'll just pop up as an infinite loop/internal error, but I really wish we could either stay away from iterating here, or enforce some idea of monotonicity in the resolver. The latter is just a lot harder, and I'm not sure how the theory of it works with the multi-staged resolver..

so concretely: do you think we could instead target the particular case and re-resolve a known block when it gets coerced, rather than iterating during resolution? That makes the intuition much simpler, we add a block, then get it "up to speed" - but I guess I don't fully know what the issue this was solving is, so I can't say for sure.

rsmmr:
> Is this because we need to re-resolve with earlier steps after introducing a new node in #2219?

Yes, right, the resolver does some things like setting the canonical ID that the new declaration needs. 

> I worry here because introducing new nodes like that may break monotonicity if we're not careful.

Actually I don't think this is very different from what we're doing elsewhere. We generally iterate the resolver after making changes until it converges. What leads to the new `iterate` parameter here is mostly a quirk of how things are currently structured: standard resolving happens inside `ASTContext::processAST()`, and that's where that iteration happens normally. Now, however, we call `resolve()` from the optimizer, which isn't doing the iteration on it side. I could have added the iteration around `resolve()` there inside the optimizer instead but thought this is nicer (maybe it's not).

However, there's an additional piece that I didn't fully realize yesterday: the resolver itself depends on the other AST processing pieces as well (scope building, type unifying). What `ASTContext::processAST()` does is interleaving them all, i.e., each iteration round calls first scope build, then type unification, then resolving—whereas the optimizer currently first does only one scope building, then one type unification, and then now iterative resolve. That leads to different semantics. I'll change that to have the optimizer do an interleaved loop similar to `processAST()`, which then also will get rid of the `iterate` parameter, making things clearer perhaps. Semantically we'll then also have very similar logic in `processAST()` and inside the optimizer, so I think that should be safe.

(In an ideal world,the optimizer could just call `processAST()` instead of implementing that loop itself. However, `processAST()` does more stuff (calling plugins, executing hooks, debug output, even triggering the optimizer; [1]), so that would need a larger refactor and probably end up with some if-called-from-optimizer branching; doesn't seem nicer to me right now.

Does this address your concern or is there more?


[1] Not claiming this is all perfect. That whole structure of the various steps over the AST has some legacy aspects that are hard to change; in particular, the plugin concept complicates things quite a bit.


> so concretely: do you think we could instead target the particular case and re-resolve a known block when it gets coerced, rather than iterating during resolution? 

I think even if doing it on per-block basis we'd run into the same dependencies, so I don't see how this would change much. Unless I'm misunderstanding?

rsmmr:
> I'll change that to have the optimizer do an interleaved loop similar to processAST()

Done. @evantypanski take another look.

evantypanski (approved):
You're completely right that `processAST` also does the iteration, so adding it to the optimizer's is strictly correct

My concerns still stand, I'm getting more and more worried that we will have some weird loop between resolver passes or the optimizer and resolver since we don't guarantee monotonicity. But, that's for another time, maybe. I guess we'll hope that it gets caught in tests if it is ever introduced

> I think even if doing it on per-block basis we'd run into the same dependencies, so I don't see how this would change much. Unless I'm misunderstanding?

I just said block in reference to the tuple case, but I mean whenever a node gets added, perform the earlier resolve steps on just that node, rather than iterating the whole AST

PR #2223, opened 2025-12-11
Add validation ensuring non-void functions returns values on all paths.

### LINKED ISSUE #1992: Function declarations with missing return are not rejected

rsmmr:
Closes #1992.

~~TODO: This currently runs into a problem with our CFG computation.
During the normal build, when compiling
`tests/benchmarks/parsers.spicy`, the new check reports:~~

~~[error] /Users/robin/work/spicy/topic/spicy/runtime/tests/benchmarks/parsers.spicy:21:21-25:1: not all paths through the function Benchmark::Regex::_parse_Benchmark__Regex_stage2 return a value
[internal-error] Optimizer::_checkState: AST is not valid after optimizer pass~~

~~Printing out the CFG at that point reveals that the validator is
technically correct: there's a branch in the CFG that does not go
through a return statement. However, it's not clear where that branch
is coming from; the function's source code is correct and has
returns/throws on all paths.~~

~~Attaching the PDF and a print of the source code of the function at
the point when the validator triggers.~~

~~(Note that his triggers only during optimization, not during the
initial AST validation right after parsing.)~~

With #2226 cherry-picked, this works now. ~~However, it does come with a noticeable performance impact for larger analyzers, due to the additional CFG computations. Need to see if I can do something about that (like passing the CFGs into the optimizer, to skip the first round of computation there).~~ (Performance impact mostly avoided, see commit messages)


rsmmr:
[x.pdf](https://github.com/user-attachments/files/24100378/x.pdf)
[x.txt](https://github.com/user-attachments/files/24100390/x.txt)

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -216,6 +216,9 @@ class CFG {
      */
     std::deque<cfg::GraphNode> postorder() const;
 
+    /** Get the starting node. */
+    const cfg::GraphNode* begin() const { return &_begin; }
```
bbannier: `GraphNode` is already a cheap to copy wrapper around `Node*` so this should probably return a value (it shouldn't cost extra).

```suggestion
    cfg::GraphNode begin() const { return _begin; }
```

This would also make the uses in `hilti/toolchain/src/compiler/validator.cc` less awkward since you can then replace `n->get()->` by `n->`.

hilti/toolchain/include/compiler/detail/optimizer/optimizer.h
```diff
@@ -58,19 +58,8 @@ class ASTState {
...
-     *
-     * @param block the block to get the CFG for, which must be part of a
-     * function or module
-     * @return the CFG (which will actually be the CFG for the outermost block
-     * containing the given block, i.e., the function or module body containing
-     * it)
-     */
-    CFG* cfg(statement::Block* block);
+    /** Returns the CFG cache in use. */
+    auto* cfgs() { return _cfgs; }
```
bbannier: Probably nit: `cfgs`/`_cfgs` does not look like a great identifier to me. How about calling this e.g., `cfg_cache`? From a user's perspective they use this more as a cache to look up a CFG than a list of CFGs.

hilti/toolchain/src/compiler/validator.cc
```diff
@@ -913,6 +915,51 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
...
 };
 
+struct VisitorCFG : visitor::PreOrder, public validator::VisitorMixIn {
+    VisitorCFG(Builder* builder, cfg::Cache* cfgs) : validator::VisitorMixIn(builder), cfgs(cfgs) {}
+
+    detail::cfg::Cache* cfgs;
+
+    // Checks whether there are return or throw statements on all paths through
+    // a CFG starting at a given node.
+    bool ensureReturns(const detail::CFG& cfg, const detail::cfg::GraphNode* n, std::unordered_set<uint64_t>* seen) {
```
bbannier: This should pass `GraphNode` by value.

hilti/toolchain/src/compiler/validator.cc
```diff
@@ -913,6 +915,51 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
...
+        const auto& successors = cfg.graph().neighborsDownstream(identity);
+        if ( successors.empty() )
+            return false;
+
+        for ( const auto& s : successors ) {
+            if ( ! ensureReturns(cfg, cfg.graph().getNode(s), seen) )
+                return false;
+        }
+
+        return true;
```
bbannier, line 946: Nit: this is just a pretty verbose way to say: Return true if all successors return, or there are no successors. One could write that also as e.g.,

```suggestion
        return std::ranges::all_of(
            successors, [&](auto n) { return ensureReturns(cfg, n, seen); },
            [&](auto s) {
                const auto* n = cfg.graph().getNode(s);
                // HERE: For consistency we should check that the CFG node is present since `getNode` could also return a nullptr.
                assert(n);
                return *n;
            });
```

(whether to use projections or do the transformation from successors to `GraphNode` in the predicate is probably a matter of taste).

If we changed the function semantics to _does *not* return_ we could even get rid of the `empty` check above by switching to `std::ranges::none_of`/`! doesNotReturn` (`none_of` of an empty range is true).

bbannier (commented):
This came out nice and compact.

Since you iterate from the beginning of a CFG, did that just feel natural or did you run into any issues with iterating from the end (that way might get us to a `throw` or `return` faster)?

bbannier (dismissed):

hilti/toolchain/src/compiler/validator.cc
rsmmr: Sorry, I really don't find that easier (it's certainly not less verbose for my eyes).

bbannier: Totally fair, this is probably an acquired taste ;)

Could you still add the `assert` for the `getNode` return value?

PR #2224, opened 2025-12-11
Fix Spicy plugin for repeated hook invocations.

bbannier:


rsmmr:
We can switch to `dynamic_cast` if it doesn't change performance (unlikely, but I can never tell), but we should then abort if the cast fails, not silently return. Could also just add `assert(dynamic_cast...)` in front of all the static_casta.

PR #2227, opened 2025-12-12
Add resolving for capture groups.

### LINKED ISSUE #2222: Properly model regexp captures on the AST
We currently do not model regexp captures in Spicy properly on the AST, e.g., in code like 

https://github.com/zeek/spicy/blob/db12edcb67a8e0fe79f423d105f3b19cc97be6c4/tests/spicy/types/regexp/parse-captures-requires.spicy#L8-L15

we already inject a local variable `_captures`, but do not translate `$1` into something referencing it on the AST (instead the connection is made during codegen making use of the magic variable name). This means that AST rewrites need to be careful to not break this code. We could probably fix the CFG generation to capture that implicit dependency which could fix a [workaround in the `remove-unused-parameters` pass](https://github.com/zeek/spicy/blob/db12edcb67a8e0fe79f423d105f3b19cc97be6c4/hilti/toolchain/src/compiler/optimizer/passes/remove-unused-params.cc#L93-L97), but the name would still be magic and unchangeable and we'd still require the [workaround in the `flatten-blocks` pass](https://github.com/zeek/spicy/blob/db12edcb67a8e0fe79f423d105f3b19cc97be6c4/hilti/toolchain/src/compiler/optimizer/passes/flatten-blocks.cc#L76-L83).

Instead we should probably just put regexp captures on the AST:

- introduce a AST node for `Captures`, or directly expose the underlying `vector<byte>`
- translate references to capture groups like `$1` into indexing into captures

bbannier:
We would previously not perform proper resolving of capture groups. Instead we emitted a magic C++ identifier so this only resolved at C++ compile time. This introduced a number of downsides:

- incorrect uses of capture groups via `$1`, `$2`, ... outside of field hooks which defined the magic captures was impossible to diagnose; instead this caused failed C++ compilation

- code rewrites needed to treat captures special since their use was not correctly reflected on the AST

This patch lowers uses of capture keywords to the identifier name much earlier so it behaves like any other use of an identifier.

Closes #2222.


tests/Baseline/spicy.types.regexp.capture-without-regexp/output
```diff
@@ -0,0 +1,3 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+[error] <...>/capture-without-regexp.spicy:10:15-10:16: unknown ID '_captures'
```
evantypanski: could we special-case resolution for this to not leak an internal implementation detail? (that this is named `_captures`)

bbannier: Let me try to add a Spicy validator which captures invalid uses of captures.

PR #2228, opened 2025-12-12
Add function return propagation pass (try2)

### LINKED ISSUE #2031: Propagate constants through function returns

evantypanski:
Supersedes #2173 (see https://github.com/zeek/spicy/pull/2173#issuecomment-3647629807)

Closes #2031 

Draft to get benchmark run, then I'll clean it up and fix bugs/tests if it's worth it. and add the explanation for all of the weirdness in here

EDIT: Seems like the perf improvement will be worth it, I'll clean this up.

### DIFF
--- a/hilti/toolchain/src/compiler/optimizer/passes/peephole.cc
+++ b/hilti/toolchain/src/compiler/optimizer/passes/peephole.cc
@@ -4,6 +4,7 @@
 
 #include <hilti/ast/builder/builder.h>
 #include <hilti/base/logger.h>
+#include <hilti/compiler/detail/optimizer/optimizer.h>
 #include <hilti/compiler/detail/optimizer/pass.h>
 
 using namespace hilti;
--- a/hilti/toolchain/src/compiler/optimizer/passes/constant-propagation.cc
+++ b/hilti/toolchain/src/compiler/optimizer/passes/constant-propagation.cc
@@ -3,6 +3,7 @@
 #include <hilti/ast/builder/builder.h>
 #include <hilti/base/logger.h>
 #include <hilti/compiler/detail/cfg.h>
+#include <hilti/compiler/detail/optimizer/optimizer.h>
 #include <hilti/compiler/detail/optimizer/pass.h>
 
 using namespace hilti;
--- a/hilti/toolchain/src/compiler/optimizer/passes/dead-code-cfg.cc
+++ b/hilti/toolchain/src/compiler/optimizer/passes/dead-code-cfg.cc
@@ -3,6 +3,7 @@
 #include <hilti/ast/builder/builder.h>
 #include <hilti/base/logger.h>
 #include <hilti/compiler/detail/cfg.h>
+#include <hilti/compiler/detail/optimizer/optimizer.h>
 #include <hilti/compiler/detail/optimizer/pass.h>
 
 using namespace hilti;
--- a/hilti/toolchain/src/compiler/optimizer/passes/feature-requirements.cc
+++ b/hilti/toolchain/src/compiler/optimizer/passes/feature-requirements.cc
@@ -3,6 +3,7 @@
 #include <hilti/ast/builder/builder.h>
 #include <hilti/ast/scope-lookup.h>
 #include <hilti/base/logger.h>
+#include <hilti/compiler/detail/optimizer/optimizer.h>
 #include <hilti/compiler/detail/optimizer/pass.h>
 
 using namespace hilti;


hilti/toolchain/src/compiler/optimizer/passes/peephole.cc
```diff
@@ -161,6 +161,54 @@ struct Mutator : public optimizer::visitor::Mutator {
...
+        if ( it == statements.end() || it == statements.begin() )
+            return;
+
+        // Loop until first not-null predecessor
+        do {
+            if ( --it == statements.begin() )
+                return;
+        } while ( ! *it );
+
+        auto* sibling = *it;
```
bbannier: Like discussed offline, this needs to take data flow facts into account, we e.g., shouldn't do this rewrite if anything here involves alias values since assigning to them could have non-local side effects.

This looks like some more general case of constant propagation, maybe we could adapt the existing pass instead.

We also definitely want a dedicated test for this.

hilti/toolchain/include/compiler/detail/optimizer/collector-callers.h
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2020-now by the Zeek Project. See LICENSE for details.
+
```
bbannier, line 2:
```suggestion

#pragma once

```

hilti/toolchain/src/compiler/validator.cc
```diff
@@ -595,7 +595,7 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
         }
 
         if ( func->ftype()->result()->type()->isA<type::Void>() ) {
-            if ( n->expression() )
+            if ( n->expression() && ! n->expression()->type()->type()->isA<type::Void>() )
```
bbannier, line 598: Let's add a test for this, e.g.,

```ruby
module foo {
function void fn() {
    return fn(); // Should pass validator.
}
}
```

hilti/toolchain/src/compiler/optimizer/passes/remove-unused-params.cc
```diff
@@ -2,36 +2,15 @@
 
 #include <hilti/ast/builder/builder.h>
 #include <hilti/base/logger.h>
+#include <hilti/compiler/detail/optimizer/collector-callers.h>
 #include <hilti/compiler/detail/optimizer/pass.h>
 
-#include "compiler/detail/optimizer/optimizer.h"
```
bbannier, line 7: Let's include this properly, e.g.,

```suggestion
#include <hilti/compiler/detail/optimizer/optimizer.h>
```

We still directly use symbols from this header, e.g., `Optimizer`.

hilti/toolchain/src/compiler/optimizer/passes/propagate-function-returns.cc
```diff
@@ -0,0 +1,633 @@
...
+            if ( ! parent )
+                return;
+            auto* fn_decl = parent->tryAs<declaration::Function>();
+            if ( ! fn_decl )
+                return;
+            placement_ids = collector_placements->fn_placements[fn_decl->functionID(context())].placements;
+        }
+
+        // Make sure at least one placement is getting removed
+        if ( ! std::ranges::any_of(placement_ids, [](const std::optional<ID>& opt) { return opt.has_value(); }) )
```
bbannier, line 464: This and more uses below need 

```cpp
#include <algorithm>
```

hilti/toolchain/src/compiler/optimizer/passes/propagate-function-returns.cc
```diff
@@ -0,0 +1,633 @@
...
+            param_names[name->id()] = (*params_iter)->id();
+
+        params_iter++;
+    }
+
+    return param_names;
+}
+
+/** Gets the field connected to a given function, if any. */
+std::optional<const declaration::Field*> getField(ASTContext* ctx, declaration::Function* n) {
```
bbannier: `const declaration::Field*` is already nullable, no need to wrap it in an `optional`. Returning just a `const declaration::Field*` would also ensure that users validate that the pointer is not-null, currently they just check whether the `optional` is engaged.

Also no need to pass `n` as pointer since we already implicitly assume it is always set, let's just pass a const ref instead. When making that change we should `assert(n)` before dereferencing in `operator()(declaration::Function* n)` below.

hilti/toolchain/src/compiler/optimizer/passes/propagate-function-returns.cc
```diff
@@ -0,0 +1,633 @@
...
+        // Since we have placements, this should be guaranteed a tuple ctor.
+        if ( ! tuple_ctor )
+            logger().internalError(util::fmt("function declaration for %s without an operator", function_id),
+                                   tuple_ctor);
+
+        removeFromTupleCtor(tuple_ctor, placements.placements);
+    }
+};
+
+
```
bbannier: Nit: stray empty line.

```suggestion
```

hilti/toolchain/src/compiler/optimizer/passes/propagate-function-returns.cc
```diff
@@ -0,0 +1,633 @@
...
+    void operator()(statement::Return* n) final {
+        auto opt_enclosing_fn = hilti::detail::Optimizer::enclosingFunction(context(), n);
+        if ( ! opt_enclosing_fn || ! n->expression() )
+            return;
+
+        auto [_, function_id] = *opt_enclosing_fn;
+        if ( ! collector_placements->fn_placements.contains(function_id) )
+            return;
+
+        auto& placements = collector_placements->fn_placements.at(function_id);
```
bbannier: Nit: we do lookup the value twice here.

```suggestion
        auto it = fn_placements.find(function_id);
        if ( it == fn_placements.end() )
            return;

        auto& placements = it->second;
```

hilti/toolchain/src/compiler/optimizer/passes/propagate-function-returns.cc
```diff
@@ -0,0 +1,633 @@
...
+            if ( auto* coerced = ctor_expr->ctor()->tryAs<ctor::Coerced>() )
+                tuple_ctor = coerced->originalCtor()->tryAs<ctor::Tuple>();
+            else
+                tuple_ctor = ctor_expr->ctor()->tryAs<ctor::Tuple>();
+        }
+
+        // Since we have placements, this should be guaranteed a tuple ctor.
+        if ( ! tuple_ctor )
+            logger().internalError(util::fmt("function declaration for %s without an operator", function_id),
+                                   tuple_ctor);
```
bbannier: `tuple_ctor` is null, we should probably pass `ctor_expr`.

hilti/toolchain/src/compiler/optimizer/passes/propagate-function-returns.cc
```diff
@@ -0,0 +1,633 @@
...
+
+        auto* ctor_expr = n->expression()->tryAs<expression::Ctor>();
+        ctor::Tuple* tuple_ctor = nullptr;
+        if ( ctor_expr ) {
+            // Skip coercion
+            if ( auto* coerced = ctor_expr->ctor()->tryAs<ctor::Coerced>() )
+                tuple_ctor = coerced->originalCtor()->tryAs<ctor::Tuple>();
+            else
+                tuple_ctor = ctor_expr->ctor()->tryAs<ctor::Tuple>();
+        }
```
bbannier: Nit: This could be simplified to

```suggestion
        if ( ! ctor_expr )
            return;

        // Skip coercion
        ctor::Tuple* tuple_ctor = nullptr;
        if ( auto* coerced = ctor_expr->ctor()->tryAs<ctor::Coerced>() )
            tuple_ctor = coerced->originalCtor()->tryAs<ctor::Tuple>();
        else
            tuple_ctor = ctor_expr->ctor()->tryAs<ctor::Tuple>();
```

hilti/toolchain/src/compiler/optimizer/passes/propagate-function-returns.cc
```diff
@@ -0,0 +1,633 @@
...
+ * is only set if all uses *also* expect 'x' in index 0.
+ *
+ * Later, we can use this information to tell that we can remove 'x'
+ * from the return, since it is never changed and just returned as-is.
+ */
+struct CollectorPlacements : public optimizer::visitor::Collector {
+    CollectorPlacements(Optimizer* optimizer, const CollectorCallers* operators)
+        : optimizer::visitor::Collector(optimizer), collector_callers(operators) {}
+
+    const CollectorCallers* collector_callers;
```
bbannier: Nit: I find the (transitive) spreading of state between the different collectors hard to follow, I'd probably pass ownership to a `CollectorCallers` here.

```suggestion
    CollectorPlacements(Optimizer* optimizer, CollectorCallers&& operators)
        : optimizer::visitor::Collector(optimizer), collector_callers(std::move(operators)) {}

    CollectorCallers collector_callers;
```

hilti/toolchain/src/compiler/optimizer/passes/propagate-function-returns.cc
```diff
@@ -0,0 +1,633 @@
...
+ *   1) Change the function's return type to not include 'x'
+ *   2) Change the field's return type, if a method, to not include 'x'
+ *   3) Change all uses to not assign 'x' to the return value of the function
+ *   4) Change all returns to not return 'x'
+ */
+struct Mutator : public optimizer::visitor::Mutator {
+    Mutator(Optimizer* optimizer, CollectorPlacements* collector_placements)
+        : optimizer::visitor::Mutator(optimizer), collector_placements(collector_placements) {}
+
+    CollectorPlacements* collector_placements = nullptr;
```
bbannier: Similar nit, but I'd even rip the data out of the passed collector so we can use it directly.

```suggestion
    Mutator(Optimizer* optimizer, CollectorPlacements&& collector_placements)
        : optimizer::visitor::Mutator(optimizer),
          collector_callers(std::move(collector_placements.collector_callers)),
          fn_placements(std::move(collector_placements.fn_placements)) {}

    CollectorCallers collector_callers;
    std::map<ID, CollectorPlacements::Placements> fn_placements;
```

hilti/toolchain/src/compiler/optimizer/passes/propagate-function-returns.cc
```diff
@@ -0,0 +1,633 @@
...
+    /**
+     * Based on the uses and parameters for the function, this calculates
+     * "placements" for the parameters. That is, which uses always
+     * assign to the same tuple position. If any use assigns to a different
+     * position, then it is not included, and its position is nullopt.
+     */
+    std::vector<std::optional<ID>> calculatePlacements(const std::vector<expression::ResolvedOperator*>* uses_of_op,
+                                                       const declaration::Parameters& params) {
+        if ( ! uses_of_op || uses_of_op->empty() )
+            return {};
```
bbannier: This doesn't need to pass by pointer

```suggestion
    std::vector<std::optional<ID>> calculatePlacements(const std::vector<expression::ResolvedOperator*>& uses_of_op,
                                                       const declaration::Parameters& params) {
        if ( uses_of_op.empty() )
            return {};
```

hilti/toolchain/src/compiler/optimizer/passes/propagate-function-returns.cc
```diff
@@ -0,0 +1,633 @@
...
+                    replaceNode(tup_assign, builder()->assign(builder()->tuple(new_tup_assign_exprs), use),
+                                "removing elements from propagated tuple");
+                    break;
+            }
+        }
+    }
+
+    void removeFromTupleCtor(ctor::Tuple* ctor, std::vector<std::optional<ID>> placements) {
+        Expressions values;
+        int i = 0;
```
bbannier: No need to pass a copy. We should check that size of the tuple ctr and `placements` match.

```suggestion
    void removeFromTupleCtor(ctor::Tuple* ctor, const std::vector<std::optional<ID>>& placements) {
        Expressions values;
        int i = 0;
        assert(ctor->value().size() == placements.size());
```

hilti/toolchain/src/compiler/optimizer/passes/propagate-function-returns.cc
```diff
@@ -0,0 +1,633 @@
...
+#include <hilti/compiler/detail/optimizer/pass.h>
+
+using namespace hilti;
+using namespace hilti::detail;
+using namespace hilti::detail::optimizer;
+
+namespace {
+
+/** Makes a map from a given use's parameters to the function params. */
+std::map<ID, ID> makeParamMap(const ctor::Tuple* tup, const declaration::Parameters& params) {
```
bbannier: Also no need to pass `tup` as pointer, let's pass by const ref.

When making that change let's `assert(call_tuple_ctor)` before derferencing in `operator()(declaration::Function* n)` below.

hilti/toolchain/src/compiler/optimizer/passes/propagate-function-returns.cc
```diff
@@ -0,0 +1,633 @@
...
+            auto* tup_assign = use->parent(parent_num)->tryAs<operator_::tuple::CustomAssign>();
+
+            auto* tup_ctor = tup_assign->op0()->as<expression::Ctor>()->ctor()->as<ctor::Tuple>();
+
+            Expressions new_tup_assign_exprs;
+            if ( tup_ctor->value().size() != placement_ids.size() )
+                logger().internalError(
+                    util::fmt("function declaration for %s calculated placements incorrectly", function_id));
+
+            auto is_unplaced = [&](auto* expr) {
```
bbannier: Let's at least `assert(expr)` here.

More on the nit side, I wouldn't mind this to be inline in the use below. Also, if you wrote this as taking a `Expression*` you wouldn't need to `template ` for the `tryAs` invocation.

evantypanski:
A bit long response, but tl;dr: I actually think that since this is so targeted, it is the simplest form. We have to care about a lot more interprocedural gotchas if we stray from the strict structure here at all. I'd argue "simplifying" means "generalizing" which is difficult with our lack of serious interprocedural guarantees (ie we can't tell if expressions have side effects, we currently have no function summaries, we have no good way of creating contexts for each caller).

> I am amazed that this works given that parser state holds alias values (e.g., `begin` which is an iterator aliasing the input stream).

I don't think this is a problem. We only propagate if the function is called with *exactly* the name that it returns. This is by design to not hit this issue, so we don't need complex aliasing logic. Aliasing is incredibly difficult in interprocedural optimizations since you have to create context for every caller (or be very coarse and assume the worst). This PR gets around that by being extremely strict: it only cares about cases like `(..., NAME, ...) = call(..., NAME, ...)`

Can you think of a case where the guarantees here aren't sufficient? ie pass in an alias which will then not get updated properly. I can't, partially because of the fact that the alias will change anyway. If we allowed a reference operator as a function argument maybe, but again, it has to be *exactly* the name. But I might not be imaginative enough

> I wonder whether it wouldn't be a better idea long term to instead implement this as a potentially simpler pass which inlines constants from function bodies, i.e.,

I would definitely like to make this a more general pass. But generality is exactly what would make this go from "gross code solving a particular problem" to "bug-prone complex interprocedural optimization" that I wanted to avoid. Namely:

1) If we don't match exactly names passed in and names returned, then we have to care about side effects. That means that we have to put any uses above the call, but we also have to put uses that came *before* above the call so they happen in the correct order.
2) Each caller has a context, which will dictate what gets aliased from parameters. Naively we can assume everything aliases each other. Then we don't optimize anything, and it's back to square one. This might be okay here, depending where we want to focus, but it doesn't really simplify anything..
3) Every path must return the same thing. Here it's guaranteed for each placement. In a general solution, especially with aliasing, that gets harder.

So I agree that it's a general problem with a "simpler" solution (the initial issue also had it as a more general problem, I focused it more narrowly here), but treating it as a general problem forces us to care about more interprocedural optimization issues. I also would like a great interprocedural optimization engine, but we're just not there, and it's kind of unclear if we need that.

This targeted approach feels simpler, and is exactly the reason we don't have to care about a lot of hard problems we would otherwise face.

I don't think that we can get a solution that is simple, general, and interprocedural. The act of generalizing here would add a bunch of complexity, in my opinion.

> TBH, I still find this pretty hard to follow, e.g., the semantics of `Placements::passthrough` and `Placements::passthrough_from`.

Concretely those just end up being tail calls. So we could rename `passthrough`->`tail_callee` and `passthrough_from`->`tail_caller`? Maybe add some extra checks to make sure it actually is a tail call.

> Since these placements are index-based that also means that we constantly need to "zip" placements with other collections which I personally find hard to follow.

Yeah. I agree it's a little nasty. I think it'd be nice if we could do the collection phase then just one zip. But, the passthrough needs to know about which parameters are in which positions, which makes it... interesting.

tests/hilti/output/optimization/propagate-function-returns.hlt
bbannier, line 1: I played some more with this and the code in this PR works like advertised. I checked behavior of aliasing values (iterators into vectors in the caller), permutations of inout parameters, and branching. Maybe worth including a test which exercises a little more of the machinery, e.g.,

```ruby
function tuple<uint<8>, uint<8>> fn(uint<8> flag, inout uint<8> a, inout uint<8> b) {
    if (flag==0)
        return (a++, b);
    else if (flag == 1)
        return ((a=1), b);
    else
        return (a, b);
}

...

    (a, b) = fn(0, a, b);


```

evantypanski: Sure, I can add a test for that. The strict structure *should* guarantee that without worrying about interprocedural aliasing! If I didn't mess up, that is

hilti/toolchain/src/compiler/optimizer/passes/propagate-function-returns.cc
evantypanski: Fair, I really don't like nullable pointers so I try not to return `nullptr`, but that way is definitely more C++

hilti/toolchain/src/compiler/optimizer/passes/peephole.cc
evantypanski: I shoved it into the constant propagator, that seems logically the correct place, just the current impl wouldn't apply since it's a function call that propagates. Then it uses the CFG, just checks if there's a gen/kill of the name decl on the CFG node before.

hilti/toolchain/src/compiler/optimizer/passes/propagate-function-returns.cc
evantypanski: sure, I kept the rvalue ref out of the ctor though for more flexibility

hilti/toolchain/src/compiler/optimizer/passes/remove-unused-params.cc
evantypanski: other passes are a bit messy with this include too, I went through and added them.

hilti/toolchain/src/compiler/optimizer/passes/propagate-function-returns.cc
```diff
@@ -0,0 +1,657 @@
...
+        // Don't change public functions
+        if ( n->linkage() == declaration::Linkage::Public )
+            return;
+
+        auto all_lookups = context()->root()->scope()->lookupAll(n->fullyQualifiedID());
+
+        // Don't change signature if there's no body or multiple implementations
+        if ( ! n->function()->body() ||
+             (all_lookups.size() > 1 && n->function()->ftype()->flavor() != type::function::Flavor::Hook) )
+            return;
```
bbannier: The check for multiple bodies seems not needed since at least after validation functions have at most one body, and no hook returns a tuple (the only hook which returns anything is `__str__`, and I am not sure the handling of hooks here is correct anyway). Maybe just


```suggestion
        // Don't change signature if there's no body.
        if ( ! fn->body() )
            return;
```

bbannier: The check for multiple bodies seems not needed since at least after validation functions have at most one body, and no hook returns a tuple (the only hook which returns anything is `__str__`, and I am not sure the handling of hooks here is correct anyway). Maybe just


```suggestion
        // Don't change signature if there's no body.
        if ( ! fn->body() )
            return;
```

This assumes that we pull out

```
        auto* fn = n->function();
```
and reuse this in the body since it is not a trivial getter.

hilti/toolchain/src/compiler/optimizer/passes/propagate-function-returns.cc
```diff
@@ -0,0 +1,657 @@
...
+
+        const auto* uses_of_op = collector_callers.uses(op);
+        if ( ! uses_of_op )
+            return;
+
+        auto& placements = fn_placements.at(function_id);
+        // First, calculate tail callers. These are important because they will
+        // dictate how the placements are calculated.
+        for ( auto* use : *uses_of_op ) {
+            if ( auto* ret = use->parent()->tryAs<statement::Return>(); ret && placements.tail_callee == nullptr ) {
```
bbannier: Nit.

```suggestion
            if ( auto* ret = use->parent()->tryAs<statement::Return>(); ret && placements.tail_callee ) {
```

hilti/toolchain/src/compiler/optimizer/passes/propagate-function-returns.cc
```diff
@@ -0,0 +1,657 @@
...
+                // This function isn't a tail caller, and its use immediately
+                // returns. Maybe the use is a tail caller.
+                auto opt_enclosing_fn = hilti::detail::Optimizer::enclosingFunction(context(), use);
+                if ( ! opt_enclosing_fn )
+                    return;
+
+                auto [func, tail_caller_function_id] = *opt_enclosing_fn;
+                if ( ! fn_placements.contains(tail_caller_function_id) ||
+                     fn_placements[tail_caller_function_id].placements.size() == 0 ) {
```
bbannier: Definitely nit since we currently do not activate clang-tidy's `readability-container-size-empty`.


```suggestion
                     fn_placements[tail_caller_function_id].placements.empty() ) {
```

bbannier (dismissed):
Sorry for taking so long. I played with this trying to figure out whether there was another way to express this, but now have seen myself that this is pretty much as simple as it gets; like we discussed, on the bright side this implements a very targeted and narrow rewrite.

PR #2232, opened 2025-12-16
Fix dead code removing methods when removing vars.

### LINKED ISSUE #2231: Creating variables makes statement get removed even with `self`
The following HILTI program has different behavior with optimizations and without:

```
module Foo {

import hilti;

type Test = struct() {
    method int<64> test();
    int<64> x;
};

method int<64> Test::test() {
    self.x = 5;
    return 1;
}

function void x() {
    local Test dat = [$x = 1];
    hilti::print(dat);
    local y = dat.test();
    hilti::print(dat);
}

x();
}
```

```
etyp:/Users/etyp/src/spicy jj:(r:zzw) $ hiltic -j test.hlt -g
[$x=1]
[$x=5]
etyp:/Users/etyp/src/spicy jj:(r:vtw) $ hiltic -j test.hlt
[$x=1]
[$x=1]
```

note that this is just because `y` is never marked as used so it gets rid of the statement. If you just do `dat.test();` then it doesn't get removed.

I'm hitting this when finishing #2228, a test is failing because a method gets removed, and it gets removed despite using `self` (as a method call).

evantypanski:
Fixes #2231

Given code like this:

```
  local y = dat.test();
```

if `y` was never used, then the whole statement would get removed. This makes it so that the RHS sticks around. Then, if the RHS should also be removed, it gets removed later.

I like this solution but didn't look super hard immediately to see if there are more cases, or like I don't think the global variable case is necessary. What do you think of the approach?

### DIFF
--- a/hilti/toolchain/src/compiler/optimizer/passes/dead-code-cfg.cc
+++ b/hilti/toolchain/src/compiler/optimizer/passes/dead-code-cfg.cc
@@ -11,18 +11,29 @@ using namespace hilti::detail::optimizer;
 
 namespace {
 
+struct NodePtrOrder {
+    bool operator()(Node* lhs, Node* rhs) const {
+        if ( ! (lhs && rhs) )
+            return false;
+
+        return lhs->identity() < rhs->identity();
+    }
+};
+
+using NodePtrSet = std::set<Node*, NodePtrOrder>;
+
 // Removes dead code based on control flow analysis.
 struct Mutator : public optimizer::visitor::Mutator {
     using optimizer::visitor::Mutator::Mutator;
 
     // Returns all nodes in the CFG that are unreachable.
-    std::vector<Node*> unreachableNodes(const CFG* cfg) const {
-        std::vector<Node*> result;
+    NodePtrSet unreachableNodes(const CFG* cfg) const {
+        NodePtrSet result;
 
         for ( const auto& [id, n] : cfg->graph().nodes() ) {
             if ( n.value.get() && ! n.value->isA<cfg::MetaNode>() && n.neighbors_upstream.empty() ) {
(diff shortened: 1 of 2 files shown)


hilti/toolchain/src/compiler/optimizer/passes/dead-code-cfg.cc
bbannier, line 1: I was about to suggest something similar, also for `unreachableNodes`, but encoded as types.

WDYT about this instead?

[0001-Remove-dead-statements-and-unreachable-code-in-deter.patch](https://github.com/user-attachments/files/24193183/0001-Remove-dead-statements-and-unreachable-code-in-deter.patch)

hilti/toolchain/src/compiler/optimizer/passes/dead-code-cfg.cc
```diff
@@ -136,12 +141,20 @@ struct Mutator : public optimizer::visitor::Mutator {
...
         if ( ! (node && node->hasParent()) )
             return false;
 
-        removeNode(node, msg);
+        if ( replace_with )
+            replaceNode(node, replace_with, msg);
+        else
+            removeNode(node, msg);
+
         cfg->removeNode(node);
```
bbannier: This will be inconsistent now, and we rebuild the CFG after the pass anyway. Could you remove this line, and from the arg list and the comment?

bbannier (approved):

hilti/toolchain/src/compiler/optimizer/passes/dead-code-cfg.cc
evantypanski: Works for me. Maybe in the future we could move away from raw `Node` pointers like that in the first place

Issue #2236, opened 2025-12-17
Odd CFG for `if`/`else`.


rsmmr:
```
module Foo {

 function bool not_5(uint<64> i) {
     if ( i == 5 )
         return False;
     else
         return True;
 }

 not_5(1);

 }
```

This yields the attached CFG for `not_5`, which doesn't seem right.

[x.pdf](https://github.com/user-attachments/files/24208071/x.pdf)

PR #2238, opened 2025-12-17
Take operand constness into account in CFG construction.

bbannier:
I looked into this since I saw a `write: [i]` in the CFG in #2236. This fixes a small issue, but there is definitely more we could do here.

### DIFF
--- a/hilti/toolchain/include/ast/operator.h
+++ b/hilti/toolchain/include/ast/operator.h
@@ -348,7 +348,7 @@ public:
     /**
      * Refines the operator's signature based on the given operands. This can
      * be used to change the signature to more specific types given concreate
-     * operands. To not change anything, return an mepty optional
+     * operands. To not change anything, return an empty optional
      */
     virtual std::optional<operator_::Operands> filter(Builder* builder, const Expressions& operands) const {
         return {};


hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -807,6 +822,10 @@ struct DataflowVisitor : visitor::PreOrder {
             // Simply flows a value but does not generate or kill any.
             transfer.read.insert(decl);
 
+        else if ( auto* operator_ = node->tryAs<expression::ResolvedOperator>();
+                  operator_ && isConstOperand(*operator_, *name) )
```
bbannier: For non-const uses we still fall through to the `else` branch which also sets the node `keep`.

Ideally we'd handle all operators here, always set `read` and `write` for non-const uses. We still would need extra handling for member-like operators which could have side effects (definitely `MemberCall`, possibly `Member` and `TryMember`).

rsmmr: We should be able to look at the operator's signature: all the method-like operators have a `self` with a `Kind` that's either `In` or `InOut`. And generally, the signatures are good to look at for all operands I think: `In` is read-only, `InOut` means operator can modify the operand.

bbannier: I now look at the signature. I had to fix a few operators which defined incorrect self kind.

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -79,6 +79,37 @@ using namespace hilti;
...
+
+    auto offset = std::distance(ops_passed.begin(), it);
+
+    // Look up operand kind in signature.
+    auto signature = operator_.operator_().signature();
+    const auto& ops_defined = signature.operands->operands();
+
+    assert(offset < ops_defined.size());
+    const auto* op = ops_defined[offset];
+    assert(op);
```
rsmmr: This could be worth a helper in ResolvedOperator: given an operand, return the signature for it.

bbannier: I added a member function to `ResolvedOperator`, but am unsure whether it fits in with what is currently there (not saying we cannot add to it). WDYT?

rsmmr: Yeah, I see what you mean. Another idea would splitting it up into two methods, with one on the signature side. I just tried that and pushed a commit. Not hugely better but maybe a bit more reusable, too. See what you think, otherwise feel free to drop it and return to the original version.

bbannier: Thanks, I went with your suggestion!

Issue #2205, opened 2025-11-21
Use of __hlt as a namespace breaks some ARM64 compilers


andyraf:
A few locations in Spicy use __hlt (with two underscores) as a namespace designation, e.g.

hilti/toolchain/include/compiler/context.h:

    std::string cxx_namespace_intern = "__hlt"; /**< CXX namespace for generated internal C++ code */

This causes compiled source code to incorporate __hlt as a namespace designation.

However, some ARM64 compilers treat __hlt as a reserved keyword for the processor "halt" instruction intrinsic. When the compiler sees __hlt as a namespace, it causes a fatal compiler error.

It is suggested that some other non-conflicting namespace string be used.

PR #2245, opened 2026-01-07
Revert change to disallow leading underscores in user identifiers.

rsmmr:
In #2205, we started disallowing leading underscores in user
identifiers as we were using them internally in generated code for
private identifiers. That's a pretty drastic change, however,
affecting existing Spicy code that's now no longer valid. It's also
generally surprising to not have access to identifiers with leading
underscores.

This reverts the change and allows leading underscores again. Instead,
we internally now use a `_t_` prefix for generated IDs. We
now disallow *that* for users but don't specially mention the
restriction in the docs as that feels quite specific.

Note: The use of `_t_` is pretty arbitrary, was thinking `t` like "temporary" ...


rsmmr:
> The original motivation was to avoid using reserved identifiers in generated code, and it looks like this would allow this again. 

Can you give me an example of how it would do that?

bbannier:
> Can you give me an example of how it would do that?

See above. One could probably also come up with code which breaks runtime behavior, e.g., by overriding internal functions (change what e.g., `__cxx_atomic_is_lock_free` resolves to? 😈).

bbannier (approved):
This turned out a shockingly small change, I guess the cleanup in #2214 was still worth it.

I assume you still want this to go into zeek-8.1. I bumped the lasted dev snapshot into Zeek yesterday, but we'd need to see whether we need any adjustments in Zeek, and do another bump. @timwoj Could you hold off creating rc2 until that has landed?

PR #2247, opened 2026-01-09
Copy propagation

evantypanski:
This changes the constant propagation pass to also do "copy propagation." Essentially if we have `y = x`, we can replace any occurrence of `y` with `x`. Before, this would only be done for constants, like `y = 15`.

This also has two extra points that are essentially bonus points on existing optimizations. They make this apply slightly more often, and when it does, makes it slightly more impactful.

I found  #2258 when trying to break this with aliasing, my theory is that it's okay if that is fixed. The crux is that the dead code removal would be what causes alias issues here, not the copy propagation. Please find examples where this could break though!

I do not suspect small parsers would see performance improvements. However, two cases come to mind:

1) User code which has some unnecessary stores
2) Large functions of generated code that can remove some declarations, preventing register spilling

I don't think 1 will happen as much as I want because it turns out fields are *way worse* for alias analysis. Ideally we would also copy-propagate field accesses. Perhaps in the future!


evantypanski:
This should be ready, it should be a nice bonus to the "supportive" optimizations that will help eliminate some stores as we add more and more optimizations

hilti/toolchain/src/compiler/optimizer/passes/dead-code-cfg.cc
```diff
@@ -160,12 +160,27 @@ struct Mutator : public optimizer::visitor::Mutator {
...
+                // If and while should have their declarations removed, but
+                // they must be constant for now since replacing the
+                // declaration with the initializer may be invalid.
+                declaration::LocalVariable* init = nullptr;
+                if ( auto* if_ = stmt->tryAs<statement::If>() )
+                    init = if_->init();
+                else if ( auto* while_ = stmt->tryAs<statement::While>() )
+                    init = while_->init();
+
+                if ( init == data && init->init() && init->init()->isConstant() )
```
rsmmr: Wondering about `isConstant()`: I don't think it's the type being constant that we want to check, but rather have a value that doesn't have side effects, right? We have a new method for that: `&& ... ! state()->cfgCache()->mayHaveSideEffects(init->init())`. (That method is a somewhat pessimistic approximation right now, but will get better in the future.)

tests/hilti/optimization/unused-while-if-init.hlt
```diff
@@ -0,0 +1,35 @@
...
+
+import hilti;
+
+function uint<64> potential_side_effect() { return 1; }
+
+function void fn(uint<64> param) {
+    # CHECK: if ( param == 10 )
+    if ( local uint<64> x = 5; param == 10 ) {
+        hilti::print(param);
+    }
```
rsmmr, line 16: (Just to acknowledge it: this use of `CHECK` I find quite readable.)

spicy/lib/spicy_rt.hlt
```diff
@@ -90,7 +90,7 @@ declare public optional<iterator<stream>> unit_find(iterator<stream> begin_, ite
 
 declare public void backtrack() &cxxname="spicy::rt::detail::backtrack" &have_prototype;
 
-declare public void initializeParsedUnit(inout ParsedUnit punit, any unit) &cxxname="spicy::rt::ParsedUnit::initialize" &have_prototype;
+declare public void initializeParsedUnit(inout ParsedUnit punit, inout any unit) &cxxname="spicy::rt::ParsedUnit::initialize" &have_prototype;
```
rsmmr: Can you elaborate what's going on here? Semantically this *is* an `in` parameter, so if something breaks without the `inout`, that would seem to be a bug somewhere.

hilti/toolchain/src/compiler/optimizer/passes/constant-propagation.cc
```diff
@@ -14,14 +14,14 @@ namespace {
 
 struct ConstantValue {
     Expression* expr = nullptr;
-    bool not_a_constant = false; // NAC
+    bool is_unknown = false; // True if not a constant or a direct name access
```
rsmmr: I find `is_unknown` quite unintuitive because it's not clear what "known" refers to.  "is_handled"? "is_of_interest" (both being the negated value of `is_unknown` I gues).

hilti/toolchain/src/compiler/optimizer/passes/constant-propagation.cc
```diff
@@ -14,14 +14,14 @@ namespace {
 
 struct ConstantValue {
```
rsmmr: `ConstantValue` doesn't sound right anymore, does it? 

Also, should the whole pass rename to `copy-propagation` now?

hilti/toolchain/src/compiler/optimizer/passes/constant-propagation.cc
```diff
@@ -32,27 +32,47 @@ struct AnalysisResult {
...
-// Marks all children that are names as not a constant in the given map.
-// This is used by function calls, since they have deeply nested names
-// that should all be marked NAC.
-struct NameNACer : optimizer::visitor::Collector {
-    NameNACer(Optimizer* optimizer, ConstantMap* constants)
+// Marks any dependencies on a modified decl as also unknown. This is necessary
+// with copy propagation if there is an assignment to the to-be-propagated value
+// between the initial copy and use (like if storing an old value then modifying
+// the new one).
+void killDependencies(const Declaration* modified_decl, ConstantMap& constants) {
```
rsmmr: Coding style nit: use pointers for out parameters, not references.

hilti/toolchain/src/compiler/optimizer/passes/dead-code-cfg.cc
rsmmr: More generally, would this notion of not-having-side-effects a more effective way of identifying expressions to propagate?

evantypanski: Ah I missed the side effect change there, will fix it

I don't think it should for propagation, though. A function call may not have side effects, then if we propagate it 5 times we just slowed the program down, from what I understand

spicy/lib/spicy_rt.hlt
evantypanski: Good catch, so here is the diff:

```
<     spicy_rt::initializeParsedUnit((*_t_gunit), default<Foo>());
---
>     spicy_rt::initializeParsedUnit((*_t_gunit), _t_unit);
```

I should have caught that the function itself wouldn't be manipulating the value, it's assigning the reference to a field within `punit`, then later changes the original.

References need a bit more care when propagating, instead

I think that hurts other optimizations I wanted to do, but it can be added selectively (eg never propagate references in function calls since it might add an alias)

hilti/toolchain/src/compiler/optimizer/passes/constant-propagation.cc
evantypanski: I spent a while trying to think of a good name, I've seen `available` (or variations thereof) used (ie it's available to propagate). Regardless these seem like they're flipped now, so I can swap that and naming would be easy.

Not-a-constant/NAC is a common term so I just adjusted that, but flipping it makes sense now

evantypanski: I went with `is_valid` but feel free to push more, I haven't found any names that make me particularly happy. `valid` almost seems like it's a validation, though...

hilti/toolchain/src/compiler/optimizer/passes/value-propagation.cc
```diff
@@ -75,29 +103,48 @@ struct TransferVisitor : optimizer::visitor::Collector {
...
+        const auto& op = static_cast<const struct_::MemberCall&>(n->operator_());
+        auto* fdecl = op.declaration();
+        auto* ft = fdecl->type()->type()->as<type::Function>();
+        auto args = n->op2()->as<expression::Ctor>()->ctor()->as<ctor::Tuple>()->value();
+        assert(args.size() == ft->parameters().size());
+
+        // Mark any arguments in inout parameters as invalid
+        for ( const auto [i, operand] : util::enumerate(ft->parameters()) ) {
+            if ( operand->kind() == hilti::parameter::Kind::InOut || operand->type()->type()->isAliasingType() ||
+                 operand->type()->type()->isA<type::Any>() )
```
evantypanski: perhaps `Any` should be an aliasing type, since it can be? I'm actually unsure. This makes `initializeParsedUnit` not need its signature changed, but its parameter is `any`. So we cannot propagate into anything which has an `any` type.

hilti/toolchain/src/compiler/optimizer/passes/dead-code-cfg.cc
```diff
@@ -160,12 +160,27 @@ struct Mutator : public optimizer::visitor::Mutator {
...
-            if ( auto* local = data->tryAs<declaration::LocalVariable>(); local && local->init() )
-                replace_with = builder()->statementExpression(local->init());
+                    // Local declarations should keep the RHS
+                    if ( auto* local = data->tryAs<declaration::LocalVariable>(); local && local->init() )
+                        replace_with = builder()->statementExpression(local->init());
+                }
+
+                // If and while should have their declarations removed, but
+                // they must be constant for now since replacing the
+                // declaration with the initializer may be invalid.
```
rsmmr, line 174:
```suggestion
                // If and while should have their declarations removed, but
                // they must they must not have side effects since replacing the
                // declaration with the initializer may be invalid.
```

hilti/toolchain/src/compiler/optimizer/passes/value-propagation.cc
```diff
@@ -12,61 +12,89 @@ using namespace hilti::detail::optimizer;
 
 namespace {
 
-struct ConstantValue {
+struct ReachingValue {
     Expression* expr = nullptr;
-    bool not_a_constant = false; // NAC
+    bool is_valid = false; // True if it may be propagated
```
rsmmr: How about simply `bool propagate` ?

hilti/toolchain/src/compiler/optimizer/passes/value-propagation.cc
rsmmr: Yeah, making `Any` an aliasing type sounds right to me.

hilti/toolchain/src/compiler/optimizer/passes/value-propagation.cc
```diff
@@ -75,29 +103,48 @@ struct TransferVisitor : optimizer::visitor::Collector {
...
+        auto* ft = fdecl->function()->type()->type()->as<type::Function>();
+        auto args = n->op1()->as<expression::Ctor>()->ctor()->as<ctor::Tuple>()->value();
+        assert(args.size() == ft->parameters().size());
+
+        // Mark any arguments in inout parameters and references as invalid
+        for ( const auto [i, operand] : util::enumerate(ft->parameters()) ) {
+            if ( operand->kind() == hilti::parameter::Kind::InOut || operand->type()->type()->isAliasingType() ||
+                 operand->type()->type()->isA<type::Any>() )
+                invalidator.run(args[i]);
+        }
```
rsmmr: The is very similar to the code for `MemberCall`. Can you factor it out into a joint helper?

evantypanski: I thought about it at first but decided against it, since the initial work is deceptively different. It'll just be the loop

but the loop is probably sufficiently complex now that it warrants it

PR #2249, opened 2026-01-15
Add new optimizer pass removing unused struct variables.

rsmmr:
This PR introduces a new optimizer pass that removes unused struct fields. It
also establishes a framework for optimizer passes that can modify the public
C++ API of generated code when helpful, meaning they can be more aggressive
about what they change.

Zeek-side PR in https://github.com/zeek/zeek/pull/5213.

This includes:

- New command line flags `--[no-]strict-public-api` allowing the optimizer
  (not) to modify the external C++ API of generated code. When disabled (the
  default for release builds), aggressive optimizations like field removal are
  permitted. When enabled (the default for debug builds), the public API of
  generated code remains stable for external interfacing. This framework will
  enable other optimizer passes to make similar trade-offs in the future.

- A new `export <TYPE_ID>;` declaration that prohibits any field-level changes
  to a unit's or struct's field by the optimizer. This is essentially
  ``--strict-public-api`` on a per-type basis. Internally, this is represented
  through a new linkage, `export`, that also implicitly makes the type
  `public`. (The reason why there's a whole new declaration, rather than just a
  new linakge during the type's own declaration, is the Zeek integration: The
  generated glue code needs to be able to `export` types that are already
  existing, to guarantee that no changes are made to types passed to Zeek that
  would confuse the Zeek side.)

- A new optimizer pass that eliminates struct fields that are never read or
  written. With strict public API off, this also changes how structs are
  printed: if fields have been removed, they now show as `(optimized out)`.

- New environment variable `HILTI_DISABLE_OPTIMIZER_PASSES` that overrides the
  command line flag, and any defaults.

- All tests that were previously building Spicy or HILTI code in release mode, but depend
  on the output of all fields being visible, switch over to using `-d` now.

- Various smaller commits preparing and adapting for these changes.

~~TODO: Documentation updates are still missing. I'll work on them once the PR is stable.~~


tests/Baseline/hilti.optimization.unused-struct-fields/output
```diff
@@ -0,0 +1,67 @@
...
+}
+
+public function void member_non_const(inout X x) {
+    assert "" == "";
+}
+
+assert "" == "";
+assert True;
+assert "" == "";
+(tmp) = (42);
```
bbannier: Removing the call to `f()` here seems unsafe to me since the machinery inspecting side effects does not take into account that `f()` could internally cause e.g., a write to a global. I wouldn't expect this to be rare for user-defined functions either.

hilti/toolchain/src/compiler/optimizer/passes/remove-unused-fields.cc
```diff
@@ -0,0 +1,387 @@
+// Copyright (c) 2020-now by the Zeek Project. See LICENSE for details.
+
+// TODO:
+//   (*self).fxx = default<Foo>();
+//   (_t_cur, _t_lah, _t_lahe, _t_error) = (*(*self).fxx)._t_parse_stage1(_t_data, _t_cur, _t_trim, _t_lah, _t_lahe,
+//   _t_error);
+//
+//   Should we change this to always work on a stack variable first so that we can tag this as a write that can be
+//   removed if unused?
```
bbannier, line 9: This sounds like a good idea.

hilti/toolchain/src/compiler/cxx/elements.cc
```diff
@@ -396,7 +403,8 @@ std::string cxx::type::Struct::str() const {
             // explicit "init" expression. Those that do will be initialized
             // through our constructors.
             cxx::Expression init = x->init ? "" : "{}";
-            return fmt("%s%s;", fmtDeclaration(x->id, x->type, x->args, x->linkage, {}), init);
+            return fmt("%s%s%s;", (x->emitted ? "" : "// "), fmtDeclaration(x->id, x->type, x->args, x->linkage, {}),
```
bbannier, line 406: We don't know that we'll end a line here so `//` seems fragile. Maybe instead

```suggestion
            auto decl = fmtDeclaration(x->id, x->type, x->args, x->linkage, {});
            if ( ! x->emitted )
                return fmt("/* %s%s; */", decl, init);
            return fmt("%s%s;", decl, init);
```

hilti/toolchain/include/base/util.h
```diff
@@ -405,8 +405,7 @@ std::vector<T> concat(std::vector<T> v1, const std::vector<T>& v2) {
 /** Appends a vector to another one. */
 template<typename T, std::ranges::input_range R>
 std::vector<T>& append(std::vector<T>& v1, R&& v2) {
-    v1.reserve(v1.size() + v2.size());
```
bbannier, line 408: Was this dropped accidentially? Given that this a base util it seems to make sense to try to be more optimal.

hilti/toolchain/include/ast/node.h
```diff
@@ -636,6 +647,21 @@ class Node {
...
+     * @param idx index of child to clear
+     * @return the old child that was cleared, which will be detached from the AST
+     */
+    Node* clearChild(size_t idx) {
+        auto* old = _children[idx];
+        old->_parent = nullptr;
+        old->release();
+        _children[idx] = nullptr;
+        return old;
+    }
```
bbannier: This minimally needs a check `idx < _children.size()`.

```suggestion
    Node* clearChild(size_t idx) {
        if ( idx >= _children.size() )
            return nullptr;

        auto* old = std::exchange(_children[idx], nullptr);
        old->_parent = nullptr;
        old->release();

        return old;
    }
```

hilti/toolchain/include/ast/ctors/tuple.h
```diff
@@ -18,6 +17,34 @@ class Tuple : public Ctor {
...
+    }
+
+    /**
+     * Removes an element from the tuple.
+     *
+     * @param i index of the element to remove
+     * @return the removed expression, now detached from the AST
+     */
+    Expression* removeElement(int i) {
+        assert(i >= 0 && std::cmp_less(i, children().size() - 1));
```
bbannier, line 43: This either needs an `assert(! children().empty())`, or a rewrite of the algorithm so it doesn't underflow if there are not children.

```suggestion
        assert(! children().empty()); // Alternatively rework below to not underflow.
        assert(i >= 0 && std::cmp_less(i, children().size() - 1));
```

hilti/runtime/include/types/bitfield.h
```diff
@@ -113,6 +113,30 @@ inline std::string render(const hilti::rt::Optional<Bitfield<Ts...>>& x, const h
...
+    Bitfield<Ts...> empty(type_info);
+    type_info::Value bitfield(&empty, type_info);
+    for ( const auto& b : type_info->bitfield->bits() ) {
+        if ( ! out.empty() )
+            out += ", ";
+
+        out += fmt("$%s=(optimized out)", b.name);
+    }
+
+    return out;
```
bbannier: Is this just this?

```suggestion
    return join(type_info->bitfield->bits() |
                    std::views::transform([](auto&& b) { return fmt("$%s=(optimized out)", b.name); }),
                ", ");
```

hilti/toolchain/include/ast/builder/node-factory.h
```diff
@@ -280,10 +283,16 @@ class NodeFactory {
         return hilti::expression::Ctor::create(context(), ctor, std::move(meta));
     }
     auto expressionGrouping(Expression* expr, Meta meta = {}) {
-        return hilti::expression::Grouping::create(context(), expr, std::move(meta));
+        return hilti::expression::Grouping::create(context(), {expr}, std::move(meta));
+    }
+    auto expressionGrouping(Expressions&& exprs, Meta meta = {}) {
```
bbannier: This is an unusual overload since we force an r-value even though a copy would also be fine (the type is copyable).

I'd suggest to just accept `Expressions expr` here, but still keep the `move` below. This would leave using `move` to the caller like for most other functions here.

hilti/toolchain/include/ast/builder/node-factory.h
```diff
@@ -280,10 +283,16 @@ class NodeFactory {
...
+        return hilti::expression::Grouping::create(context(), std::move(exprs), std::move(meta));
     }
     auto expressionGrouping(declaration::LocalVariable* local, Expression* expr, Meta meta = {}) {
-        return hilti::expression::Grouping::create(context(), local, expr, std::move(meta));
+        return hilti::expression::Grouping::create(context(), local, {expr}, std::move(meta));
+    }
+    auto expressionGrouping(declaration::LocalVariable* local, Expressions&& exprs, Meta meta = {}) {
```
bbannier: Same.

hilti/toolchain/include/ast/expressions/grouping.h
```diff
@@ -11,25 +11,39 @@
...
+
+    QualifiedType* type() const final {
+        if ( auto* last = child<Expression>(-1) )
+            return last->type();
+        else
+            return nullptr;
+    }
 
-    void setExpression(ASTContext* ctx, Expression* expr) { setChild(ctx, 0, expr); }
+    void setExpressions(ASTContext* ctx, Expressions&& exprs) {
```
bbannier: Same.

hilti/toolchain/include/ast/expressions/grouping.h
```diff
@@ -11,25 +11,39 @@
...
+        removeChildren(1, {});
+        addChildren(ctx, std::move(exprs));
+    }
 
-    static auto create(ASTContext* ctx, Expression* expr, Meta meta = {}) {
-        return ctx->make<Grouping>(ctx, {expr, nullptr}, std::move(meta));
+    static auto create(ASTContext* ctx, Expressions&& exprs, Meta meta = {}) {
```
bbannier: Same.

hilti/toolchain/include/ast/expressions/grouping.h
```diff
@@ -11,25 +11,39 @@
...
+        Nodes nodes = {nullptr};
+        nodes.insert(nodes.end(), exprs.begin(), exprs.end());
+        return ctx->make<Grouping>(ctx, std::move(nodes), std::move(meta));
     }
 
-    static auto create(ASTContext* ctx, declaration::LocalVariable* local, Expression* expr, Meta meta = {}) {
-        return ctx->make<Grouping>(ctx, {expr, local}, std::move(meta));
+    static auto create(ASTContext* ctx, declaration::LocalVariable* local, Expressions&& exprs, Meta meta = {}) {
```
bbannier: Same.

hilti/toolchain/include/ast/builder/builder.h
```diff
@@ -440,6 +442,7 @@ class Builder : public builder::NodeFactory {
 
     auto grouping(Expression* e, const Meta& m = Meta()) { return expressionGrouping(e, m); }
 
+    auto grouping(Expressions&& exprs, const Meta& m = Meta()) { return expressionGrouping(std::move(exprs), m); }
```
bbannier: Same.

hilti/toolchain/include/ast/builder/builder.h
```diff
@@ -75,6 +75,8 @@ class Builder : public builder::NodeFactory {
 
     //////// Declarations
 
+    auto export_(ID id, const Meta& m = Meta()) { return declarationExport(std::move(id), m); }
```
bbannier, line 78: Nit: most functions here pass `m` by const ref for historic reasons, and we then copy them somewhere below in the callstack.

```suggestion
    auto export_(ID id, Meta m = Meta()) { return declarationExport(std::move(id), std::move(m)); }
```

hilti/toolchain/src/compiler/driver.cc
```diff
@@ -133,6 +137,10 @@ void Driver::usage() {
            "produced HLTO file. Can be given multiple times.\n"
            "       --skip-standard-imports      Do not automatically import standard library modules (for debugging "
            "only).\n"
+           "       --strict-public-api          Skip optimizations that change the public C++ API of generated code.  "
```
bbannier, line 140: We also have a number of files `doc/examples/_usage-*.output` which should be updated.

Same elsewhere where these flags were added.

hilti/toolchain/src/compiler/optimizer/passes/remove-unused-fields.cc
```diff
@@ -0,0 +1,387 @@
...
+        type::Struct* struct_ = nullptr;
+        std::vector<Node*> reads;
+        std::vector<Node*> writes;
+        std::vector<Node*> unsets;
+    };
+
+    std::map<ID, Field> fields;
+
+    void done() override {
+        if ( ! logger().isEnabled(logging::debug::OptimizerPasses) )
```
bbannier: Nit: Can we put the rest of the function into a `if ( logger().isEnabled(logging::debug::OptimizerPasses) )` block? That way things should work as expected if we naively add non-logging code here later.

hilti/toolchain/src/compiler/optimizer/passes/remove-unused-fields.cc
```diff
@@ -0,0 +1,387 @@
...
+        for ( const auto& [id, field] : fields ) {
+            assert(field.decl && field.struct_);
+            HILTI_DEBUG(logging::debug::OptimizerPasses,
+                        util::fmt("    %s  #reads=%zu #writes=%zu #unsets=%zu", field.decl->fullyQualifiedID(),
+                                  field.reads.size(), field.writes.size(), field.unsets.size()));
+        }
+    }
+
+    // Returns true if a field should be considered for removal.
+    bool considerField(declaration::Field* field) {
```
bbannier: Let's either make this `const` or move it to a standalone function.

```suggestion
    bool considerField(declaration::Field* field) const {
```

hilti/toolchain/src/compiler/validator.cc
```diff
@@ -300,6 +301,22 @@ struct VisitorPost : visitor::PreOrder, public validator::VisitorMixIn {
         hilti::visitor::visit(VisitExpressions(this), n);
     }
 
+    void operator()(declaration::Export* n) final {
+        if ( ! n->parent() || ! n->parent()->isA<declaration::Module>() ) {
+            error("export declaration can be used only at module scope", n);
+            return;
+        }
```
bbannier, line 308: We have no tests for this, but I wonder whether this isn't impossible by (parser) construction.

If we keep this, I don't think it needs an early `return` which would keep us from emitting diagnostics for the remaining checks.

spicy/toolchain/bin/spicy-dump/printer-json.cc
```diff
@@ -103,27 +103,36 @@ nlohmann::json JSONPrinter::convert(const hilti::rt::type_info::Value& v) {
...
-                    for ( const auto& [b, val] : f.type->bitfield->iterate(y) )
-                        j[b.name] = convert(val);
+                    if ( f.isEmitted() && y ) {
+                        for ( const auto& [b, val] : f.type->bitfield->iterate(y) )
+                            j[b.name] = convert(val);
+                    }
+                    else {
+                        for ( const auto& b : f.type->bitfield->bits() )
+                            j[b.name] =
+                                json(); // TODO: This the right choice to use? Skip non-emitted fields altogether?
```
bbannier: I think so.

```suggestion
                            j[b.name] = {};
```

spicy/toolchain/bin/spicy-dump/printer-json.cc
```diff
@@ -103,27 +103,36 @@ nlohmann::json JSONPrinter::convert(const hilti::rt::type_info::Value& v) {
...
+                    }
 
                     continue;
                 }
 
-                j[f.name] = convert(y);
+                if ( f.isEmitted() && y )
+                    j[f.name] = convert(y);
+                else
+                    j[f.name] = json(); // TODO: This the right choice to use? Skip non-emitted fields altogether?
```
bbannier: Same.

spicy/toolchain/src/compiler/codegen/unit-builder.cc
```diff
@@ -58,7 +61,8 @@ struct FieldBuilder : public visitor::PreOrder {
             // This field will never make it into the C++ struct. We still
             // carry it around though as that makes type inference easier at
             // times, and also can improve error messages.
-            attrs->add(context(), builder()->attribute(hilti::attribute::kind::NoEmit));
+            attrs->add(context(),
+                       builder()->attribute(hilti::attribute::kind::NoEmit, builder()->string("private", true)));
```
bbannier: Nit: we tried to get rid of magic strings for attributes, but now introduce `private` and `optimized`. Could you introduce compiler globals for that?

Nit: we have a dedicated builder function for string literals `stringLiteral`, this would fit the bill.

tests/hilti/optimization/unused-struct-fields.hlt
bbannier, line 1: This test does a lot of stuff. We have tried to move away from that elsewhere, so I wonder whether we could break it up into multiple tests. That would also make it simpler to use `FileCheck` to validate specific behavior instead of baselining potentially unrelated information.

hilti/toolchain/src/compiler/printer.cc
```diff
@@ -335,6 +338,11 @@ struct Printer : visitor::PreOrder {
         _out.endLine();
     }
 
+    void operator()(declaration::Export* n) final {
+        _out << "export " << n->id() << ';';
```
bbannier, line 342: Nit: We are duplicating the `export` string here, maybe


```suggestion
        _out << n->displayName() << ' ' << n->id() << ';';
```

hilti/toolchain/src/compiler/printer.cc
```diff
@@ -492,7 +500,7 @@ struct Printer : visitor::PreOrder {
         if ( auto* local = n->local() )
             _out << "{" << local << ";} ";
 
-        _out << n->expression() << ')';
+        _out << std::make_pair(n->expressions(), ", ") << ')';
```
bbannier, line 503: Absolute nit, but maybe FYI: with C++17 CTAD there's much less need for `std::make_pair`, and this could be written as


```suggestion
        _out << std::pair(n->expressions(), ", ") << ')';
```

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -262,6 +262,7 @@ class CFG {
         return r;
     }
 
+    ASTContext* _context;
```
bbannier: Let's be paranoid and always initialize `_context`.


```suggestion
    ASTContext* _context = nullptr;
```

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -337,6 +373,8 @@ class Cache {
     void checkValidity() const;
 
 private:
+    ASTContext* _context;
```
bbannier:
```suggestion
    ASTContext* _context = nullptr;
```

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -1145,24 +1146,25 @@ static std::string dataflowDot(const hilti::Statement& stmt) {
 
 // Helper class to print CFGs to a debug stream.
 class PrintCfgVisitor : public visitor::PreOrder {
+    ASTContext* _context;
```
bbannier:
```suggestion
    ASTContext* _context = nullptr;
```

hilti/toolchain/include/compiler/detail/cfg.h
```diff
@@ -290,10 +298,38 @@ class Cache {
...
+     * available
+     */
+    const cfg::Transfer* dataflow(const Expression* expr);
+
+    /**
+     * Returns true if an expression may have side effects. This includes
+     * situations where we cannot prove that it doesn't.
+     *
+     * TODO: This relies partially on `dataflow()`, so comes with the same
+     * caveat about expression-level dataflows  for now.
```
bbannier:
```suggestion
     * caveat about expression-level dataflows for now.
```

hilti/toolchain/include/ast/declaration.h
```diff
@@ -20,15 +20,18 @@ enum class Linkage {
     PreInit, /// executes automatically at load time, even before the runtime library is fully set up
     Struct,  /// method inside a method
     Private, /// accessible only locally
-    Public,  /// accessible across modules
+    Public,  /// accessible across modules (note: prefer to test for this with `isPublic()` rather than direct
+             /// comparison)
+    Export,  /// accessible across modules and guaranteed not to be modified by the optimizer
```
bbannier: Nit: Let's keep this list sorted.

hilti/toolchain/include/ast/declarations/field.h
```diff
@@ -59,6 +59,24 @@ class Field : public Declaration {
     auto isStatic() const { return attributes()->find(hilti::attribute::kind::Static) != nullptr; }
     auto isNoEmit() const { return attributes()->find(hilti::attribute::kind::NoEmit) != nullptr; }
 
+    /**
+     * Returns true if the field has a `&no-emit` attribute with a string
+     * value of `private`.
+     */
+    auto isNoEmitPrivate() const {
+        auto* attrs = attributes()->find(hilti::attribute::kind::NoEmit);
+        return attrs && *attrs->valueAsString() == "private";
```
bbannier, line 68: Nit.


```suggestion
        auto* attr = attributes()->find(hilti::attribute::kind::NoEmit);
        return attr && *attr->valueAsString() == "private";
```

Ideally we'd refer to a constant for `"private"`.

hilti/toolchain/include/ast/declarations/field.h
```diff
@@ -59,6 +59,24 @@ class Field : public Declaration {
...
+    }
+
+    /**
+     * Returns true if the field has a `&no-emit` attribute with a string
+     * value of `optimized`.
+     */
+    auto isNoEmitOptimized() const {
+        auto* attrs = attributes()->find(hilti::attribute::kind::NoEmit);
+        return attrs && *attrs->valueAsString() == "optimized";
```
bbannier, line 77: Nit.


```suggestion
        auto* attr = attributes()->find(hilti::attribute::kind::NoEmit);
        return attr && *attr->valueAsString() == "optimized";
```

Same comment re: constant.

hilti/toolchain/src/compiler/optimizer/pass.cc
```diff
@@ -73,7 +73,19 @@ void optimizer::visitor::Mutator::recordChange(const Node* old, const std::strin
     hilti::visitor::MutatingVisitorBase::recordChange(old, msg);
 }
 
+optimizer::Registry::Registry() {
+    const auto& disabled = rt::getenv("HILTI_DISABLE_OPTIMIZER_PASSES");
+    if ( disabled && ! disabled->empty() ) {
```
bbannier: Nit.

```suggestion
    if ( auto disabled = rt::getenv("HILTI_DISABLE_OPTIMIZER_PASSES"); disabled && ! disabled->empty() ) {
```

doc/toolchain.rst
bbannier, line 1: Attaching comment here since it relates to docs: we probably want to document

- `export` in statement listing
- reference `export` from custom host applications docs

hilti/toolchain/src/compiler/optimizer/passes/remove-unused-fields.cc
rsmmr: This is stale, it's actually done by the PR (to the degree possible)

hilti/toolchain/src/compiler/cxx/elements.cc
rsmmr: > We don't know that we'll end a line 

Pretty sure we do know this here, the whole type generation is line based.

hilti/toolchain/include/base/util.h
rsmmr: No, on purpose: it now led to compiler errors that were beyond my skills of fixing.

hilti/toolchain/include/ast/node.h
rsmmr: I'll go with just the range check. I'll also update to support a negative index as other methods now do, for consistency.

hilti/toolchain/include/ast/declaration.h
rsmmr: Ack, though it wasn't sorted before either.

hilti/toolchain/include/base/util.h
bbannier: Could you share the error? This should be totally safe, and if it causes errors we might have more serious issues elsewhere.

hilti/toolchain/include/ast/declarations/field.h
rsmmr: That's why it's a method, so that it's centralized.

hilti/toolchain/include/ast/declarations/field.h
rsmmr: Same response.

hilti/toolchain/include/ast/ctors/tuple.h
rsmmr: I'll add the assert but note that children not being empty is a general assumption across the class.

hilti/runtime/include/types/bitfield.h
rsmmr: Yes, probably, it's adapted from other places that all use the same pattern. Let me see if I can clean those all up.

hilti/toolchain/include/ast/builder/builder.h
rsmmr: prefer the historic consistency

hilti/toolchain/src/compiler/optimizer/passes/remove-unused-fields.cc
rsmmr: Sure ...

hilti/toolchain/src/compiler/driver.cc
rsmmr: Will come with the doc changes.

hilti/toolchain/src/compiler/validator.cc
rsmmr: Just a defensive check. The early return is because the other checks won't make sense (resolve) anyways.

spicy/toolchain/src/compiler/codegen/unit-builder.cc
rsmmr: The existing change was for the attributes' kinds. For checking the values here, I moved the magic strings into methods to have them central there. For adding the values, the validator will catch unexpected values. I don't really see another place where it would be more natural to put these.

> stringLiteral

Ack.

hilti/toolchain/src/compiler/printer.cc
rsmmr: The printer generally hardcodes most of these things, on purpose: the actual source code representation doesn't necessarily need to match how we refer to pieces elsewhere.

hilti/toolchain/src/compiler/printer.cc
rsmmr:
```
$ git grep std::make_pair | wc -l
     118
```

I'll leave as is. :-)

tests/hilti/optimization/unused-struct-fields.hlt
rsmmr: > We have tried to move away from that elsewhere?

Have we really? I just looked and I don't think the various optimizer tests generally break things down to testing all their individual cases separately. It's a lot of work both to create and to maintain. I really prefer just a single test source file here covering multiple cases; plus, it also tests things in combination with each other.

tests/Baseline/hilti.optimization.unused-struct-fields/output
rsmmr: Good catch, this was a bug. When checking the output I didn't realize we aren't treating `f()` as side-effect free.

PR #2253, opened 2026-01-21
Implement optimizer pass to `move` last uses.

### LINKED ISSUE #1960: Add optimizer pass to avoid copies of variables no longer used
Most of them time when generating code we do not know whether a HILTI value will be used later and consequentially do not use `std::move` when emitting C++ code. This means that we might that we could e.g., needlessly create copies when calling functions which parameters pass by `copy`, or create copies when storing results in another variable.

The work in #1690 should gives us control flow information to decide whether a reference to a variable (a use of a `hilti::ID` referring to a variable) is the last use. If we persisted that information with e.g., the expression we could make use of it during C++ codegen.

evantypanski:
This adds a copy-last-use optimization pass. We don't have many cases when generating parsers that would hit this, but user code certainly might.

I realized making this that C++ is pretty lax with expression ordering. For example, function call arguments can be evaluated in any order. So, we can't move the last argument of a call if it's used twice. That limits a bit of what we can do, but it's not like that would happen very much.

Took a bit of time to make this due to hitting a couple other bugs. I included the CFG one here - basically, switch statements are just large `if -> else if` but they were modelled as parallel paths. That would break liveness analysis, so each expression would say that the variable can be moved after, even if it's used in other case expressions.

I don't believe I used LLMs much when initially making this. I used claude to remember what the heck my mess of code was doing when I picked it back up, then had it review parts (it suggested the unordered set for existence in the worklist to avoid an extra `N` loops each worklist element ¯\\\_(ツ)_/¯)

Closes #1960


evantypanski:
I'm reviving this; not quite done but I think it could be simple enough that it's worth it enough for user code

evantypanski:
gonna do another pass tomorrow before requesting review, but I think this is good enough to be marked ready

evantypanski:
just a heads up in the PR: I chatted with @bbannier about this. Two points:

1) The `isTriviallyCopyable` helper in types is really coupled with C++ codegen. I already want to throw out other coupling, so I'd rather not add more. Think the suggestion was to make a runtime helper that determines if it's trivially copyable, and it seems functional from my initial approach. This seems reasonable in this PR.
2) these dataflow analyses could probably be better in the CFG. Especially in this case, since liveness analysis is useful for other passes. I think we could generalize the problem, so if we have a dataflow pass, add it to the CFG (or some `DataflowSolver` or something) and lazily-evaluate it. Then just request the analysis you want, get the results, and use them. Simple! If the CFG was unchanged between 2 passes, it'll cache the result automatically (which shouldn't be too hard with the CFG caching mechanism). But, I think this should be punted for a future PR, since that might also be overengineering. Maybe we just want to expose liveness to all CFG users. So I'll defer to later on this where we can make a more informed decision with more thought.

I'll get to that tomorrow probably. Other stuff was occupying me today and Friday

evantypanski:
okay, I kept the lats commit separate to see how it'd change with a separate RT function for moves. Essentially:

- Move nodes now are merely a suggestion
- Runtime decides when to move
- Optimizer can then emit as it sees fit (though it still avoids return value optimization cases)
- Initializers in `switch` now emit as `auto` not `auto&&`. This prevents that `switch` case from being a reference and silently getting moved-from. This seems to me like a logic bug: if we treat it as a value in the AST, it should not suddenly be able to be a reference. I don't think it'll be a meaningful slowdown.

If we like it I'll move the `move_if_not_trivial` function first, then base the optimization commit on top of it without the `isTriviallyCopyable` method. Just kept it at the end to show what may change.

Thoughts?

bbannier:
> Thoughts?

Thanks for having a look, I do like this better. Instead of having to be very careful with each type to not break `isTriviallyCopyable` semantics we now decide this where it matters, when actually generating C++ code. The helper function is compact and we can improve on it in one place, e.g., I was wondering whether we could be less conservative by also taking available move ctrs into account, something like this:

```cpp
template<typename T>
[[nodiscard]] constexpr std::remove_cvref_t<T> move_if_not_trivial(T&& t) {
    using RawType = std::remove_reference_t<T>;

    // Trivially movable and const types should not be moved.
    if constexpr ( std::is_trivially_move_constructible_v<RawType> || std::is_const_v<RawType> )
        // Forward here since P2266 is only fixed in C++23.
        return std::forward<T>(t);
    else
        // We should not emit cases that this is actually bugprone.
        // A forward here would not do anything.
        // // FIXME: actually it would forward if passed an lvalue ref, but we always want the move anyway.
        return std::move(t); // NOLINT(bugprone-move-forwarding-reference)
}
```

We probably also should put it into a `detail` namespace so people do not get a wrong idea.

Adding smoke tests for the helper is currently not so easy since it couples detection and actually moving, we could also break that out into a trait like the following and add some unit tests/static asserts to the hilti-rt testsuite:

```cpp
template<typename T>
    requires(! std::is_reference_v<T>)
inline constexpr bool should_move_v = ! std::is_trivially_move_constructible_v<T> && ! std::is_const_v<T>;
```

> * Optimizer can then emit as it sees fit (though it still avoids return value optimization cases)

Curious what you had in mind here. Wouldn't the new helper either internally perform NRVO, or produce an rvalue which would be a candidate for RVO?

>  * Initializers in `switch` now emit as `auto` not `auto&&`. This prevents that `switch` case from being a reference and silently getting moved-from. This seems to me like a logic bug: if we treat it as a value in the AST, it should not suddenly be able to be a reference. I don't think it'll be a meaningful slowdown.

I suspect I added that from muscle memory, fine with me and seems to fit well here.

evantypanski:
Cool I'll go through and replace move with this helper, it seems nice to me as well. From what I can tell, implementation of the pass won't change, too. That's a nice touch

> Curious what you had in mind here. Wouldn't the new helper either internally perform NRVO, or produce an rvalue which would be a candidate for RVO?

I more of meant that the pass still [had the check for RTO](https://github.com/zeek/spicy/blob/0441af31e9fe93eefc4631b8ef54133aa6cef83f/hilti/toolchain/src/compiler/optimizer/passes/move-last-uses.cc#L189-L191). I can experiment with removing it and see, I just didn't.

hilti/toolchain/src/compiler/optimizer/passes/move-last-uses.cc
```diff
@@ -0,0 +1,255 @@
...
+    // An ID is generally only moveable if it's the only occurrence within a
+    // node.
+    void _walk(Node* n, Declaration*& target, std::optional<uint64_t>& to_move) {
+        if ( ! n )
+            return;
+
+        if ( auto* move = n->tryAs<expression::Move>() ) {
+            // If we already are moving this name, do not add any more moves
+            if ( auto* name = move->expression()->tryAs<expression::Name>();
+                 name && to_move && name->resolvedDeclaration() == target ) {
```
bbannier: I am struggling a bit what including `to_move` here means. Is it meant to just cancel if we already discovered a move in this walk? Thinking of incorrect code like 

```ruby
foo(move(x), x);
```

With `to_move` we'd add another move, equally broken, but maybe we should just not include it here to stay closer to user-dictated brokenness (users don't write HILTI though, so maybe just for us).

hilti/toolchain/src/compiler/optimizer/passes/move-last-uses.cc
```diff
@@ -0,0 +1,255 @@
...
+
+        // Function and method calls should only mark their non-inout
+        // arguments as movable. This gathers those facts and only
+        // recurses on arguments which are potentially moveable.
+        if ( auto* resolved = n->tryAs<operator_::struct_::MemberCall>() ) {
+            const auto& op = static_cast<const struct_::MemberCall&>(resolved->operator_());
+            auto* ft = op.declaration()->type()->type()->as<type::Function>();
+            auto args = resolved->op2()->as<expression::Ctor>()->ctor()->as<ctor::Tuple>()->value();
+
+            _walkArgs(args, ft, target, to_move);
```
bbannier, line 208: For methods we should also inspect `op0`, otherwise we might add an incorrect arg move for weird code like `x.foo(x)`.

Separately, I am wondering whether moving from `op0` would make sense in code like `x.foo()` so it becomes `move(x).foo()`. For HILTI types we don't really declare C++ methods which make that useful, and I think with wrapping `std::string` recently we also shouldn't expose raw C++ types anymore. If it can fall automatically out of including `op0` handling not bad, but probably not worth going out of our way for.

hilti/runtime/include/util.h
```diff
@@ -844,4 +842,20 @@ struct scope_exit {
     EF _f;
 };
 
+namespace detail {
+template<typename T>
+concept should_move =
```
bbannier, line 847: The roaring twenties are here 🥳.

hilti/runtime/src/tests/move.cc
bbannier, line 1: I would have naively expected this in `hilti/runtime/src/tests/util.cc` for symmetry, but it already is pretty large, so separate file is also fine.

hilti/toolchain/src/compiler/optimizer/passes/move-last-uses.cc
```diff
@@ -0,0 +1,255 @@
...
+
+    // Walks args from a function or method call, skips anything we shouldn't move
+    // from (inout, alias, etc).
+    void _walkArgs(const node::Range<Expression> args,
+                   type::Function* ft,
+                   Declaration*& target,
+                   std::optional<uint64_t>& to_move) {
+        assert(args.size() == ft->parameters().size());
+
+        // Do not walk any inout arguments or aliasing types.
```
bbannier, line 146: This is probably where the giant TODO is hidden, would you mind creating a ticket capturing possible improvements? I could think of tracking that if all users of a function would have liked to move an `inout` arg we could rewrite the function to instead take it as `copy`. Since you wrote this comment your can probably think of more 😄.

hilti/toolchain/src/compiler/optimizer/passes/move-last-uses.cc
```diff
@@ -0,0 +1,255 @@
...
+
+        return false;
+    }
+
+    // Standard liveness analysis
+    //
+    // Returns how long it took to converge
+    int _analyzeLiveness(AnalysisResult& result, const CFG* cfg) {
+        auto worklist = cfg->postorder();
+        // Parallel set for existence checks
```
bbannier:
```suggestion
        // Shadow set for existence checks.
```

hilti/toolchain/src/compiler/optimizer/passes/move-last-uses.cc
```diff
@@ -0,0 +1,255 @@
...
+            return true;
+        }
+
+        return false;
+    }
+
+    // Standard liveness analysis
+    //
+    // Returns how long it took to converge
+    int _analyzeLiveness(AnalysisResult& result, const CFG* cfg) {
```
bbannier: We don't usually return the number of required iterations, but this seems useful in your debug message. Let's fix the comment though since this doesn't return a time, but an iteration count.

hilti/toolchain/src/compiler/optimizer/passes/move-last-uses.cc
evantypanski: The intent is that moving a name means that name will never be moved in this expression. I think I copied what was in the `Name` case without considering there's a subtle difference: no matter what, if we see a move of the `target`, then we should not move anything. Regardless of `to_move`.

So yeah I think this is just a bug, thanks 

& I mostly stayed away from moving if there is another use of the name in one expression. Though, I *think* that aggregate init in C++ has a definite expression ordering, so we could move the last appearances there. I just didn't. Seeing a move is definitely "nonono definitely don't move again" though :)

hilti/toolchain/src/compiler/optimizer/passes/move-last-uses.cc
evantypanski: yeah the only reason I made it return was to remove `n` as an argument to liveness, since it was only used in that debug message. So I ripped `n` out and put the debug message in the caller. Feels odd to pass in an argument only for a debug message when I could have it return the result :)

(will update the comment though)

hilti/toolchain/src/compiler/optimizer/passes/move-last-uses.cc
evantypanski: sure I'll make an issue when this is merged. I think it's just more "not worth it to give much effort" than a big gap, though

hilti/runtime/src/tests/move.cc
evantypanski: just moved it to `util.cc`, it's fine there

(also added back the comments to the move function)

hilti/runtime/include/util.h
evantypanski: I asked Claude how it'd be implemented in C++14, this is what it said:

```
  // remove_cvref_t is C++20 — trivial to define
  template<typename T>
  using remove_cvref_t = std::remove_cv_t<std::remove_reference_t<T>>;

  // the concept becomes a trait
  template<typename T>
  struct should_move
      : std::integral_constant<bool,
            ! std::is_const<std::remove_reference_t<T>>::value &&
            ! std::is_trivially_move_constructible<remove_cvref_t<T>>::value> {};

  // if constexpr becomes overload resolution via enable_if
  template<typename T, std::enable_if_t<should_move<T>::value, int> = 0>
  constexpr remove_cvref_t<T> move_non_trivial(T&& t) {
      return std::move(t); // NOLINT(bugprone-move-forwarding-reference)
  }

  template<typename T, std::enable_if_t<! should_move<T>::value, int> = 0>
  constexpr remove_cvref_t<T> move_non_trivial(T&& t) {
      return std::forward<T>(t);
  }
```

Man I'm glad those days are done. In a sense, `if constexpr` adds more readability than the `concept`. And I really hate the `::value`..

bbannier (approved):

Issue #2129, opened 2025-07-29
Fix kill set construction for RHSs which can branch


bbannier:
Related to short-circuiting, this shouldn't always kill the previous declarations, since it might be context-dependent on whether it gets updated or not in assignment expressions:

```
function void f(int<64> i) {
    local x = "bad";
    local x2 = (i == 10) || ((x = "good") == "good");
    hilti::print(x);
}
```

Depending on `i`, `x` may end up as `good` or `bad`. It looks like `x = "bad"` is added to the kill set after `x2`, which would be wrong. Happy to punt that for later, but just pointing it out

<img width="2968" height="705" alt="Image" src="https://github.com/user-attachments/assets/6540712a-4508-425b-9673-94ba1aaa8d78" />

_Originally posted by @evantypanski in https://github.com/zeek/spicy/pull/2126#discussion_r2240205405_

evantypanski:
There's something hidden here: `||` and `&&` are not really best thought of as expressions - they are control flow. If we just modeled them within the CFG as control flow (ie `||` expression has a branching path) then this issue is solved with current infrastructure

Each expression may then still correspond to a given CFG node that is its closest parent, even if the entire expression itself has multiple control flow nodes.

Then we don't need any special handling for assignment expressions or anything - `||` will just create a branching path in the CFG like an `if` (well, because it *is* an `if`)

PR #2255, opened 2026-01-22
Add explicit CFG handling for member accesses.

bbannier:


tests/Baseline/hilti.cfg.struct_member_access/output
```diff
@@ -0,0 +1,43 @@
...
+    1 [label="end <...>/struct_member_access.hlt:4:1-20:1" shape=triangle xlabel="in: [] out: []"];
+    2 [label="end <...>/struct_member_access.hlt:4:1-20:1" shape=triangle xlabel="in: [] out: []"];
+    0 -> 2 [label="0"];
+    2 -> 1 [label="1"];
+}
+[debug/cfg-initial] Function 'f'
+digraph {
+    0 [label="local Foo f" xlabel="gen: [f: local Foo f] in: [] out: [f: local Foo f]"];
+    1 [label="f.x;" xlabel="read: [f] in: [f: local Foo f] out: [f: local Foo f]"];
+    2 [label="g(f);" xlabel="read: [f] write: [f] in: [f: local Foo f] out: [f: local Foo f] keep"];
```
rsmmr: `g(f)` remains a write to `f`. That isn't right, is it?

hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -821,6 +822,10 @@ struct DataflowVisitor : visitor::PreOrder {
                 transfer.maybe_alias.insert(decl);
         }
 
+        else if ( node->isA<operator_::struct_::MemberNonConst>() || node->isA<operator_::struct_::MemberConst>() ||
+                  node->isA<operator_::union_::MemberNonConst>() || node->isA<operator_::union_::MemberConst>() )
+            transfer.read.insert(decl);
```
rsmmr, line 860: For my understanding, why are even the `MemberNonConst` always just a read here? I've tried this branch and it actually does seem to work for cases where it ends up being a write, too, so I suppose it's correct this way. I just don't quite follow why. For example, nested structs: something like `b.foo.bar = "bar"` does correctly report both a read and a write on `b`; where's the write coming from?

bbannier: The code added here is only for the member exprs like `b.foo` which on their own are just reads (potentially going nowhere). Somewhere above we could see that expr in bigger exprs, e.g., an assign expr `b.foo = 42` which additionally marks the LHS as write (if it were on the RHS it would often still be a read unless we have funky exprs there like in #2129).

tests/Baseline/hilti.cfg.struct_member_access/output
bbannier: Yes, something seems wrong about the code we added in https://github.com/zeek/spicy/pull/2238 in that we don't take into account that we cannot directly index into operands (there might be an extra wrapping in a tuple).

hilti/toolchain/src/compiler/cfg.cc
rsmmr: I guess what I'm wondering about is that "above": `b.foo.bar` would be grouped as `(b.foo).bar = "bar"`, meaning `b.foo` itself isn't the LHS of the assigment. But it still infers a write to `b`. Can you point me to the code where that write is recorded? (As I said, it's all working correctly, just want to understand how it works.)

bbannier: I believe it would be 

https://github.com/zeek/spicy/blob/ffb70ca366b401a47ad71ae61bc16b0f1a5f9921/hilti/toolchain/src/compiler/cfg.cc#L774-L776

rsmmr: Ah, that climbs up the list of parents! Got it, thanks.

PR #2303, opened 2026-04-08
Fix potential memory leak splitting strings

evantypanski:
In practice I don't think this happens, the code was technically correct. But, Coverity is right that there is a case where `wordfree` is not called. Just always call it on scope exit for safety.

Technically this could be suppressed since only `WRDE_NOSPACE` allocates of the error conditions from what I can tell. But, this code is cleaner IMO so probably worth it. Before it was ripped straight from the manual


bbannier (approved):
Thanks! 🚀

PR #2304, opened 2026-04-09
Produce better C++ code for `switch` statements.

rsmmr:
We now emit actual C++ `switch` statements when the condition is of
integral or enum type and all case expressions are constant values.
Otherwise we continue to emit an `if-else` chain as before.

Note that `break` inside a `case` exits an enclosing loop, if any; not
the `switch`. We were potentially emitting broken code there before,
although that situation should have been rare (`break` inside `case`
inside `while/for`).

- **Emit C++ `goto` instead of `break` to exit a loop prematurely.**
- **Produce better C++ code for `switch` statements.**


hilti/toolchain/src/compiler/codegen/statements.cc
```diff
@@ -233,55 +258,139 @@ struct Visitor : hilti::visitor::PreOrder {
...
+                    cond = util::join(exprs | std::views::transform([&](auto e) { return cg->compile(e); }), " || ");
 
-            auto body = cg->compile(c->body());
+                auto body = cg->compile(c->body());
 
-            if ( first ) {
-                block->addIf(fmt("%s %s = %s", cxx_type, cxx_id, cxx_init), std::move(cond), std::move(body));
-                first = false;
+                if ( first ) {
+                    block->addIf(fmt("%s %s = %s", cxx_type, cxx_id, cxx_init), std::move(cond), std::move(body));
```
bbannier: Preexisting, but do you recall why we need to create a copy here? We only seem to use the generated binding in the generated if condition.


```suggestion
                    block->addIf(fmt("auto&& %s = %s", cxx_id, cxx_init), std::move(cond), std::move(body));
```

hilti/toolchain/src/compiler/cxx/elements.cc
```diff
@@ -300,16 +301,29 @@ void cxx::Block::addForRange(bool const_, const ID& id, const Expression& seq, c
...
+    x.setEnsureBracesforBlock();
 
-    for ( const auto& c : cases_ )
-        x._stmts.emplace_back(fmt("case %s:", c.first), c.second, 0);
+    for ( const auto& c : cases_ ) {
+        auto conditions =
+            util::toVector(c.first | std::views::transform([](const auto& e) { return fmt("case %s:", e); }));
+
+        x._stmts.emplace_back(util::join(conditions, " "), c.second, 0);
+    }
```
bbannier: Nit: no need to allocate into a temporary vector, `util::join` can work with ranges. We could also use structured bindings to make this more self-documenting.

```suggestion
    for ( const auto& [values, block] : cases_ ) {
        auto conditions = values | std::views::transform([&](const auto& e) { return fmt("case %s:", e); });
       
        x._stmts.emplace_back(util::join(conditions, " "), block, 0);
    }
```

hilti/toolchain/src/compiler/codegen/statements.cc
```diff
@@ -233,55 +258,139 @@ struct Visitor : hilti::visitor::PreOrder {
...
+                return true;
 
-        if ( first )
-            block->addBlock(std::move(default_));
+            return false;
+        };
+
+        bool can_use_cxx_switch = is_integral_type(cond->type()->type());
+
+        for ( const auto& c : n->cases() ) {
```
bbannier:
```suggestion
        for ( const auto* c : n->cases() ) {
```

tests/hilti/statements/break-fail.hlt
bbannier, line 1: Could you add test with a valid use of `break` with `switch` (`switch` in a loop) to the tests in `switch.hlt`?

hilti/toolchain/src/compiler/codegen/statements.cc
```diff
@@ -233,55 +258,139 @@ struct Visitor : hilti::visitor::PreOrder {
...
-            default_ = cg->compile(d->body());
-        else
-            default_.addStatement(
-                fmt("throw ::hilti::rt::UnhandledSwitchCase(::hilti::rt::to_string_for_print(%s), \"%s\")",
-                    (first ? cxx_init : cxx_id), n->meta().location()));
+        static auto is_integral_and_constant_value = [](const Expression* e) {
+            if ( ! is_integral_type(e->type()->type()) )
+                return false;
+
+            if ( const auto* ctor = e->tryAs<expression::Ctor>(); ctor && ctor->isConstant() )
```
bbannier, line 389: For the following code I still see `if`/`else` generated, shouldn't this cover that? I could have sworn we would inline the const but we don't, but that would be after we decided on `if` or `switch` anyway. 

```ruby
const ONE = 1;

switch (1) {
    case ONE: print 1;
    default: print "default";
}
```

Maybe we should add a test for this.

I also checked function calls and like expected they don't behave like literals.

hilti/toolchain/src/compiler/cxx/elements.cc
```diff
@@ -676,6 +696,16 @@ cxx::Formatter& cxx::operator<<(cxx::Formatter& f, const cxx::Block& x) {
...
                 f << separator();
 
+            if ( fl & flags::SkipBlock ) {
+                f << s;
+                f << eol();
+
+                if ( fl & flags::AddSeparatorAfter )
+                    f << separator();
+
+                continue;
```
bbannier, line 704: Nit: this function is pretty big and does a lot of things after this block. When working near the bottom one has probably already forgotten about this `continue`.

WDYT about instead wrapping all code below in `if ! SkipBlock` and dropping the `continue`?

hilti/toolchain/src/compiler/resolver.cc
```diff
@@ -995,6 +995,21 @@ struct VisitorPass2 : visitor::MutatingPostOrder {
         }
     }
 
+    void operator()(statement::Break* n) final {
+        if ( ! n->linkedStatement() ) {
+            for ( auto* loop = n->parent(); loop; loop = loop->parent() ) {
+                if ( loop->isA<statement::For>() || loop->isA<statement::While>() ) {
+                    recordChange(n, "recording the loop for break statement");
+                    n->setLinkedStatement(loop->as<Statement>());
```
bbannier: Nit: this makes me feel that this should be called "linked loop" instead of "linked statement" here and in `Block`.

hilti/toolchain/src/compiler/codegen/statements.cc
rsmmr: Don't recall, let's try without.

hilti/toolchain/src/compiler/codegen/statements.cc
rsmmr: That is surprising and I suppose indeed a missing const propagation. That should be done by the optimizer right? Then codegen would make the right if/switch decision. I'll take a look.

hilti/toolchain/src/compiler/cxx/elements.cc
rsmmr: Actually given the overall messiness of the function, I find the early-exit easier to comprehend. (The function, and the whole CXX elements output code, needs a cleanup sometime ..)

hilti/toolchain/src/compiler/codegen/statements.cc
rsmmr: Ah, I forgot, that we don't actually do general constant propagation during optimization; we leave global constants to the C++ compiler to figure out. I'll add an explicit check for constants to the if-vs-switch decision logic.

rsmmr:
@bbannier This got a couple of additional commits (commit 2  and 3) to fix issues that came up, one unrelated and pre-existing (the operator resolving). Plus, incorporated the review feedback.

bbannier (approved):

bbannier:
I had to revert the last patch from this PR, see https://github.com/zeek/spicy/issues/2332.

PR #2308, opened 2026-04-09
Fix construction of tuples with const element types.

bbannier:
The previous implementation which was introduced with 2dc54d2 did not properly handle tuples with const element types, e.g., `std::pair` of `const` values which we encounter when iterating `map` values. This patch adds handling for that.

Closes #2300.


rsmmr:
I believe I have an alternative fix that seems easier, see #2310.

(I had already started looking into this before I realized you were working on it too. I now just went back to what I was trying then.)

PR #2309, opened 2026-04-10
Fix recent issues reported by Coverity

bbannier:


hilti/toolchain/include/base/logger.h
```diff
@@ -227,16 +227,9 @@ class Logger {
...
-    static std::unique_ptr<Logger> _singleton;
 };
 
-inline Logger& logger() {
-    if ( ! Logger::_singleton )
-        Logger::_singleton = std::make_unique<Logger>();
-
-    return *Logger::_singleton;
-}
+Logger& logger();
```
evantypanski: this looks redundant

hilti/toolchain/include/ast/location.h
```diff
@@ -98,7 +98,7 @@ inline std::ostream& operator<<(std::ostream& out, const Location& l) {
 
 namespace location {
 /** Sentinel value indicating that no location information is available. */
-extern const Location None;
+const Location& none();
```
evantypanski: there are a couple places that still reference `location::None` (in this file) and `Location::None` (in `production.h`). They're comments, so not a big deal, but probably worth cleaning up

evantypanski (dismissed):
just a couple nits

hilti/toolchain/include/base/logger.h
```diff
@@ -227,17 +227,8 @@ class Logger {
     int _errors = 0;
 
     std::map<logging::DebugStream, size_t> _debug_streams;
-
-    static std::unique_ptr<Logger> _singleton;
 };
 
-inline Logger& logger() {
-    if ( ! Logger::_singleton )
-        Logger::_singleton = std::make_unique<Logger>();
```
rsmmr, line 236: I don't quite get why this isn't safe? What's the difference between this and the new version putting the unique ptr into a static local?

hilti/toolchain/include/ast/location.h
rsmmr: I'm not a fan of this frankly. Turning this into a function makes it look like the none location is something special/magical, hiding that it's just another value; and then having to call that function for all those default arguments just seems awkward. I guess *technically* I can see that construction order could matter here, but `Location` is a very simple object internally that (I'm guessing) is probably just initialized correctly from the get-go through its memory contents. So unless we have evidence that this is actually causing a problem, I'd definitely prefer to leave things as they are.

hilti/toolchain/include/base/logger.h
bbannier: This function is `inline` so it can be embedded into any TU making use of it, but the `static` var `Logger::_singleton` is declared in its own TU `logger.cc`. This means that conceptually static initialization could race on the initialization of this global.

A place where this shows up is when other static variables use logging in their constructor.

hilti/toolchain/include/ast/location.h
bbannier: I agree it is not great. I can check again whether we can make the `Location` constructor `constexpr` so the sentinel has defined semantics with `constinit`.

hilti/toolchain/include/base/logger.h
rsmmr: I understand the general issue with these things but I thought the existing code *already* is the fix: there's only a single instance of `_singleton` and whoever gets into `logger()` first (inlined or not), gets to initialize it.

bbannier: The empty `unique_ptr` itself might not have been constructed when somebody runs the function; the current code makes sure that we set a value only once.

rsmmr: ... but the `unique_ptr` default constructor is `constexpr`, there shouldn't anything needed to be done there.

bbannier: If the logger is declared `constinit` (available in C++20) the code would be safe. I implemented that so the patch should be much smaller now.

rsmmr (approved):

PR #2310, opened 2026-04-10
Fix construction of tuples from map keys.

rsmmr:
The previous implementation did not properly handle construction
from `std::map::value_type`, which has a `const` first element.
We need this for iterations over maps.

Closes #2300.

### DIFF
--- a/hilti/runtime/include/types/tuple.h
+++ b/hilti/runtime/include/types/tuple.h
@@ -143,7 +143,12 @@ public:
     template<typename... Us>
     explicit Tuple(tuple::OptionalTag, Optional<Us>&&... us) : Tuple() {
         [&]<size_t... Is>(std::index_sequence<Is...>) {
-            ((us.hasValue() ? std::get<Is>(*this) = std::forward<Us>(*us), TupleBase::set(Is) : void()), ...);
+            // Use construct_at rather than assignment so this works even when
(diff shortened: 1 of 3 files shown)


hilti/runtime/include/types/tuple.h
```diff
@@ -143,7 +143,12 @@ class Tuple : public TupleBase, protected std::tuple<Ts...> {
     template<typename... Us>
     explicit Tuple(tuple::OptionalTag, Optional<Us>&&... us) : Tuple() {
         [&]<size_t... Is>(std::index_sequence<Is...>) {
-            ((us.hasValue() ? std::get<Is>(*this) = std::forward<Us>(*us), TupleBase::set(Is) : void()), ...);
+            // Use construct_at rather than assignment so this works even when
+            // an element type has a deleted assignment operator (e.g.,
+            // std::map::value_type).
+            ((us.hasValue() ? (std::construct_at(&std::get<Is>(*this), std::forward<Us>(*us)), TupleBase::set(Is)) :
```
bbannier: Since `us` is not a "universal reference" this `std::forward` (already there before) just does a `std::move`, let's just directly do that.


```suggestion
            ((us.hasValue() ? (std::construct_at(&std::get<Is>(*this), std::move(*us)), TupleBase::set(Is)) :
```

bbannier (dismissed):

PR #2312, opened 2026-04-12
verify hlto scope consistency (fix linker missmatch)

mamaorha:


hilti/toolchain/src/compiler/codegen/expressions.cc
```diff
@@ -77,8 +77,8 @@ struct Visitor : hilti::visitor::PreOrder {
             case expression::keyword::Kind::Self: result = {cg->self(), Side::LHS}; break;
             case expression::keyword::Kind::DollarDollar: result = {cg->dollardollar(), Side::LHS}; break;
             case expression::keyword::Kind::Scope: {
-                auto scope = fmt("%s_hlto_scope", cg->options().cxx_namespace_intern);
-                auto extern_scope = cxx::declaration::Global(cxx::ID(scope), "uint64_t", {}, {}, "extern");
+                auto scope = cxx::ID(fmt("%s_hlto_scope", cg->options().cxx_namespace_intern));
```
mamaorha, line 80: here the logic was good, i just modified it to have same "writing style" like the 2 other places

mamaorha:
found this issue while using "zeek", added a test and explanation at zeek as well
https://github.com/zeek/zeek/pull/5335/changes/8abcb261b5969495df90389ef983cb88df0403f2#diff-f2cc938f96affb70f7652d0d241c773fc719a21296af2e98f856b5d8082dbabb

bbannier:
> found this issue while using "zeek", added a test and explanation at zeek as well [zeek/zeek@8abcb26#diff-f2cc938f96affb70f7652d0d241c773fc719a21296af2e98f856b5d8082dbabb](https://github.com/zeek/zeek/commit/8abcb261b5969495df90389ef983cb88df0403f2#diff-f2cc938f96affb70f7652d0d241c773fc719a21296af2e98f856b5d8082dbabb)

In the future, let's tests features in Spicy directly here instead of in zeek/zeek. I'll add a test while merging.

bbannier (dismissed):

tests/hilti/codegen/scope-name.hlt
```diff
@@ -0,0 +1,17 @@
+# @TEST-DOC: Verify we codegen consistent hlto_scope symbol names across translation units.
```
mamaorha: this will pass without the fix, as its "internally" when running via spicy, it happens on "spicyz" in zeek as its flagged "external" - TLDR the test pass without the "fix" above

mamaorha:
> > found this issue while using "zeek", added a test and explanation at zeek as well [zeek/zeek@8abcb26#diff-f2cc938f96affb70f7652d0d241c773fc719a21296af2e98f856b5d8082dbabb](https://github.com/zeek/zeek/commit/8abcb261b5969495df90389ef983cb88df0403f2#diff-f2cc938f96affb70f7652d0d241c773fc719a21296af2e98f856b5d8082dbabb)
> 
> In the future, let's tests features in Spicy directly here instead of in zeek/zeek. I'll add a test while merging.

i tried to add it to but it seems only "spicyz" caused the issue, when u ran it from spicy repo you dont get "__" as its internal (the test pass without the "fix")

tests/hilti/codegen/scope-name.hlt
bbannier: Yeah, you are right, the inconsistent symbol would have come from code generated by `spicyz`. I am not sure how to test either and most tests already rely on consistent symbol names, so let me drop this test again.

Issue #2322, opened 2026-04-16
Task: Unify all the linkage, calling conventions, and attributes that affect external visibility.


rsmmr:
Over time, various ways to express & constrain external visibility have evolved in HILTI and Spicy; it's become quite a mess, and partially redundant: we have at least `extern`, `public`, `&public`, `export`, `&always-emit`, and perhaps more. 

We should take stock what we need and then clean that all up, and in particular rethink the notion of linkage vs calling convention. We probably should also stop using `Linkage::Struct` for methods, because that prevents methods having other linkages (like `public`).

bbannier:
Closed by #2323.

PR #2323, opened 2026-04-16
Add optimizer logic promoting unmodified `copy`/`inout` parameters to `in`.

rsmmr:
This extends and generalizes the existing remove-unused-parameters
pass to apply further parameters optimizations. Specifically, we look
for `inout` and `copy` parameters that are never modified (per
dataflow analysis) and promote them to `in` parameters. 

Turns out this was a bit more tricky to achieve than I had hoped, due
to existing problems/inconsistencies. One concerns how we were
treating `extern` functions and methods; the PR fixes that, but
there's certainly some need for more cleanup in that whole space (see
#2322). I also noticed a couple other things that I will create
tickets for; but they did not end up affecting this PR in the end. 

- **When computing dataflow, apply same logic for member calls that we do already for function calls.**
- **Fix optimization of `extern` functions.**
- **Make a couple optimizer state management functions public.**
- **Rename `RemoveUnusedParameters` optimizer pass to `OptimizeParameters.`**
- **Add optimizer logic promoting unmodified `copy`/`inout` parameters to `in`.**


bbannier (approved):
Sorry for letting this linger for so long. This looks good to me, and we already see it work with other passes to simplify code in one of the baselines. I rebased this on `main` and it seems ready to go.

PR #2328, opened 2026-04-16
Fix regression with val prop names and coercions

### LINKED ISSUE #2327: Spicy compiler bug when &parse-from references a parameter
This code used to work earlier
```
module Reproducer;

type Inner = unit(data: bytes) {
    val: uint8 &parse-from=data;
};

public type Outer = unit {
    raw: bytes &size=1;
    parsed: Inner(self.raw);
};
```
Now it does not compile and `hilti-jit-error.log` gives the below output
```
$ cat hilti-jit-error.log
...
HILTI version 1.16.0-dev.140 (ee97bbe2a)

## [job 3] /usr/bin/c++ -c -I/usr/local/zeek/include -fPIC -std=c++20 -g -O0 -fvisibility=hidden -Wno-invalid-offsetof -I/usr/local/zeek/include -Wall -Wno-unused -funsigned-char -Wno-c2y-extensions -Wno-register -Werror=vla -funsigned-char -Wno-c2y-extensions -I /usr/local/zeek/include -I /usr/include -I /usr/include -I /usr/include -I /usr/include -o /tmp/Reproducer_2b11a61a2bc24be1-ecf2dbdb3a165015.cc_2b11a61a2bc24be1.o /tmp/Reproducer_2b11a61a2bc24be1-ecf2dbdb3a165015.cc

/tmp/Reproducer_2b11a61a2bc24be1-ecf2dbdb3a165015.cc: In member function 'hilti::rt::Optional<hilti::rt::RecoverableFailure> hlt_internal::Reproducer::Inner::_t_parse_Reproducer__Inner_stage2(hilti::rt::Optional<hilti::rt::RecoverableFailure>)':
/tmp/Reproducer_2b11a61a2bc24be1-ecf2dbdb3a165015.cc:196:81: error: conversion from 'hilti::rt::Bytes' to non-scalar type 'hilti::rt::ValueReference<hilti::rt::Stream>' requested
  196 |     ::hilti::rt::ValueReference<::hilti::rt::Stream> _t_parse_from_2 = _t_self->_t_p_data;
      |                                                                        ~~~~~~~~~^~~~~~~~~
...
```

evantypanski:
Fixes #2327

(this is not in a release so no need for patch releases or anything)

Constant propagation added special handling to replace coercions if the given expression was coerced. This helps for cases that will then need further "work" after constant propagation in order to trigger another coerceion cycle.

But, when value propagation gained more abilities, this became a liability. Coercions no longer go through the same path and end up missing cases. These cases are necessary to stay the same, and we get no value in replacing the coercion (eg for `! Null` coercions, those are only possible with `Ctor`).

We need to be a bit pickier about replacing coercions when propagating.

### DIFF
--- a/hilti/toolchain/src/compiler/optimizer/passes/value-propagation.cc
+++ b/hilti/toolchain/src/compiler/optimizer/passes/value-propagation.cc
@@ -225,8 +225,11 @@ struct Replacer : optimizer::visitor::Mutator {
 
         if ( source_expr.propagate ) {
             Node* to_replace = n;
-            // Replace the coercion, too, so that the coercer reruns.
-            if ( auto* coerced = n->parent()->tryAs<expression::Coerced>() )
+
+            // Only replace the coercion for ctors, since that gets different handling.
+            // Names (and anything else possibly propagated) can keep the coercion.
+            if ( auto* coerced = n->parent()->tryAs<expression::Coerced>();
+                 coerced && source_expr.expr->isA<expression::Ctor>() )
                 to_replace = coerced;
 
             replaceNode(to_replace, source_expr.expr, "propagating constant value");
--- a/tests/spicy/optimization/name-coercions.spicy
+++ b/tests/spicy/optimization/name-coercions.spicy
@@ -1,0 +1,14 @@
+# @TEST-DOC: Ensures coerced names from value propagation can compile. Regression test for #2327
+#
+# @TEST-EXEC: spicyc -j %INPUT
+
+module Test;
+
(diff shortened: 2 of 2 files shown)


bbannier (approved):
LGTM. The test case could probably reduced more by moving from Spicy units to HILTI code, but better than nothing.

evantypanski:
I think I prefer tests like this in Spicy actually, since it's a user-facing issue (& found by a user). If I discovered it developing another optimization or something I'd probably make a HILTI test. Just my preference though

PR #2330, opened 2026-04-17
Add draft v1.16.0 release notes. [skip CI]

bbannier:


NEWS.rst
```diff
@@ -7,12 +7,75 @@ Version 1.16 (in progress)
...
+  throughput improvements of 20-60% in microbenchmarks. These improvements are
+  powered by the existing machinery and new passes, mainly
+
+  - GH-2249: Add new optimizer pass removing unused struct fields.
+
+    Add optimizer pass that removes struct fields with no productive
+    accesses. The pass marks unused fields with ``&no-emit="optimized"``,
+    allowing them to remain in the AST through codegen, which then emits
+    code to render them as ``(optimized out)``. Fields with ``&always-emit``,
+    or inside types with ``&always-emit``, are never removed.
```
evantypanski: This feels really technical and nitty-gritty, I would prefer just cutting the middle part. It's important, but maybe not to most people


```suggestion
    Add optimizer pass that removes struct fields with no productive
    accesses. Fields with ``&always-emit``, or inside types with
    ``&always-emit``, are never removed.
```

then combine with following paragraph. They can get the other stuff from our favorite `CHANGES` file

NEWS.rst
```diff
@@ -7,12 +7,75 @@ Version 1.16 (in progress)
...
+  - GH-2247 to extend constant propagation, and also added copy propagation as
+    an extension which took advantage of improved handling of non-mutating data
+    access in CFG (GH-2238, GH-2254)
+
+- Support for using Spicy on Windows
+
+  We are grateful to Maor Hamami for his contributions to this release to
+  make Spicy work under Windows. This included numerous changes for compiling
+  with MSVC, fixes to codegen and to allow Windows processes to load HLTO files
+  (which are always used even if not safed). He also added CI support for
```
evantypanski:
```suggestion
  (which are always used even if not safe). He also added CI support for
```

though not 100% sure what "not safe" means and it's a little scary :)

NEWS.rst
bbannier: Streamlined this a bit, but wanted to keep the "`(optimized out)`" in since that is the breaking change users might see.

NEWS.rst
bbannier: _saved_, what a cursed language 😅

evantypanski (approved):

PR #2331, opened 2026-04-20
Fixup and partially revert code from #2304

bbannier:
In #2304 we both fixed the way we emit C++ constants as well as improved how we emit HILTI switch statements. This patch fixes the our constants fix, and reverts the new switch rendering since it has more issues than we can fix today with an Zeek RC1 due.


evantypanski (approved):
LGTM. I think it's fine to be a bit heavy-handed with the pending RC1

PR #2334, opened 2026-04-21
Reapply and fix "Produce better C++ code for `switch` statements."

### LINKED ISSUE #2332: Reapply reverted changes from #2304
When preparing a bump of Spicy into Zeek a bunch of changes from #2304 broke the build there and needed to fixed or reverted, in particular https://github.com/zeek/spicy/pull/2304/changes/324aee958ec5dd513685faac9ef0d6b94bca2529 had to go out again for the time being. This patched failed for [two reasons](https://github.com/zeek/spicy/pull/2331/changes/7867adc7bea30103d87b37fdf44d26dc1885ea44):

- it didn't properly emit local variables when also emitting a C++ `switch`, i.e., the branch emitting a switch never made use of `cxx_id`. This shouldn't be too hard to fix, but needs to make sure conditions are only evaluated once (we have a test for that)
- HILTI allows duplicate `case` labels, but C++ does not. We could either update the logic determining C++ `switch` eligibility to reject constructs with duplicate cases, or emit them from the C++.

bbannier:
Closes #2332


bbannier:
Matching Zeek PR: https://github.com/zeek/zeek/pull/5383

rsmmr:
@bbannier  Take a look at https://github.com/zeek/spicy/pull/2341 for an alternative solution that leaves the hard work to resolver. I kept the 2 first commits from your PR.

PR #2336, opened 2026-04-21
    Inside optimizer, index per-function state by canonical ID.

### LINKED ISSUE #2325: Optimizer state should index functions by canonical ID
Several of our optimizer passes collect state about functions in maps or tables. Currently they all use `Declaration::fullyQualifiedID()` as the index (and/or `declaration::Function::functionID()`, which also returns fully qualified IDs, but prefers returning one from a linked prototype, if available [1]).

However, FQIDs aren't necessarily unique in an AST, there are ways to create ambigious names. We have one test for that in `spicy/types/unit/field-name-clash-fail.spicy` and #2324 shows another example. 

Right now, everything seems to be working ok nevertheless. I suspect that's a combination of luck and [one specific check](https://github.com/zeek/spicy/blob/501e3137ae6bd53816394c6c4ce280f1c709684e/hilti/toolchain/src/compiler/optimizer/passes/remove-unused-params.cc#L111) for ambiguities in one pass. That's just speculation, though. In fact #2324 was my attempt to trigger the issue, but then I ran into a different  problem there first. 

In the end, I think we should change the passes to index by `Declaration::canonicalD()` instead, that's guaranteed to be unique. Note, however, that canonical IDs *will* be different for implementation and prototype, so we need a similar lookup as in `declaration::Function::functionID()` to index consistently (i.e., by prototype's canonical ID if available, implementation's otherwise). Probably that means a similar helper function to factor this out.

[1] I'm actually not sure if that lookup to the prototype is needed. Would the prototype not have the same FQID as the function itself?

rsmmr:
- **Link free-function implementations to their prototypes.**
- **Inside optimizer, index per-function state by canonical ID.**

### DIFF
--- a/hilti/toolchain/include/ast/declarations/function.h
+++ b/hilti/toolchain/include/ast/declarations/function.h
@@ -88,12 +88,6 @@ public:
         _linked_prototype_index = index;
     }
 
-    /**
-     * Returns the ID for the function declaration, regardless of the module which
-     * it was implemented in. Only valid once the AST has been resolved.
-     */
-    ID functionID(ASTContext* ctx) const;
-
     std::string_view displayName() const final { return "function"; }
 
     node::Properties properties() const final;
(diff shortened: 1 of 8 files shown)


bbannier (approved):
Sorry for not looking into this earlier! This looks good, let me push a rebase and then merge.

PR #2339, opened 2026-04-22
Fix codegen to emit missing `goto` labels.

rsmmr:
If loop statements changed during AST resolving, we wouldn't update
any linked break statements.

Also render linked loop statements' identities in debug output in hex,
for easier comparison to node identities.


bbannier (approved):
Thanks! 🚀

PR #2341, opened 2026-04-22
Reapply "Produce better C++ code for `switch` statements."

rsmmr:
(Needs another look and some testing would be good.)

- **Reapply "Produce better C++ code for `switch` statements."**
- **Ensure emitting of variable bindings in `switch` statements.**
- **Extend the API for `case` AST nodes.**
- **Add new style to constant folder that folds all named constants.**
- **Extend resolver to unify duplicate case handles.**


hilti/toolchain/src/compiler/resolver.cc
```diff
@@ -1033,8 +1036,46 @@ struct VisitorPass2 : visitor::MutatingPostOrder {
...
+        // Find cases where we can statically determine that they are
+        // duplicates. If we see duplicates across cases, that's an error.
+        // Inside a single case, we filter them down to a single instance of
+        // that value.
+        std::unordered_map<std::string, statement::switch_::Case*> seen;
+        std::set<std::pair<statement::switch_::Case*, Expression*>> to_remove;
+
+        for ( auto* case_ : n->cases() ) {
+            for ( auto* cond : case_->expressions() ) {
+                // std::cerr << "    " << *cond << std::endl;
```
bbannier:
```suggestion
```

hilti/toolchain/src/compiler/resolver.cc
```diff
@@ -1033,8 +1036,46 @@ struct VisitorPass2 : visitor::MutatingPostOrder {
...
+                    continue;
+
+                auto ctor =
+                    detail::constant_folder::foldExpression(builder(), cond,
+                                                            detail::constant_folder::Style::InlineAllConstants |
+                                                                detail::constant_folder::Style::FoldTernaryOperator);
+                if ( ! ctor && *ctor )
+                    continue;
 
+                auto key = (*ctor)->print();
```
bbannier, line 1064: Like discussed offline, this is a slight hack since we do not have a way to compare expressions for equivalence, and also implicitly rely on stringification to produce a non-empty string. Let's `assert(! key.empty())` for the latter, for the first one would could add a TODO.

hilti/toolchain/src/compiler/codegen/statements.cc
```diff
@@ -257,52 +258,154 @@ struct Visitor : hilti::visitor::PreOrder {
...
+
+            for ( const auto& c : n->cases() ) {
+                if ( c->isDefault() )
+                    continue; // will handle below
+
+                auto conditions = util::toVector(c->preprocessedExpressions() | std::views::transform([&](auto e) {
+                                                     return normalize_cxx_switch_value(cg->compile(e), true);
+                                                 }));
+
+                cases.emplace_back(conditions, cg->compile(c->body()));
```
bbannier: Nit since effectively preexisting.


```suggestion
                cases.emplace_back(std::move(conditions), cg->compile(c->body()));
```

hilti/toolchain/src/compiler/codegen/statements.cc
```diff
@@ -257,52 +258,154 @@ struct Visitor : hilti::visitor::PreOrder {
...
+
+        bool can_use_cxx_switch = is_integral_type(cond->type()->type());
+
+        for ( const auto* c : n->cases() ) {
+            if ( c->isDefault() )
+                continue;
+
+            for ( const auto* e : c->expressions() ) {
+                if ( ! is_integral_and_constant_value(e) )
+                    can_use_cxx_switch = false;
```
bbannier: Nit since preexisting.


```suggestion
                if ( ! is_integral_and_constant_value(e) ) {
                    can_use_cxx_switch = false;
                    break;
                }
```

bbannier (dismissed):
Thanks, this makes sense to me. Like I mentioned I was wondering whether pulling the new validation into zeek-8.2.0-rc2 is fair, but I think that's okay since we would have landed it with the original PR also. I opened a full CI run for Zeek [here](https://github.com/zeek/zeek/pull/5383).

PR #2342, opened 2026-04-24
Remove reduction of entropy when computing unique IDs.

bbannier:
When computing unique IDs we were internally hashing locations to
`uint64_t`, but then reducing to `uint16_t`'s range for more readable
identifiers in generated code. This removes reduces entropy by about
75%.

Given how much of a reduction this was it still seems to have worked
surprisingly well up until de8db50 where we now suddenly see
deterministic hash collisions for a single test in the
`validate_release_tarball` task; in that case it manifests as two
distint hooks being generated with identical C++ identifier. Since this
seemed so obscure I initially though this to be some memory corruption
bug, but it turned out mundane in the end.

### DIFF
--- a/hilti/toolchain/include/compiler/detail/codegen/codegen.h
+++ b/hilti/toolchain/include/compiler/detail/codegen/codegen.h
@@ -127,7 +127,7 @@ public:
      * @param n node to generate the ID for
      *
      */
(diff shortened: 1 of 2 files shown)


evantypanski (approved):
it's not that much ~more~ less readable now honestly

PR #2349, opened 2026-04-29
Fix transitive aliases

### LINKED ISSUE #2258: Dead code elimination erroneously removes aliased values
The following code has different behavior with and without optimizations:

```
module Foo;

function f() {
    local a = new 10;
    local b = a;

    local p: uint64& = a;
    *p = 20;

    print b;
}

f();
```

Outputs:

```
$ spicyc -j test.spicy
10
$ spicyc -j -g test.spicy
20
```

It seems like `b` is only marked as aliasing `a` and `p` is only marked as aliasing `a`. We need to make sure that every alias such that:

- `b` aliases `a`
- `c` aliases `a`
- Then: `b` aliases `c`

I think not only at use do the two `maybe_alias` sets get unioned, they should *also* union whatever is aliased by both nodes, aggressively.

An alternative is to use a [union-find](https://llvm.org/docs/AliasAnalysis.html#the-aliassettracker-implementation) algorithm instead. Then it's just one set that contains all the things that alias, and something doesn't alias if it is in a set alone.

evantypanski:
Fixes #2258

Consider a program where `a` has two direct aliases: `b` and `p`. Then, `b` has an alias `c`. `p` and `c` must alias, or else writes to `p` might be removed. This forces that by unioning alias sets until convergence.

There are other ways to ensure aliases are unioned (see: union-find) but this is the simplest way, albeit a little slow.

Second commit is just for legibility and efficiency: theoretically it should be faster, but I didn't see much change in my test. I'm just scared of an obvious O(n^3) algorithm. It was suggested by Gemini, and I agree since we already have a pass that uses reverse postorder

### DIFF
--- a/hilti/toolchain/src/compiler/cfg.cc
+++ b/hilti/toolchain/src/compiler/cfg.cc
@@ -1013,40 +1013,47 @@ void CFG::_populateDataflow() {
 
     { // Encode aliasing information.
 
-        // First make aliasing symmetric: to handle the case of e.g.,
-        // references aliasing is stored symmetrically, i.e., if `a` aliases
-        // `b`, `b` will also alias `a`.
-        bool aliases_changed;
-        do {
-            // Fixed point iteration
-            aliases_changed = false;
-            for ( auto& [n, transfer] : _dataflow ) {
-                for ( const auto& alias : transfer.maybe_alias ) {
-                    const auto* stmt = _graph.getNode(alias->identity());
-                    if ( ! stmt || ! stmt->get() || ! _dataflow.contains(*stmt) )
-                        util::detail::internalError(
-                            util::fmt("could not find CFG node for '%s' aliased in '%s'", alias->print(), n->print()));
+        // Aliases must be symmetric and transitive, i.e., if `a` aliases
+        // `b`, `b` will also alias `a`. If `c` aliases `b`, then `a` also
+        // aliases `c`.
+        auto worklist = postorder();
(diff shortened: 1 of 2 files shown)


hilti/toolchain/src/compiler/cfg.cc
```diff
@@ -994,11 +1013,20 @@ void CFG::_populateDataflow() {
...
-        // First make aliasing symmetric: to handle the case of e.g.,
-        // references aliasing is stored symmetrically, i.e., if `a` aliases
-        // `b`, `b` will also alias `a`.
-        for ( const auto& [n, transfer] : _dataflow ) {
-            for ( const auto& alias : transfer.maybe_alias ) {
+        // Aliases must be symmetric and transitive, i.e., if `a` aliases
+        // `b`, `b` will also alias `a`. If `c` aliases `b`, then `a` also
+        // aliases `c`.
+        auto worklist = postorder();
+        std::ranges::reverse(worklist);
```
bbannier, line 1022: Why do we reverse here? If we want to keep it, let's mention why we do it in a comment.

bbannier (dismissed):

hilti/toolchain/src/compiler/cfg.cc
evantypanski: Since I'm treating this as a forward dataflow analysis, reverse postorder will converge quickest (I think)

Though I haven't tested it, I'm almost certain it's true though.

I can add a comment

PR #2353, opened 2026-05-07
Fix setting up of return value stack space in fiber stack switcher.

bbannier:
Usually fiber switching goes through a trampoline function which saves
the stack of the originator fiber and switches to the target fiber.

There was a bug in how we set up stack space for the result of the
target fiber: in a shared stack setup both fibers (well) shared a stack,
and calling `::fiber_push_return` could clobber the stack of the
originator. This patch fixes the order of calls so now the switcher
sets up the stack space on the correct fiber. This required tracking the
state of a fiber across a switch boundary, so we now track whether a
target fiber needs setup in `SwitchArgs`.

Closes zeek/zeek#5434.

PR #2354, opened 2026-05-11
Fix Fiber constructor clobbering shared stack.

awelzel:
Previously, when constructing a Fiber with type SharedStack, we'd initialize the underlying fiber with the global shared stack. This isn't safe: The fiber_init() call writes to the shared stack, but there's no guarantee that the shared stack isn't being used by another fiber when this happens.

Circumvent this by moving the underlying fiber initialization into __fiber_switch_trampoline(), where it is guaranteed that the previously executing fiber has saved() the shared stack so that it is now free to be used by a different fiber. Use the nullptr fiber_stack() return as the indicator, that the initialization still needs to happen.

Alternative to #2353 , I think.

### DIFF
--- a/hilti/runtime/src/fiber.cc
+++ b/hilti/runtime/src/fiber.cc
@@ -143,8 +143,32 @@ void __fiber_switch_trampoline(void* argsp) {
     if ( from->_type == detail::Fiber::Type::SharedStack )
         from->_stack_buffer.save();
 
-    if ( to->_type == detail::Fiber::Type::SharedStack )
-        to->_stack_buffer.restore();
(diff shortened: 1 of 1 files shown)


bbannier (approved):
Thanks, this makes sense to me. I pushed a bunch of nit fixes, but AFAICT this can go in. Anything else you'd like to add here @awelzel?

awelzel:
> Anything else you'd like to add here @awelzel?

Nope. :shipit:

bbannier:
Backported to `release/1.16` and `release/1.14`.

PR #2355, opened 2026-05-11
Switch back to upstream version of 3rdparty/SafeInt.

bbannier:
We moved to our own fork of SafeInt because of #1829. It looks like upstream has fixed enough for us to switch back to them. It might be that more related changes are incoming, but we can follow up on them separately.


evantypanski (approved):
heck yeah

PR #2361, opened 2026-05-18
Fix potential nullptr dereference in remove-unused-fields pass.

bbannier:
This was introduced in 79d44f786c so needs to be backported to release/1.16.


evantypanski (approved):

bbannier:
Also backported to `release/1.16`.

PR #2366, opened 2026-05-20
Add Github Actions CI

bbannier:


.github/workflows/check.yml
```diff
@@ -0,0 +1,375 @@
...
+        persist-credentials: false
+
+    - name: Setup ccache
+      uses: hendrikmuhs/ccache-action@d62db5f07c26379fc4b4e0916f098a92573c3b03 # v1.2.23
+      with:
+        key: ${{ github.workflow }}-${{ github.job }}-${{ matrix.release }}
+
+    - name: Install dependencies
+      run: |
+        brew install bison cmake flex ninja ccache
```
evantypanski: seems like there are a few ways to cache from `brew`, could we?

.github/workflows/check.yml
```diff
@@ -0,0 +1,375 @@
...
+    - name: Setup ccache
+      uses: hendrikmuhs/ccache-action@d62db5f07c26379fc4b4e0916f098a92573c3b03 # v1.2.23
+      with:
+        key: ${{ github.workflow }}-${{ github.job }}-${{ matrix.id }}
+
+    - name: Configure environment
+      run: |
+        echo "NPROCS=$(nproc)" >> "$GITHUB_ENV"
+
+    - name: Run script
```
evantypanski: This name is a little ambiguous in the Github UI, it says "Run script" and ends up doing almost all of the work there. But, that's where we would look for anything CI related.

What about "Build and test" (even though no-toolchain doesn't test) or "Perform CI" or something like that? I think "Run script" blends in with the other steps ("Configure environment", "setup ccache") even though it does *far* more

ci/run-ci
```diff
@@ -15,6 +15,8 @@ root=$(cd $(dirname $0)/.. && pwd)
 build=
 install=
 
+J=$(nproc)
```
evantypanski: nit, while `J` is obvious in the context of shell scripts, I'd prefer the variable be named `jobs` or similar

nothing needed for the `-j` argument, I think that's good, just the variable name

```suggestion
jobs=$(nproc)
```

evantypanski (dismissed):
Seems good to me, just left a few small comments

.github/workflows/check.yml
bbannier: I started with _Run script_ when I was thinking of moving this to a reusable workflow, and we might eventually still do that (the GH actions CI for zeek/zeek uses the same pattern). Went with _Perform CI_ 👯‍♂️ for now.

evantypanski (dismissed):

.github/workflows/check.yml
```diff
@@ -0,0 +1,421 @@
+# Copyright (c) 2020-now by the Zeek Project. See LICENSE for details.
+
+name: Check
+
+permissions:
+  contents: read
+
+on:
+  push:
+    branches: [master]
```
awelzel, line 10: Should this have been main? (total fly-by)

bbannier: 🤦

PR #2371, opened 2026-05-28
Remove Cirrus CI setup.

bbannier:
With the introduction of a GH actions CI setup we updated some scripts used by the Cirrus CI setup so they work on CI images provided by GH actions; this caused Windows CI to break on Cirrus CI. In any case Cirrus CI will go away in a couple days anyway.


evantypanski (commented):
Couple questions:

1) I see a couple Cirrus references in the README for build status, should that swap over or be removed?
2) `scripts/store-release-artifacts` has a few references to Cirrus. That script seems pretty Cirrus based. Should that be rewritten/removed in the meantime?

bbannier:
> 1. I see a couple Cirrus references in the README for build status, should that swap over or be removed?

Yes, we can move that over now even though we don't have a build for `releases/1.16`, yet. I opened #2372 to add a GH actions CI to that branch as well. I'd expect the links to the previous Cirrus build status to be dead on Monday.

> 2. `scripts/store-release-artifacts` has a few references to Cirrus. That script seems pretty Cirrus based. Should that be rewritten/removed in the meantime?

That is slightly more work which I planned to tackle later. Right now the jobs upload a single artifact tarball which also would include distro packages, and this probably needs to be disentangled. Above script pulls stuff from Cirrus' artifact store so we can host it with the release, and I'd expect that can be simpler going forward.

Additionally we also link release artifacts in Cirrus from the docs which should be migrated.

evantypanski (approved):

PR #2373, opened 2026-05-29
Stop requiring git commit titles to end in `.` [skip ci]

bbannier:
This deviated from the convention in all other Zeek projects and always tripped me up. In the past the trailing dot also made downstream use harder, e.g., proper PR titles would have it autoadded which I always tried to remove, or when assembling release notes.

### DIFF
--- a/.gitlint
+++ b/.gitlint
@@ -10,7 +10,7 @@
 # used in here for clarity.
 #
 [general]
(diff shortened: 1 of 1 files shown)


awelzel (approved):
Thanks!

evantypanski (approved):
agreed

PR #2378, opened 2026-06-01
Add centos-stream-10 to CI

bbannier:
This was added by Zeek, but never added here.

### DIFF
--- a/docker/Dockerfile.centos-stream-10
+++ b/docker/Dockerfile.centos-stream-10
@@ -1,0 +1,35 @@
+# Copyright (c) 2020-now by the Zeek Project. See LICENSE for details.
+
+FROM quay.io/centos/centos:stream10
+
+WORKDIR /root
+
+ENV PATH="${PATH}"
+
+RUN echo 'LC_CTYPE="C"' >> /etc/locale.conf \
+ && echo 'LC_ALL="C"' >> /etc/locale.conf \
+ && echo 'LANG="C"' >> /etc/locale.conf
+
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.centos-stream-10
```diff
@@ -0,0 +1,36 @@
...
+# dnf config-manager isn't available at first, and
+# we need it to install the CRB repo below.
+RUN dnf -y install 'dnf-command(config-manager)'
+
+# What used to be powertools is now called "CRB".
+# We need it for some of the packages installed below.
+# https://docs.fedoraproject.org/en-US/epel/
+RUN dnf config-manager --set-enabled crb
+RUN dnf -y install \
+    https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm \
```
awelzel: Should this be 10?

docker/Dockerfile.centos-stream-10
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) 2020-now by the Zeek Project. See LICENSE for details.
+
+FROM quay.io/centos/centos:stream9
```
awelzel: Should this be 10?

docker/Dockerfile.centos-stream-10
bbannier: Definitely. I copied the file over and then tweaked it to pass locally, but must somehow lost the changes.

awelzel (approved):
Assuming it passes.

PR #2379, opened 2026-06-01
Some tweaks to clang-format style

bbannier:


evantypanski (dismissed):
Cool, I like both more

maybe time to give it another look too, I saw a couple that we should (?) be doing that aren't present, like:

```
QualifierAlignment: Left
EmptyLineBeforeAccessModifier: Always
```

but those would be new and don't think they apply much since we mostly follow them

bbannier:
> Cool, I like both more
> 
> maybe time to give it another look too, I saw a couple that we should (?) be doing that aren't present, like:
> 
> ```
> QualifierAlignment: Left
> EmptyLineBeforeAccessModifier: Always
> ```
> 
> but those would be new and don't think they apply much since we mostly follow them

Yes, these would be great to add, especially since they reflect our current style and get rid of inconsistencies, adding them as well.

PR #2382, opened 2026-06-02
Handle overflows in `atoi_n`

bbannier:


hilti/runtime/include/util.h
```diff
@@ -492,7 +493,14 @@ inline Iter atoi_n(Iter s, Iter e, uint8_t base, Result* result) {
         else
             break;
 
-        n = n.value_or(Result()) * base + d;
+        constexpr auto max_value = std::numeric_limits<Result>::max();
+
+        Result val = n.value_or(Result());
+
+        if ( (val > 0 && base > 0 && val > max_value / base) || (d > max_value - val * base) ) [[unlikely]]
```
awelzel: `base > 0` is guaranteed by guard at top of function.


```suggestion
        if ( (val > 0 && val > max_value / base) || (d > max_value - val * base) ) [[unlikely]]
```

hilti/runtime/include/util.h
```diff
@@ -492,7 +493,14 @@ inline Iter atoi_n(Iter s, Iter e, uint8_t base, Result* result) {
         else
             break;
 
-        n = n.value_or(Result()) * base + d;
+        constexpr auto max_value = std::numeric_limits<Result>::max();
```
awelzel: Using `::max()` here introduces a off-by-one for the negative parsing, no?

No strong opinion if we should handle this correctly, but e.g. strtoll does magic:

https://github.com/gcc-mirror/gcc/blob/master/libiberty/strtoll.c#L128

awelzel (changes requested):
I think there's an off by one for the limit of negative numbers.

It would be good to extend the test cases in `tests/util.cc` to cover the extremes and extremes +/- 1 of some or all stdints? That would have made UBSAN trigger just by running tests and also show the off by one for the negative limit.an

hilti/runtime/include/util.h
bbannier: Yeah, thanks for catching that. I have reorganized the code slightly to make this easier to track, let me know what you think.

hilti/runtime/include/util.h
```diff
@@ -479,10 +482,32 @@ inline Iter atoi_n(Iter s, Iter e, uint8_t base, Result* result) {
...
+
+    using Unsigned = std::conditional_t<std::is_signed_v<Result>, std::make_unsigned_t<Result>, Result>;
+    constexpr auto max_value =
+        std::is_unsigned_v<Result> ? Unsigned(max) : static_cast<Unsigned>(0) - static_cast<Unsigned>(min);
+
+    // Sanity check since e.g., our own `hilti::rt::integer::safe` type does
+    // not implement `std::is_signed_v`, `std::make_unsigned_t`, or
+    // `std::numeric_limits`. Since we require a builtin integral type above we
+    // should never reach this.
+    static_assert(min < max);
```
awelzel, line 496: Not sure I understand this one. This just sanitizes min() / max() from lines 488 and 489 above which is just from numeric_limits. Should it just come right after those or is that superfluous?

hilti/runtime/include/util.h
```diff
@@ -479,10 +482,32 @@ inline Iter atoi_n(Iter s, Iter e, uint8_t base, Result* result) {
...
+    // Since we handle any sign above we can accumulate into an unsigned
+    // integer when iterating digits. We pick `uint64_t` since it is large
+    // enough to handle any type we need to handle.
+    static_assert(max_value <= std::numeric_limits<uint64_t>::max());
+    uint64_t n = 0;
     for ( ; it != e; ++it ) {
         auto c = *it;
 
-        Result d;
+        // `uint8_t` is sufficient to store any digit up to base 32.
```
awelzel:
```suggestion
        // `uint8_t` is sufficient to store any digit up to max base 36.
```

?

awelzel (dismissed):
Just two nits minor things, but :shipit:

hilti/runtime/include/util.h
bbannier: Yeah exactly, I move it to there. The issue for our safe integers is that there we'd get `min == max == 0` which makes everything below misbehave. I also added some more context to the comment.

bbannier:
Also backported the `release/1.14` and `release/1.16`.

PR #2388, opened 2026-06-03
Reduce number of warnings the msvc build emits

bbannier:


mamaorha:
This looks promising but maybe worth rerunning windows tests? (currently in this check it was skipped) - this can introduce a lof of runtime failures and it will be better to veirfy before merging (by running the btests on windows, there is a PR to re-enable it on ci)

hilti/runtime/include/types/any.h
```diff
@@ -8,6 +8,6 @@
 #include <hilti/rt/extension-points.h>
 
 namespace hilti::rt::detail::adl {
-inline std::string to_string(const hilti::rt::any& x, adl::tag /*unused*/) { return "<any value>"; }
+inline std::string to_string(const hilti::rt::any& /*x*/, adl::tag /*unused*/) { return "<any value>"; }
```
awelzel, line 11: Might be a bit more consistent to replace `/* unused */` with `/* tag */` or `/* x */` with `/* unused */` and the same elsewhere, but not sure it's worth the churn and total nitpicky.

awelzel (approved):
LGTM - pretty much skipped the "unused" commit.

hilti/toolchain/include/ast/types/bitfield.h
```diff
@@ -186,7 +186,7 @@ class Bitfield : public UnqualifiedType, public node::WithUniqueID {
         return ctx->make<Bitfield>(ctx, node::flatten(attrs, bits, value), width, m);
     }
 
-    static auto create(ASTContext* ctx, Wildcard _, const Meta& m = Meta()) {
+    static auto create(ASTContext* ctx, Wildcard /*_*/, const Meta& m = Meta()) {
```
awelzel, line 189: Just skimmed over this all, but C++26 should make this okay again, right?

https://www.sandordargo.com/blog/2025/01/08/cpp26-unnamed-placeholders

bbannier: Yes, which means that we can roll this back when we migrate to C++26 or later in the mid to late 2030s.

hilti/runtime/include/types/any.h
bbannier: Yeah, skipping this, `unused` is the current pattern and used consistently.

PR #2391, opened 2026-06-03
[windows] fix btests failures

mamaorha:


hilti/toolchain/src/compiler/driver.cc
```diff
@@ -679,7 +679,10 @@ Result<Nothing> Driver::addInput(const hilti::rt::filesystem::path& path) {
                 if ( ! redirect_done ) {
                     redirect_done = true;
                     auto current_exe = rt::filesystem::canonical(util::currentExecutable());
-                    auto redirect_dir = rt::filesystem::temp_directory_path() / "hilti-hlto-redirect";
+                    // Place redirect directory on the same drive as the executable
+                    // so that hard links always succeed (they require the same
+                    // filesystem). We use a subdirectory next to the exe's bin dir.
+                    auto redirect_dir = current_exe.parent_path() / "hilti-hlto-redirect";
                     std::error_code ec;
                     rt::filesystem::create_directories(redirect_dir, ec);
```
bbannier, line 687: This does zero error checking which is not great and made this failure pretty obscure before. Suggest to add error handling to both `redirect_directories` and `copy_file`, fine to abort on error.

spicy/runtime/src/driver.cc
```diff
@@ -407,6 +407,10 @@ Result<hilti::rt::Nothing> Driver::processPreBatchedInput(std::istream& in) {
     std::string magic;
     std::getline(in, magic);
 
+    // Strip trailing \r in case the input has Windows-style (CRLF) line endings.
```
bbannier: Both here and the instance below: this seems like a pretty late point to clean this up, and also litters processing with low level adjustments, could we instead open the stream as text in the driver? Then all of this shouldn't be needed.

bbannier (changes requested):
Thanks, just minor requests for cleanup.

hilti/toolchain/src/compiler/driver.cc
```diff
@@ -855,6 +858,16 @@ Result<Nothing> Driver::compile() {
         if ( _driver_options.output_path.empty() ) {
             // OK if not available.
             if ( _library ) {
+#ifdef _WIN32
+                // Register the directories of any --cxx-link libraries so that
+                // the loader can find their DLL dependencies at runtime.
+                for ( const auto& lib : _compiler_options.cxx_link ) {
+                    if ( lib.empty() )
+                        continue;
+                    auto dir = rt::filesystem::absolute(rt::filesystem::path(lib)).parent_path();
```
bbannier: Let's us the non-throwing variant here and check for errors.

spicy/runtime/src/driver.cc
bbannier: I tweaked your fixup to solve this on the `ifstream` level like I had proposed.

bbannier (approved):

PR #2394, opened 2026-06-04
Avoid overflows when checking coercion from `real` to integer type

bbannier:
The casts from `double` to integral type could previously trigger UB at compile time if the `double` value was outside of the range of the integral type, NaN or infinity. To handle this correctly we add a range check in this patch. Due to the way comparisons against float NaN and infinity work they are also handled.

### DIFF
--- a/hilti/toolchain/src/compiler/coercer.cc
+++ b/hilti/toolchain/src/compiler/coercer.cc
@@ -1,6 +1,9 @@
 // Copyright (c) 2020-now by the Zeek Project. See LICENSE for details.
 
+#include <cmath>
+#include <cstdint>
 #include <limits>
+#include <type_traits>
 #include <utility>
 
 #include <hilti/ast/builder/builder.h>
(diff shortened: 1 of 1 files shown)


hilti/toolchain/src/compiler/coercer.cc
```diff
@@ -33,6 +34,15 @@ inline const DebugStream Coercer("coercer");
 
 namespace {
 
+// Helper to check whether a double value is representable in some integral type `T`.
+template<typename T>
+bool isRepresentable(double x) {
+    constexpr auto min = static_cast<double>(std::numeric_limits<T>::min());
+    constexpr auto max = static_cast<double>(std::numeric_limits<T>::max());
```
evantypanski: so for `uint64` and `int64`, double can't represent the max value, so this cast ends up failing. I wasn't able to get an ubsan build up for some weird reason so I can't verify, but I know that's [a common pitfall](https://stackoverflow.com/questions/63686765/double-to-uint64-t-conversion)

so in order to get that max number we might have to do something funky with `std::ldexp` and `std::numeric_limits<T>::digits` in order to get `2^64` directly and avoid the cast? IDK I did not run it with an undefined sanitizer so I might be wrong

I think there are a few solutions, I don't know if you considered that though so I won't go much further for now. Should probably see it with conversions to `uint64_t`'s max value, right?

bbannier: Yeah, I didn't tackle that existing behavior at all and focused on getting to a stage where the test suite passes with ubsan.

evantypanski: Ah yeah. For some reason I thought that `static_cast<double>(std::numeric_limits<uint64_t>::max())` might also be UB, but it's valid, just maybe not always correct. So some values will convert weirdly near the top of `uint64_t`'s range, but that's existing. Should probably note that for the future, though

evantypanski (dismissed):

hilti/toolchain/src/compiler/coercer.cc
bbannier: Let me add a `TODO` here, definitely something worth looking into in the future.

bbannier: I actually went in and added proper range checks, PTAL.

hilti/toolchain/src/compiler/coercer.cc
```diff
@@ -1,6 +1,10 @@
 // Copyright (c) 2020-now by the Zeek Project. See LICENSE for details.
 
+#include <cmath>
+#include <cstdint>
 #include <limits>
+#include <ranges>
```
evantypanski: my editor says this is unused

evantypanski (dismissed):
LGTM

PR #2396, opened 2026-06-05
Skip leak tests unless running on Linux

bbannier:
Clang's leak sanitizer is only supported on Linux at the moment, so skip tests relying on it on all other platforms.

### DIFF
--- a/tests/Scripts/is-linux
+++ b/tests/Scripts/is-linux
@@ -1,0 +1,5 @@
+#!/bin/sh
+#
+# Returns true if running on Linux.
+
+test "$(uname -o 2>/dev/null)" = "Linux"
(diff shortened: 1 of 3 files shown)


evantypanski (commented):
or Mac? From [here](https://clang.llvm.org/docs/LeakSanitizer.html#supported-platforms)

is this for Windows?

bbannier:
> or Mac? From [here](https://clang.llvm.org/docs/LeakSanitizer.html#supported-platforms)
> 
> is this for Windows?

I run into failures with these tests on macos with a build with any sanitizer enabled. There is still the issue that we equate `have-sanitizer` with "asan enabled", but I left that be for the time being.

evantypanski (approved):
My Mac builds have been failing with some dependencies (eg benchmark) when sanitizers are enabled so tough for me to check, maybe I should figure that out soon too. Haven't pinned down if it's me or the cmake config yet.

So that's fine

PR #2397, opened 2026-06-08
Avoid underflow in `atoi_n`

bbannier:
We had previously fixed some overflows in this function, but missed a case when extracting `INT64_MIN`. We were already parsing the absolute value with enough range, but could underflow when negating it for the final result since the negation was done in the target type which did not have enough range.

With this patch we instead perform the negation in the unsigned type which has enough range, and only cast to the target type after.

### DIFF
--- a/hilti/runtime/src/tests/util.cc
+++ b/hilti/runtime/src/tests/util.cc
@@ -6,6 +6,7 @@
 #include <cstdint>
 #include <cstdlib>
 #include <ctime>
+#include <limits>
 #include <locale>
 #include <string>
 #include <vector>
@@ -84,9 +85,6 @@ TEST_CASE("atoi_n") {
         }
 
         CHECK_EQ(x, -42);
-
-        // Parsing an a signed value into an unsigned type underflows.
-        CHECK_EQ(atoi_n_<uint8_t>("-1", 10, 2), 255);
     }
 
     SUBCASE("parse something") {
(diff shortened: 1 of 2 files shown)


hilti/runtime/src/tests/util.cc
```diff
@@ -136,6 +134,19 @@ TEST_CASE("atoi_n") {
...
+        CHECK_EQ(atoi_n_<int64_t>(std::to_string(std::numeric_limits<int64_t>::max()), 10, 19),
+                 std::numeric_limits<int64_t>::max());
+
+        CHECK_EQ(atoi_n_<uint64_t>(std::to_string(std::numeric_limits<uint64_t>::min()), 10, 1),
+                 std::numeric_limits<uint64_t>::min());
+        CHECK_EQ(atoi_n_<uint64_t>(std::to_string(std::numeric_limits<uint64_t>::max()), 10, 20),
+                 std::numeric_limits<uint64_t>::max());
+
+        // Parsing an a signed value into an unsigned type underflows.
+        CHECK_EQ(atoi_n_<uint8_t>("-1", 10, 2), 255);
```
awelzel, line 149: I don't think this is sensible behavior, but you said it's pre-existing, right?

awelzel (approved):

hilti/runtime/src/tests/util.cc
awelzel: Actually, seems strtoull() behaves similarly, so maybe it is.

bbannier: IMO this is not sensible and surprising for anyone expecting proper parsing and not `strtoull`, but yes, existing behavior.

bbannier:

PR #2400, opened 2026-06-10
Additional fixes for #2399

leres:
Here are more changes needed to compile spicy (zeek) on FreeBSD 16 current after 86326398b73b. Thanks to @bbannier.


bbannier (approved):
Thank you @leres!

PR #2402, opened 2026-06-12
safe-int: Ignore -Wpedantic for SafeInt.hpp

awelzel:
Compiling Zeek with -Wpedantic produces a lot of the following warnings whenever safe-int.h is included.

    SafeInt.hpp:1999:54: warning: ISO C++ does not support ‘__int128’ for ‘type name’ [-Wpedantic]

Skip passing -Wpedantic to SafeInt.hpp, didn't seem like there's anything more specific.

### DIFF
--- a/hilti/runtime/include/safe-int.h
+++ b/hilti/runtime/include/safe-int.h
@@ -2,8 +2,18 @@
 
 #pragma once
 
+#if defined(__clang__) || defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpedantic"
+#endif
+
 #define SAFEINT_DISABLE_ADDRESS_OPERATOR
 #include <hilti/rt/3rdparty/SafeInt/SafeInt.hpp>
+
+#if defined(__clang__) || defined(__GNUC__)
+#pragma GCC diagnostic pop
+#endif
+
 #include <hilti/rt/exception.h>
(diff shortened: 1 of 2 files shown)


awelzel:
I'm not actually sure if this is a useful change. We probably won't compile with `-pedantic`, but wanted to see what squelching these could look like. Feel free to close if it's not reasonable. Reference/motivation was zeek/zeek#5573.

bbannier:
> I'm not actually sure if this is a useful change. We probably won't compile with `-pedantic`, but wanted to see what squelching these could look like. Feel free to close if it's not reasonable. Reference/motivation was [zeek/zeek#5573](https://github.com/zeek/zeek/pull/5573).

Yeah, I think this is not that useful. The current approach in Spicy is not ideal:

1. we directly include some 3rdparty files from a directory which is available both at build as well as JIT time, e.g., `hilti/rt/3rdparty/SafeInt/SafeInt.hpp` here
2. for compilation during JIT there is nothing which declares e.g., `hilti/rt/3rdparty` as a system header which would suppress diagnostics
3. we compile these 3rdparty headers with our own flags (e.g., warning flags), even though we do not control them

The current approach is to add add one-off suppression like you did here for warning flags we support; since `-Wpedantic` is not a flag we useit seems weird to add it here, but it would make sense if we added the flag for the warning flags of Spicy; that would also ensure that we validate it in CI.

PR #2403, opened 2026-06-12
Change `hilti::rt::atoi_n` to take a `string_view`

bbannier:
This function previously accepted and returned generic iterators. This
not only made calling it potentially error prone, but also lead to some
awkwardness in using its return value (e.g., use of iterator math to
decide whether anything was consumed). On the implementation side, it
also restricted what operations were possible on the passed iterators
which would make e.g., rewriting the function in terms of
`std::from_chars` harder than needed.

This patch changes the function to instead accept `std::string_view`
instead, and return the number of consumed bytes.

### DIFF
--- a/hilti/runtime/include/util.h
+++ b/hilti/runtime/include/util.h
@@ -452,26 +452,24 @@ class OutOfRange;
  * @pre The input sequence must not be empty, i.e., we require `s != e`.
  * @pre Base must be in the inclusive range [2, 36].
  *
- * @par s beginning of the input range.
- * @par e end of the input range.
+ * @par input buffer to parse number from.
  * @par base base of the input range.
  * @par result address of the memory location to used for storing
  *     a possible parsed result.
- * @return iterator to the first character not used in value
- *     extraction.
+ * @return number of bytes consumed in parsing.
  */
-template<class Iter, typename Result>
-inline Iter atoi_n(Iter s, Iter e, uint8_t base, Result* result)
-    requires std::is_integral_v<Result> && (sizeof(Result) <= sizeof(uint64_t)) && std::contiguous_iterator<Iter>
+template<typename Result>
+inline uint8_t atoi_n(std::string_view input, uint8_t base, Result* result)
+    requires std::is_integral_v<Result> && (sizeof(Result) <= sizeof(uint64_t))
 {
     if ( base < 2 || base > 36 )
         throw OutOfRange("base for numerical conversion must be between 2 and 36");
 
-    if ( s == e )
+    if ( input.empty() )
         throw InvalidArgument("cannot decode from empty range");
 
     bool neg = false;
-    auto it = s;
+    const auto* it = input.begin();
 
     if ( *it == '-' ) {
         neg = true;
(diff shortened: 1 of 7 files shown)


bbannier:
I added a benchmark to this which I now moved to the front so we can check the effect on throughput for each commit. The results for a release build on my machine were:

- baseline:

  ```
  Load Average: 6.26, 11.54, 7.81
  --------------------------------------------------------------------
  Benchmark                          Time             CPU   Iterations
  --------------------------------------------------------------------
  atoi_n_uint64_t/digits:1        2.53 ns         2.53 ns    277228821
  atoi_n_uint64_t/digits:8        8.78 ns         8.78 ns     79979891
  atoi_n_uint64_t/digits:20       19.5 ns         19.5 ns     34824311
  ```

- take `std::string_view`:

  ```
  Load Average: 6.51, 10.80, 7.73
  --------------------------------------------------------------------
  Benchmark                          Time             CPU   Iterations
  --------------------------------------------------------------------
  atoi_n_uint64_t/digits:1        2.51 ns         2.51 ns    278653546
  atoi_n_uint64_t/digits:8        8.75 ns         8.75 ns     80632156
  atoi_n_uint64_t/digits:20       19.4 ns         19.4 ns     36338341
  ```

- use `std::from_chars`:

  ```
  Load Average: 5.47, 10.29, 7.61
  --------------------------------------------------------------------
  Benchmark                          Time             CPU   Iterations
  --------------------------------------------------------------------
  atoi_n_uint64_t/digits:1        4.68 ns         4.68 ns    149206647
  atoi_n_uint64_t/digits:8        10.2 ns         10.2 ns     67408156
  atoi_n_uint64_t/digits:20       18.3 ns         18.3 ns     38226717
  ```

The `std::string_view` patch seems like a no brainer, but I am torn about `std::from_chars`. I'd suspect integers encountered in parsers to not be evenly distributed, but instead be more likely smaller.

bbannier:
Re: the patch with the refactor to use `std::from_chars`, WDYT @evantypanski? The current code is very hard to follow and I would be tempted to replace it, but I do not think we can or should make things on the parsing path slower, so dropping the patch seems sensible.

evantypanski (commented):
my view is drop the `from_chars` commit, though I like the impl more. Random disparate thoughts:

1) I don't mind the behavior changes, but I think there's also one if none match, right? Before it'd return `0`, now it throws invalid argument. Also it looks like this would accept `++123` too, which is a slight behavior change, and the recursion is a little ugly IMO
2) I would imagine `atoi_n` is more likely to be used with small numbers, *and* if it's in a hot path then small numbers are more likely. Makes me think it's not worth it
3) The code it's replacing is definitely more verbose, but I actually find it quite readable. Maybe I've just done digit parsing a lot, but it's clear where the core loop is. The min/max stuff is a bit annoying but you can mostly gloss over that
4) `string_view` looks like a no brainer to me, too

just my opinion though

bbannier:
The build failure under windows seems to be fixed by https://github.com/zeek/spicy/pull/2426.

evantypanski (approved):

PR #2408, opened 2026-06-15
Include coercions in detection of expressions with side effects

### LINKED ISSUE #2404: Generated code for widening integer assignment in a field hook produces redefinition of '_t_dd'
/tmp/vex/vex.spicy:
  module vex;

  public type Msg = unit {
      var wide: uint32;
      narrow: uint16 { self.wide = $$; }   # uint16 -> uint32 widening in a field hook
  };

  spicyc -j vex.spicy

  .../vex_*.cc:226:41: warning: parentheses were disambiguated as redundant
     parentheses around declaration of variable named '_t_dd' [-Wvexing-parse]
    226 |     ::hilti::rt::integer::safe<uint32_t>(_t_dd);
  .../vex_*.cc:226:42: error: redefinition of '_t_dd' with a different type:
     '::hilti::rt::integer::safe<uint32_t>' vs 'const ::hilti::rt::integer::safe<uint16_t> &'

Spicy:     v1.16.0 (version-number 11600), release build
           (bundled with Homebrew zeek 8.2.0, prefix /opt/homebrew/Cellar/zeek/8.2.0)
Spicy built with C++ compiler:  clang++
JIT C++ compiler (at runtime):  Apple clang 21.0.0 (clang-2100.1.1.101)
OS:        macOS 26.5 (build 25F71), Darwin 25.5.0, arm64

[hilti-jit-error.log](https://github.com/user-attachments/files/28900577/hilti-jit-error.log)

bbannier:
We previously would always consider expressions with coercions as having
side effects which in at least #2404 lead to us generating code which
caused a most vexing parse. Since coercions do not have side effects
themself, extend our detection to include them by querying the behavior
of the coerced expression. This should fix the majority of new failures
introduced by recent rewrites which removed the LHS of assignments; we
will do more structural fixes in follow-up patches.

Closes #2404.


bbannier:
The idea here would be to backport the first patch to `release/1.16`, and merge the latter only to `main` to prevent further issues from similar automatic rewrites.

evantypanski (dismissed):
yeah, this is good to me

did you look at other files like `operators.cc`? I think it's fine as-is, just curious.

bbannier:
> did you look at other files like `operators.cc`? I think it's fine as-is, just curious.

Yeah I didn't and I mechanically assumed there'd be no value construction there, but there is, mainly in more constructor and cast definitions; I tweaked these as well. There are probably more instances of `T(..)` across the codebase, but we can fix these as we go, and try to remember to prefer `T{..}` when adding new ones.

bbannier:
Fix for #2404 (`d9ecb209f2593b038c10ea6d8d05e4a96873d2e9`) also backported to `release/1.16`.

PR #2409, opened 2026-06-16
Switch runner to `windows-2022`

bbannier:
On 2026-06-15 Github switched the `windows-2025` runner to mean instead `windows-2025-vs2026`. Since we want to keep building with vs2022 switch image to `windows-2022` instead.


bbannier:
Folded into #2408.

PR #2410, opened 2026-06-16
Fix build for GH actions windows runner updates

bbannier:
On 2026-06-15 [Github switched the `windows-2025` runner to mean instead `windows-2025-vs2026`](actions/runner-images#14017). This caused our Windows CI job to break since it hardcoded the path to the Visual Studio installation.

Since this migration working automatically in https://github.com/zeek/zeek bring this CI in line with the status there.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -369,7 +369,7 @@ jobs:
 
     - name: Install dependencies
       run: |
-        choco install -y --no-progress winflexbison3 ccache msysgit python ccache
+        choco install -y --no-progress winflexbison3 ccache msysgit python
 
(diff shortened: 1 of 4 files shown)


.github/workflows/check.yml
```diff
@@ -369,7 +369,7 @@ jobs:
 
     - name: Install dependencies
       run: |
-        choco install -y --no-progress winflexbison3 ccache msysgit python ccache
+        choco install -y --no-progress winflexbison3 ccache msysgit python
```
awelzel, line 372: Sort an multi-line makes these easier to spot.

awelzel (approved):

Issue #2414, opened 2026-06-24
Configuring Spicy with GCC on Mac gets configuration errors in 3rdparty


evantypanski:
When I configure Spicy on Mac with GCC, I get the following error:

```
CMake Error at 3rdparty/benchmark/CMakeLists.txt:325 (message):
  Failed to determine the source files for the regular expression backend
```

I did a simple fix to just say we have it:

```diff
diff --git i/3rdparty/CMakeLists.txt w/3rdparty/CMakeLists.txt
index 68f9c4ed3..7ab847bbc 100644
--- i/3rdparty/CMakeLists.txt
+++ w/3rdparty/CMakeLists.txt
@@ -57,6 +57,7 @@ else()
     set(BENCHMARK_ENABLE_TESTING OFF)
     set(BENCHMARK_ENABLE_INSTALL OFF)
     set(BENCHMARK_ENABLE_WERROR OFF)
+    set(HAVE_STD_REGEX TRUE CACHE BOOL "" FORCE)
     add_subdirectory(benchmark)
     set_target_properties(benchmark PROPERTIES EXCLUDE_FROM_ALL ON)
     set_target_properties(benchmark_main PROPERTIES EXCLUDE_FROM_ALL ON)
```

... but then I get a bunch of other issues when building:

```
$ ninja
[3/391] Building C object 3rdparty/repro...oc/CMakeFiles/reproc.dir/src/redirect.c.
FAILED: [code=1] 3rdparty/reproc/reproc/CMakeFiles/reproc.dir/src/redirect.c.o
/opt/homebrew/bin/ccache /opt/homebrew/bin/gcc-15  -I/Users/etyp/src/spicy/3rdparty/reproc/reproc/include -g -std=c99 -arch arm64 -isysroot /Applications/Xcode_26.1_rc.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX26.1.sdk -fPIC -MD -MT 3rdparty/reproc/reproc/CMakeFiles/reproc.dir/src/redirect.c.o -MF 3rdparty/reproc/reproc/CMakeFiles/reproc.dir/src/redirect.c.o.d -o 3rdparty/reproc/reproc/CMakeFiles/reproc.dir/src/redirect.c.o -c /Users/etyp/src/spicy/3rdparty/reproc/reproc/src/redirect.c
In file included from /Users/etyp/src/spicy/3rdparty/reproc/reproc/include/reproc/reproc.h:6,
[...]

PR #2416, opened 2026-06-25
Allow disabling of building of benchmarks

bbannier:
We have a reoccurring and hard to diagnose issue with the google/benchmark dependency failing to detect the regex backend, #2414. Since the benchmarks and their dependencies are not needed during regular use, add a configure option `--disable-benchmarks` and a CMake variable `SPICY_ENABLE_BENCHMARKS` to allow turning benchmarks off.


evantypanski (approved):
annoyingly large number of places that needed changed IMO but looks good

timwoj:
Can this be ported into the releases for zeek 8.0 and 8.2 as well?

bbannier:

PR #2417, opened 2026-06-26
Update docs for removal of binary package artifacts

bbannier:


doc/installation.rst
```diff
@@ -20,104 +19,80 @@ platforms to support and test.
...
-      - :repo:`Dockerfile <docker/Dockerfile.ubuntu-24>`
+We provide a ``./configure`` script to set up and parameterize the build and to
+produce e.g., a ``Makefile``. See ``./configure --help`` for the available
+options. The general workflow is::
 
-macOS
-~~~~~
+    ./configure
+    make -C build
+    make -C build install
```
evantypanski:
```suggestion
    # ./configure
    # make -C build
    # make -C build install
```

doc/installation.rst
```diff
@@ -20,104 +19,80 @@ platforms to support and test.
...
-      - :repo:`Dockerfile <docker/Dockerfile.fedora-41>`
+    # port install flex bison cmake ninja python310 py310-pip
+    # pip install btest
 
-    * - Fedora 42
-      - :package-release-rpm:`RPM <linux_fedora42>`
-      - :package-dev-rpm:`RPM <docker_fedora42>`
-      - :repo:`Dockerfile <docker/Dockerfile.fedora-42>`
+If you want to build the documentation as well, also install
+``sphinx_rtd_theme`` and ``diagrams`` through *pip*.
```
evantypanski: preexisting but the italics read weird to me for a binary:


```suggestion
``sphinx_rtd_theme`` and ``diagrams`` through ``pip``.
```

evantypanski (changes requested):
then `doc/programming/guidelines.rst` mentions binary packages, so that needs changed, too. Probably switch that to just mention the binaries that ship with Zeek?

bbannier:
> then `doc/programming/guidelines.rst` mentions binary packages, so that needs changed, too. Probably switch that to just mention the binaries that ship with Zeek?

This looks factually fine to me since this is about some artifact which got a compiler baked in during prep. These artifacts could come from distros, Zeek binary packages or Homebrew, and the FAQ still makes sense. Does that make sense?

evantypanski (approved):
Yeah the `spicy-config` point stands, just the "Binary packages are most often built with a system compiler" feels odd. Though it does still stand in multiple instances so I'm probably just reading it wrong

PR #2420, opened 2026-06-30
Fix regression causing removal of used fields

bbannier:
In 09c092a9733 from #2336 we changed how a number of optimizer passes reference functions. A cleanup there caused a regression in the `RemoveUnusedFields` pass so that we now remove fields even though they are used. This leads to compilation failures, e.g., in Zeek's builtin PostgreSQL parser.

Since that pass works on fields instead of general functions this patch simply reverts the cleanup in that pass.


evantypanski (approved):
I also wouldn't mind an assertion in the `functionID` that makes sure the field is a method if possible, but this is good for the fix

PR #2421, opened 2026-07-03
Update release workflow

bbannier:
We previously dropped publishing of binary artifacts, but hadn't updated the release workflow for that yet. This patch does that, and also performs a bunch of bumps and pins of actions that workflow uses.


evantypanski (approved):

bbannier:

PR #2425, opened 2026-07-07
Bump doctest and clean up CMake build

bbannier:


3rdparty/CMakeLists.txt
```diff
@@ -1,3 +1,14 @@
+# Override `cmake_minimum_required` to suppress deprecation warnings from
+# 3rdparty projects using a lower version.
+#
+# NOTE: We use our own version here instead of the absolute minimum required
+# one which could cause policies to be enabled which are not expected by the
+# 3rdparty project; we likely will catch any incompatiblities via build
+# failures.
+macro(cmake_minimum_required)
+    _cmake_minimum_required(VERSION ${CMAKE_MINIMUM_REQUIRED_VERSION})
```
evantypanski, line 9: this makes me really uneasy since cmake policies change behavior based on this. Maybe we could pass `-Wno-deprecated` or something in the configure script? I see that [`CMAKE_WARN_DEPRECATED`](https://cmake.org/cmake/help/latest/variable/CMAKE_WARN_DEPRECATED.html) is itself deprecated

or did you consider that already?

cmake/TypeErase.cmake
```diff
@@ -1,43 +0,0 @@
-# Copyright (c) 2020-now by the Zeek Project. See LICENSE for details.
```
evantypanski, line 1: is `FindGoldLinker.cmake` also unused? I just see comments in `CHANGES`

bbannier: It is pulled in by `CMakeLists.txt` and enables the Gold linker globally when it is found. 🤷

3rdparty/CMakeLists.txt
bbannier: The downside of hardcoding flags in `./configure` is that it would not be active when Spicy is bundled in the Zeek build. I also rarely run it and instead let CMake reconfigure automatically, so I'd still see the warnings.

I tried getting an overview of what policies can do which I why I put that comment. Are there any policies you'd be worried about?

cmake/TypeErase.cmake
evantypanski: oh oops!

3rdparty/CMakeLists.txt
evantypanski: not really, but that makes sense. Having a broad window for cmake behavior changes causing subtle issues just seems like a debugging nightmare a year from now. It's more theoretical though so I guess it's mostly fine

evantypanski (approved):

PR #2426, opened 2026-07-08
Update CI to llvm-22

bbannier:
This updates CI to llvm-21. We clean up where we manage LLVM versions, and address or at least mute newly raised clang-tidy diagnostics. The sanitizer CI job reports no new failures which is something new. The benchmark differences appear minor or incidental; they should only be due to code changes since the benchmark job's compiler setup did not change.

### DIFF
--- a/hilti/toolchain/include/base/logger.h
+++ b/hilti/toolchain/include/base/logger.h
@@ -101,7 +101,7 @@ public:
     Stream(logging::Level level) : std::ostream(&_buf), _buf(level) {}
 
     /** Creates a stream that sends output to a given debug stream. */
-    Stream(logging::DebugStream dbg) : std::ostream(&_buf), _buf(std::move(dbg)) {}
+    Stream(logging::DebugStream dbg) : std::ostream(&_buf), _buf(dbg) {}
 
 private:
     Buffer _buf;
--- a/hilti/toolchain/include/base/logger.h
+++ b/hilti/toolchain/include/base/logger.h
@@ -80,7 +80,7 @@ constexpr auto from_string(std::string_view s) { return util::enum_::from_string
 } // namespace level
 
 /** Ostream-variant that forwards output to the central logger. */
-class Stream : public std::ostream {
+class Stream : public std::ostream { // NOLINT(misc-multiple-inheritance)
 private:
     class Buffer : public std::stringbuf {
     public:
--- a/hilti/toolchain/include/base/logger.h
+++ b/hilti/toolchain/include/base/logger.h
@@ -29,22 +29,32 @@ namespace logging {
 class DebugStream {
 public:
     /**
-     * @param name name of the stream, which must be unique across all stream
+     * @param name string literal for name of the stream, which must be unique across all stream
+     *
      */
-    explicit DebugStream(const std::string& name);
+    template<size_t N>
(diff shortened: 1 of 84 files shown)


hilti/runtime/src/type-info.cc
```diff
@@ -13,100 +13,126 @@ using namespace hilti::rt;
...
 const TypeInfo hilti::rt::type_info::uint32{std::nullopt,
                                             "uint32",
                                             TO_STRING(hilti::rt::integer::safe<uint32_t>),
                                             new hilti::rt::type_info::UnsignedInteger<uint32_t>()};
+// NOLINTNEXTLINE(bugprone-throwing-static-initialization)
 const TypeInfo hilti::rt::type_info::uint64{std::nullopt,
                                             "uint64",
                                             TO_STRING(hilti::rt::integer::safe<uint64_t>),
                                             new hilti::rt::type_info::UnsignedInteger<uint64_t>()};
+// NOLINTNEXTLINE(bugprone-throwing-static-initialization)
```
evantypanski: wait we're disabling it in `.clang-tidy`, are these `NOLINTNEXTLINE` necessary? Am I missing it?

hilti/toolchain/include/base/logger.h
```diff
@@ -80,7 +80,7 @@ constexpr auto from_string(std::string_view s) { return util::enum_::from_string
 } // namespace level
 
 /** Ostream-variant that forwards output to the central logger. */
-class Stream : public std::ostream {
+class Stream : public std::ostream { // NOLINT(misc-multiple-inheritance)
```
evantypanski, line 83: should we just disable this check altogether? this line confused me a lot, at a certain point it's just more effort than it's worth

evantypanski (commented):
Just a couple minor comments

Also: do we know if Clang 9 works?

https://github.com/zeek/spicy/blob/fd21eab41a60e052384e90b0158854d4d14749ca/doc/installation.rst?plain=1#L36

Seems like that number was added 6 years ago and never changed. I would be shocked if it worked, considering where we are now. May be worth updating. We have more than doubled that version!

hilti/runtime/src/type-info.cc
bbannier: Not missing anything, I just didn't properly clean this up after attempting to fix it. IMO it would be great to not have code like this since it is pretty bad, but cleaning it up turned into too much work so I went for a global suppression. I now drop these local suppressions.

hilti/toolchain/include/base/logger.h
bbannier: I agree this is confusing. TBH, I still find this lint useful, and using multiple inheritance should be rare, give us pause and be carefully considered. Having marked up the 13 places were we currently do this, I'd expect it to be rare for us to run into this again, but it adds a useful guardrail, especially as people let bots loose on this codebase. WDYT about keeping this?

bbannier:
> Also: do we know if Clang 9 works?

I created https://github.com/zeek/spicy/issues/2428 for that work.

hilti/toolchain/include/base/logger.h
evantypanski: Yeah we can keep it. I think I would always catch the obvious cases in review and be pretty critical, but you're right that explicit opt-out could be better for future contributions

evantypanski (approved):

PR #2427, opened 2026-07-09
Fix inout param and add helper for move last uses pass

evantypanski:
I was looking at #2253 again and these two commits came up. Figured I'd put them in a separate commit while I get that other one good enough to get in.

Do you like the new `isOneOf`? I remember wanting it but never bothered adding it. We could also change `isA` to take variadic args, or just drop that, but I like it.


doc/autogen/architecture.svg
bbannier, line 1: Let's not add this, it seems incompatible with the version of `diagrams` pinned in `doc/pyproject.toml`.

hilti/lib/hilti.hlt
```diff
@@ -43,7 +43,7 @@ declare public void throw_unset_optional() &cxxname="hilti::rt::optional::detail
 declare public void throw_attribute_not_set() &cxxname="hilti::rt::struct_::detail::throw_attribute_not_set" &have_prototype;
 
 declare public optional<Profiler> profiler_start(string name, optional<uint<64>> size = Null) &cxxname="hilti::rt::profiler::start" &have_prototype;
-declare public void profiler_stop(optional<Profiler> p, optional<uint<64>> size = Null) &cxxname="hilti::rt::profiler::stop" &have_prototype;
+declare public void profiler_stop(inout optional<Profiler> p, optional<uint<64>> size = Null) &cxxname="hilti::rt::profiler::stop" &have_prototype;
```
bbannier, line 46: Uff, since this should have generated a function taking a `const std::optional<Profiler>&` I wonder how this ever linked.

hilti/toolchain/include/ast/node.h
```diff
@@ -797,6 +797,12 @@ class Node {
             return nullptr;
     }
 
+    /** Checks if this is one of many possible node types. */
+    template<typename... Ts>
+    bool isOneOf() const {
```
bbannier: Yes, I like this since it makes uses easier to read and less error prone. I think this makes more sense than a variadic `isA` (any, one, all?), but we should call this one `isAnyOf`.

While at it, would you mind adding a `requires` clause that each `Ts` is a `Node`, something like

```cpp
        requires(std::is_base_of_v<Node, Ts> && ...)
```

bbannier (dismissed):

doc/autogen/architecture.svg
evantypanski: ah for some reason I thought changing `Node` did that!

FTR it wasn't `diagrams`, it was with `graphviz` which doesn't seem pinned anywhere. I upgraded from 14.1.1 to 15.1.0 and it fixed it.

hilti/lib/hilti.hlt
evantypanski: honestly no idea, and there are a couple optimizations that rely on `inout` so I was surprised we didn't catch it till here, too.

PR #2437, opened 2026-07-14
Fix FP cycle in types

### LINKED ISSUE #2433: FP cycle in struct
The following Spicy code should compile, but gets an error:

```
module Test;

global x = 0;

public type S = struct {
    i: uint64;
};

function take_two(s1: S, s2: S): void {
    x = s1.i;
    x = s2.i;
}

function f() {
    local one: S = [$i = 0];
    local two: S = [$i = 0];

    take_two(one, two);
}

f();
```

error:

```
$ spicyc -c test.spicy
[error] test.spicy:5:17-7:1: cycle detected in definition of type 'Test::S'
```

Weirdly I started making a test in hilti and it asserts, may be related?:

```
module Test {

import hilti;

global x = 0;

public type S = struct {
	uint<64> i;
};

public function void take_two(S s1, S s2) {
	x = s1.i;
	x = s2.i;
}

function void f() {
	local S one = [$i = 1];
	local S two = [$i = 1];

	#take_two(two, one);
}

f();

}
```

hilti error:

```
$ hiltic -p test.hlt
- type::Struct <declaration="-" type="T17" unified="<unset>" wildcard="false"> [parent @q:1b3a] (test.hlt:17:16-17:23) (resolved) [@t:1b3b]
 - declaration::Expression <canonical-id="" declaration="-" fqid="" id="self" linkage="private"> [parent @t:1b3b] (test.hlt:17:16-17:23) [@d:1b3c]
  - expression::Keyword <kind="self"> [parent @d:1b3c] (test.hlt:17:16-17:23)  (non-const)  (resolved) [@e:1b3d]
   - QualifiedType <const="false" extern="false" side="rhs"> [parent @e:1b3d] (test.hlt:17:16-17:23) [@q:1b3e]
    - type::ValueReference <declaration="-" type="-" unified="<unset>" wildcard="false"> [parent @q:1b3e] (resolved) [@t:1b3f]
     - QualifiedType <const="false" extern="true" side="rhs"> [parent @t:1b3f] [@q:1b40]
  - AttributeSet [parent @d:1b3c] (test.hlt:17:16-17:23) [@a:1b41]
 - declaration::Field <canonical-id="" declaration="-" fqid="" id="i" linkage="struct" linked-type="T17"> [parent @t:1b3b] (test.hlt:17:16-17:23) (resolved) [@d:1b42]
  - QualifiedType <const="true" extern="false" side="rhs"> [parent @d:1b42] (test.hlt:17:22-17:22) [@q:1b43]
   - type::UnsignedInteger <declaration="-" type="-" unified="uint64" width="64" wildcard="false"> [parent @q:1b43] (test.hlt:17:22-17:22) (resolved) [@t:1b44]
  - AttributeSet [parent @d:1b42] (test.hlt:17:16-17:23) [@a:1b45]
  - <empty>
[internal-error] declaration without canonical ID found: self

--- Aborting
# 0   libhilti.dylib                      0x0000000102a9dc34 _ZN5hilti2rt9BacktraceC2Ev + 128
# 1   libhilti.dylib                      0x0000000102a9dd90 _ZN5hilti2rt9BacktraceC1Ev + 28
# 2   libhilti.dylib                      0x0000000102785e60 _ZN5hilti4util18abortWithBacktraceEv + 52
# 3   libhilti.dylib                      0x0000000102775c80 _ZNSt3__13mapIN5hilti7logging11DebugStreamEmNS_4lessIS3_EENS_9allocatorINS_4pairIKS3_mEEEEE4findB9nqe220106ERS8_ + 0
# 4   libhilti.dylib                      0x00000001024865c0 _ZN12_GLOBAL__N_115VisitorCheckIDsclEPN5hilti11DeclarationE + 164
# 5   libhilti.dylib                      0x000000010276f9bc _ZN5hilti11declaration10Expression8dispatchERNS_7visitor10DispatcherE + 64
# 6   libhilti.dylib                      0x0000000102484c6c _ZN5hilti7visitor7VisitorILNS0_5OrderE0ENS0_10DispatcherEE8dispatchEPNS_4NodeE + 60
# 7   libhilti.dylib                      0x000000010245137c _ZN5hilti7visitor5visitIN12_GLOBAL__N_115VisitorCheckIDsENS_7ASTRootEEEDaOT_PT0_NSt3__117basic_string_viewIcNS9_11char_traitsIcEEEE + 248
# 8   libhilti.dylib                      0x00000001024503e0 _ZNK5hilti10ASTContext8checkASTEb + 1216
# 9   libhilti.dylib                      0x0000000102960edc _ZN5hilti6detail9optimizer8ASTState8checkASTENS1_6PassIDE + 136
# 10  libhilti.dylib                      0x0000000102961ad4 _ZN5hilti6detail9Optimizer8_runPassERKNS0_9optimizer8PassInfoEj + 1376
# 11  libhilti.dylib                      0x000000010296213c _ZN5hilti6detail9Optimizer3runEv + 480
# 12  libhilti.dylib                      0x0000000102450a24 _ZN5hilti10ASTContext9_optimizeEPNS_7BuilderEPNS_6detail3cfg5CacheE + 292
# 13  libhilti.dylib                      0x000000010244eff8 _ZN5hilti10ASTContext10processASTEPNS_7BuilderEPNS_6DriverE + 2052
# 14  libhilti.dylib                      0x00000001029269a4 _ZN5hilti6Driver12compileUnitsEv + 160
# 15  libhilti.dylib                      0x00000001029279ec _ZN5hilti6Driver7compileEv + 68
# 16  libhilti.dylib                      0x00000001029274c8 _ZN5hilti6Driver3runEv + 436
# 17  hiltic                              0x0000000100655628 main + 412
# 18  dyld                                0x0000000183707e00 start + 6992
[1]    5404 abort      hiltic -p test.hlt
```

evantypanski:
Fixes #2433 

This is minor, but I caught it trying to test #2253 

Function parameters with 2 of the same aggregate type would say they're "cycles" even though they are separate parameters.

Note I couldn't easily reproduce the tuple case, but I included the fix there too, just in case. I feel like there's a way to hit it that I just didn't try.

Also not super sure how I feel about the getter/setter for the cycle detector.. but meh


hilti/toolchain/src/compiler/type-unifier.cc
```diff
@@ -53,9 +53,11 @@ class VisitorSerializer : public visitor::PostOrder {
     void operator()(type::Function* n) final {
         unifier->add("function(result:");
         unifier->add(n->result());
+        auto snapshot = unifier->getCycleDetector();
         for ( const auto& p : n->parameters() ) {
             unifier->add(", ");
             unifier->add(p->type());
+            unifier->setCycleDetector(snapshot);
```
bbannier: This seems both pretty low-level and also very repetitive, and I'd prefer something which steers us to the same pattern consistently, e.g., also for above `Bitfield` handling.

What do you think about instead capturing this pattern with a dedicated `Unifier` method, e.g.,

```cpp
    template<std::ranges::range R, typename Elem = std::ranges::range_reference_t<R>, std::invocable<Unifier&, Elem> Op>
    void forEach(const R& elements, Op op) {
        auto snapshot = _cd;

        for ( auto&& t : elements ) {
            op(*this, t);
            _cd = snapshot;
        }
    }
```
Maybe you have a better idea for a name. We'd use this here than like

```cpp
    void operator()(type::Function* n) final {
        unifier->add("function(result:");
        unifier->add(n->result());
        unifier->forEach(n->parameters(), [](auto&& unifier, auto&& p) {
            unifier.add(", ");
            unifier.add(p->type());
        });
        unifier->add(")");
    }
```

With that we wouldn't need the new getter and setter anymore.

hilti/toolchain/src/compiler/type-unifier.cc
bbannier, line 49: I think we should use the same pattern here, and if not we should call out in a comment why we deviate from it locally.

tests/hilti/optimization/dead-code-cfg-resolve.hlt
```diff
@@ -0,0 +1,19 @@
+# @TEST-DOC: Ensures removing unused declarations reresolves 'self' in the left-behind ctor without an internal error
+#
+# @TEST-EXEC: spicyc -p %INPUT
```
bbannier: Let's use `hiltic` for hilti tests.

tests/hilti/optimization/dead-code-cfg-resolve.hlt
```diff
@@ -0,0 +1,19 @@
...
+
+module Test {
+
+public type S = struct {
+	uint<64> i;
```
bbannier: We don't use tab here.

tests/hilti/optimization/dead-code-cfg-resolve.hlt
bbannier, line 1: This test currently does not have any explicit assertions. It documents the faulty behavior, but I wonder how useful it will be down the line.

If we cannot come up with anything dedicated to assert, let's at least add a comment `Regression test for #ABCD` so we know that this merely documents a previous bug, and might be a candidate for removal at a later point.

evantypanski: yeah it'll be tough to test because the expression later goes away in a later pass, it's just checking the AST between passes that asserts

I'll add the regression test marker, I didn't since the issue turned out to be separate. Seemed like there was enough there to me to show this is a very targeted test.

hilti/toolchain/src/compiler/type-unifier.cc
evantypanski: oh, I think I found a better solution, I was being a bit silly. There's already a mechanism for using an already-unified type with qualified types, but the one place I hit it used `type()->type()`, so that's just a typo. I'll fix it that way rather than exposing the cycle detector.

That's why the other cases worked fine! 🤦

bbannier: I like this so much better, thanks for digging in!

bbannier (approved):
🚀 

(module CI passing 🤓)

PR #2440, opened 2026-07-17
Fix a pile of Windows compiler warnings

timwoj:
Zeek builds the `spicyz` code with /W4 on Windows, which enables a bunch of compiler warnings not seen as part of the main Spicy build. This fixes a lot of them. The biggest set of the warnings are the unused arguments. I fixed these in three different ways depending on context:

- Removed the argument name altogether, if the method was a virtual definition with an empty body in a base class
- Commented the argument name, if the name was in a child class or if the argument was documented in the method's comment
- Added `[[maybe_unused]]` if the argument was used in a child class as an override of the method

If you'd like to stick to one of these specifically, I can adapt the rest. I could see making all of them `[[maybe_unused]]` but I'm not sure how that works for virtual methods (do all of the overrides need it too?).

It also fixes a bunch of random one-off kind of things. The majority the remaining warnings actually come from various uses of `tinyformat`. I think there should be a way to mark the `spicy/3rdparty` includes as `external`/`system` includes in the Zeek build, but all of my attempts to do so via CMake failed (marking the call to `add_subdirectory(3rdparty)` as `SYSTEM`, marking the individual includes in Zeek as `SYSTEM`).

### DIFF
--- a/hilti/toolchain/include/ast/declarations/module-uid.h
+++ b/hilti/toolchain/include/ast/declarations/module-uid.h
@@ -107,9 +107,9 @@ struct UID {
     explicit operator bool() const { return ! id.empty(); }
 
 private:
-    ID _makeUnique(const ID& id) const {
-        auto& x = _id_to_counter[id];
-        return ++x > 1 ? ID(util::fmt("%s_%" PRIu64, id, x)) : id;
+    ID _makeUnique(const ID& id_) const {
+        auto& x = _id_to_counter[id_];
+        return ++x > 1 ? ID(util::fmt("%s_%" PRIu64, id_, x)) : id_;
     }
 
     inline static uint64_t _no_file_counter = 0; // global counter for creating unique paths for modules without any
--- a/hilti/toolchain/include/ast/operator.h
+++ b/hilti/toolchain/include/ast/operator.h
@@ -357,7 +357,7 @@ public:
      * Performs semantics validation of an instantiated operator. To record any
      * errors, add them to the given AST node.
      */
-    virtual void validate(expression::ResolvedOperator* n) const {};
+    virtual void validate(expression::ResolvedOperator* /*n*/) const {};
 
     /** Instantiates the operator as an AST node, given specific operand expressions. */
     virtual Result<expression::ResolvedOperator*> instantiate(Builder* builder,
--- a/hilti/toolchain/include/ast/types/union.h
+++ b/hilti/toolchain/include/ast/types/union.h
@@ -39,7 +39,7 @@ public:
     unsigned int index(const ID& id) const {
         for ( const auto&& [i, f] : util::enumerate(fields()) ) {
             if ( f->id() == id )
-                return i + 1;
+                return static_cast<unsigned int>(i + 1);
         }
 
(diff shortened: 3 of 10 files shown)


hilti/toolchain/include/ast/declarations/module-uid.h
```diff
@@ -107,9 +107,9 @@ struct UID {
     explicit operator bool() const { return ! id.empty(); }
 
 private:
-    ID _makeUnique(const ID& id) const {
-        auto& x = _id_to_counter[id];
-        return ++x > 1 ? ID(util::fmt("%s_%" PRIu64, id, x)) : id;
+    ID _makeUnique(const ID& new_id) const {
```
bbannier: Nit: `new_id` suggests to me that this creates or updates state somewhere while it is strictly passthrough. Would using `id_` instead work for you?

hilti/toolchain/include/ast/operator.h
```diff
@@ -357,7 +357,7 @@ class Operator {
      * Performs semantics validation of an instantiated operator. To record any
      * errors, add them to the given AST node.
      */
-    virtual void validate(expression::ResolvedOperator* n) const {};
+    virtual void validate(expression::ResolvedOperator*) const {};
```
bbannier: Nit: for local consistency this should be

```suggestion
    virtual void validate(expression::ResolvedOperator* /*n*/) const {};

File: hilti/toolchain/include/ast/types/union.h [hunk 3]
```diff
@@ -39,7 +39,7 @@ class Union : public UnqualifiedType {
     unsigned int index(const ID& id) const {
         for ( const auto&& [i, f] : util::enumerate(fields()) ) {
             if ( f->id() == id )
-                return i + 1;
+                return static_cast<int>(i + 1);
```
bbannier: This seems wrong. `i` is an `unsigned long` for me, and the function returns an `unsigned int`. Casting either `i` or the sum to `unsigned int` would make sense to me, but I don't understand why we'd want `int`.

bbannier (changes requested):

hilti/toolchain/include/ast/visitor.h
bbannier, line 1: Introducing this seems correct, but as fallout we now need to mark the existing `virtual` dtr in a child class `override`. Could you take care of that before merging? If unsure simply rerun CI on your final version, we expect everything to pass.

```
/home/runner/work/spicy/spicy/hilti/toolchain/include/hilti/ast/visitor.h:179:13: error: prefer using 'override' or (rarely) 'final' instead of 'virtual' [modernize-use-override,-warnings-as-errors]
  179 |     virtual ~Visitor() = default;
      |     ~~~~~~~ ^
      |                        override
```

bbannier (dismissed):
LGTM but for the lint failure, could you fix that before merging?

hilti/toolchain/include/ast/visitor.h
bbannier: Actually just let me fix this up while I make the merge for you. That is, unless you were eager to do your first merge to Spicy.

PR #2444, opened 2026-07-22
Add support for unconditionally putting struct or unit types onto the heap

bbannier:
In #1781 we tweaked when a unit would be put on the heap with most of the behavior there driven by heuristics. This PR adds an explicit know as well.


doc/programming/parsing.rst
```diff
@@ -923,6 +923,24 @@ Unit types support the following type attributes:
...
+    Forces the unit's internal storage to be allocated on the heap rather than
+    the stack. This can be useful for large units to avoid stack pressure.
+    Example:
+
+    .. spicy-code::
+
+        type Foo = unit {
+            a: uint32;
+            data: bytes &eod;
+        } &on-heap;
```
bbannier, line 938: How does this look to you @evantypanski? It fits attributes like e.g., `&max-size` for units, and the same syntax also works for struct types, but it could also be a unit attribute here. I have the feeling that this is obscure enough that the approach here is fine.

evantypanski: Hm, I don't hate it, but it feels like we could make something with a bit more granularity that mimics other languages a bit more.

For example: same idea, but just with a new type:

```
module Test;

type Foo = unit {
	a: uint32;
	data: bytes &eod;
};

type FooBox = Box<Foo>;

public type X = unit {
	: FooBox;
};
```

Now callers could use either the `Box` version themselves, the provided alias, or `Foo` if they want it on the stack. TBH it could basically be syntactic sugar, too.

Though, that might just be too much work to be worth it for this rare case. I'd be fine if you went down the attribute approach, with a preference towards doing this as an explicitly "boxed type" instead.

bbannier: Currently (even before this patch) the `&on-heap`-ness of a type is a global property, and we automatically always allocate every value of a type on the heap for two cases. What I did here was add a way to tweak that automatic behavior as an escape hatch.

https://github.com/zeek/spicy/blob/0f5101255394d427e03ca6a9bb4b3f89553e3aff/spicy/toolchain/src/compiler/codegen/codegen.cc#L185-L198

Before we introduced that, every unit value was always allocated on the heap which caused potentially expensive heap allocations and also forced each access through an indirection, so moving away from that was well motivated. The slightly flawed framing was that we could decide statically where heap allocation "made sense", but did not take stack exhaustion into account. That said, I do not like how users have to think about allocation when writing their grammar (via boxing, an attribute, ...) since with that things suddenly become very "programmery" at a surprising layer which is not only subtle, but probably also pretty unintuitive at that point (I was unsure where to even document it since it doesn't naturally fit into thinking about grammar rules).

Still, I think in our current world where heap-ness is a type, not value property, we need a type-level way to express this. Whatever we do should work with entry-point `public` units, so working on just fields is insufficient. Creating alias with a new box type or just already existing reference syntax[^ref] (I think this should be equivalent, but maybe I miss something) seems clean, but the need to put a type on the heap seems already dictated by the type's structure (e.g., has many big fields), so putting that elsewhere in an alias seems awkward and too decoupled.

[^ref]: We could make references parseable types to support syntax like this
    
    ```ruby
    public type X = unit {
        : Foo&; // UNSUPPORTED
    };
    
    # Also could allow introducing references
    public type Foop = Foo&;
    ```
    
    I suspect that even if we'd introduce this, the alias part could still hit #2447.

All that said, the current global approach seems a bit heavy-handed, and we probably should separately look into how we could make this more granular. I suspect that to be a bigger chunk of work.
All this to say, we should probably make references parseable types to simplify this. The only way to make this work for entry point units would be a type alias which I find a bit surprising.

evantypanski: Okay, I think I've convinced myself that exposing `&on-heap` seems good. You're right that the type-level way to express heapness is necessary. Expressing it as a value property is useful for some contexts but not all (eg if you know it's deeply recursive or something). The more likely scenario is people will want to express it as a type property, so that makes sense to me.

then it's already easy to do since it's supported internally.

I was held up because of how leaky this feels. But, we're already leaking the small stack size anyway (via the error). I guess the preferred knob might have been to configure what size we would put objects on the heap, but that is more complicated and could be complementary to the attribute.

One point worth noting, I'd like if we mentioned it in the `"not enough stack space remaining"` error message. That way it's more visible to users if they hit the issue, I don't want this documentation front-and-center because it really is just a nitty-gritty detail. Only caveat is we don't know what units would most benefit from `&on-heap` at that point, just that we're under the minimum stack size required to continue. Bleh.

tl;dr: I'm on board with the attribute. Actionable part would be to express this to users if they hit "not enough stack space remaining" errors.

bbannier: I added it to the error message, but like you wrote, we can neither tell the user which unit to adjust nor even make sure this is really due to a unit and not just plain user-written recursion. I am resolving this thread, PTAL so we can get this merged.

evantypanski (dismissed):
LGTM

PR #2446, opened 2026-07-23
Fix parsing of regexps with capture groups in lookahead

### LINKED ISSUE #2438: IndexError when using capture groups inside unit inside switch
I wanted to suggest using capture groups on https://github.com/zeek/zeek/pull/5669, but there's something going on:

The following parser fails with:
```
$ echo -n "123 " | ../../build/bin/spicy-driver ./x.spicy 
[error] processing failed with exception of type hilti::rt::IndexError: vector index 1 out of range (./x.spicy:16:9-16:41)
```

When the `c: C` unit is commented out of the switch:
```
$ echo -n "123 " | ../../build/bin/spicy-driver ./x.spicy 
got it, 123 , 123 , 123
A%done
```

It also succeeds parsing when removing the `$1` from the `print` in unit `B`, so it's something with the capture group access.

Expectation: Accessing `$1` within the hook of the `len` field works.

---

```
$ cat x.spicy 
module CaptureGroupInSwitch;

public type A = unit {
    switch {
        -> b: B;
        -> c: C;  # commenting this out makes the capture group in `B` work.
    };

    on %done {
        print "A%done";
    }
};

type B = unit {
    len: /([1-9][0-9]*) / {
        print "got it", self.len, $$, $1;
    }
};

type C = unit {
    open_bracket: b"<" {
        print "open_bracket";
    }
};
```

bbannier:
The lookahead fast path consumed the look-ahead token and skipped the regex loop, so capture groups were never populated. For regexp fields that have captures, we now skip the fast path and let the regex loop run instead so captures are filled normally.

Fixes #2438.


tests/Baseline/spicy.types.unit.synchronize-on-gap/gap_between_matches.log
```diff
@@ -23,13 +22,11 @@
 [spicy-verbose]     successfully synchronized
 [spicy-verbose]     - state: type=sync::X1 input="BC" stream=0xXXXXXXXX offsets=1026/0/1026/1028/1028 chunks=1 frozen=no mode=default trim=yes lah=1 lah_token="B" recovering=no
 [spicy-verbose]     - parsing production: Ctor: _anon -> /(A|B|C)/ (regexp) (container 'xs')
-[spicy-verbose]       - consuming look-ahead token
 [spicy-verbose]       - trimming input
 [spicy-verbose]     - got container item
 [spicy-verbose]     - state: type=sync::X1 input="C" stream=0xXXXXXXXX offsets=1027/0/1027/1028/1028 chunks=1 frozen=no mode=default trim=yes lah=1 lah_token="C" recovering=no
 [spicy-verbose]     - state: type=sync::X1 input="C" stream=0xXXXXXXXX offsets=1027/0/1027/1028/1028 chunks=1 frozen=no mode=default trim=yes lah=1 lah_token="C" recovering=no
 [spicy-verbose]     - parsing production: Ctor: _anon -> /(A|B|C)/ (regexp) (container 'xs')
-[spicy-verbose]       - consuming look-ahead token
```
evantypanski, line 32: do you think we could change the test to not capture instead? It doesn't seem to use the capture

evantypanski (dismissed):

tests/Baseline/spicy.types.unit.synchronize-on-gap/gap_between_matches.log
bbannier: Good idea. I added a commit doing that before the code changes. This also causes a baseline change, but it is unrelated to the code changes.

PR #2451, opened 2026-07-28
Unvendor nlohmann/json again

bbannier:
In f7d3d6d786df655a9b9c8d3b7f2ca8c98c05b9ff we switched to a vendored nlohmann/json to work around its very fat Git history. This was never really needed since Git submodules can be declared `shallow = True` so they are not pulled with their full history. Additionally it also made maintenance of this dependency much, much harder since we cannot simply bump a submodule anymore. Undo that here, and bump us to the latest release.


bbannier:
@timwoj since you raised the original concern which made us switch to vendoring, does this look reasonable to you? AFAICT this should still work fine while simplifying maintenance for us.

evantypanski (dismissed):
Cool, I like it

we don't have any users of the 3rdparty headers in Zeek or anything, correct? So no need to worry about the different paths?

bbannier:
> we don't have any users of the 3rdparty headers in Zeek or anything, correct? So no need to worry about the different paths?

Not intentionally, but I created https://github.com/zeek/zeek/pull/5726 to check. If we find something there I'd rather fix the bump than making things here more complicated, these are internal implementation details anyway.

timwoj:
Looks fine to me. I don't remember why I made the argument for the submodule being too big, but if it works now the change is good :shrug:

timwoj (dismissed):

bbannier:
> > we don't have any users of the 3rdparty headers in Zeek or anything, correct? So no need to worry about the different paths?
> 
> Not intentionally, but I created [zeek/zeek#5726](https://github.com/zeek/zeek/pull/5726) to check. If we find something there I'd rather fix the bump than making things here more complicated, these are internal implementation details anyway.

Okay, this broke a bunch of stuff since the way Spicy or Zeek set up their drivers was fragile. I have added a commit here, and another one to the bump PR and this seems to work now.

bbannier:
I need to iterate on this. While installing 3rdparty dependencies into the global prefix works, it begs for path conflicts especially for popular dependencies, so unless we have a way to unbundle each dependency we do that. I'll check whether there is a cleaner approach.

bbannier:
> I'll check whether there is a cleaner approach.

I now changed this to install into `hilti/rt/3rdparty` again. I stayed with using "bare" includes like e.g., `#include <nlohmann/json.hpp>` so we loose the visual distinction of 3rdparty deps in installed headers we previously had. To make that work safely with systems which already have these headers in their prefix (possibly with different versions), I needed to tweak the include order slightly. This is a bit more magical than I had initially hoped, but IMO not worse and easier to maintain than the previous approach. For the [dummy Zeek bump](https://github.com/zeek/zeek/pull/5726) CI also seems content. Can you have another look @evantypanski?

evantypanski (dismissed):
I'm happy with this. That `initLocation` change makes me think `Configuration` might be able to drop that bool constructor, but it seems iffy and like it might take some work so not worth considering now.

TBH I think it's more obvious they're 3rdparty at a glance since they're no longer under `hilti/`...

PR #2453, opened 2026-07-28
Do not include debug info in non-debug HLTO files

### LINKED ISSUE #62: Optimize JIT code
Currently we compile individual C++ units with `-O3` when JITing in non-debug mode, we don't however apply any cross-module optimizations after linking them all together. We should add that when the `optimize` flag is set. 

We should probably also make the `-O3` dependent on the `optimize` flag as well and/or join the `debug`/`optimize` flag into one.

### LINKED ISSUE #1102: Possibly omit debug information from non-debug HLTO files
With #1092 we now omit debug information from Spicy and HILTI libraries in `Release` builds.

We should as a follow-up also omit debug information from HLTO files generated by e.g., `spicyc` and `spicyz` if Spicy itself was built without symbols (i.e., in `Release` mode). It looks like we might be able to achieve this by passing `-s` to the link step.

bbannier:
Including debug information significantly increases the size of these artifacts, something users have often complained about. Additionally, compiling with both `-O3` and `-g` can make release builds even slower.

Since no user-visible feature depends on this, codegen is mostly identical between between, and I do not recall ever being unable to debug an issue which required debug symbols in release HLTO files, simply drop the flag. If this is really needed, we could always patch it back in temporarily while debugging without affecting all users.

Closes #62.
Closes #1102.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -9,6 +9,11 @@ Version 1.17 (in progress)
 
 .. rubric:: Changed Functionality
 
+- GH-62, GH-1102: Release builds no longer include debug information in
+  JIT-compiled code and the hilti-rt library which significantly reduces
+  the size of ``.hlto`` files. Users who need debug information in JIT-compiled
+  code can use ``spicyc -d``, which retains full debug symbols.
+
 .. rubric:: Bug fixes
 
 .. rubric:: Documentation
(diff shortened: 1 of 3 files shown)


evantypanski (approved):
If I ever debug I wouldn't want `-O3` anyway, this is only positive IMO

PR #2457, opened 2026-07-29
Fix namespace for type code emission during C++ codegen

### LINKED ISSUE #2456: Compilation failure when trying to convert a field value to a reference
Attempting to put a unit field into a references in the following example produces a compilation failure when I would have expected it to work:

```ruby
# foo.spicy
module Consumer; # works for me if named `foo`.

public type Bar = unit {
    a: Outer;
};

type Outer = unit {
    a: Foo &convert=new Foo($$);
};

type Foo = unit {};
```

```console
$ spicyc -dj foo.spicy -o foo.hlto

=== Uh oh ... C++ compilation failed ===

This is unexpected and should not happen. Please help us fix the issue
by opening a ticket at https://github.com/zeek/spicy/issues/new.

Please include the content of the following file with your report:

  "/private/tmp/repro/hilti-jit-error.log"

That file contains diagnostics from the C++ compiler. Please also
include your Spicy code that triggered this error, ideally reduced to
a small snippet still exhibiting the issue if you can.

===

make: *** [foo.hlto] Error 1
```

```
Apple clang version 17.0.0 (clang-1700.0.13.5)
Target: arm64-apple-darwin24.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
HILTI version 1.16.1

## [job 3] ccache clang++ -c -I/opt/homebrew/Cellar/zeek/8.2.1/include -fPIC -std=c++20 -g -O0 -fvisibility=hidden -Wno-invalid-offsetof -I/opt/homebrew/Cellar/zeek/8.2.1/include -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk -Wall -Wno-unused -funsigned-char -Wno-c2y-extensions -Wno-register -Werror=vla -funsigned-char -Wno-c2y-extensions -o /var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/Consumer_1a8c9128dba77178-410b2ecd5a942a10.cc_1a8c9128dba77178.o /private/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/Consumer_1a8c9128dba77178-410b2ecd5a942a10.cc

/private/var/folders/ht/vtb0rkbd4ws15vx02xdb84zw0000gn/T/Consumer_1a8c9128dba77178-410b2ecd5a942a10.cc:149:5: error: use of undeclared identifier 'Foo'
  149 |     Foo::Foo() {
      |     ^
1 error generated.
```

That changing naming changes behavior reads suspicious to me. This first broke for me with 370254bf89720377e050c50fc0d5bac0cfdd588c which appeared with spicy-1.12.0.

This came from @mohandhawan.

bbannier:
We now enter the correct namespace before emitting a type declaration's implementation code during `Phase::Functions`. Previously the formatter stayed in whatever namespace the previous declaration left active, causing C++ compilation failures when declaration ordering placed a type's code after an unrelated namespace block.

The sensitivity to module naming (e.g., "Consumer" failing while "foo" works) comes from declaration insertion order: the dependency graph traversal places different declarations between a type's struct code and its implementation depending on names, which changes whether an intervening declaration switches the formatter's active namespace away.

Closes #2456.

> [!NOTE]
> All code changes in this PR were generated with Claude Sonnet 4.6. I reviewed all changes and am accountable.

### DIFF
--- a/hilti/toolchain/src/ast/ast-context.cc
+++ b/hilti/toolchain/src/ast/ast-context.cc
@@ -142,8 +142,17 @@ public:
     void operator()(declaration::Type* n) final { insert(n); }
 
     void operator()(QualifiedType* n) final {
-        if ( n->isExternal() )
-            follow(n->type());
+        if ( n->isExternal() ) {
+            auto* t = n->type();
+            if ( auto* decl = t->typeDeclaration() ) {
+                // Skip self-references, e.g., a unit's parse methods referencing its own type.
+                if ( ! cd.haveSeen(decl) )
+                    insert(decl);
+                follow(decl);
+            }
+            else
+                follow(t);
+        }
     }
 
     void operator()(expression::Name* n) final {
--- a/hilti/toolchain/src/compiler/cxx/unit.cc
+++ b/hilti/toolchain/src/compiler/cxx/unit.cc
@@ -186,6 +186,7 @@ void Unit::_emitDeclarations(const cxxDeclaration& decl,
 
             else if ( phase == Phase::Functions ) {
                 if ( d.code.size() && ! prototypes_only ) {
+                    f.enterNamespace(d.id.namespace_());
                     f << d.code << eol();
                 }
             }
--- a/tests/spicy/types/unit/convert-field-to-ref.spicy
+++ b/tests/spicy/types/unit/convert-field-to-ref.spicy
@@ -4,7 +4,14 @@
 # introduced the issue, validate tests still reproduce the issue before
 # changing names.
 #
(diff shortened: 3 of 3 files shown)


bbannier:
Moving this back to draft status since @mohandhawan reported seeing more, similar issues with the current state on his unminimized input.

Mohan-Dhawan (commented):
The latest fix works for me.

evantypanski (approved):

PR #2461, opened 2026-07-31
Cover all cases when handling arguments to `new`

### LINKED ISSUE #2460: Object allocation on the heap via `&convert=new T($$)` does not work
The `&convert new Foo($$)` idiom to put elements on the heap does not work. Here is a reproducer `repro.spicy`:
```
module Test;

type Inner = unit {
    n: uint8;
};

public type Outer = unit {
    buggy: Inner &convert=new Inner($$);
    good: Inner;

    on %done {
        print "buggy", self.buggy;
        print "buggy.n", self.buggy.n;
        print "good", self.good;
        print "good.n", self.good.n;
    }
};
```
When invoked via
```
$ printf '\x41\x41' | spicy-driver repro.spicy
```
it gives
```
buggy, [$n=(not set)]
[error] processing failed with exception of type hilti::rt::UnsetOptional: unset optional value (repro.spicy:13:9-13:38)
```
Commenting out the `print` statement for the `buggy` field, gives
```
$ printf '\x41\x41' | spicy-driver repro.spicy
good, [$n=65]
good.n, 65
```

bbannier:
This started out as a bugfix for #2460. Along the way I figured out that our implementation of operator `new` was silently discarding arguments for unsupported uses, so this now does two things:

- Add a new feature to support copy construction with `new T(t)`. The docs did not clearly rule this out, and I think users reasonably expected support for this.
- Add validation to reject remaining unsupported uses instead of silently discarding arguments.

This does not technically need the changes from #2456, but I stacked it so things can be tested together easily.

Closes #2460.

> [!NOTE]
> All code changes in this PR were generated with Claude Opus 4.6. I reviewed all changes and am accountable.


evantypanski (approved):
this looks good to me

just one point: this is copied, so I would be reluctant to use this as a fix for stack pressure (unless I'm mistaken). I believe we should support it, though (and especially not silently discard the params..)

PR #2463, opened 2026-08-02
Require GCC 12 for C++20 builds

### LINKED ISSUE #2428: Check that documented minimum required compiler versions are accurate
We document minimum compiler version requirements and check the same numbers in https://github.com/zeek/spicy/blob/fd21eab41a60e052384e90b0158854d4d14749ca/cmake/CheckCompiler.cmake#L4-L16

If we keep using these numbers we should make sure they actually reflect our real requirements.

> Also: do we know if Clang 9 works?
> 
> https://github.com/zeek/spicy/blob/fd21eab41a60e052384e90b0158854d4d14749ca/doc/installation.rst?plain=1#L36
> 
> Seems like that number was added 6 years ago and never changed. I would be shocked if it worked, considering where we are now. May be worth updating. We have more than doubled that version!

_Originally posted by @evantypanski in https://github.com/zeek/spicy/pull/2426#pullrequestreview-4672398536_

aouxwoux:
## what changed

the docs and ci already say gcc 12 is the supported gcc baseline, but the cmake guard still allowed gcc 8.3. this updates the guard to 12.0 and removes the old 9.1.1 note.

## why

with the old check, an unsupported gcc version could get past configure and fail later while compiling c++20 code.

## checks

- git diff --check passes.
- the local machine does not have cmake or a c++ toolchain installed, so the native build is left to repository ci.

fixes #2428

### DIFF
--- a/cmake/CheckCompiler.cmake
+++ b/cmake/CheckCompiler.cmake
@@ -9,10 +9,8 @@ set(clang_minimum_version "9.0")
 # 11.0 comes with 10.15 (Catalina) and works
 set(apple_clang_minimum_version "11.0")
 
-# 7.x is not compiling HILTI/Spicy, although the standard library seems to be recent enough.
-# 8.x is untested.
-# 9.1.1 works.
-set(gcc_minimum_version "8.3")
+# GCC 12 is the minimum supported version for compiling HILTI/Spicy with C++20.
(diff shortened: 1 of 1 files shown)


aouxwoux:
this was already fixed on main in d2e1cd157, so closing this rather than asking for review on a duplicate.

PR #2469, opened 2026-08-10
Switch from checkbashisms.pl to shellcheck

bbannier:
We were previously linting shell scripts with checkbashisms.pl which
implements a subset of functionality already present in shellcheck at
most. It also is a pretty specific tool and editors might more often
come with shellcheck integration instead.

This commit switches us over to checking with shellcheck instead. In
order to make that work we fix issues reported by it, either by
implementing a proper fix or by targetted local suppressions.


evantypanski (approved):
👍 from me

PR #2471, opened 2026-08-11
Incremental perf improvements to runtime functions used during parsing

bbannier:


evantypanski (approved):
cool these seem nice. I've considered adding branch hints here and there, I was also wondering if we could use them in generated code and guard off the slow paths. Here you use branch hints with an `inline` function and a slow function, which is probably equivalent, but I can't say for sure :)

Another point, the debug stream stuff should help. I actually noticed somewhere from 3-5% of time in the SSL parser was spent with those a while ago. Wonder if there's another way to completely cut that time out, but this should help.

PR #2474, opened 2026-08-16
Make toolchain headers compatible with C++23

### LINKED ISSUE #2473: Host application headers do not compile as C++23 with libc++
We are working on embedding Spicy into [Tenzir](https://github.com/tenzir/tenzir) through the host application API. Tenzir uses C++23, while Spicy itself is currently built as C++20.

Including the Spicy host headers from a C++23 translation unit fails with libc++. The minimal reproducer is:

```cpp
#include <spicy/spicy.h>

int main() {}
```

I tested it with Spicy 1.16.0 installed through Homebrew:

```console
$ spicy-config --version
1.16.0 (b19c31d8d)

$ c++ $(spicy-config --cxxflags) -std=c++20 \
    -fsyntax-only spicy-cxx23.cc
# succeeds

$ c++ $(spicy-config --cxxflags) -std=c++23 \
    -fsyntax-only spicy-cxx23.cc
# fails
```

`-std=c++23` is deliberately placed after the flags from `spicy-config`, which currently include `-std=c++20`.

I reproduced this with Apple Clang 21 and Homebrew LLVM 22.1.8. Both use libc++.

There are two errors.

First, `hilti/base/logger.h` defines an inline static `std::unique_ptr<Logger>` inside `Logger`:

```cpp
class Logger {
  // ...
  constinit inline static std::unique_ptr<Logger> _singleton;
};
```

libc++ marks the `std::unique_ptr` destructor constexpr in C++23. This causes its deleter to be instantiated while `Logger` is still incomplete:

```text
error: invalid application of 'sizeof' to an incomplete type 'hilti::Logger'
note: in instantiation of member function
      'std::unique_ptr<hilti::Logger>::~unique_ptr' requested here
note: definition of 'hilti::Logger' is not complete until the closing '}'
```

Second, `hilti/rt/fiber.h` uses `std::exception_ptr` without including `<exception>`:

```text
error: field has incomplete type 'std::exception_ptr'
note: forward declaration of 'std::exception_ptr'
```

I tested these local changes:

```cpp
class Logger {
  // ...
  static std::unique_ptr<Logger> _singleton;
};

constinit inline std::unique_ptr<Logger> Logger::_singleton;
```

and:

```cpp
#include <exception>
```

With both changes, the minimal program compiles successfully as C++23. The corresponding patterns also appear to remain on the current `main` branch.

Is compiling a host application against the Spicy headers as C++23 intended to be supported, even while Spicy itself remains built as C++20? Is mixing a C++23 host translation unit with the C++20-built Spicy libraries considered supported?

We currently work around this by isolating all Spicy headers and calls in a narrow C++20 bridge library. If direct C++23 consumption is intended to work, I would be happy to prepare a patch for the two header issues.

mavam:
## Summary

- Include `<exception>` where `std::exception_ptr` is used.
- Define `Logger::_singleton` after `Logger` is complete.
- Define `printer::detail::State::current` after `State` is complete.
- Add a conditional C++23 header compilation test.

This does not change Spicy’s required C++ standard or claim official C++23 support. The regression test is skipped when the configured compiler does not accept `-std=c++23`.

Fixes #2473.

## Validation

- Normal C++20 build succeeded.
- Complete Spicy/HILTI toolchain build as C++23 succeeded.
- All five unit-test suites passed under the C++23 build.
- Full BTest suite passed: 876 successful, 36 skipped.
- Installed `<spicy/spicy.h>` compiled successfully as C++23.
- Pre-commit checks passed.

### DIFF
--- a/tests/spicy/doc/toolchain-headers-cxx23.cc
+++ b/tests/spicy/doc/toolchain-headers-cxx23.cc
@@ -1,0 +1,9 @@
+// @TEST-REQUIRES: ! is-windows
+// @TEST-REQUIRES: have-cxx23
+// @TEST-REQUIRES: using-build-directory
+// @TEST-EXEC: "$(spicy-config --cxx)" $(spicy-config --cxxflags) $(printf -- '-I%s ' $(spicy-config --include-dirs-toolchain)) -std=c++23 -fsyntax-only %INPUT
+
+#include <hilti/compiler/printer.h>
+#include <spicy.h>
+
+int main() {}
--- a/tests/spicy/doc/toolchain-headers-cxx23.cc
+++ b/tests/spicy/doc/toolchain-headers-cxx23.cc
@@ -1,9 +1,0 @@
-// @TEST-REQUIRES: ! is-windows
-// @TEST-REQUIRES: have-cxx23
-// @TEST-REQUIRES: using-build-directory
-// @TEST-EXEC: "$(spicy-config --cxx)" $(spicy-config --cxxflags) $(printf -- '-I%s ' $(spicy-config --include-dirs-toolchain)) -std=c++23 -fsyntax-only %INPUT
-
-#include <hilti/compiler/printer.h>
-#include <spicy.h>
-
-int main() {}
--- a/hilti/runtime/include/fiber.h
+++ b/hilti/runtime/include/fiber.h
@@ -3,6 +3,7 @@
 #pragma once
 
 #include <csetjmp>
+#include <exception>
 #include <memory>
 #include <optional>
 #include <string>
--- a/hilti/toolchain/include/base/logger.h
+++ b/hilti/toolchain/include/base/logger.h
@@ -240,8 +240,10 @@ private:
 
     std::map<logging::DebugStream, size_t> _debug_streams;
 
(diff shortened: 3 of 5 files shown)


tests/spicy/doc/toolchain-headers-cxx23.cc
bbannier, line 1: Let's not add this test, we do not want to commit to C++23 support. The helper can also go away.

bbannier (commented):
Thanks! Am I correct to assume that this was prepared without an LLM? Else please follow [Zeek's AI policy](https://github.com/zeek/zeek/blob/master/AI_POLICY.md).

mavam:
Yes, this was AI-assisted. I used OpenAI Codex’s `gpt-5.6-sol` model through the pi coding agent to reproduce the C++23 failures, identify and implement the three header fixes, add the regression test, and draft the issue and PR text. I reviewed the changes and test results and understand what each change does. Apologies for not disclosing this initially.

tests/spicy/doc/toolchain-headers-cxx23.cc
mavam: Removed the test and helper in 62ff8eef0. The patch now contains only the three header fixes.

bbannier (approved):

PR #2478, opened 2026-08-24
Treat `void` unit fields as epsilon

### LINKED ISSUE #2477: Adding : void { } "field" skips parsing preceding array/vector
This one is strange. I'm working on the PostgreSQL parser in Zeek and wanted to report a weird from a `: void { }` block only to eventually find that adding that `: void { ... }` block somehow results in skipping  parsing of the prior array.

Minimal reproducer:
```
module PostgreSQL;

type ErrorIdentifiedField = unit {
    code: uint8;
    value: bytes &until=b"\x00";
};

public type ErrorResponse = unit {
    : ErrorIdentifiedField[] foreach {
        print "error", $$;
    }

    # Having this here seems to make it skip the array above.
    : void {}

    : skip b"\x00";
};
```
```
# two "errors" a and c with null terminated strings b and d.
$ printf "ab\x00cd\x00\x00" | ../../build/bin/spicy-driver ./error.spicy 
[error] processing failed with exception of type spicy::rt::ParseError: expected bytes literal "\x00" but input starts with "a" (./error.spicy:16:12-16:18)
```

As if the the input was never consumed.. Removing the `: void {}` "field" or moving it past the `: skip b"\x00"` produces the following expected output.
```
$ printf "ab\x00cd\x00\x00" | ../../build/bin/spicy-driver ./error.spicy 
error, [$code=(optimized out), $value=b"b"]
error, [$code=(optimized out), $value=b"d"]
```

```
$ ../../build/bin/spicy-driver --version
spicy-driver v1.17.0-dev.241 (89a1ea05a)
```

This seems to go back to Zeek 8.0 as well (tested also with spicy-driver v1.14.1 (6e88dbd98)).

bbannier:
We would previously treat `void` fields as producing normal variables. Since they consume no data this is incorrect and also confuses lookahead parsing.

This patch switches `void` fields to instead generate epsilon productions.

Closes #2477.

### DIFF
--- a/spicy/toolchain/src/compiler/codegen/grammar-builder.cc
+++ b/spicy/toolchain/src/compiler/codegen/grammar-builder.cc
@@ -319,6 +319,8 @@ struct Visitor : public visitor::PreOrder {
 
     void operator()(hilti::declaration::Type* n) final { result = pf->createProduction(n->type()); }
 
+    void operator()(hilti::type::Void*) final { result = std::make_unique<production::Epsilon>(context()); }
+
     void operator()(type::Unit* n) final {
         // Note: We can't use the cache's getOrCreate() here because of the
         // unique_ptr storage semantics.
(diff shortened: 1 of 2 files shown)


evantypanski (approved):
yeup looks right to me assuming CI passes

PR #2482, opened 2026-08-25
Fix Clang-23 RC build

evantypanski:
draft since I'm going to look at other stuff too (and so I make sure to come back to this), this isn't urgent.

Just tested Clang 23 RC3 and hit an error building Zeek:

```
In file included from /Users/etyp/src/zeek/auxil/spicy/hilti/runtime/src/backtrace.cc:6:
/Users/etyp/src/zeek/auxil/spicy/hilti/runtime/include/hilti/rt/backtrace.h:57:9: error: use of undeclared identifier 'free'
   57 |         free(dname); // NOLINT
      |         ^~~~
1 error generated.
```

Also changed this to fix some warnings, as well.

Note this fixes warnings in fiber/ - I did that as patches (see https://github.com/zeek/fiber/pull/1). Unless they respond upstream to the other patch, I don't see much reason to not mildly maintain it ourselves.

### DIFF
--- a/hilti/runtime/CMakeLists.txt
+++ b/hilti/runtime/CMakeLists.txt
@@ -71,16 +71,17 @@ foreach (lib hilti-rt hilti-rt-debug)
     target_include_directories(${lib}-objects BEFORE
                                PUBLIC $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>)
     target_include_directories(
-        ${lib}-objects BEFORE PUBLIC $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/3rdparty>
-                                     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/3rdparty/json/include>)
+        ${lib}-objects SYSTEM BEFORE
+        PUBLIC $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/3rdparty>
+               $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/3rdparty/json/include>)
 
     add_dependencies(${lib}-objects jrx-objects)
-    target_include_directories(${lib}-objects BEFORE
+    target_include_directories(${lib}-objects SYSTEM BEFORE
                                PRIVATE ${PROJECT_SOURCE_DIR}/3rdparty/justrx/include)
 
     add_dependencies(${lib}-objects fiber)
     target_include_directories(
-        ${lib}-objects
+        ${lib}-objects SYSTEM BEFORE
         PRIVATE ${PROJECT_SOURCE_DIR}/3rdparty/fiber/include
                 ${PROJECT_SOURCE_DIR}/3rdparty/fiber/deps/cxx-header-utils/include
                 ${PROJECT_SOURCE_DIR}/3rdparty/utfcpp/source)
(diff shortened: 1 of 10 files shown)


evantypanski:
Marked this as ready. Tested brew's LLVM-23 and it works with this branch without warnings.

only iffy spot is the fiber bump. We could drop that since it was there before, it just cleans up some warnings (and I think @ckreibich complained about spicy warnings to me relatively recently ;) )

hilti/runtime/CMakeLists.txt
```diff
@@ -70,9 +70,15 @@ foreach (lib hilti-rt hilti-rt-debug)
                                PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>)
     target_include_directories(${lib}-objects BEFORE
                                PUBLIC $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>)
+    target_include_directories(${lib}-objects BEFORE
+                               PUBLIC $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/3rdparty>)
```
bbannier, line 74: Why don't we mark all of `3rdparty` `SYSTEM`? It seems with that and by disabling warnings for `fiber/` a la

```cmake
+set_property(TARGET fiber APPEND PROPERTY COMPILE_OPTIONS "-w")  # GCC/Clang option
````

we should be able to avoid adding more patches to our fork (might also want something like `-isystem` for JIT flags, not sure).

File: 3rdparty/fiber
Comment (bbannier) [inline 3rdparty/fiber:1]: I'd really would like to avoid accumulating more patches in our fork since I don't think we want to maintain it longterm. I still hope that upstream will come back and merge https://github.com/simonfxr/fiber/pull/5.

File: spicy/toolchain/src/compiler/codegen/parsers/literals.cc [hunk 2]
```diff
@@ -50,10 +50,6 @@ struct Visitor : public visitor::PreOrder {
     auto context() { return pb()->context(); }
     auto pushBuilder(std::shared_ptr<Builder> b) { return pb()->pushBuilder(std::move(b)); }
     auto pushBuilder() { return pb()->pushBuilder(); }
-    template<typename Func>
-    auto pushBuilder(std::shared_ptr<Builder> b, Func&& func) {
-        return pb()->pushBuilder(std::move(b), std::forward(func));
-    }
```
bbannier, line 56: I think this should be the only version of this we use, but sadly we use `pushBuilder/popBuilder` everywhere here which is much harder to use correctly. Unless we really need to get rid of this, I personally would prefer keeping this in, for better code we write in the future or just as a reminder to try harder.

spicy/toolchain/src/compiler/codegen/parsers/types.cc
```diff
@@ -37,10 +37,6 @@ struct TypeParser {
     auto context() { return pb->context(); }
     auto pushBuilder(std::shared_ptr<Builder> b) { return pb->pushBuilder(std::move(b)); }
     auto pushBuilder() { return pb->pushBuilder(); }
-    template<typename Func>
-    auto pushBuilder(std::shared_ptr<Builder> b, Func&& func) {
-        return pb->pushBuilder(std::move(b), std::forward(func));
-    }
```
bbannier, line 43: Same here.

spicy/toolchain/src/compiler/codegen/parsers/literals.cc
evantypanski: marked both `[[maybe_unused]]`

hilti/runtime/CMakeLists.txt
evantypanski: rather than spend an hour remembering cmake syntax I asked GPT-5.6 to do this. The `-w` is annoying and it kept the runtime changes, but all looks good IMO.

bbannier: Totally worth it, thanks! The changes look alright to me.

bbannier (approved):
Looks good to me assuming CI passes.

PR #2483, opened 2026-08-28
add the sub-offset operator

### LINKED ISSUE #2335: Add `bytes.sub(end)` and `stream.sub(end)` without an iterator
Both `bytes` and `stream` have a `sub` operator which they can use to return a subrange from offsets:

https://github.com/zeek/spicy/blob/9903fc6c26434319484950a3c3aaadc10ff1314a/hilti/toolchain/src/ast/operators/bytes.cc#L742-L769

https://github.com/zeek/spicy/blob/9903fc6c26434319484950a3c3aaadc10ff1314a/hilti/toolchain/src/ast/operators/stream.cc#L649-L677

But, there is no `sub(end)` operator, where `end` is an offset. Users either have to do `bytes.sub(0, end)` or `bytes.sub(bytes.at(end))`.

I believe these are already implemented in the runtime library:

https://github.com/zeek/spicy/blob/9903fc6c26434319484950a3c3aaadc10ff1314a/hilti/runtime/include/types/bytes.h#L398-L404

https://github.com/zeek/spicy/blob/9903fc6c26434319484950a3c3aaadc10ff1314a/hilti/runtime/include/types/stream.h#L1437-L1443

So, all this needs is to expose those to users, output the correct overload (like [here](https://github.com/zeek/spicy/blob/9903fc6c26434319484950a3c3aaadc10ff1314a/hilti/toolchain/src/compiler/codegen/operators.cc#L229-L232)), and document it. 

It's just mildly surprising that this is not exposed to users

mbKaleb:
Fixes #2335

## Added

+ sub-offset operator and corresponding documentation.
+ corresponding test-cases under bytes,stream, and operators, matching convention over coverage.

## Tested

- Complete btest suite: 880 passed, 0 failed, 33 skipped
- New btest coverage: single-offset `sub` on bytes and stream views with
  boundary cases (`0`, mid, full length), a const-operand case, and a
  Spicy-syntax assert (`tests/hilti/types/bytes/ops.hlt`,
  `tests/hilti/types/stream/view.hlt`, `tests/spicy/types/bytes/operators.spicy`)


mbKaleb:
@evantypanski Any thoughts?

evantypanski:
Hey sorry I haven't gotten to this, I'll get to it either today or tomorrow. I've had some pressing stuff I wanted to get out of the way first :) thanks for your patience!

While waiting, to what extent did you use AI for this PR? Zeek has an [AI policy](https://github.com/zeek/zeek/blob/master/AI_POLICY.md), we don't have it in Spicy yet, but probably should.

hilti/toolchain/include/ast/node-tag.h
```diff
@@ -202,6 +202,7 @@ constexpr Tag ToUIntAscii = 930;
 constexpr Tag ToUIntBinary = 931;
 constexpr Tag Unequal = 932;
 constexpr Tag UpperCase = 933;
+constexpr Tag SubOffset = 934;
```
evantypanski, line 205: It's kind of annoying, but could you move this next to `SubOffsets`? So this'd be `921`, then you bump the number for all the subsequent tags.

That's what we've done eg here: bf9277176f8956a30b231f09b43cd1f17eee3044

hilti/toolchain/include/ast/node-tag.h
```diff
@@ -476,6 +477,7 @@ constexpr Tag SubIterators = 3314;
 constexpr Tag SubOffsets = 3315;
 constexpr Tag UnequalBytes = 3316;
 constexpr Tag UnequalView = 3317;
+constexpr Tag SubOffset = 3318;
```
evantypanski, line 480: same as before, can this be 3316 then bump the other 2?

tests/hilti/types/bytes/ops.hlt
```diff
@@ -44,6 +44,9 @@ assert y.find(b"890") == (False, y.at(8));
 
 global auto y2 = b"abcdedede";
 assert y2.sub(y2.find(b"def")[1], end(y2)) == b"de";
+assert y.sub(3) == b"012";
+assert y.sub(0) == b"";
+assert y.sub(|y|) == y;
```
evantypanski, line 49: We currently lack a test from hilti/spicy for an offset that's too large, mind adding one here? It should clamp. Untested suggestion:

```
assert y.sub(|y| + 1337) == y;
```

tests/hilti/types/stream/view.hlt
```diff
@@ -27,6 +27,11 @@ assert ! v.starts_with(b"XYZ");
 
 assert |v| == 9;
 
+assert v.sub(3) == b"123";
+assert v.sub(0) == b"";
+assert v.sub(|v|) == b"123456789";
+assert all.sub(5) == b"01234";
```
evantypanski, line 33: similar to my comment above on `bytes`, but views are different, it'd be nice to enshrine that in a test. `v.sub(1337)` will grow with a cap at `1337` if the underlying stream grows. For example, run this, it'll pass:

```
module Test {

global stream s = stream(b"abc");
global view<stream> my_view = s;
global view<stream> subbed = my_view.sub(1337);

assert subbed == b"abc";

s += b"def";
assert subbed == b"abcdef";
}
```

with:

```
hiltic -j test.hlt
```

We should test something along those lines. That's tricky behavior, and different from `bytes`.

EDIT: Oh, and what if the view `sub` was within range, but then appending puts it out? Like this:

```
module Test {

import hilti;

global stream s = stream(b"abc");
global view<stream> my_view = s;
global view<stream> subbed = my_view.sub(5);

assert subbed == b"abc";

s += b"a lot more than 5";
assert subbed == b"abca ";
}
```

Testing this behavior would be nice, since I don't actually know if we have runtime tests for this case. The more the merrier

evantypanski (commented):
The core looks right to me, thanks! I had a couple comments on tests because when playing with it, I found it's actually slightly tricky for views. Some tests would help. Feel free to add more edge cases, too.

And my question before will still stand: did you use AI for this PR, and to what extent? Just so we can keep track

mbKaleb:
> The core looks right to me, thanks! I had a couple comments on tests because when playing with it, I found it's actually slightly tricky for views. Some tests would help. Feel free to add more edge cases, too.
> 
> And my question before will still stand: did you use AI for this PR, and to what extent? Just so we can keep track

- I used Claude _Fable 5.1_ to find this issue and explain the codebase+downstream tools. 

- In addition, all written documents and code were reviewed by Fable 5.1 but the language model **did not make any final decisions or changes**

evantypanski:
@mbKaleb were you going to go through and address the comments? It's mainly ensuring proper testing.

mbKaleb:
> @mbKaleb were you going to go through and address the comments? It's mainly ensuring proper testing.

Sorry for delay, almost ready.

tests/hilti/types/stream/view.hlt
mbKaleb: Screenshot 2026-09-21 at 1 37 16 PM

tests/hilti/types/bytes/ops.hlt
mbKaleb: added, but I found sub already has this sort of test under 
[hilti/runtime/src/tests/bytes.cc:354](https://github.com/zeek/spicy/blob/main/hilti/runtime/src/tests/bytes.cc#L354-L363)

evantypanski (approved):
LGTM, thanks!

PR #2487, opened 2026-09-04
Enable the fiber runtime on PowerPC64

pkubaj:
The three stack-inspection helpers (Fiber::Fiber, StackBuffer::activeRegion,
StackBuffer::liveRemainingSize) are already portable -- they use
__builtin_frame_address(0) and the fiber's saved stack pointer -- but were
gated behind a fixed architecture whitelist. Add __powerpc64__ so they are
enabled on PowerPC64, whose fiber context switching the bundled fiber library
supports (ELFv2, both endiannesses).

Verified on FreeBSD 15.1/powerpc64 (big-endian, POWER9).


bbannier (approved):
Thanks!

PR #2503, opened 2026-09-23
Fix codegen for modules with clashing names

### LINKED ISSUE #2324: Incorrect resolution of ambiguous function IDs.
The attached btest produces this output:

```
$ btest -d spicy/optimization/function-with-name-clash.hlt
[...]
  % cat .stderr
  a/foo(): a
  a/foo(): b
```

Expected  would be `b/foo(): b` for the second line.

The AST contains two separate functions of the name `foo:f()`, and it seems name resolution picks the wrong one in the 2nd case.

[function-with-name-clash.hlt.txt](https://github.com/user-attachments/files/26784010/function-with-name-clash.hlt.txt)

bbannier:
The C++ codegen mixed `fullyQualifiedID()` (which uses the original
module name) with `module->id()` (which uses the unique module name)
when deciding whether to emit function implementations and when
generating C++ namespaces for dependencies. For modules with clashing
names (e.g., two `foo` modules in different directories), this caused
the function body to be omitted and references to use a mismatched
namespace, resulting in undefined symbol errors at link time.

Use `canonicalID()` for the module-identity checks in
`addCxxDeclarationsFor`, and remap the module prefix in
`expression::Name` to the resolved declaration's parent module unique
name for qualified cross-module function calls. With this, the second
`foo` module's extern declarations and call sites consistently use the
`foo_2` namespace, matching the definition side.

Closes #2324.

> [!NOTE]
> All code changes in this PR were generated with Claude Opus 4.6. I reviewed all changes and am accountable.


hilti/toolchain/src/compiler/optimizer/passes/feature-requirements.cc
```diff
@@ -215,7 +215,7 @@ struct Collector : public optimizer::visitor::Collector {
         while ( type_->type()->isReferenceType() )
             type_ = type_->type()->dereferencedType();
 
-        auto type_id = type_->type()->typeID();
+        const auto& type_id = type_->type()->canonicalID();
```
evantypanski, line 218: this is used in `scope::lookupID` a few lines down

wouldn't the canonical ID result in a wrong lookup if 2 modules clash?

evantypanski (changes requested):

hilti/toolchain/src/compiler/optimizer/passes/feature-requirements.cc
bbannier: Oops of course, I switched that use in the `scope::lookupID` to use the type ID again. This should still be fine since it explicitly performs a scope walk.

I also added a cleanup commit at the end changing the bindings of `canonicalID()` from `type_id` to `id`, but it still unfortunate how different ID kinds can be mixed so freely.

evantypanski (approved):
LGTM

PR #2504, opened 2026-09-23
Remove CI deviations in Windows job

bbannier:
The Windows job was explicitly disabling ccache during test execution, and also running failed tests with retries. All other do the former to speed up running of the test suite, and disable retries since we are not interested in hiding flaky tests.

This patch removes these deviations.


bbannier:
Abandoning this as the way we compile Spicy/HILTI code on Windows is pretty complicated with an intermediary wrapper executable which uses GCC-style flags not understood by ccache when running with msvc.

